PR #30, opened 2019-03-29
Add dialogue-serde, remove OkHttp dependency

uschi2000:


uschi2000:
Adapts the serialization framework from conjure-undertow

PR #34, opened 2019-03-29
Add dialogue-java-client

uschi2000:


uschi2000:
FYI, @ellisjoe @markelliot , this adds a Java11 client to Dialogue. I refactored this repo a fair amount to match conjure-undertow in terms of how serialization works (https://github.com/palantir/dialogue/pull/30 ) and on the way removed the okhttp coupling. This PR isn't quite done, but it should be easy enough.

ellisjoe:
What's the plan for how things like backoff, retry, blacklisting, limiting, etc are going to work here?

uschi2000:
There's no concrete plan.

ellisjoe:
Alright. I feel like that's where the interesting stuff is? Most of the issues we have today are less related to okhttp and more related to that logic.

All for using the java http client though.

PR #49, opened 2019-04-11
Add HttpUrl class

uschi2000:


uschi2000:
fyi @ellisjoe

uschi2000:
nvm, will have to tweak this to be correct.

dialogue-java-client/src/main/java/com/palantir/dialogue/HttpUrl.java
```diff
@@ -0,0 +1,124 @@
...
+import com.palantir.logsafe.Preconditions;
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.net.URLEncoder;
+import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
+import java.util.List;
+
+/** A simplistic URL builder, not tuned for performance. */
+public final class HttpUrl {
```
markelliot: maybe `UrlBuilder`?

dialogue-java-client/src/main/java/com/palantir/dialogue/HttpUrl.java
```diff
@@ -0,0 +1,124 @@
...
+    public static Builder https() {
+        return new Builder("https");
+    }
+
+    static class Builder {
+        private final String protocol;
+        private String host;
+        private int port = -1;
+        private List<String> pathSegments = new ArrayList<>();
+        private List<String> queryNamesAndValues = new ArrayList<>(); // alternating (name, value) pairs
```
markelliot: might make more sense for this to be a Multimap

dialogue-java-client/src/main/java/com/palantir/dialogue/HttpUrl.java
```diff
@@ -0,0 +1,124 @@
...
+        this.url = new URL(builder.protocol, builder.host, builder.port, file);
+    }
+
+    private static String encode(String string) {
+        return URLEncoder.encode(string, StandardCharsets.UTF_8);
+    }
+
+    private static String encodePath(List<String> pairs) {
+        StringBuilder result = new StringBuilder();
+        for (int i = 0; i < pairs.size(); i += 1) {
```
markelliot: possibly nicer to write this as:

```java
Joiner.on("/").join(paths)
```

dialogue-java-client/src/main/java/com/palantir/dialogue/HttpUrl.java
```diff
@@ -0,0 +1,124 @@
...
+        file = path.isEmpty() ? file : "/" + file;
+
+        this.url = new URL(builder.protocol, builder.host, builder.port, file);
+    }
+
+    private static String encode(String string) {
+        return URLEncoder.encode(string, StandardCharsets.UTF_8);
+    }
+
+    private static String encodePath(List<String> pairs) {
```
markelliot: `pairs` -> `paths`

dialogue-java-client/src/main/java/com/palantir/dialogue/HttpUrl.java
markelliot: also might consider just making the top-level class the builder and have the 'build' method be `toUrl`

markelliot:
you may want to consider looking at URI, [the docs on URL say this](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/net/URL.html):

> The recommended way to manage the encoding and decoding of URLs is to use URI, and to convert between these two classes using toURI() and URI.toURL().

uschi2000:
problem with the URI class is that it cannot distinguish between `url.pathSegment("foo").pathSegment("bar")` (which gives http://host/foo/bar) and `url.pathSegment("foo/bar")` (which gives http://host/foo%2Fbar).

dialogue-java-client/src/main/java/com/palantir/dialogue/HttpUrl.java
uschi2000: done

dialogue-java-client/src/main/java/com/palantir/dialogue/HttpUrl.java
uschi2000: I suspect the list construction has fewer allocations; I think it's OK since it's just used internally?

dialogue-java-client/src/main/java/com/palantir/dialogue/HttpUrl.java
uschi2000: done here, but not below to save object allocations

dialogue-java-client/src/main/java/com/palantir/dialogue/HttpUrl.java
uschi2000: done

uschi2000:
take another look please, @markelliot

dialogue-java-client/src/main/java/com/palantir/dialogue/UrlBuilder.java
```diff
@@ -0,0 +1,189 @@
...
+     * {@code 192.168.0.1}), and IPv6 addresses of the form
+     * {@code [2010:836B:4179::836B:4179]} (note the enclosing square brackets).
+     */
+    public UrlBuilder host(String theHost) {
+        this.host = theHost;
+        return this;
+    }
+
+    public UrlBuilder port(int thePort) {
+        this.port = thePort;
```
markelliot, line 67: Consider a validation of range

dialogue-java-client/src/main/java/com/palantir/dialogue/UrlBuilder.java
```diff
@@ -0,0 +1,189 @@
...
+/** A simplistic URL builder, not tuned for performance. */
+public final class UrlBuilder {
+
+    private static final Joiner SLASH_JOINER = Joiner.on('/');
+
+    private final String protocol;
+    private String host;
+    private int port = -1;
+    private List<String> pathSegments = new ArrayList<>();
+    private List<String> queryNamesAndValues = new ArrayList<>(); // alternating (name, value) pairs
```
markelliot, line 42: I think it may be clearer to use a Multimap

uschi2000: sorry, thought I had commented on your previous comment; I used a list to save some object allocations. maybe it doesn't matter to much. switching to multimap.

uschi2000: actually, due to the encode logic the joining code becomes either a lot more costly or much harder to read with a multimap, so I think i'll leave it as is.

dialogue-java-client/src/main/java/com/palantir/dialogue/UrlBuilder.java
uschi2000: done

uschi2000:
take another look please, @markelliot

dialogue-java-client/src/main/java/com/palantir/dialogue/UrlBuilder.java
markelliot: ok

markelliot (approved):

PR #54, opened 2019-04-17
Add UrlBuilder#encodedPathSegments

uschi2000:


uschi2000:
fyi @markelliot ; will move on, but let me know if you have any comments

dialogue-target/src/main/java/com/palantir/dialogue/UrlBuilder.java
```diff
@@ -59,35 +61,50 @@ private UrlBuilder(String protocol) {
...
         return this;
     }
 
+    /**
+     * Adds the given URL-encoded path segment (one or more) to the list of segments, or fails if the given segments
+     * contain forbidden characters. Note that leading or trailing slashes are preserved, for instance {@code url
+     * .pathSegment("foo").encodedPathSegments{"/bar/"}.pathSegment("baz")} yields a URL with path
+     * {@code foo//bar//baz} (note the empty segments).
+     */
+    public UrlBuilder encodedPathSegments(String segments) {
```
markelliot, line 81: the method naming here is a little confusing. I think this "encoded" in the sense that the argument is already encoded? Not sure I have a better suggestion, but seems like it's going to catch someone offguard down the road.

uschi2000: yeah, I agree, but I also don't have a better proposal. maybe `preEncodedPathSegments`? I added a fair amount of docs and an example above, at least.

PR #55, opened 2019-04-18
Add support for query params

uschi2000:


uschi2000:
ping, @ellisjoe

markelliot (approved):
Would be good to think about parameterizing the tests instead of duplicating code.

uschi2000:
yeah, I have a half-balked PR already that harmonizes the tests.

PR #56, opened 2019-04-18
Use UrlBuilder to construct all internal URLs; fix empty case handling

uschi2000:


uschi2000:
maybe not so interesting, but feel free to take a look, @ellisjoe

PR #66, opened 2019-04-24
Simplify and harmonize tests for Java11- and OkHttp-based clients

uschi2000:


uschi2000:
as discussed, @markelliot

markelliot:
Thanks!

PR #71, opened 2019-04-25
Add support for PLAIN serialization

uschi2000:


dialogue-example/src/main/java/com/palantir/dialogue/example/SampleServiceClient.java
```diff
@@ -80,28 +84,31 @@ public HttpMethod httpMethod() {
...
 
             @Override
-            public String stringToString(String objectId, String header, String body) {
-                Preconditions.checkNotNull(objectId, "objectId parameter must not be null");
+            public SampleObject objectToObject(
+                    String path, OffsetDateTime header, List<ResourceIdentifier> query, SampleObject body) {
+                Preconditions.checkNotNull(path, "objectId parameter must not be null");
                 Preconditions.checkNotNull(header, "header parameter must not be null");
                 Preconditions.checkNotNull(body, "body parameter must not be null");
                 Request request = Request.builder()
```
uschi2000, line 100: this is the interesting bit

uschi2000:
Mark, I tagged the interesting bit.

dialogue-example/src/main/java/com/palantir/dialogue/example/SampleService.java
```diff
@@ -16,8 +16,43 @@
...
+import com.palantir.ri.ResourceIdentifier;
+import java.time.OffsetDateTime;
+import java.util.List;
+import java.util.Objects;
+
 // Example of the interface code conjure would generate for a simple SampleService.
 public interface SampleService {
-    String stringToString(String objectId, String header, String body);
+
+    final class SampleObject {
```
markelliot: Conjure probably doesn't end up nesting this object in generated code?

markelliot (approved):

dialogue-example/src/main/java/com/palantir/dialogue/example/SampleService.java
uschi2000: Correct. I'll move it up.

PR #74, opened 2019-04-30
Use ListenableFuture rather than Call

ellisjoe:
## Before this PR
We passed an `Observer` into the `execute` method which has made it difficult to chain things together. Also, the call interface forces the Channel implementations to deal with conjure specific error decoding, which I'd prefer done at a higher level (eg: transform the `ListenableFuture<Response>` into something that looks like an `Either<Response, RemoteException>`)

## After this PR
We now return a `ListenableFuture<Response>` which means `Channel`s no longer need to deal with error decoding and also makes it easer to chain together Channels that would like to transform the response in some way (eg: retry on failure)


markelliot (approved):
nice

dialogue-client-test-lib/src/main/java/com/palantir/dialogue/AbstractChannelTest.java
```diff
@@ -107,21 +103,21 @@ public void before() {
     @Test
     public void respectsBasePath_emptyBasePath() throws InterruptedException {
         channel = createChannel(server.url("").url(), MoreExecutors.newDirectExecutorService());
-        channel.createCall(endpoint, request).execute(observer);
+        channel.createCall(endpoint, request);
```
uschi2000, line 106: this method should now be called "call" or "execute" or something

dialogue-client-test-lib/src/main/java/com/palantir/dialogue/AbstractChannelTest.java
```diff
@@ -196,165 +192,122 @@ public void fillsQueryParameters() throws Exception {
...
-        call.execute(observer);
-        assertThatThrownBy(() -> call.execute(observer))
-                .hasMessage("Error, this call was already executed");
-    }
-
-    @Test
-    public void canCancelNonExecutedCalls() {
-        assertThatCode(() -> channel.createCall(endpoint, request).cancel()).doesNotThrowAnyException();
+        ListenableFuture<Response> call = channel.createCall(endpoint, request);
+        Futures.addCallback(call, callback, MoreExecutors.directExecutor());
```
uschi2000, line 270: this is a bit odd now; why not assert that the future completed successfully?

dialogue-client-test-lib/src/main/java/com/palantir/dialogue/AbstractChannelTest.java
```diff
@@ -196,165 +192,122 @@ public void fillsQueryParameters() throws Exception {
...
     public void callCancellationIsObservedAsException() throws InterruptedException {
-        channel.createCall(endpoint, request).execute(mock(Observer.class));  // drain enqueued response
+        channel.createCall(endpoint, request);  // drain enqueued response
 
         channel = createChannel(server.url("").url(), Executors.newCachedThreadPool());
-        Call call = channel.createCall(endpoint, request);
-        call.execute(observer);
-        call.cancel();
+        ListenableFuture<Response> call = channel.createCall(endpoint, request);
+        Futures.addCallback(call, callback, MoreExecutors.directExecutor());
```
uschi2000, line 288: as above, this doesn't make much sense now

dialogue-client-test-lib/src/main/java/com/palantir/dialogue/AbstractChannelTest.java
```diff
@@ -196,165 +192,122 @@ public void fillsQueryParameters() throws Exception {
...
-                .setResponseCode(500)
-                .addHeader("Content-Type", "application/json"));
-        channel.createCall(endpoint, request).execute(observer);
-        verify(observer, never()).success(any());
-        verify(observer, never()).exception(any());
-        verify(observer).failure(any());
-    }
-
-    @Test
-    public void unparseableNonConjureErrorsSurfaceExceptions() throws IOException {
```
uschi2000, line 349: how do you think you will handle conjure vs application vs qos errors now?

dialogue-client-test-lib/src/main/java/com/palantir/dialogue/AbstractChannelTest.java
```diff
@@ -196,165 +192,122 @@ public void fillsQueryParameters() throws Exception {
...
     public void connectionErrorsSurfaceAsExceptions() throws IOException {
         server.shutdown();
-        channel.createCall(endpoint, request).execute(observer);
-        verify(observer, never()).success(any());
-        verify(observer).exception(any());
-        verify(observer, never()).failure(any());
-    }
-
-    @Test
-    public void conjureErrorsSurfaceAsFailures() throws IOException {
```
uschi2000, line 333: what happened to this behavior?

dialogue-client-test-lib/src/main/java/com/palantir/dialogue/AbstractChannelTest.java
ellisjoe: done

dialogue-client-test-lib/src/main/java/com/palantir/dialogue/AbstractChannelTest.java
ellisjoe: done

dialogue-client-test-lib/src/main/java/com/palantir/dialogue/AbstractChannelTest.java
ellisjoe: done

dialogue-client-test-lib/src/main/java/com/palantir/dialogue/AbstractChannelTest.java
ellisjoe: Going to have something that maps a `Response` into a `ResponseOrError` sort of thing

dialogue-client-test-lib/src/main/java/com/palantir/dialogue/AbstractChannelTest.java
ellisjoe: As above, I think you have something that maps the `Response` into something that differentiates an error from a conjure error you can allow callers to deal with that type. There's no need to deal with conjure errors down here, because nothing we're doing at the http layer here needs to know anything about conjure

ellisjoe:
PR here: https://github.com/palantir/dialogue/pull/79

PR #75, opened 2019-05-01
[improvement] Add RetryingChannel

ellisjoe:

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
@@ -34,8 +34,8 @@ import java.util.function.Supplier;
  * Retries calls to the underlying channel upon failure.
  */
 public final class RetryingChannel implements Channel {
-    private static final Executor direct = MoreExecutors.directExecutor();
-    private static final int DEFAULT_MAX_RETRIES = 3;
+    private static final Executor DIRECT_EXECUTOR = MoreExecutors.directExecutor();
+    private static final int DEFAULT_MAX_RETRIES = 4;
 
     private final Channel delegate;
     private final int maxRetries;
@@ -54,15 +54,15 @@ public final class RetryingChannel implements Channel {
     public ListenableFuture<Response> createCall(Endpoint endpoint, Request request) {
         SettableFuture<Response> future = SettableFuture.create();
 
-        ListenableFuture<Response> call = delegate.createCall(endpoint, request);
-        FutureCallback<Response> retryer = new RetryingCallback<>(() -> delegate.createCall(endpoint, request), future);
-        Futures.addCallback(call, retryer, direct);
+        Supplier<ListenableFuture<Response>> callSupplier = () -> delegate.createCall(endpoint, request);
+        FutureCallback<Response> retryer = new RetryingCallback<>(callSupplier, future);
+        Futures.addCallback(callSupplier.get(), retryer, DIRECT_EXECUTOR);
 
         return future;
     }
 
     private final class RetryingCallback<T> implements FutureCallback<T> {
-        private final AtomicInteger failures = new AtomicInteger();
+        private final AtomicInteger failures = new AtomicInteger(0);
         private final Supplier<ListenableFuture<T>> runnable;
         private final SettableFuture<T> delegate;
 
@@ -79,7 +79,7 @@ public final class RetryingChannel implements Channel {
         @Override
         public void onFailure(Throwable throwable) {
             if (failures.incrementAndGet() < maxRetries) {
-                Futures.addCallback(runnable.get(), this, direct);
+                Futures.addCallback(runnable.get(), this, DIRECT_EXECUTOR);
             } else {
                 delegate.setException(throwable);
             }
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
@@ -73,11 +73,11 @@ final class RetryingChannel implements Channel {
         }
 
         @Override
-        public void onFailure(Throwable t) {
+        public void onFailure(Throwable throwable) {
             if (failures.incrementAndGet() < maxRetries) {
                 Futures.addCallback(runnable.get(), this, direct);
             } else {
-                delegate.setException(t);
+                delegate.setException(throwable);
             }
         }
     }
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
@@ -1,0 +1,84 @@
+/*
+ * (c) Copyright 2019 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.dialogue.core;
+
+import com.google.common.util.concurrent.FutureCallback;
+import com.google.common.util.concurrent.Futures;
+import com.google.common.util.concurrent.ListenableFuture;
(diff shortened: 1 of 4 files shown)


ellisjoe:
should be rebased and merged after https://github.com/palantir/dialogue/pull/74

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -0,0 +1,83 @@
...
+import com.palantir.dialogue.Channel;
+import com.palantir.dialogue.Endpoint;
+import com.palantir.dialogue.Request;
+import com.palantir.dialogue.Response;
+import java.util.concurrent.Executor;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.Supplier;
+
+final class RetryingChannel implements Channel {
+    private static final Executor direct = MoreExecutors.directExecutor();
```
markelliot: nit: DIRECT_EXECUTOR

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -0,0 +1,83 @@
...
+    private static final int DEFAULT_MAX_RETRIES = 3;
```
markelliot: cjr sets this to 4 by default

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -0,0 +1,83 @@
...
+
+        ListenableFuture<Response> call = delegate.createCall(endpoint, request);
+        FutureCallback<Response> retryer = new RetryingCallback<>(() -> delegate.createCall(endpoint, request), future);
+        Futures.addCallback(call, retryer, direct);
+
+        return future;
+    }
+
+    private final class RetryingCallback<T> implements FutureCallback<T> {
+        private final AtomicInteger failures = new AtomicInteger();
```
markelliot: I know this defaults to 0 but it'd be nice to be explicit, in the call so that others reading the code know that without having to hit the docs, too.

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -0,0 +1,83 @@
...
+        this.delegate = delegate;
+        this.maxRetries = maxRetries;
+    }
+
+    @Override
+    public ListenableFuture<Response> createCall(Endpoint endpoint, Request request) {
+        SettableFuture<Response> future = SettableFuture.create();
+
+        ListenableFuture<Response> call = delegate.createCall(endpoint, request);
+        FutureCallback<Response> retryer = new RetryingCallback<>(() -> delegate.createCall(endpoint, request), future);
```
markelliot: aesthetics:
```java
Supplier<ListenableFuture<Response>> callSupplier = () -> delegate.createCall(endpoint, request);
FutureCallback<Response> retryer = new RetryingCallback<>(callSupplier, future);
Futures.addCallback(callSupplier.get(), retryer, direct);
```

dialogue-core/src/test/java/com/palantir/dialogue/core/RetryingChannelTest.java
```diff
@@ -0,0 +1,118 @@
...
+import java.util.concurrent.ExecutionException;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.Mock;
+import org.mockito.junit.MockitoJUnitRunner;
+
+@RunWith(MockitoJUnitRunner.class)
+public class RetryingChannelTest {
+    private static final TestResponse expectedResponse = new TestResponse();
```
markelliot: expectedResponse -> EXPECTED_RESPONSE

markelliot (approved):
nits

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
ellisjoe: done

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
ellisjoe: done

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
ellisjoe: done

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
ellisjoe: done

dialogue-core/src/test/java/com/palantir/dialogue/core/RetryingChannelTest.java
ellisjoe: done

markelliot (approved):

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -0,0 +1,88 @@
...
+            this.delegate = delegate;
+        }
+
+        @Override
+        public void onSuccess(T result) {
+            delegate.set(result);
+        }
+
+        @Override
+        public void onFailure(Throwable throwable) {
```
uschi2000, line 80: is the plan to only retry on certain errors here later?

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -0,0 +1,88 @@
...
+            if (failures.incrementAndGet() < maxRetries) {
+                Futures.addCallback(runnable.get(), this, DIRECT_EXECUTOR);
```
uschi2000, line 82: not all requests can be retried, probably need to worry about that here

ellisjoe: so I think the retryability of an endpoint should be specified in the `Endpoint` definition above, at which point we can either wire up this call back or not [here](https://github.com/palantir/dialogue/pull/75/files#diff-3b3ed2f3bd0b0b9d1adce9a9fa7ab891R59)

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
ellisjoe: yeah. what sort of things shouldn't be retried?

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
uschi2000: actually, there are two reasons a request may not be retryable: first, if the endpoint statically delcares it (could be derived from the API def), or second when the request carries read-once data like streams. the latter cannot be derived from the endpoint alone.

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
uschi2000: everything that's caused by a server-side ServiceException.

uschi2000: today, conjure-java only retries IOExceptions (connection refused etc) and 429 and 503.

ellisjoe: ah, so anything that would be a server side exception or 429/503 would actually go through the `onSuccess` codepath because we actually received a response. So I think that mostly means we want to add some retrying to the onSuccess method, and should be fine retrying everything here?

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
ellisjoe: I think we're going to need to define whether or not the request body is streamed in the endpoint, or possibly as part of the `RequestBody`. We should also make it part of the contract of `RequestBody#content` that each time it's called it must hand back a new `InputStream`

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
uschi2000: ah, yeah thanks didn't read carefully enough. yep.

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
uschi2000: yes it could be part of the request body.

PR #76, opened 2019-05-01
[improvement] Add concurrency limits and load balancing

ellisjoe:
## After this PR
Adds a ConcurrencyLimitingChannel that only allows calls to be executed if there is sufficient capacity available on the underlying channel. Also adds a RoundRobinChannel which rotates through channels when sending requests while skipping over channels that don't have any available capacity.

`LimitedChannel` may be of some interest.

Think we'll want to add another thing around the RoundRobinChannel which has some fallback behavior (eg: send the request to a random host when non of the LimitedChannels have any capacity, or queue it, etc.)


dialogue-core/src/main/java/com/palantir/dialogue/core/ConcurrencyLimitedChannel.java
```diff
@@ -0,0 +1,99 @@
...
+                .build();
+        Limiter<Object> limiter = SimpleLimiter.newBuilder()
+                .limit(WindowedLimit.newBuilder().build(aimdLimit))
+                .build();
+        return new ConcurrencyLimitedChannel(delegate, limiter);
+    }
+
+    @Override
+    public Optional<ListenableFuture<Response>> maybeCreateCall(Endpoint endpoint, Request request) {
+        return limiter.acquire(CONTEXT).map(listener -> {
```
markelliot: limiters should be keyed by endpoint

dialogue-core/src/main/java/com/palantir/dialogue/core/ConcurrencyLimitedChannel.java
```diff
@@ -0,0 +1,99 @@
...
+import com.palantir.dialogue.Channel;
+import com.palantir.dialogue.Endpoint;
+import com.palantir.dialogue.Request;
+import com.palantir.dialogue.Response;
+import java.util.Optional;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * A channel that monitors the successes and failures of requests in order to determine the number of concurrent
+ * requests allowed to a particular channel. If the channel's concurrency limit has been reach, the call will not be
```
markelliot: if the channel's concurrency limit has been reached, the maybeCreateCall method returns empty

dialogue-core/src/main/java/com/palantir/dialogue/core/ConcurrencyLimitedChannel.java
ellisjoe: done

dialogue-core/src/main/java/com/palantir/dialogue/core/ConcurrencyLimitedChannel.java
ellisjoe: done

dialogue-core/src/main/java/com/palantir/dialogue/core/RoundRobinChannel.java
```diff
@@ -0,0 +1,58 @@
...
+    public RoundRobinChannel(List<LimitedChannel> delegates) {
+        this.delegates = delegates;
+    }
+
+    @Override
+    public Optional<ListenableFuture<Response>> maybeCreateCall(Endpoint endpoint, Request request) {
+        int host = currentHost.getAndUpdate(value -> toIndex(value + 1));
+
+        for (int i = 0; i < delegates.size(); i++) {
+            LimitedChannel channel = delegates.get(toIndex(host + i));
```
uschi2000, line 50: should take an immutable copy of this list to avoid the race condition

dialogue-core/src/main/java/com/palantir/dialogue/core/ConcurrencyLimitedChannel.java
```diff
@@ -0,0 +1,113 @@
...
+            if (DROP_CODES.contains(result.code())) {
+                listener.onDropped();
+            } else {
+                listener.onSuccess();
+            }
+        }
+
+        @Override
+        public void onFailure(Throwable throwable) {
+            listener.onIgnore();
```
uschi2000, line 110: what would a channel timeout or a connection refused exception mean for the limiter?

dialogue-core/src/main/java/com/palantir/dialogue/core/RoundRobinChannel.java
uschi2000: (or if you want the list to be mutable you should document that and remove the race)

uschi2000: (to be explicit: you can get a out of bound exception if the host list shrinks under you)

dialogue-core/src/test/java/com/palantir/dialogue/core/ConcurrencyLimitedChannelTest.java
```diff
@@ -0,0 +1,118 @@
...
+        when(delegate.createCall(endpoint, request)).thenReturn(responseFuture);
+    }
+
+    @Test
+    public void testLimiterAvailable_successfulRequest() {
+        limitAvailable();
+        responseCode(200);
+
+        assertThat(channel.maybeCreateCall(endpoint, request)).contains(responseFuture);
+
```
uschi2000, line 61: nit: would remove this line break (same below)

dialogue-core/src/test/java/com/palantir/dialogue/core/ConcurrencyLimitedChannelTest.java
```diff
@@ -0,0 +1,118 @@
...
+    @Test
+    public void testUnavailable() {
+        limitUnvailable();
+
+        assertThat(channel.maybeCreateCall(endpoint, request)).isEmpty();
+
+        verifyZeroInteractions(listener);
+    }
+
+    private void responseCode(int code) {
```
uschi2000: nit: would call this mockResponseCode (same below)

dialogue-core/src/main/java/com/palantir/dialogue/core/LimitedChannel.java
```diff
@@ -0,0 +1,31 @@
...
+import com.palantir.dialogue.Response;
+import java.util.Optional;
+
+/**
+ * Similar to {@link com.palantir.dialogue.Channel}, but may not actually execute the call (eg: when the channel is
+ * overloaded).
+ */
+public interface LimitedChannel {
+    Optional<ListenableFuture<Response>> maybeCreateCall(Endpoint endpoint, Request request);
+}
```
carterkozak, line 31: I'm not sure this should be public since we allow pre-flight request limiting to be disabled entirely.

dialogue-core/src/main/java/com/palantir/dialogue/core/RoundRobinChannel.java
```diff
@@ -0,0 +1,58 @@
...
+
+/**
+ * Round robins requests across many channels, attempting to choose a channel that has some available capacity.
+ */
+public final class RoundRobinChannel implements LimitedChannel {
+
+    private final AtomicInteger currentHost = new AtomicInteger(0);
+    private final List<LimitedChannel> delegates;
+
+    public RoundRobinChannel(List<LimitedChannel> delegates) {
```
carterkozak, line 37: I think we're missing a concept of a `ChannelSelector`. We will need to force rotation when requests fail in certain ways, but this is only aware of pre-flight limiting, and doesn't seem like the right place for exceptional response interpretation.

dialogue-core/src/main/java/com/palantir/dialogue/core/RoundRobinChannel.java
```diff
@@ -0,0 +1,58 @@
...
+import com.palantir.dialogue.Request;
+import com.palantir.dialogue.Response;
+import java.util.List;
+import java.util.Optional;
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * Round robins requests across many channels, attempting to choose a channel that has some available capacity.
+ */
+public final class RoundRobinChannel implements LimitedChannel {
```
carterkozak, line 32: Should this implement channel?

dialogue-core/src/main/java/com/palantir/dialogue/core/ConcurrencyLimitedChannel.java
markelliot: this matches today's behavior, which is that timeouts + connection refused do not actually adjust the limiter

dialogue-core/src/main/java/com/palantir/dialogue/core/LimitedChannel.java
markelliot: I think the idea here is that you'd wire the thing differently rather than configure the channels to behave differently

dialogue-core/src/main/java/com/palantir/dialogue/core/RoundRobinChannel.java
ellisjoe: I'd like to have something wrap this and turn it into a channel, and in there you can do things like queue requests when there isn't enough space on the underlying channel to execute the request at this moment.

dialogue-core/src/main/java/com/palantir/dialogue/core/RoundRobinChannel.java
ellisjoe: Think you just end up with another implementation of the `LimitedChannel` which returns Optional.empty() for some period of time after a particular exception. The goal here is to back each of these things into very separate layers to avoid entangling things like they are today in remoting

dialogue-core/src/main/java/com/palantir/dialogue/core/RoundRobinChannel.java
ellisjoe: 👍 done

dialogue-core/src/test/java/com/palantir/dialogue/core/ConcurrencyLimitedChannelTest.java
ellisjoe: done

dialogue-core/src/test/java/com/palantir/dialogue/core/ConcurrencyLimitedChannelTest.java
ellisjoe: done

dialogue-core/src/main/java/com/palantir/dialogue/core/ConcurrencyLimitedChannel.java
```diff
@@ -0,0 +1,113 @@
...
+                // Don't count slow calls as a sign of the server being overloaded
+                .timeout(Long.MAX_VALUE, TimeUnit.DAYS)
+                .build();
+        return SimpleLimiter.newBuilder()
+                .limit(WindowedLimit.newBuilder().build(aimdLimit))
+                .build();
+    }
+
+    @Override
+    public Optional<ListenableFuture<Response>> maybeCreateCall(Endpoint endpoint, Request request) {
```
uschi2000: maybeExecutor (as per other PR).

dialogue-core/src/main/java/com/palantir/dialogue/core/ConcurrencyLimitedChannel.java
```diff
@@ -0,0 +1,113 @@
...
+import com.palantir.dialogue.Request;
+import com.palantir.dialogue.Response;
+import java.time.Duration;
+import java.util.Optional;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Supplier;
+
+/**
+ * A channel that monitors the successes and failures of requests in order to determine the number of concurrent
+ * requests allowed to a particular channel. If the channel's concurrency limit has been reached, the maybeCreateCall
```
uschi2000: `{@link #maybeExecute}`

dialogue-core/src/main/java/com/palantir/dialogue/core/RoundRobinChannel.java
```diff
@@ -0,0 +1,59 @@
...
+
+    private final AtomicInteger currentHost = new AtomicInteger(0);
+    private final List<LimitedChannel> delegates;
```
uschi2000: nit: could make this an ImmutableList to make the contract clear

dialogue-core/src/test/java/com/palantir/dialogue/core/RoundRobinChannelTest.java
```diff
@@ -0,0 +1,78 @@
...
+    @Test
+    public void testIgnoresUnavailableChannels() {
+        when(channelA.maybeCreateCall(endpoint, request)).thenReturn(UNAVAILABLE);
+
+        assertThat(loadBalancer.maybeCreateCall(endpoint, request)).isEqualTo(CHANNEL_B_RESPONSE);
+        assertThat(loadBalancer.maybeCreateCall(endpoint, request)).isEqualTo(CHANNEL_B_RESPONSE);
+    }
+
+    @Test
+    public void testNoChannelsAvailable() {
```
uschi2000, line 72: nit: could also test the empty case `new RoundRobinChannel(ImmutableList.of())`

dialogue-core/src/test/java/com/palantir/dialogue/core/ConcurrencyLimitedChannelTest.java
```diff
@@ -0,0 +1,113 @@
...
+        verify(listener).onIgnore();
+    }
+
+    @Test
+    public void testUnavailable() {
+        mockLimitUnavailable();
+
+        assertThat(channel.maybeCreateCall(endpoint, request)).isEmpty();
+        verifyZeroInteractions(listener);
+    }
```
uschi2000, line 99: let's also add one test that uses a standard ConcurrencyLimitingChannel#create channel to test the default behavior.

uschi2000 (approved):
let's run with this and see where we end up.

dialogue-core/src/main/java/com/palantir/dialogue/core/RoundRobinChannel.java
ellisjoe: done

dialogue-core/src/test/java/com/palantir/dialogue/core/RoundRobinChannelTest.java
ellisjoe: good call. done

dialogue-core/src/test/java/com/palantir/dialogue/core/ConcurrencyLimitedChannelTest.java
ellisjoe: done

PR #82, opened 2019-05-07
Fix develop

ellisjoe:


uschi2000:
https://eng.uber.com/research/keeping-master-green-at-scale/

uschi2000:
but the root cause is the same, right? (CI against an outdated develop branch)

ellisjoe:
well yeah, it's just that in our case there was no cost to the failed build, in their case devs can't do work

ellisjoe:
lol I don't think we have the same problem here: `concurrently committing code changes by thousands of engineers to a big repository, with millions of lines of code`

PR #90, opened 2019-05-14
[improvement] Remove executor from HttpClient

ellisjoe:
## Before this PR
We'd pass an `ExecutorService` into the `HttpClient` which would in turn execute the request on its own underlying `ExecutorService`, which means we'd end up with two threads per request. 

## After this PR
Just rely on the underlying client's thread pool.


markelliot (commented):
Does this end up running everything on the forkJoin pool?

ellisjoe:
The request itself gets run on a cached thread pool that's created per client: https://github.com/AdoptOpenJDK/openjdk-jdk11/blob/master/src/java.net.http/share/classes/jdk/internal/net/http/HttpClientImpl.java#L275

Think the only thing the common fork join pool is used for is for executing callbacks on the returned future, but that should be avoided by passing your own when transforming the future yourself.

uschi2000:
so on a typical server we'd then want to share across all conjure clients a single HttpClient instance and thus a single dialogue Channel, ja?

ellisjoe:
that seems like a reasonable place to start

dialogue-java-client/src/main/java/com/palantir/dialogue/CompletableToListenableFuture.java
```diff
@@ -0,0 +1,66 @@
...
+package com.palantir.dialogue;
+
+import com.google.common.util.concurrent.ListenableFuture;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Executor;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+class CompletableToListenableFuture<T> implements ListenableFuture<T> {
```
uschi2000, line 26: looks a bit different from https://github.com/spotify/futures-extra/blob/master/src/main/java/com/spotify/futures/CompletableToListenableFutureWrapper.java

uschi2000:
Cool. This weird contraption was added to work around the HttpClient thread handling, in particular the inability to execute everything on a direct executor. this means that the callObservesSuccessfulResponses test could be made synchronous, but it seems that the test refactor in 6bc5bba made that problem go away. much better after this PR, thanks.

uschi2000:
looks good modulo questions on how exactly to implement the adapter

dialogue-java-client/src/main/java/com/palantir/dialogue/CompletableToListenableFuture.java
ellisjoe: So it looks like the main difference is in the exception handling they need to do. We definitely don't need to deal with the `CancellationException` because we delegate cancellation to the underlying future, whereas they have a completely separate ListenableFuture that tries to mirror the original one.

It looks like the `CompletionException` was handled in [here](https://github.com/spotify/futures-extra/commit/be9d8e4bd1c58efb2d8d137b737c28ebced4366b), so I added some tests, but can't seem to find any issues with the current implementation. Lemme know if you see something I'm missing here.

uschi2000: let's try it

uschi2000 (approved):

PR #92, opened 2019-05-15
Add a QueuedChannel

ellisjoe:


dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
```diff
@@ -0,0 +1,119 @@
...
+
+        @Override
+        public void onFailure(Throwable throwable) {
+            response.setException(throwable);
+            schedule();
+        }
+    }
+
+    @Value.Immutable
+    interface CallComponents {
```
markelliot: `DeferredCall`?

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
```diff
@@ -0,0 +1,119 @@
...
+     */
+    @Override
+    public ListenableFuture<Response> execute(Endpoint endpoint, Request request) {
+        CallComponents components = ImmutableCallComponents.of(endpoint, request, SettableFuture.create());
+
+        if (!queuedCalls.offer(components)) {
+            throw QosException.unavailable();
+        }
+
+        schedule();
```
markelliot, line 94: it's a little weird to me that we 'steal' this thread to possibly cause the scheduling/processing of other in-flight calls; maybe we should just eat the cost of having a thread around to do the processing of the queue and have it poll/wait for work to show up?

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
```diff
@@ -0,0 +1,123 @@
...
+     * tasks may be able to be scheduled, and false otherwise.
+     */
+    private boolean scheduleNextTask() {
+        CallComponents components = queuedCalls.poll();
+        if (components == null) {
+            return false;
+        }
+
+        Optional<ListenableFuture<Response>> response =
+                delegate.maybeExecute(components.endpoint(), components.request());
```
markelliot, line 120: if this throws (I guess it can throw?) and the response didn't complete, I think you lose the item from the queue.

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
ellisjoe: yeah I'm +1 to keeping a thread around for scheduling. I've been trying to figure out the right way to signal to that thread that it should run. Don't think I want to be in the business fo `wait`/`notify`, but that's basically the behavior you want (scheduling thread `wait`s until a request is enqueued or completed, and then should be `notified`).

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
ellisjoe: hmm, yeah I'd like to say that the LimitedChannel interface can't throw, given it should either not execute the request or forward the exception to the Future. I'm not sure what you should even do in the case when it throws? Maybe you catch and fail the Future exceptionally?

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
ellisjoe: done

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
```diff
@@ -0,0 +1,123 @@
...
+    private static final Executor DIRECT = MoreExecutors.directExecutor();
+    private final Deque<DeferredCall> queuedCalls = new ConcurrentLinkedDeque<>();
+    private final LimitedChannel delegate;
+
+    QueuedChannel(LimitedChannel delegate) {
+        this.delegate = delegate;
+    }
+
+    /**
+     * Enqueue the call and try to schedule as many queued tasks as possible.
```
uschi2000: nit: Enqueues and tries

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
```diff
@@ -0,0 +1,123 @@
...
+     */
+    @Override
+    public ListenableFuture<Response> execute(Endpoint endpoint, Request request) {
+        DeferredCall components = ImmutableDeferredCall.of(endpoint, request, SettableFuture.create());
+
+        if (!queuedCalls.offer(components)) {
+            throw QosException.unavailable();
```
uschi2000: the Channel javadoc says you must not throw here

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
uschi2000: a BlockingQueue should implement all of this already, no? i.e., just kick off a thread that does:
```
while(true) {
  call = queue.take(); // blocks
  handle(call);
  // need to handle interruption and other exceptions gracefully
}
```

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
```diff
@@ -0,0 +1,123 @@
...
+
+        Optional<ListenableFuture<Response>> response =
+                delegate.maybeExecute(components.endpoint(), components.request());
+
+        if (response.isPresent()) {
+            Futures.addCallback(response.get(), new ForwardAndSchedule(components.response()), DIRECT);
+            return true;
+        } else {
+            queuedCalls.addFirst(components);
+            return false;
```
uschi2000, line 127: this doesn't work as written. if my call doesn't execute the first time around, then it ends up in the queue and will only run the next time i submit a call on this channel, right?

uschi2000: ah, you have the schedule thing below, i see. I think this whole contraption is a lot easier with a dedicated scheduling thread and a blocking queue as described above.

uschi2000 (changes requested):

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
ellisjoe: the issues is that `handle(call)` may not be able to execute (the LimitedChannel can return `Optional.empty()`)

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
ellisjoe: changed to return an immediate failed future with the same exception. probably worth thinking about how this case should be handled (eg: other tasks that have already been submitted should be cancelled)

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
ellisjoe: issue mentioned above. the scheduling thread can't always do the next bit of work, so you either have to
1. schedule it to run periodically
2. spin and keep trying to execute the next request
3. signaling to the scheduling thread that it needs to run (eg: when a request completes)
4. have the underlying channel tell the scheduling thread to retryAfter some period of time

uschi2000: can we talk over VTC, I don't think I understand the flow you have in mind here.

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
ellisjoe: done

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
```diff
@@ -0,0 +1,157 @@
...
+ * A {@link Channel} that queues requests while the underlying {@link LimitedChannel} is unable to accept any new
+ * requests. This is done by enqueueing requests on submission, and then running the schedule loop in one of 3 ways:
+ * <ol>
+ *     <li>On submission - allows execution when there is available capacity</li>
+ *     <li>On request completion - allows execution when capacity has now become available</li>
+ *     <li>Periodically (eg: every 100ms) - allows execution when there may have been no capaciy and no in-flight
+ *     requests</li>
+ * </ol>
+ *
+ * This implementation was chose over alternatives for the following reasons:
```
uschi2000: chosen

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
```diff
@@ -0,0 +1,157 @@
...
+    private final Deque<DeferredCall> queuedCalls = new ConcurrentLinkedDeque<>();
+    private final LimitedChannel delegate;
+    private final ScheduledExecutorService backgroundScheduler =
+            Executors.newSingleThreadScheduledExecutor(new ThreadFactoryBuilder()
+                    .setNameFormat("dialogue-request-scheduler")
+                    .setUncaughtExceptionHandler((t, e) ->
+                            log.error("Uncaught exception while scheduling request. This is a programming error.", e))
+                    .setDaemon(false)
+                    .build());
+
```
uschi2000, line 70: nit: extra \n

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
```diff
@@ -0,0 +1,157 @@
...
+final class QueuedChannel implements Channel {
+
+    private static final Logger log = LoggerFactory.getLogger(QueuedChannel.class);
+    private static final Executor DIRECT = MoreExecutors.directExecutor();
+    private final Deque<DeferredCall> queuedCalls = new ConcurrentLinkedDeque<>();
+    private final LimitedChannel delegate;
+    private final ScheduledExecutorService backgroundScheduler =
+            Executors.newSingleThreadScheduledExecutor(new ThreadFactoryBuilder()
+                    .setNameFormat("dialogue-request-scheduler")
+                    .setUncaughtExceptionHandler((t, e) ->
```
uschi2000: what happens when schedule() throws? this thing will logged and then no subsequent schedule() will get scheduled on this executor, right? is that what you want?

uschi2000 (approved):
few clarifying questions, but feel free to merge whenever you're ready.

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
carterkozak: This implementation will work fine with the concurrency limited channel because completion of one request always allows the next request to be processed. That doesn't quite align with the LimitedChannel API, which provides no such guarantee.
Using a rate-limited channel will result in 100ms hiccups where no requests are sent because response completion is not directly tied to the rate limiter.

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
ellisjoe: ah good call. moved this into a try catch around the `schedule` call. removed this as well since it's now redundant.

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
ellisjoe: done

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
ellisjoe: done

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
ellisjoe: Yeah, you'd end up with bursts of request sent every 100ms rather than a smooth request rate. I'm going to merge this for now and can revisit going forward.

carterkozak: 👍

PR #96, opened 2019-05-20
[improvement] Add static constructor for JavaHttp client

ellisjoe:
## Before this PR
No standard way to construct a client.

## After this PR
Can now construct a properly configured `Channel` backed by the java `HttpClient` given a `ClientConfiguration`. This should act as the entry point when migrating existing clients to Dialogue.

## Other considerations
We should consider how many of the existing conjure features we want to continue to support (there are TODOs for the ones not currently supported). eg: cipher suites may not need to be configured anymore and backoff slot size may not make sense, given we're relying mostly on Concurrency limiting to cause client size backpressure.

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/Channels.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/Channels.java
@@ -1,0 +1,36 @@
+/*
+ * (c) Copyright 2019 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.dialogue.core;
+
+import static java.util.stream.Collectors.toList;
+
+import com.palantir.dialogue.Channel;
+import java.util.Collection;
+import java.util.List;
+
+public final class Channels {
+
+    private Channels() {}
+
+    public static Channel create(Collection<? extends Channel> channels) {
+        List<LimitedChannel> limitedChannels = channels.stream()
+                .map(ConcurrencyLimitedChannel::create)
+                .collect(toList());
+
+        return new RetryingChannel(new QueuedChannel(new RoundRobinChannel(limitedChannels)));
+    }
+}
--- a/dialogue-java-client/src/main/java/com/palantir/dialogue/JavaChannels.java
+++ b/dialogue-java-client/src/main/java/com/palantir/dialogue/JavaChannels.java
@@ -1,0 +1,77 @@
+/*
+ * (c) Copyright 2019 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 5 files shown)


dialogue-core/src/main/java/com/palantir/dialogue/core/Channels.java
```diff
@@ -0,0 +1,36 @@
...
+public final class Channels {
+
+    private Channels() {}
+
+    public static Channel create(Collection<? extends Channel> channels) {
+        List<LimitedChannel> limitedChannels = channels.stream()
+                .map(ConcurrencyLimitedChannel::create)
+                .collect(toList());
+
+        return new RetryingChannel(new QueuedChannel(new RoundRobinChannel(limitedChannels)));
```
markelliot, line 34: it'd kind of be nice if this were something that was flattenable

dialogue-java-client/src/main/java/com/palantir/dialogue/JavaChannels.java
```diff
@@ -0,0 +1,77 @@
...
+
+    private JavaChannels() {}
+
+    public static Channel create(ClientConfiguration conf) {
+        // TODO(jellis): read/write timeouts
+        // TODO(jellis): record metrics
+        // TODO(jellis): gcm cipher toggle
+        // TODO(jellis): proxy creds + mesh proxy
+        // TODO(jellis): configure node selection strategy
+        // TODO(jellis): failed url cooldown
```
markelliot, line 42: ideally this would go away

dialogue-core/src/main/java/com/palantir/dialogue/core/Channels.java
```diff
@@ -0,0 +1,36 @@
...
+
+import com.palantir.dialogue.Channel;
+import java.util.Collection;
+import java.util.List;
+
+public final class Channels {
+
+    private Channels() {}
+
+    public static Channel create(Collection<? extends Channel> channels) {
```
uschi2000, line 29: let's make this package-private

dialogue-java-client/src/main/java/com/palantir/dialogue/JavaChannels.java
```diff
@@ -0,0 +1,77 @@
...
+public final class JavaChannels {
+
+    private JavaChannels() {}
+
+    public static Channel create(ClientConfiguration conf) {
+        // TODO(jellis): read/write timeouts
+        // TODO(jellis): record metrics
+        // TODO(jellis): gcm cipher toggle
+        // TODO(jellis): proxy creds + mesh proxy
+        // TODO(jellis): configure node selection strategy
```
markelliot, line 41: ideally this would go away

dialogue-java-client/src/main/java/com/palantir/dialogue/JavaChannels.java
```diff
@@ -0,0 +1,77 @@
...
+
+    private static URL url(String uri) {
+        try {
+            return new URL(uri);
+        } catch (MalformedURLException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    private static SSLContext createSslContext(TrustManager trustManager) {
```
uschi2000, line 68: how is this different from SslSocketFactories#createSslContext?

uschi2000: ah, nevermind, just saw you have a clientconfig already

uschi2000 (approved):

dialogue-core/src/main/java/com/palantir/dialogue/core/Channels.java
ellisjoe: Yeah, going to revisit a builder pattern or something of that sort. The issue right now is that none of these things have the same constructor signatures:

RetryingChannel takes a Channel
QueuedChannel takes a LimitedChannel
RoundRobinChannel takes a List<LimitedChannel>

dialogue-core/src/main/java/com/palantir/dialogue/core/Channels.java
ellisjoe: I can't do that right now, because the constructors that people should really use are in different projects. eg: if you want to create an HttpClient backed Channel then you need to use the constructor in the `dialogue-java-client` project, which in turn will call this. This is only the case right now because we're supporting both the Java and OkHttp clients.

PR #97, opened 2019-05-20
[improvement] Add BlacklistingChannel

ellisjoe:
## Before this PR
Errors seen in a channel would not affect future decisions to use that channel.

## After this PR
After an error, we will wait for some cooldown period before using that Channel again.

## Other Considerations
We don't currently have any mechanism for forcing requests to execute when none of the underlying `LimitedChannels` have capacity. That seems desirable, though we may want some escape hatch in the BlacklistingChannel to reset them when all Channels have been blacklisted.


dialogue-core/src/main/java/com/palantir/dialogue/core/MoreFutures.java
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+package com.palantir.dialogue.core;
+
+import com.google.common.util.concurrent.FutureCallback;
+import com.google.common.util.concurrent.Futures;
+import com.google.common.util.concurrent.ListenableFuture;
+import com.google.common.util.concurrent.MoreExecutors;
+
+final class MoreFutures {
```
markelliot: maybe we should call this `DialogueFutures`?

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -56,7 +54,7 @@ public RetryingChannel(Channel delegate) {
 
         Supplier<ListenableFuture<Response>> callSupplier = () -> delegate.execute(endpoint, request);
         FutureCallback<Response> retryer = new RetryingCallback<>(callSupplier, future);
-        Futures.addCallback(callSupplier.get(), retryer, DIRECT_EXECUTOR);
+        Futures.addCallback(callSupplier.get(), retryer, MoreExecutors.directExecutor());
```
markelliot: did you intend to also refactor these to call MoreFutures.addDirectCallback?

dialogue-core/src/main/java/com/palantir/dialogue/core/BlacklistingChannel.java
```diff
@@ -0,0 +1,77 @@
...
+                    .map(future -> MoreFutures.addDirectCallback(future, new BlacklistingCallback<>()));
+        }
+    }
+
+    private class BlacklistingCallback<T> implements FutureCallback<T> {
+        @Override
+        public void onSuccess(T result) {}
+
+        @Override
+        public void onFailure(Throwable throwable) {
```
markelliot, line 77: does onFailure actually cover all the scenarios we care about?

uschi2000: ... this may lead back to the discussion about the `Call<T>` interface that gives you more fine-grained control over the type of error you're seeing

dialogue-core/src/main/java/com/palantir/dialogue/core/MoreFutures.java
ellisjoe: 👍 done

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
ellisjoe: I refactored and then got all the warnings about ignoring a returned Future so flipped em back. The main benefit of the MoreFutures method is the fact that it returns the future which isn't necessary here. switched these back to have a constant for the executor

dialogue-core/src/main/java/com/palantir/dialogue/core/BlacklistingChannel.java
ellisjoe: If there are certain responses we'd like to blacklist for (eg: on 503s) we can do that in the onSuccess handler. @uschi2000 my claim is that we can do all the dispatching in our channels based on response code (and maybe will need headers) but don't need anything conjure specific. Basically, I'd like to keep our http channel stack completely conjure agnostic, given nothing about it should be conjure specific

uschi2000: that's note quite true. for instance, this library might do something with 308, but not with 307, simply because conjure only cares about the former.

ellisjoe: I guess I'm fine saying that this library implements a subset of the http spec that is sufficient for the conjure use case. I mainly don't think it's worth taking a dependency on any sort of serialization logic in this library.

markelliot (approved):

PR #100, opened 2019-05-23
[improvement] Make all impl classes package private

ellisjoe:


carterkozak (approved):
I'd like to split dialogue-api and dialogue-runtime into separate modules as well so that it's easier to track what our generated code can compile against vs what we can fix with a library bump (without recompilation). I'm happy with how that has worked in conjure-undertow.
That's a task for a separate PR though.

ellisjoe:
I think `dialogue-target` is the set of things the generated code can depend on. On board with that plan, so we should figure out if that's not quite true.

uschi2000:
dialogue-target is (or at least was) intended to be the code gen api (-target).

uschi2000 (approved):

PR #101, opened 2019-05-23
[improvement] Fix rate limited response handling

ellisjoe:
## Before this PR
Handle the inability to enqueue a request of completing the response exceptionally.

## After this PR
Return a 429 in the standard way. Exceptions should only be thrown for non standard response issues.


markelliot:
tests?

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
```diff
@@ -152,6 +152,25 @@ public void onFailure(Throwable throwable) {
...
+        INSTANCE;
+
+        @Override
+        public InputStream body() {
+            return InputStream.nullInputStream();
+        }
+
+        @Override
+        public int code() {
+            return 429;
```
uschi2000, line 172: this is different from unavailable

ellisjoe: yeah I think this is representative of what you want. you're seeing this because you've reached your concurrency limit and you'd expect that downstream service to response with a 429 if you were to execute a request. You also wouldn't expect going to another server to help much here because it's the underlying service that's overloaded.

markelliot (approved):

PR #104, opened 2019-05-24
[improvement] Include headers in response object

ellisjoe:
## After this PR
Headers are now accessible in the `Response` object.

### DIFF
--- a/dialogue-java-client/src/main/java/com/palantir/dialogue/HttpChannel.java
+++ b/dialogue-java-client/src/main/java/com/palantir/dialogue/HttpChannel.java
@@ -28,6 +28,7 @@ import java.net.URL;
 import java.net.http.HttpClient;
 import java.net.http.HttpRequest;
 import java.net.http.HttpResponse;
+import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
@@ -116,6 +117,11 @@ public final class HttpChannel implements Channel {
             @Override
             public int code() {
                 return response.statusCode();
+            }
+
+            @Override
+            public Map<String, List<String>> headers() {
+                return response.headers().map();
             }
 
             @Override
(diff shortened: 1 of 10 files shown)


dialogue-target/src/main/java/com/palantir/dialogue/Response.java
```diff
@@ -26,6 +28,9 @@
     /** The HTTP response code for this response. */
     int code();
 
+    /** The HTTP headers for this response. */
+    Map<String, List<String>> headers();
```
carterkozak, line 34: Does it make sense to provide both this and `Optional<String> contentType();`?

dialogue-java-client/src/main/java/com/palantir/dialogue/HttpChannel.java
```diff
@@ -118,6 +119,11 @@ public int code() {
                 return response.statusCode();
             }
 
+            @Override
+            public Map<String, List<String>> headers() {
+                return response.headers().map();
```
markelliot, line 123: idk ... is it worth a defensive copy here?

dialogue-target/src/main/java/com/palantir/dialogue/Response.java
```diff
@@ -26,6 +28,9 @@
     /** The HTTP response code for this response. */
     int code();
 
+    /** The HTTP headers for this response. */
```
carterkozak: Can we specify that this map uses a case-insensitive comparison?
https://tools.ietf.org/html/rfc7540#section-8.1.2

dialogue-target/src/main/java/com/palantir/dialogue/Response.java
ellisjoe: I replaced it with `Optional<String> getFirstHeader(String header)` which seems like the nice behavior you'd want

dialogue-java-client/src/main/java/com/palantir/dialogue/HttpChannel.java
ellisjoe: Can verify that this is an unmodifiable map: https://github.com/AdoptOpenJDK/openjdk-jdk11/blob/master/src/java.net.http/share/classes/java/net/http/HttpHeaders.java#L117

dialogue-target/src/main/java/com/palantir/dialogue/Response.java
ellisjoe: done

markelliot (approved):

PR #105, opened 2019-05-27
Inject user agent headers

uschi2000:


dialogue-core/src/main/java/com/palantir/dialogue/core/UserAgentChannel.java
```diff
@@ -0,0 +1,62 @@
...
+
+    private static UserAgent augmentUserAgent(UserAgent baseAgent, Endpoint endpoint) {
+        String maybeDialogueVersion = Channel.class.getPackage().getImplementationVersion();
+        UserAgent.Agent dialogueAgent = UserAgent.Agent.of(
+                "dialogue",
+                maybeDialogueVersion != null ? maybeDialogueVersion : "0.0.0");
+
+        return baseAgent
+                .addAgent(UserAgent.Agent.of(endpoint.serviceName(), endpoint.version()))
+                .addAgent(UserAgent.Agent.of(endpoint.endpointName(), endpoint.version()))
```
ellisjoe: let's take this out for now. in particular because it's nice to aggregate by user agent and see who are the biggest callers of a particular service. you're also able to sub-aggregate by path (eg: endpoint) at that point anyway.

dialogue-core/src/main/java/com/palantir/dialogue/core/UserAgentChannel.java
```diff
@@ -0,0 +1,62 @@
...
+        Request newRequest = Request.builder()
+                .from(request)
+                .putHeaderParams("user-agent", UserAgents.format(augmentUserAgent(baseAgent, endpoint)))
+                .build();
+        return delegate.execute(endpoint, newRequest);
+    }
+
+    private static UserAgent augmentUserAgent(UserAgent baseAgent, Endpoint endpoint) {
+        String maybeDialogueVersion = Channel.class.getPackage().getImplementationVersion();
+        UserAgent.Agent dialogueAgent = UserAgent.Agent.of(
```
ellisjoe: can make this construction all static

dialogue-core/src/main/java/com/palantir/dialogue/core/UserAgentChannel.java
uschi2000: done

dialogue-core/src/main/java/com/palantir/dialogue/core/UserAgentChannel.java
uschi2000: done

dialogue-core/src/main/java/com/palantir/dialogue/core/UserAgentChannel.java
```diff
@@ -31,6 +31,15 @@
  */
 final class UserAgentChannel implements Channel {
 
+    private static final UserAgent.Agent DIALOGUE_AGENT;
+
+    static {
```
ellisjoe: slight preference to stick this into a static method and then initialize `DIALOGUE_AGENT` by calling the method

uschi2000: done

ellisjoe (approved):

PR #106, opened 2019-05-27
Add InstrumentedChannel

uschi2000:


dialogue-core/src/main/java/com/palantir/dialogue/core/Channels.java
```diff
@@ -20,20 +20,26 @@
...
+            Collection<? extends Channel> channels,
+            UserAgent userAgent,
+            TaggedMetricRegistry metrics) {
         List<LimitedChannel> limitedChannels = channels.stream()
                 .map(ConcurrencyLimitedChannel::create)
                 .collect(toList());
 
         return new UserAgentChannel(
-                new RetryingChannel(new QueuedChannel(new RoundRobinChannel(limitedChannels))),
+                new InstrumentedChannel(
```
ellisjoe: think you want to stick this around the bottom most channel so we record just time observed in the remote server. then we can other metrics for things like time spent in the queue

dialogue-core/build.gradle
```diff
@@ -9,6 +9,7 @@ dependencies {
     compile 'com.netflix.concurrency-limits:concurrency-limits-core'
     compile 'com.palantir.conjure.java.runtime:client-config'
     compile 'com.palantir.safe-logging:preconditions'
+    compile 'io.dropwizard.metrics:metrics-core'
```
ellisjoe, line 12: pulling in tritium isn't sufficient?

dialogue-core/src/main/java/com/palantir/dialogue/core/Channels.java
markelliot: we'd probably benefit from a method which flattens the channel initialization

uschi2000: can you elaborate, @markelliot ?

dialogue-core/build.gradle
uschi2000: fixed, thanks

dialogue-core/src/main/java/com/palantir/dialogue/core/Channels.java
uschi2000: specifically, most of the channels have different arguments, so it's not clear to me how to canonicalize anything here. let me know.

dialogue-core/build.gradle
schlosna: ahh, I'm assuming this is what triggered https://github.com/palantir/tritium/issues/274

ellisjoe: @schlosna yep!

PR #107, opened 2019-05-28
Add support for gzip-compressed responses

uschi2000:


dialogue-java-client/src/main/java/com/palantir/dialogue/HttpChannel.java
```diff
@@ -110,7 +112,16 @@ private Response toResponse(HttpResponse<InputStream> response) {
         return new Response() {
             @Override
             public InputStream body() {
-                return response.body();
+                boolean isGzipped = response.headers()
+                        .firstValue("content-encoding").orElse("")
+                        .toLowerCase()
+                        .equals("gzip");
+                if (isGzipped) {
+                    // TODO(rfink): Think about removing the okio dependency at some point?
```
markelliot, line 119: why do we need Okio here?

markelliot: maybe we should use the Java built-in https://docs.oracle.com/javase/8/docs/api/java/util/zip/GZIPInputStream.html?
```java
return new GZIPInputStream(response.body());
```

dialogue-java-client/src/main/java/com/palantir/dialogue/HttpChannel.java
```diff
@@ -110,7 +112,16 @@ private Response toResponse(HttpResponse<InputStream> response) {
         return new Response() {
             @Override
             public InputStream body() {
-                return response.body();
+                boolean isGzipped = response.headers()
+                        .firstValue("content-encoding").orElse("")
+                        .toLowerCase()
```
ellisjoe: `equalsIgnoringCase(...)`

dialogue-java-client/src/main/java/com/palantir/dialogue/HttpChannel.java
uschi2000: That thing unfortunately reads the GZIP header in the constructor and thus blocks the otherwise async wiring.

dialogue-java-client/src/main/java/com/palantir/dialogue/HttpChannel.java
uschi2000: done

ellisjoe (approved):

PR #115, opened 2019-05-30
Exception handling

ellisjoe:


dialogue-example/src/main/java/com/palantir/dialogue/example/SampleServiceClient.java
```diff
@@ -164,11 +158,8 @@ public void voidToVoid() {
                         call,
                         r -> voidToVoidDeserializer.deserialize(r),
                         MoreExecutors.directExecutor());
-                try {
-                    deserializedResponse.get(callTimeout.toMillis(), TimeUnit.MILLISECONDS);
-                } catch (Throwable t) {
-                    throw Exceptions.unwrapExecutionException(t);
-                }
+
+                Exceptions.getUnchecked(deserializedResponse);
```
uschi2000: call it response, just like above

dialogue-target/src/main/java/com/palantir/dialogue/Exceptions.java
```diff
@@ -16,30 +16,47 @@
...
-                        : new RuntimeException(cause);
+    public static <T> T getUnchecked(Future<T> future) {
+        try {
+            return future.get();
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            throw new RuntimeException(e);
+        } catch (ExecutionException e) {
+            if (e.getCause() instanceof RemoteException) {
+                RemoteException remoteException = (RemoteException) e.getCause();
```
uschi2000: can you leave a comment explaining what this magic thing is here?

dialogue-target/src/main/java/com/palantir/dialogue/Exceptions.java
```diff
@@ -16,30 +16,47 @@
...
+     * {@link RemoteException}s directly, rather than wrapping them in an {@link UncheckedExecutionException}.
      */
-    public static RuntimeException unwrapExecutionException(Throwable throwable) {
-        if (throwable instanceof ExecutionException) {
-            if (throwable.getCause() != null) {
-                Throwable cause = throwable.getCause();
-                throw throwable.getCause() instanceof RuntimeException
-                        ? (RuntimeException) cause
-                        : new RuntimeException(cause);
+    public static <T> T getUnchecked(Future<T> future) {
```
uschi2000: it's a bit funny to call this Exceptions.getUnchecked...

dialogue-target/src/test/java/com/palantir/dialogue/ExceptionsTest.java
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    @Test
+    public void testRemoteException() {
+        RemoteException remoteException = remoteException(new ServiceException(ErrorType.INVALID_ARGUMENT));
+        ListenableFuture<Object> failedFuture = Futures.immediateFailedFuture(remoteException);
+
+        assertThatThrownBy(() -> Exceptions.getUnchecked(failedFuture))
+                .isInstanceOf(RemoteException.class)
+                .hasFieldOrPropertyWithValue("status", ErrorType.INVALID_ARGUMENT.httpErrorCode());
```
uschi2000: can you also something about the stacktrace? (not sure if possible)

dialogue-example/src/main/java/com/palantir/dialogue/example/SampleServiceClient.java
ellisjoe: done

dialogue-target/src/main/java/com/palantir/dialogue/Exceptions.java
ellisjoe: done

dialogue-target/src/main/java/com/palantir/dialogue/Exceptions.java
ellisjoe: renamed to `RemoteExceptions`

dialogue-target/src/test/java/com/palantir/dialogue/ExceptionsTest.java
ellisjoe: not a particularly nice way to assert this because the exception is created here, and the get is done here, which means both exceptions have essentially the same stacktrace.

PR #123, opened 2019-06-24
[WIP] Add client side tracing

ellisjoe:

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
@@ -28,6 +28,7 @@ import com.palantir.dialogue.Channel;
 import com.palantir.dialogue.Endpoint;
 import com.palantir.dialogue.Request;
 import com.palantir.dialogue.Response;
+import com.palantir.tracing.DeferredTracer;
 import java.io.InputStream;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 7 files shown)


dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
```diff
@@ -132,6 +133,9 @@ private boolean scheduleNextTask() {
                 delegate.maybeExecute(components.endpoint(), components.request());
 
         if (response.isPresent()) {
+            // Complete the queued span
+            components.tracer().withTrace(() -> null);
```
ellisjoe, line 137: this isn't great

dialogue-java-client/src/main/java/com/palantir/dialogue/JavaChannels.java
```diff
@@ -75,4 +81,13 @@ private static SSLContext createSslContext(TrustManager trustManager) {
             throw new IllegalStateException(e);
         }
     }
+
+    // Same default thread pool with tracing enabled
+    private static ExecutorService createExecutor() {
```
ellisjoe, line 86: implementing this here, rather than higher up means we need to also implement it in other Channel impls (eg: OkHttp).

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
ellisjoe: there's also currently no way to properly create a span with the request path included in the span name. it's known here, but we have no way to pass that to the span created when the call is actually executed.

ellisjoe:
putting this here to illustrate some of the current issues with tracing async calls. related to: https://github.com/palantir/tracing-java/pull/124

PR #151, opened 2019-09-11
Add tracing around queue and execution time

ellisjoe:
## Before this PR
There was no tracing of requests

## After this PR
We trace how long the request was enqueued

### DIFF
--- a/dialogue-core/build.gradle
+++ b/dialogue-core/build.gradle
@@ -9,6 +9,7 @@ dependencies {
     compile 'com.netflix.concurrency-limits:concurrency-limits-core'
     compile 'com.palantir.conjure.java.runtime:client-config'
     compile 'com.palantir.safe-logging:preconditions'
+    compile 'com.palantir.tracing:tracing'
     compile 'io.dropwizard.metrics:metrics-core'
 
     testCompile 'com.palantir.safe-logging:preconditions-assertj'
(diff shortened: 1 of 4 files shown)


dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
```diff
@@ -116,7 +118,8 @@
      */
     @Override
     public ListenableFuture<Response> execute(Endpoint endpoint, Request request) {
-        DeferredCall components = ImmutableDeferredCall.of(endpoint, request, SettableFuture.create());
+        DetachedSpan span = DetachedSpan.start("queued");
+        DeferredCall components = DeferredCall.of(endpoint, request, span, SettableFuture.create());
 
         if (!queuedCalls.offer(components)) {
             return Futures.immediateFuture(RateLimitedResponse.INSTANCE);
```
carterkozak, line 125: Should we complete the DetachedSpan here? Or are we happy with not recording it at all in this case?

carterkozak: Suppose queued doesn't make sense unless we actually do queue something. Disregard!

dialogue-core/build.gradle
```diff
@@ -9,6 +9,7 @@ dependencies {
     compile 'com.netflix.concurrency-limits:concurrency-limits-core'
     compile 'com.palantir.conjure.java.runtime:client-config'
     compile 'com.palantir.safe-logging:preconditions'
+    compile 'com.palantir.tracing:tracing'
```
carterkozak, line 12: Can we use the `java-library` plugin and use an `implementation` dependency here?

carterkozak (approved):
lgtm, dependency type changes probably make sense in a separate refactor.

PR #224, opened 2020-01-06
[improvement] Add TracedChannel

markelliot:
## Before this PR
Dialogue does not trace any requests.

## After this PR
Users of Dialogue may use a TracedChannel in the channel hierarchy in order to trace specific channel executions with a named operation.


dialogue-core/src/main/java/com/palantir/dialogue/core/TracedChannel.java
```diff
@@ -0,0 +1,39 @@
...
+import com.palantir.dialogue.Endpoint;
+import com.palantir.dialogue.Request;
+import com.palantir.dialogue.Response;
+import com.palantir.tracing.Tracers;
+
+public final class TracedChannel implements Channel {
+    private final Channel delegate;
+    private final String operationName;
+
+    TracedChannel(Channel delegate, String operationName) {
```
carterkozak, line 30: I think this either needs a public factory, or the class should be package private

markelliot: made package private to match other *Channel classes.

carterkozak:
lgtm 👍

markelliot:
unsure of the procedure for cutting a release here

markelliot:
also looks like your thumbs up wasn't enough here, @carterkozak

ferozco:
👍

carterkozak:
I don't appear to have perms, sorry

ferozco:
Fixed it so that @carterkozak should be able to approve moving forward

PR #230, opened 2020-01-07
[improvement] Add tracing headers to outbound requests

markelliot:
## Before this PR
Dialogue requests do not include Zipkin compatible tracing headers.

## After this PR
Dialogue requests include Zipkin compatible tracing headers.


ferozco:
👍

PR #262, opened 2020-01-24
Provide client-side warnings on deprecated endpoint invocations

markelliot:
## Before this PR
Clients invoking deprecated endpoints have no local signaling about those deprecations

## After this PR
Clients invoking deprecated endpoints will log a warning indicating which service class and, when available, the details of the server indicating access to a deprecated endpoint.

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/DeprecationWarningChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/DeprecationWarningChannel.java
@@ -1,0 +1,91 @@
+/*
+ * (c) Copyright 2020 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.dialogue.core;
+
+import com.github.benmanes.caffeine.cache.Cache;
+import com.github.benmanes.caffeine.cache.Caffeine;
+import com.google.common.util.concurrent.ListenableFuture;
+import com.palantir.dialogue.Channel;
+import com.palantir.dialogue.Endpoint;
+import com.palantir.dialogue.Request;
+import com.palantir.dialogue.Response;
+import com.palantir.logsafe.SafeArg;
+import com.palantir.tritium.metrics.registry.MetricName;
+import com.palantir.tritium.metrics.registry.TaggedMetricRegistry;
+import java.time.Duration;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+final class DeprecationWarningChannel implements Channel {
+    private static final Logger log = LoggerFactory.getLogger(DeprecationWarningChannel.class);
+    private static final Object SENTINEL = new Object();
+
+    private final Channel delegate;
+    private final TaggedMetricRegistry registry;
+    private final Cache<String, Object> loggingRateLimiter = Caffeine.newBuilder()
+            .expireAfterWrite(Duration.ofMinutes(1))
+            .build();
+
+    DeprecationWarningChannel(Channel delegate, TaggedMetricRegistry registry) {
+        this.delegate = delegate;
+        this.registry = registry;
+    }
+
+    @Override
+    public ListenableFuture<Response> execute(Endpoint endpoint, Request request) {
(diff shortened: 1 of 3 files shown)


markelliot:
see also https://github.com/palantir/conjure-java-runtime/pull/1438

dialogue-core/src/main/java/com/palantir/dialogue/core/DeprecationWarningChannel.java
```diff
@@ -0,0 +1,91 @@
...
+import com.palantir.dialogue.Request;
+import com.palantir.dialogue.Response;
+import com.palantir.logsafe.SafeArg;
+import com.palantir.tritium.metrics.registry.MetricName;
+import com.palantir.tritium.metrics.registry.TaggedMetricRegistry;
+import java.time.Duration;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+final class DeprecationWarningChannel implements Channel {
```
uschi2000, line 41: can you add a one-liner class comment, please?

dialogue-core/src/main/java/com/palantir/dialogue/core/DeprecationWarningChannel.java
```diff
@@ -0,0 +1,91 @@
...
+    private static final Logger log = LoggerFactory.getLogger(DeprecationWarningChannel.class);
+    private static final Object SENTINEL = new Object();
+
+    private final Channel delegate;
+    private final TaggedMetricRegistry registry;
+    private final Cache<String, Object> loggingRateLimiter = Caffeine.newBuilder()
```
uschi2000, line 47: why not use the Guava rate limiter?

dialogue-core/src/main/java/com/palantir/dialogue/core/DeprecationWarningChannel.java
```diff
@@ -0,0 +1,91 @@
...
+                            if (tryAcquire(endpoint.serviceName())) {
+                                log.warn(
+                                        "Using a deprecated endpoint when connecting to service",
+                                        SafeArg.of("serviceName", endpoint.serviceName()),
+                                        SafeArg.of("endpointHttpMethod", endpoint.httpMethod()),
+                                        SafeArg.of("endpointName", endpoint.endpointName()),
+                                        SafeArg.of("endpointClientVersion", endpoint.version()),
+                                        SafeArg.of(
+                                                "service",
+                                                response.getFirstHeader("server").orElse("no server header provided")));
```
uschi2000, line 78: "server" or "service"?

dialogue-core/src/main/java/com/palantir/dialogue/core/DeprecationWarningChannel.java
```diff
@@ -0,0 +1,91 @@
...
+                            }
+                        });
+                    }
+                }));
+    }
+
+    /**
+     * Returns true when the loggingRateLimiter permits logging for the provided serviceName, and false otherwise.
+     *
+     * <p>Note: this method is not synchronized because the throttling is best effort rather than guaranteed -- the
```
uschi2000, line 88: thank you!

dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueFutures.java
```diff
@@ -28,4 +30,16 @@ private DialogueFutures() {}
         Futures.addCallback(future, callback, MoreExecutors.directExecutor());
         return future;
     }
+
+    static <T> FutureCallback<T> onSuccess(Consumer<T> onSuccess) {
```
uschi2000, line 34: would maybe call this `onSuccessOnly`?

uschi2000: nah, dunno. could go either way.

dialogue-core/src/main/java/com/palantir/dialogue/core/DeprecationWarningChannel.java
markelliot: I need/want a RateLimiter that's namespaced to a particular serviceName, Guava RateLimiters are limiting per instance. One option would be to have the cache be String -> RateLimiter, but Caffeine is perhaps sufficient. Maybe I don't need to be so cautious with memory use/object allocation?

dialogue-core/src/main/java/com/palantir/dialogue/core/DeprecationWarningChannel.java
markelliot: the header is "server", but for logging purposes I think we want to call this parameter "service".

dialogue-core/src/main/java/com/palantir/dialogue/core/DeprecationWarningChannel.java
markelliot: done

dialogue-core/src/main/java/com/palantir/dialogue/core/DeprecationWarningChannel.java
uschi2000: probably fine either way

uschi2000 (approved):

PR #267, opened 2020-01-27
Begin to adopt metric-schema

carterkozak:


dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
```diff
@@ -96,7 +88,7 @@
     QueuedChannel(
             LimitedChannel delegate,
             int maxQueueSize,
-            TaggedMetricRegistry metrics) {
+            TaggedMetricRegistry registry) {
```
ferozco: Can this take a `DispatcherMetrics` instead

dialogue-core/src/main/metrics/dialogue-core-metrics.yml
```diff
@@ -0,0 +1,14 @@
...
+  com.palantir.conjure.java.dispatcher:
+    shortName: Dispatcher
+    docs: Dialogue client dispatcher metrics
+    metrics:
+      calls.queued:
+        type: gauge
+        docs: Number of queued calls waiting to execute.
+      calls.running:
+        type: gauge
+        docs: Number of active outgoing requests.
```
ferozco: nit: newline

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
carterkozak: 👍

ferozco (approved):
Thanks carter

dialogue-core/src/test/java/com/palantir/dialogue/core/QueuedChannelTest.java
```diff
@@ -44,6 +44,11 @@
 @SuppressWarnings("FutureReturnValueIgnored")
 public class QueuedChannelTest {
 
+    private static final MetricName NUM_QUEUED_METRIC =
```
ellisjoe, line 47: would it be possible to have the generated code either make these `MetricName`s accessible or provide getters for the particular metrics? I think the latter is preferable

carterkozak: We provide getters for everything except gauges, as always gauges are special in all sorts of dangerous ways.

Metric names would require similar builders to those we provide to register(gauge)/access(everything else) metrics, so exposing those directly doesn’t feel like the right complexity trade off.
Accessing a gauge is difficult. We can’t provide generic type information, and outside of testing we don’t want users to load a gauge<?> from the registry.
Thoughts?

PR #288, opened 2020-02-05
[fix][break] Dialogue `Request` header names are case insensitive

carterkozak:
This is a break because immutables does not provide a clean way
to use a case insensitive sorted map. This breaks ABI, but
retains API in order to give us more options for similar changes
moving forward.

## Before this PR
Request headers were not compared in a case-insensitive manner, so Channel implementations that depend on specific header parameters risk bugs depending on inputs.

## After this PR
This conforms more closely to the http specification.


dialogue-target/src/main/java/com/palantir/dialogue/Request.java
```diff
@@ -16,37 +16,203 @@
...
+import java.util.Objects;
 import java.util.Optional;
-import org.immutables.value.Value;
 
 /** Defines the parameters of a single {@link Call} to an {@link Endpoint}. */
-@DialogueImmutablesStyle
-@Value.Immutable
-public interface Request {
-    /** The HTTP headers for this request, encoded as a map of {@code header-name: header-value}. */
-    Map<String, String> headerParams();
```
markelliot, line 29: perhaps a slightly smaller change and one with less hand-written code would be changing the build method of the builder class to make a TreeMap and then copy it back into the object, something like

```java
Request request = super.build();
Map<String, String> caseInsensitiveHeaderParams = new TreeMap<>(Case...);
caseInsensitiveHeaderParams.putAll(request.headerParams());
return request.toBuilder().headerParams(caseInsensitiveHeaderParams).build();
```

markelliot: or alternatively, we could replace the builder methods for adding headerParams to run toLowerCase on the strings at build time?

markelliot: one more alternative: we could leave this interface as-is and manually write the implementation class + Builder and return the implemented interface vs. making the entire thing concrete

carterkozak: The problem is consumers writing code along the lines of `request.headerParams().containsKey("Content-Type")` would fail when `"content-type"` is present. I'm worried that the asymmetry between Request and Response headers will make it easy to introduce subtle bugs that aren't easily caught in review.

carterkozak: >  we could leave this interface as-is and manually write the implementation class + Builder

That's true, but is it desirable in this case? Do we want to allow custom Request implementations? We could avoid the break, but I think now is the time to make breaks before we roll this out more broadly.

dialogue-target/src/test/java/com/palantir/dialogue/RequestTest.java
```diff
@@ -0,0 +1,43 @@
...
+import com.palantir.tokens.auth.AuthHeader;
+import com.palantir.tokens.auth.BearerToken;
+import org.junit.Test;
+
+public final class RequestTest {
+
+    @Test
+    public void testRequestHeaderInsensitivity() {
+        Request request = Request.builder().putHeaderParams("Foo", "bar").build();
+        assertThat(request.headerParams()).containsKey("foo");
```
uschi2000: please also test `.containsKey("FOO")`

carterkozak:
I'm curious why request query parameters use a guava `Multimap<String, String>` but response headers use `Map<String, List<String>>`. Should response headers use a Multimap (break)?

It's odd that request headers can have a single value, but response headers can have multiple values, it makes sense from the conjure perspective, but that may mean we have to push implementation into individual clients which we're actively trying to avoid.

uschi2000:
> I'm curious why request query parameters use a guava `Multimap<String, String>` but response headers use `Map<String, List<String>>`. Should response headers use a Multimap (break)?
> 
> It's odd that request headers can have a single value, but response headers can have multiple values, it makes sense from the conjure perspective, but that may mean we have to push implementation into individual clients which we're actively trying to avoid.

uschi2000:
Re map/multimap: probably my mistake for not catching in CR. I wrote the initial Request interface and Joe wrote the Response one, maybe just taking a short-cut by forwarding the okhttp API.

Re single multiple: I don't quite follow, is it not the case that both can have multiple values?

dialogue-target/src/main/java/com/palantir/dialogue/Request.java
```diff
@@ -16,37 +16,203 @@
...
+public final class Request {
```
iamdanfox, line 32: :+1: to hand-rolling - it seems much easier to control exactly which methods exist (and maintain back compat) when we write them ourselves vs have immutables magically generate them!

dialogue-target/src/main/java/com/palantir/dialogue/Request.java
```diff
@@ -16,37 +16,203 @@
...
+    }
+
+    public static Builder builder() {
         return new Builder();
     }
 
-    class Builder extends ImmutableRequest.Builder {}
+    public static final class Builder {
+        private ImmutableSortedMap.Builder<String, String> headerParams =
+                ImmutableSortedMap.orderedBy(String.CASE_INSENSITIVE_ORDER);
```
iamdanfox: this is neat!

carterkozak:
> Re single multiple: I don't quite follow, is it not the case that both can have multiple values?

Request headers use type `Map<String, String>` and response headers use `Map<String, List<String>`. It's the queryParams that are stored in a Multimap.

uschi2000:
ah, I was looking at the query params multimap, sorry.
yes, the request header params always have multiplicity 1 per conjure construction.

carterkozak:
> yes, the request header params always have multiplicity 1 per conjure construction.

The argument for Multimap<String,String> is that we may need multiple header values for something at the infrastructure level in the future (not conjure params, but something like `X-Forwarded-For`). I don't have a good example off the top of my head where multiple header values are required over a single value with a delimiter, so this may be a non-issue, but if an need came up it would require special code in all client implementations rather than a single Channel wrapper.

dialogue-target/src/main/java/com/palantir/dialogue/Request.java
```diff
@@ -16,37 +16,204 @@
...
+    public Map<String, String> headerParams() {
+        return headerParams;
+    }
 
     /**
      * The URL query parameters headers for this request. These should *not* be URL-encoded and {@link Channel}s are
      * responsible for performing the appropriate encoding if necessary.
      */
-    Multimap<String, String> queryParams();
+    public Multimap<String, String> queryParams() {
```
carterkozak: This is ambiguous, I'm going to pile one more break into this change to use a ListMultimap here.

dialogue-target/src/main/java/com/palantir/dialogue/Request.java
```diff
@@ -16,37 +16,205 @@
...
 
 /** Defines the parameters of a single {@link Call} to an {@link Endpoint}. */
-@DialogueImmutablesStyle
-@Value.Immutable
-public interface Request {
-    /** The HTTP headers for this request, encoded as a map of {@code header-name: header-value}. */
-    Map<String, String> headerParams();
+public final class Request {
+
+    private final ImmutableSortedMap<String, String> headerParams;
```
ferozco: Although conjure does enforce a single value per header, I don't see a reason why the underlying client should be so strict. I think being able to to express any API (including non-conjurized ones) within dialogue is important if we want to build a full replacement for CJR since it currently supports any valid JAXRS interface.

As a concrete example, Github allows you to opt into API previews using multiple `Accept` headers

carterkozak: Updated, I've also filed #343 to update the Response API to use a ListMultimap, but I'd prefer not to do that in this PR because it targets the Request side.

Regarding CJR, once dialogue is in a good place I'd like to replace our feign client with a dialogue channel (implement the feign client interface delegating to a dialogue Channel) as it should capture what we've tried to do with okhttp nicely.

iamdanfox (approved):
I'm happy this makes `Request` a bit more immutable (otherwise in theory people could try to build some 'clever' mutable implementation of the interface).

Hopefully it won't make testing with mocks too much of a pain.

PR #289, opened 2020-02-05
[improvement] Implement a generalized ContentDecodingChannel for gzip support

carterkozak:
## Before this PR
Gzip support implemented per-client with risk that the client transparently attempts to use other forms of compression.

## After this PR
Standard content-encoding support across all clients.


carterkozak:
This is the change that prompted the Request change (this PR is targeting that branch to reduce the diff)

dialogue-core/src/main/java/com/palantir/dialogue/core/ContentDecodingChannel.java
```diff
@@ -0,0 +1,236 @@
...
+
+        @Override
+        public boolean markSupported() {
+            return getDelegateSafely().markSupported();
+        }
+
+        @Override
+        public long transferTo(OutputStream out) throws IOException {
+            return getDelegate().transferTo(out);
+        }
```
carterkozak: This won't compile when #290 merges since it's a newer InputStream API

dialogue-client-test-lib/src/main/java/com/palantir/dialogue/AbstractChannelTest.java
```diff
@@ -308,30 +306,6 @@ public void connectionErrorsSurfaceAsExceptions() throws IOException {
...
-        Buffer gzipBytes = new Buffer();
-        Buffer rawBytes = new Buffer();
-        rawBytes.writeString(content, StandardCharsets.UTF_8);
-        rawBytes.flush();
-
-        GzipSink gzip = new GzipSink(gzipBytes);
-        gzip.write(rawBytes, 3);
-        gzip.close();
-        return gzipBytes;
-    }
```
carterkozak: I think we should keep this test, but test all clients wrapped with Channels.create to inherit generalized behavior.

dialogue-target/src/main/java/com/palantir/dialogue/Request.java
```diff
@@ -16,37 +16,203 @@
...
-    Optional<RequestBody> body();
+    public Optional<RequestBody> body() {
+        return body;
+    }
+
+    @Override
+    public String toString() {
+        return "Request{"
+                + "headerParamsSize="
+                + headerParams.size()
```
iamdanfox: :+1: nice - no leaking tokens here

dialogue-core/src/main/java/com/palantir/dialogue/core/ContentDecodingChannel.java
```diff
@@ -0,0 +1,236 @@
...
+
+    private static Response decompress(Response input) {
+        Optional<String> contentEncoding = input.getFirstHeader(CONTENT_ENCODING);
+        if (contentEncoding.isPresent() && GZIP.equals(contentEncoding.get())) {
+            return new ContentDecodingResponse(input);
+        }
+        return input;
+    }
+
+    private static final class ContentDecodingResponse implements Response {
```
iamdanfox, line 87: The naming of ContentDecodingChannel and ContentDecodingResponse kinda suggests this might at some point end up handling other kinds of encodings (e.g. brotli)? Is this the intention here or should we name this GzipContentDecodingChannel etc?

dialogue-core/src/main/java/com/palantir/dialogue/core/ContentDecodingChannel.java
```diff
@@ -0,0 +1,236 @@
...
+                .putHeaderParams(ACCEPT_ENCODING, GZIP)
+                .build();
+    }
+
+    private static Response decompress(Response input) {
+        Optional<String> contentEncoding = input.getFirstHeader(CONTENT_ENCODING);
+        if (contentEncoding.isPresent() && GZIP.equals(contentEncoding.get())) {
+            return new ContentDecodingResponse(input);
+        }
+        return input;
```
iamdanfox, line 84: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Encoding#Syntax shows a slightly funky response that servers can apparently send: `Content-Encoding: gzip, identity`... do you think it's worth having a debug log in here in case we ever want to look into why this isn't running?

dialogue-core/src/main/java/com/palantir/dialogue/core/ContentDecodingChannel.java
```diff
@@ -0,0 +1,236 @@
...
+
+        ContentDecodingResponse(Response delegate) {
+            this.delegate = delegate;
+            this.headers = Maps.filterKeys(delegate.headers(), ContentDecodingResponse::allowHeader);
+            this.body = new DeferredGzipInputStream(delegate.body());
+        }
+
+        @Override
+        public InputStream body() {
+            if (body == null) {
```
iamdanfox, line 101: I don't really understand this little guy - if the `body` field is final and assigned in the constructor, how can we end up in the body of this `if`?

carterkozak: oops, I don't think this conditional is necessary

dialogue-core/src/main/java/com/palantir/dialogue/core/ContentDecodingChannel.java
carterkozak: Good call on debug logging, I've never seen multiple Content-Encoding values in practice, but it would be helpful to have logging in this case.

dialogue-core/src/main/java/com/palantir/dialogue/core/ContentDecodingChannel.java
carterkozak: That's correct, this class is designed not to set an Accept-Encoding value if a custom value is already set elsewhere so it should be responsible for all conjure-supported response encodings, of which we currently have only one.

dialogue-target/src/main/java/com/palantir/dialogue/Request.java
carterkozak: In a later commit I think I updated this to include header names, not values, for a better debugging experience. Should be in #288

dialogue-core/src/main/java/com/palantir/dialogue/core/ContentDecodingChannel.java
iamdanfox: interesting that nothing about our static analysis tooling complained though!

iamdanfox (approved):
Looks sensible to me!

iamdanfox:
Happy to `merge when ready` this if you are!

carterkozak:
This depends on #288, I'll re-target to develop once that merges

PR #290, opened 2020-02-05
[improvement] dialogue-core targets jdk8

carterkozak:
## Before this PR
dialogue-core required jdk11, which isn't always available.

## After this PR
Dialogue clients can be built for jdk8

### DIFF
--- a/dialogue-core/build.gradle
+++ b/dialogue-core/build.gradle
@@ -2,8 +2,6 @@ apply from: "$rootDir/gradle/publish-jar.gradle"
 apply plugin: 'com.palantir.revapi'
 apply plugin: 'com.palantir.metric-schema'
 
-sourceCompatibility = 11
-
 dependencies {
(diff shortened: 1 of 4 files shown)


dialogue-core/build.gradle
```diff
@@ -2,8 +2,6 @@ apply from: "$rootDir/gradle/publish-jar.gradle"
 apply plugin: 'com.palantir.revapi'
 apply plugin: 'com.palantir.metric-schema'
 
-sourceCompatibility = 11
```
ferozco, line 5: you might want to set `sourceCompatibility = 1.8` explicitly to avoid an accidental regression

carterkozak: Should we do that in all the modules? The other modules which are jdk8-compatible inherit sourceCompatibility from the root build.gradle.

ferozco:
👍

PR #292, opened 2020-02-06
Move verb+body validation into an isolated Channel

carterkozak:
Fix: channel.execute must not throw, a future should be returned.
Follow-up: implement an outer channel to catch RuntimeExceptions
and Errors thrown and convert them into immediate failed futures.

### DIFF
--- a/dialogue-java-client/src/main/java/com/palantir/dialogue/HttpChannel.java
+++ b/dialogue-java-client/src/main/java/com/palantir/dialogue/HttpChannel.java
@@ -84,7 +84,6 @@ public final class HttpChannel implements Channel {
         // Fill request body and set HTTP method
         switch (endpoint.httpMethod()) {
             case GET:
-                Preconditions.checkArgument(!request.body().isPresent(), "GET endpoints must not have a request body");
                 httpRequest.GET();
                 break;
             case POST:
@@ -94,8 +93,6 @@ public final class HttpChannel implements Channel {
                 httpRequest.PUT(toBody(request));
                 break;
             case DELETE:
-                Preconditions.checkArgument(
-                        !request.body().isPresent(), "DELETE endpoints must not have a request body");
                 httpRequest.DELETE();
                 break;
         }
--- a/dialogue-client-test-lib/src/main/java/com/palantir/dialogue/AbstractChannelTest.java
+++ b/dialogue-client-test-lib/src/main/java/com/palantir/dialogue/AbstractChannelTest.java
@@ -26,6 +26,7 @@ import static org.mockito.Mockito.when;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableMultimap;
+import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
 import com.palantir.logsafe.exceptions.SafeIllegalArgumentException;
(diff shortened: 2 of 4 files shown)


carterkozak:
WIP: Needs its own tests, AbstractChannelTest.*failsWhenBodyIsGiven should be removed as we don't need to implement this validation in every client.

uschi2000:
ACK on the future-vs-throw change.
I find the version with a verifying channel harder to read. Is this in preparation for another channel implementation?

carterkozak:
> Is this in preparation for another channel implementation?

I started hacking on channel implementations for a couple other http clients for a direct comparison. This makes it easier to write a new client by reducing the required code duplication.

carterkozak:
#289 has a similar motivation since not all clients support content compression by default.

dialogue-java-client/src/main/java/com/palantir/dialogue/HttpChannel.java
```diff
@@ -84,7 +84,6 @@ public static HttpChannel of(HttpClient client, URL baseUrl, Duration requestTim
         // Fill request body and set HTTP method
         switch (endpoint.httpMethod()) {
             case GET:
-                Preconditions.checkArgument(!request.body().isPresent(), "GET endpoints must not have a request body");
```
iamdanfox, line 87: +1 to rfink's suggestion (co-locating the GET related validation with the actual mention of GET in this switch seems nice), could we just pop a line in here saying:

```java
if (request.body().isPresent()) {
  return Futures.immediateFailedFuture(
                    new SafeIllegalArgumentException("GET endpoint must not have a request body"));
}
```

I know java's control flow with async stuff is a bit crap (would love rust's `?` operator here), but maybe the verbosity is fine?

carterkozak: This means every client must re-implement the same validation, I’d like to generalize as much as possible to avoid similar issues to our okhttp mess. happy to leave both in place here, but I think the generalized handler makes it easier to build additional client implementations with equivalent semantics.

carterkozak: To rephrase; I don’t think this validation is a concern of client implementations, but a general dialogue concept.

uschi2000: As an alternative, could we consider stronger types that avoid optional fields where they don't make sense semantically?

iamdanfox: as long as we don't have to include derive4j ;)

carterkozak: Always in favor of moving validation to the type system :-)
That would be a larger break than the proposed request change, is that something we’re amenable to?

uschi2000: I think it's now or never.

carterkozak: Ack, before we change this I'd like to put some thought into the request body API. I put together quick netty client implementation (very, very naive at this point), but it exposes some api limitations that are worth considering before we lock things down.

carterkozak:
Closing, DELETE validation is unnecessary because we can send bodies.

PR #295, opened 2020-02-06
Backstop against exceptions thrown in Channel impls

iamdanfox:
## Before this PR

This repo uses the paradigm of ListenableFutures everywhere, and tries to avoid exceptions. Unfortunately java doesn't make it easy to completely avoid exceptions using static analysis, and we've already had some unintentional preconditions that would blast through the nice abstractions we set up in Channel.

## After this PR

Whatever strange implementations get passed into `Channels#create`, clients should always be able to assume that failures will come out as failed listenable futures, not as the _occasional_ exception!

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/NeverThrowChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/NeverThrowChannel.java
@@ -1,0 +1,53 @@
+/*
+ * (c) Copyright 2020 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.dialogue.core;
+
+import com.google.common.util.concurrent.Futures;
+import com.google.common.util.concurrent.ListenableFuture;
+import com.palantir.dialogue.Channel;
+import com.palantir.dialogue.Endpoint;
+import com.palantir.dialogue.Request;
+import com.palantir.dialogue.Response;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * The contract of {@link Channel} requires that the {@link Channel#execute} method never throws. This is a defensive
+ * backstop so that callers can rely on this invariant.
+ */
+final class NeverThrowChannel implements Channel {
+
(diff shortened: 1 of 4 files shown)


dialogue-core/src/main/java/com/palantir/dialogue/core/NeverThrowChannel.java
```diff
@@ -0,0 +1,53 @@
...
+
+    NeverThrowChannel(Channel delegate) {
+        this.delegate = delegate;
+    }
+
+    @Override
+    public ListenableFuture<Response> execute(Endpoint endpoint, Request request) {
+        try {
+            return delegate.execute(endpoint, request);
+        } catch (RuntimeException e) {
```
carterkozak: I could go either way on handling Error as well, OutOfMemoryError isn't terribly uncommon, and throwing from unexpected calls when things go poorly could result in additional bugs being surfaced (resource leaks). Thoughts?

carterkozak:
Thanks for putting this together!

dialogue-core/src/main/java/com/palantir/dialogue/core/NeverThrowChannel.java
iamdanfox: I think we set [CrashOnOutOfMemoryError](https://github.com/palantir/sls-packaging/blob/07fbbe8e57eb59771d29eff0fda1423d33ec7829/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/tasks/LaunchConfigTask.java#L57) but I think isolating NoSuchMethodError so that it only breaks specific codepaths is probably a good idea.

carterkozak: CrashOnOutOfMemoryError doesn't crash when you run out of direct memory or (I might be mis-remembering this one) attempt to create an array that's larger than the maximum length. I think ByteArrayOutputStream explicitly thows OME in some cases which doesn't trigger a crash.

carterkozak:
Odd that javaChannels and HttpChannels disagree, aren’t they the same thing with different wrappers? Should we require the channel wrapper for all these tests?

iamdanfox:
- HttpChannel is the lowest level - it's pretty much just a wrapper around java's stdlib client (i.e. it doesn't involve the NeverThrowChannel because it doesn't go via the Channels#create method)
- JavaChannels#create includes all the opinionated logic around tracing and multiple urls and _does_ include my new NeverThrowChannel because it calls Channels#create.

They're both required to pass the AbstractChannelTests I think just to make sure that all the 'opinionated logic' doesn't accidentally change the semantics of making a straightforward request to one url.

Maybe this just means the `get_failsWhenBodyIsGiven` and `delete_failsWhenBodyIsGiven` things should be moved to the dedicated `JavaChannelsTest` (especially if your PR goes in)?

carterkozak:
I ran into a similar issue working on gzip content decoding when I generalized it as well, perhaps we want different set of test for generic raw clients as opposed to wrapper clients, I think it's helpful to run tests against all clients with the standard wrappers.

carterkozak (approved):
lgtm, thanks!

carterkozak:
dozer?

Issue #299, opened 2020-02-07
QueuedChannel leaks threads


carterkozak:
Each new instance of a QueuedChannel creates a new scheduled executor that is never shut down.

carterkozak:
Fixed by #303

Issue #300, opened 2020-02-07
QueudChannel executor prevents the JVM from stopping


carterkozak:
QueuedChannel executor should use daemon threads, otherwise the JVM does not exit when main completes.

Relatedly, we may want to consider making Channel implement Closeable to clean up resources when we're done with it. Our infrastructure around OkHttp doesn't do this because the okhttp client is not closeable, however most http clients are.

carterkozak:
Fixed by #303

PR #303, opened 2020-02-07
[improvement] QueuedChannel avoids queue contention in the fast path

carterkozak:
## Before this PR
The queue was used for every request

## After this PR
The QueuedChannel queue is avoided unless an initial attempt to execute the request is limited.
This can produced unexpected order of execution when a limit is reached and recedes, but worth the trade-off to limit load.

When requests are limited, other requests are less likely to prompt the queue to allow them through as there's no backpressure signal from the delegate LimitedChannel, however this is no different from a request followed by no subsequent requests, which is a bug that should also be fixed. Ultimately LimitedChannels should be able to signal when they're ready to accept requests again.

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
@@ -44,8 +44,6 @@ import org.immutables.value.Value;
  * <ol>
  *     <li>On submission - allows execution when there is available capacity</li>
  *     <li>On request completion - allows execution when capacity has now become available</li>
- *     <li>Periodically (eg: every 100ms) - allows execution when there may have been no capaciy and no in-flight
- *     requests</li>
  * </ol>
  *
  * This implementation was chosen over alternatives for the following reasons:
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
@@ -23,7 +23,6 @@ import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
 import com.google.common.util.concurrent.MoreExecutors;
 import com.google.common.util.concurrent.SettableFuture;
-import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import com.palantir.dialogue.Channel;
 import com.palantir.dialogue.Endpoint;
 import com.palantir.dialogue.Request;
@@ -35,14 +34,9 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.BlockingDeque;
 import java.util.concurrent.Executor;
-import java.util.concurrent.Executors;
 import java.util.concurrent.LinkedBlockingDeque;
-import java.util.concurrent.ScheduledExecutorService;
-import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 import org.immutables.value.Value;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 /**
  * A {@link Channel} that queues requests while the underlying {@link LimitedChannel} is unable to accept any new
@@ -66,18 +60,12 @@ import org.slf4j.LoggerFactory;
  */
 final class QueuedChannel implements Channel {
 
-    private static final Logger log = LoggerFactory.getLogger(QueuedChannel.class);
     private static final Executor DIRECT = MoreExecutors.directExecutor();
 
     private final BlockingDeque<DeferredCall> queuedCalls;
     private final LimitedChannel delegate;
     // Tracks requests that are current executing in delegate and are not tracked in queuedCalls
     private final AtomicInteger numRunningRequests = new AtomicInteger(0);
-    private final ScheduledExecutorService backgroundScheduler = Executors.newSingleThreadScheduledExecutor(
-            new ThreadFactoryBuilder()
-                    .setNameFormat("dialogue-request-scheduler")
-                    .setDaemon(false)
-                    .build());
 
     QueuedChannel(LimitedChannel channel, DispatcherMetrics metrics) {
         this(channel, 1_000, metrics);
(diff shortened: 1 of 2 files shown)


dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
```diff
@@ -109,6 +109,16 @@
      */
     @Override
     public ListenableFuture<Response> execute(Endpoint endpoint, Request request) {
+        // Optimistically avoid the queue in the fast path.
+        // Queuing adds contention between threads and should be avoided unless we need to shed load.
+        Optional<ListenableFuture<Response>> maybeResult = delegate.maybeExecute(endpoint, request);
+        if (maybeResult.isPresent()) {
+            ListenableFuture<Response> result = maybeResult.get();
+            numRunningRequests.incrementAndGet();
+            result.addListener(numRunningRequests::decrementAndGet, MoreExecutors.directExecutor());
```
carterkozak: This needs to call schedule as well.

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
```diff
@@ -88,17 +76,6 @@
...
-                () -> {
-                    try {
-                        schedule();
-                    } catch (Exception e) {
-                        log.error("Uncaught exception while scheduling request. This is a programming error.", e);
-                    }
-                },
-                100,
-                100,
-                TimeUnit.MILLISECONDS);
```
iamdanfox, line 101: Deleting this background scheduler means that if there are 10 calls running and 5 more queued and the delegate suddenly decides it has more capacity available but this is _not_ aligned with an existing call starting or finishing then the queue of 5 won't be drained quickly.

A reason we might be able to suddenly gain more capacity is if we live-reloaded a list of urls and a new URL appeared (to which we had zero in flight requests).

carterkozak: I’m not sure a 100ms delay is acceptable in that case either, that’s about 300 full requests+responses/thread. I think we need a way for the delegate channel to signal when it’s hungry.

At the moment reloading urls requires us to recreate the entire channel chain, so queued requests cannot be tried against new targets.

iamdanfox: Oof yep looks like the RefreshingChannelFactory is currently being used as a wrapper around everything:

```java
    private static RefreshingChannelFactory channelFactory(
            Witchcraft witchcraft,
            Refreshable<EmailRuntimeConfiguration> runtime) {
        return new RefreshingChannelFactory(
                runtime.map(c -> c.servicesBlock().orElse(EMPTY_CONF)),
                c -> JavaChannels.create(c, witchcraft.userAgent(), witchcraft.taggedMetrics()));
    }
```

So that actually means that every time something got live-reloaded we'd be re-creating that `dialogue-request-scheduler`!

Probably worth deleting mentions of the 100ms timeout from this class' javadoc then, as it'll be confusing to see a description that's out of sync with the actual impl.

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
```diff
@@ -120,6 +107,11 @@
         return components.response();
     }
 
+    private void onCompletion() {
+        numRunningRequests.decrementAndGet();
```
iamdanfox, line 109: just to make sure I'm understanding this stuff correctly - the book-keeping we do to update numRunningRequests is purely for metrics purposes right? it doesn't actually influence any of the behaviour?

carterkozak: That's correct, I have a ticket filed for that as well since it's used for a gauge and we have multiple QueuedChannel instances. Replacing the atomiclongs with metric Counter instances should do the trick though.

carterkozak:
Fixes #299 and #300

iamdanfox (approved):

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
```diff
@@ -109,6 +84,16 @@
...
     @Override
     public ListenableFuture<Response> execute(Endpoint endpoint, Request request) {
+        // Optimistically avoid the queue in the fast path.
+        // Queuing adds contention between threads and should be avoided unless we need to shed load.
+        Optional<ListenableFuture<Response>> maybeResult = delegate.maybeExecute(endpoint, request);
+        if (maybeResult.isPresent()) {
+            ListenableFuture<Response> result = maybeResult.get();
+            numRunningRequests.incrementAndGet();
+            result.addListener(this::onCompletion, DIRECT);
+            return result;
```
ellisjoe, line 94: can we pull this logic out since it looks about the same as what's in `scheduleNextTask`?

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
```diff
@@ -109,6 +84,16 @@
      */
     @Override
     public ListenableFuture<Response> execute(Endpoint endpoint, Request request) {
+        // Optimistically avoid the queue in the fast path.
+        // Queuing adds contention between threads and should be avoided unless we need to shed load.
```
ellisjoe, line 88: you now can queue jump in a pretty weird way? probably need to at least check the queue is empty

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
```diff
@@ -88,17 +74,6 @@
     QueuedChannel(LimitedChannel delegate, int maxQueueSize, DispatcherMetrics metrics) {
         this.delegate = delegate;
         this.queuedCalls = new LinkedBlockingDeque<>(maxQueueSize);
-        this.backgroundScheduler.scheduleWithFixedDelay(
```
ellisjoe, line 91: Without this you can end up with requests that are queued forever:

1. Submit 2 requests, but only have capacity for 1
2. Request 1 finishes and tries to schedule next request but still no capacity
3. Request 2 is stuck

carterkozak: Forever and 100ms are roughly equivalent here. See the comments on signaling availability for a better long-term solution.

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
carterkozak: An empty check would also result in a race. I don't think we make any guarantees that the first request submitted will be the first one we attempt to execute.

ellisjoe: I guess the situation that is now possible (albeit unlikely) is that I submit a request that gets queued and every subsequent request preempts it such that it is never run. That seems like an undesirable structural thing?

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
ellisjoe: I'm not sure I understand how "forever" and "100ms" are roughly equivalent?

As far as having LimitedChannels signal that they can accept requests: I'd have to see a proposal for how this works. It sounds like the sort of thing that could get a bit hairy.

PR #310, opened 2020-02-11
Match CJR okhttp version

carterkozak:
A dependency on dialogue no longer brings an unexpected okhttp
version.

## Before this PR
Dialogue okhttp channel included an unexpected okhttp upgrade which breaks reliance on unrepeatable request bodies.

## After this PR
Same version as other components.


iamdanfox (approved):

iamdanfox:
cc @jkozlowski 😢

PR #314, opened 2020-02-11
fix #312: RoundRobinChannel fails when no targets are available

### LINKED ISSUE #312: RoundRobinChannel should fail when no targets exist
## What happened?

Currently it returns a 'limited' response despite being impossible for the request to ever work. When no targets are present we cannot assume that they will be added. At the moment RoundRobinChannel is immutable and the list of hosts cannot change.

The channel should return a failed response rather than a limited response.

carterkozak:
## Before this PR
Previously it would respond as though it was not ready and allow
requests to potentially queue.

## After this PR
Requests fail when there's nowhere to send them

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/RoundRobinChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/RoundRobinChannel.java
@@ -17,10 +17,12 @@
 package com.palantir.dialogue.core;
 
 import com.google.common.collect.ImmutableList;
+import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
 import com.palantir.dialogue.Endpoint;
 import com.palantir.dialogue.Request;
 import com.palantir.dialogue.Response;
+import com.palantir.logsafe.exceptions.SafeIoException;
 import java.util.List;
 import java.util.Optional;
 import java.util.concurrent.atomic.AtomicInteger;
@@ -40,7 +42,7 @@ final class RoundRobinChannel implements LimitedChannel {
     @Override
     public Optional<ListenableFuture<Response>> maybeExecute(Endpoint endpoint, Request request) {
         if (delegates.isEmpty()) {
-            return Optional.empty();
+            return Optional.of(Futures.immediateFailedFuture(new SafeIoException("No targets are available")));
         }
 
         int host = currentHost.getAndUpdate(value -> toIndex(value + 1));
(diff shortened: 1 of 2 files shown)


markelliot:
This doesn't seem like obviously good behavior, especially if the targets become live-reloadable. Instead, a channel that consumes this one should encode a finite delay. If you want to key on a 3-state response here ("no channels", "channels but none ready for picking", "dispatched") then the return type should be a union rather than an optional

carterkozak:
@markelliot I'm not sure that makes sense in the context of the 'LimitedChannel' abstraction, but the RetryingChannel could be improved to support backoff.

markelliot:
I think it's already kind of clear from the docs here that the RoundRobinChannel _attempts_ to schedule a future and the fact that one cannot be scheduled is handled by the Optional return. Something which calls the RoundRobinChannel and gets an empty instead of a Future can surely use that signal to make a decision about what to do, whereas this change _forces_ the behavior to be a hard failure...that seems wrong to me.

carterkozak:
Our documentation on LimitedChannel is sparse, but I think we're litigating the precise meaning of an empty return from LimitedChannel.maybeExecute.

I'm not confident we can adequately estimate whether an endpoint will be discovered or how long we should wait before failing.

markelliot:
Sure -- my argument is that RoundRobinChannel (or any LimitedChannel) is not the appropriate place to decide the wait criteria; that should be handled by a regular Channel which consumes a LimitedChannel and handles the case of an empty Optional.

Your argument seems to be that RoundRobinChannel is a special kind of LimitedChannel that chooses to handle certain limits in a way which prevents consumers of the LimitedChannel from deciding what action to take.

ellisjoe:
Yeah this does not seem like the right thing. In particular, you'll start failing requests when all downstreams are rate limited rather than queueing the requests.

carterkozak:
I think we’re mostly in agreement, I’m planning a replacement for the empty optional that can signal when the channel is ready for inputs, but that requires us to implement live reloadable targets, otherwise we wait forever in this case when we know it’s impossible to complete requests.

carterkozak:
> Yeah this does not seem like the right thing. In particular, you'll start failing requests when all downstreams are rate limited rather than queueing the requests

That’s not quite correct, this doesn’t change behavior when delegates are limited, only when they do not exist, and cannot be added to an immutable list.

carterkozak:
To be clear, this fixes a bug. If we implement a feature for live reloadable channels we can design changes to this, but at the moment it's simply broken.

I would like to get this project into a state that it's no longer dangerous before we implement additional features.

carterkozak:
fixed elsewhere

PR #331, opened 2020-02-14
fix #326: Requests write to output streams, they don't create inputstreams

### LINKED ISSUE #326: RequestBody should write to an OutputStream rather than produce an InputStream
## What happened?
The binary request body code (implementation is a TODO) exposes cracks in the request body API.

Consider inverting the API from getting a byte source (inpubstream) to consuming an outputstream that can be written to.

Currently there's a lot of glue code, allocation, and unnecessary memory copying, let's consider the lifecycle of a standard json request:

1: Pojo object is serialized into a RequestBody:

```java
        @Override
        public RequestBody serialize(T value) {
            Preconditions.checkNotNull(value, "cannot serialize null value");
            ByteArrayOutputStream bytes = new ByteArrayOutputStream();
            encoding.serializer.serialize(value, bytes);
            try {
                bytes.flush();
                bytes.close();
            } catch (IOException e) {
                throw new SafeRuntimeException("Failed to close or flush ByteStream. This is a bug.", e);
            }

            return new RequestBody() {
                @Override
                public Optional<Long> length() {
                    return Optional.of((long) bytes.size());
                }

                @Override
                public InputStream content() {
                    return new ByteArrayInputStream(bytes.toByteArray());
                }

                @Override
                public String contentType() {
                    return encoding.encoding.getContentType();
                }
            };
        }
    }
```

2: Body is converted into a format the underlying http client understands

```java
    private RequestBody toOkHttpBody(com.palantir.dialogue.RequestBody body) {
        return new RequestBody() {
            @Nullable
            @Override
            public MediaType contentType() {
                return MediaType.parse(body.contentType());
            }

            @Override
            public void writeTo(BufferedSink sink) throws IOException {
                ByteStreams.copy(body.content(), sink.outputStream());
            }
        };
    }
```

If we were to use a body API that writes to an OutputStream (or other data sink) we could vastly simplify the code as well as remove the need for several copies of the serialized object on heap, allowing the jvm to garbage collect what we do create sooner.

The first snippet becomes:
```java
        @Override
        public RequestBody serialize(T value) {
            Preconditions.checkNotNull(value, "cannot serialize null value");
            return new RequestBody() {
                @Override
                public Optional<Long> length() {
                    return Optional.empty();
                }

                @Override
                public void content(OutputStream out) throws IOException {
                    encoding.serializer.serialize(value, out);
                }

                @Override
                public String contentType() {
                    return encoding.encoding.getContentType();
                }
            };
        }
    }
```

And the second:
```java
    private RequestBody toOkHttpBody(com.palantir.dialogue.RequestBody body) {
        return new RequestBody() {
            @Nullable
            @Override
            public MediaType contentType() {
                return MediaType.parse(body.contentType());
            }

            @Override
            public void writeTo(BufferedSink sink) throws IOException {
                body.content(sink.outputStream());
            }
        };
    }
```

carterkozak:
See #326 for details.

### DIFF
--- a/dialogue-client-test-lib/src/main/java/com/palantir/dialogue/AbstractChannelTest.java
+++ b/dialogue-client-test-lib/src/main/java/com/palantir/dialogue/AbstractChannelTest.java
@@ -28,9 +28,8 @@ import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableMultimap;
 import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
-import java.io.ByteArrayInputStream;
 import java.io.IOException;
-import java.io.InputStream;
+import java.io.OutputStream;
 import java.net.ConnectException;
 import java.net.URL;
 import java.nio.charset.StandardCharsets;
@@ -72,8 +71,8 @@ public abstract class AbstractChannelTest {
         }
 
         @Override
-        public InputStream content() {
-            return new ByteArrayInputStream(CONTENT);
+        public void writeTo(OutputStream output) throws IOException {
+            output.write(CONTENT);
         }
 
         @Override
--- a/dialogue-client-test-lib/src/main/java/com/palantir/dialogue/AbstractChannelTest.java
+++ b/dialogue-client-test-lib/src/main/java/com/palantir/dialogue/AbstractChannelTest.java
@@ -35,7 +35,6 @@ import java.net.URL;
 import java.nio.charset.StandardCharsets;
 import java.util.Map;
 import java.util.Optional;
-import java.util.OptionalLong;
 import java.util.Set;
 import java.util.concurrent.CancellationException;
 import java.util.concurrent.ExecutionException;
@@ -65,11 +64,6 @@ public abstract class AbstractChannelTest {
     public final MockWebServer server = new MockWebServer();
 
     private final RequestBody body = new RequestBody() {
-        @Override
-        public OptionalLong length() {
-            return OptionalLong.of(CONTENT.length);
-        }
-
         @Override
         public void writeTo(OutputStream output) throws IOException {
             output.write(CONTENT);
--- a/dialogue-java-client/src/main/java/com/palantir/dialogue/HttpChannel.java
+++ b/dialogue-java-client/src/main/java/com/palantir/dialogue/HttpChannel.java
@@ -20,6 +20,7 @@ import com.google.common.base.Suppliers;
 import com.google.common.util.concurrent.ListenableFuture;
 import com.palantir.logsafe.Preconditions;
 import com.palantir.logsafe.exceptions.SafeRuntimeException;
+import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.URISyntaxException;
@@ -135,7 +136,13 @@ public final class HttpChannel implements Channel {
 
     private static HttpRequest.BodyPublisher toBody(Request request) {
         if (request.body().isPresent()) {
-            return HttpRequest.BodyPublishers.ofInputStream(request.body().get()::content);
+            ByteArrayOutputStream bytes = new ByteArrayOutputStream();
+            try {
+                request.body().get().writeTo(bytes);
+            } catch (IOException e) {
+                throw new SafeRuntimeException("Failed to create a BodyPublisher", e);
+            }
+            return HttpRequest.BodyPublishers.ofByteArray(bytes.toByteArray());
         } else {
             return HttpRequest.BodyPublishers.noBody();
         }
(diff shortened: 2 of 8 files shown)


dialogue-java-client/src/main/java/com/palantir/dialogue/HttpChannel.java
```diff
@@ -154,7 +155,13 @@ public int code() {
 
     private static HttpRequest.BodyPublisher toBody(Request request) {
         if (request.body().isPresent()) {
-            return HttpRequest.BodyPublishers.ofInputStream(request.body().get()::content);
+            ByteArrayOutputStream bytes = new ByteArrayOutputStream();
+            try {
+                request.body().get().writeTo(bytes);
+            } catch (IOException e) {
+                throw new SafeRuntimeException("Failed to create a BodyPublisher", e);
+            }
```
carterkozak, line 144: This is the same block from ConjureBodySerDe, but it's only required for HttpChannel, not socket based transports.

dialogue-java-client/src/main/java/com/palantir/dialogue/HttpChannel.java
```diff
@@ -154,7 +155,13 @@ public int code() {
...
+            return HttpRequest.BodyPublishers.ofByteArray(bytes.toByteArray());
```
carterkozak, line 145: Here we're able to use the more efficient ofByteArray publisher rather than ofInputStream.

carterkozak: Note that before, octet-stream requests would throw (not-implemented) where here they're fully buffered on heap like our feign client. This can be dangerous, but I don't think we should spend the time to optimize that case unless we decide to go all in on the jdk11 client.

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ConjureBodySerDe.java
```diff
@@ -100,8 +99,8 @@ public RequestBody serialize(BinaryRequestBody value) {
             }
 
             @Override
-            public InputStream content() {
-                throw new UnsupportedOperationException("TODO(rfink): implement this");
```
carterkozak, line 105: todone.

carterkozak: todo: implement tests for binary requests in the abstract channel test.

ferozco (commented):
It feels weird that we're requiring a request body to advertise its size but also providing an API which encourages the implementer to omit it. I worry that avoiding the duplication of bytes might be pre-mature optimization and we should sit on this until we have some better benchmarks or real life usage

carterkozak:
> It feels weird that we're requiring a request body to advertise its size but also providing an API which encourages the implementer to omit it.

I'd be in favor of removing request body size entirely.

carterkozak:
> I worry that avoiding the duplication of bytes might be pre-mature optimization and we should sit on this until we have some better benchmarks or real life usage

I think this both simplifies the API, ignoring the performance benefit entirely.

ferozco:
> I'd be in favor of removing request body size entirely.

Would you intend to omit the header entirely? It seems we'd be putting ourself in a bind to try and populate the header from within the channel if we decided to not buffer binary requests.

> I think this both simplifies the API, ignoring the performance benefit entirely.

I agree that it does simplify the API but I would still prefer to focus on getting the current client out before making too many changes

carterkozak:
The http client libraries I'm aware of tend to either take chunks of data from a callback/consumer (async-io) or provide an OutputStream to write data to. I don't think we should invent a new model for our API that makes it more difficult and less performant to adapt to clients unless we have a strong reason for it.

> Would you intend to omit the header entirely?

The Content-Length header? Many clients will buffer up to N bytes (often 8 or 16k) and send content-length if the data is fully buffered, otherwise send chunked requests. I think that's a reasonable approach for us to take.

>  It seems we'd be putting ourself in a bind to try and populate the header from within the channel if we decided to not buffer binary requests.

There's not really a downside to sending chunked requests, especially compared to buffering. Binary request buffering gets a little bit funky, but I think it's the same problem that we have a TODO to implement already, that's only going to be special for async-io clients (e.g. netty and jdk9 client) where InputStream has similar problems, the jdk9 input stream body producer is complicated and we're better off avoiding it if we can -- this will give us a better idea of the real performance of the client.

> I would still prefer to focus on getting the current client out before making too many changes

Fully supportive of getting what we have into a good state but I'm also worried that we'll gather metrics to pick our target client using data from a subpar implementation. Once we start to roll dialogue out I imagine API changes like this will be harder to make.

ferozco:
👍 ok, lets move ahead with this

dialogue-client-test-lib/src/main/java/com/palantir/dialogue/AbstractChannelTest.java
```diff
@@ -67,13 +65,8 @@
 
     private final RequestBody body = new RequestBody() {
         @Override
-        public OptionalLong length() {
```
markelliot, line 70: when we know the length we probably want to write the Content-Length header

carterkozak: Clients which buffer requests prior to sending them do send a content-length. This change allowed us to avoid fully buffering requests in many cases as we write directly to a socket.
HttpChannel (java 9 client) could set a content-length header, though it never did before.

PR #335, opened 2020-02-16
Implement a HttpUrlConnection and Apache HC4 based channels

carterkozak:
Gives us a data point to compare other clients with.


dialogue-httpurlconnection-client/src/main/java/com/palantir/dialogue/httpurlconnection/HttpUrlConnectionBlockingChannel.java
```diff
@@ -0,0 +1,171 @@
...
+
+final class HttpUrlConnectionBlockingChannel implements BlockingChannel {
+
+    private final ClientConfiguration config;
+    private final UrlBuilder baseUrl;
+
+    HttpUrlConnectionBlockingChannel(ClientConfiguration config, URL baseUrl) {
+        this.config = config;
+        // Sanitize path syntax and strip all irrelevant URL components
+        Preconditions.checkArgument(
```
ferozco: You can get rid of all of this since the `UrlBuilder` now does all the validation for you

dialogue-httpurlconnection-client/src/main/java/com/palantir/dialogue/httpurlconnection/HttpUrlConnectionBlockingChannel.java
```diff
@@ -0,0 +1,171 @@
...
+        this.baseUrl = UrlBuilder.from(baseUrl);
+    }
+
+    @Override
+    public Response execute(Endpoint endpoint, Request request) throws IOException {
+        // Create base request given the URL
+        UrlBuilder url = baseUrl.newBuilder();
+        endpoint.renderPath(request.pathParams(), url);
+        request.queryParams().forEach(url::queryParam);
+        URLConnection urlConnection = url.build().openConnection();
```
ferozco: probably just want to cast directly to the HTTP connection

dialogue-httpurlconnection-client/src/main/java/com/palantir/dialogue/httpurlconnection/HttpUrlConnectionBlockingChannel.java
```diff
@@ -0,0 +1,171 @@
...
+        if (connection instanceof HttpsURLConnection) {
+            HttpsURLConnection httpsConnection = (HttpsURLConnection) connection;
+            httpsConnection.setSSLSocketFactory(config.sslSocketFactory());
+            // TODO(ckozak): hostname verifier, otherwise the default is used.
+            // It's likely fine for most scenarios, but we want consistency
+            // across all channels.
+        }
+
+        if (request.body().isPresent()) {
+            if (endpoint.httpMethod() == HttpMethod.GET) {
```
ferozco: Preconditions.checkArgument

dialogue-httpurlconnection-client/src/main/java/com/palantir/dialogue/httpurlconnection/HttpUrlConnectionBlockingChannel.java
```diff
@@ -0,0 +1,171 @@
...
+                throw new SafeIllegalArgumentException("GET endpoints must not have a request body");
+            }
+            if (endpoint.httpMethod() == HttpMethod.DELETE) {
```
ferozco: J9 client has this same limitation which is unfortunate given that Conjure does allow for DELETEs with bodies

dialogue-httpurlconnection-client/src/test/java/com/palantir/dialogue/httpurlconnection/HttpUrlConnectionChannelsTest.java
```diff
@@ -0,0 +1,47 @@
...
+import com.palantir.conjure.java.client.config.ClientConfigurations;
+import com.palantir.dialogue.AbstractChannelTest;
+import com.palantir.dialogue.Channel;
+import com.palantir.tritium.metrics.registry.DefaultTaggedMetricRegistry;
+import java.net.URL;
+import java.nio.file.Paths;
+import org.junit.runner.RunWith;
+import org.mockito.junit.MockitoJUnitRunner;
+
+@RunWith(MockitoJUnitRunner.class)
```
ferozco: why do we need the mockito runner?

dialogue-httpurlconnection-client/src/test/java/com/palantir/dialogue/httpurlconnection/HttpUrlConnectionChannelsTest.java
```diff
@@ -0,0 +1,47 @@
...
+public final class HttpUrlConnectionChannelsTest extends AbstractChannelTest {
+
+    private static final SslConfiguration SSL_CONFIG = SslConfiguration.of(
+            Paths.get("src/test/resources/trustStore.jks"), Paths.get("src/test/resources/keyStore.jks"), "keystore");
```
ferozco, line 31: could we use `ca.cert` instead?

dialogue-hc4-client/src/main/java/com/palantir/dialogue/hc4/ApacheHttpClientChannel.java
```diff
@@ -0,0 +1,217 @@
...
+
+final class ApacheHttpClientChannel implements BlockingChannel {
+
+    private final CloseableHttpClient client;
+    private final UrlBuilder baseUrl;
+
+    ApacheHttpClientChannel(CloseableHttpClient client, URL baseUrl) {
+        this.client = client;
+        // Sanitize path syntax and strip all irrelevant URL components
+        Preconditions.checkArgument(
```
ferozco: can also remove this validation

dialogue-hc4-client/src/main/java/com/palantir/dialogue/hc4/ApacheHttpClientChannel.java
```diff
@@ -0,0 +1,217 @@
...
+    }
+
+    @Override
+    public Response execute(Endpoint endpoint, Request request) throws IOException {
+        // Create base request given the URL
+        UrlBuilder url = baseUrl.newBuilder();
+        endpoint.renderPath(request.pathParams(), url);
+        request.queryParams().forEach(url::queryParam);
+        URL target = url.build();
+        RequestBuilder builder = buildFor(endpoint, target);
```
ferozco:
```suggestion
        RequestBuilder builder = RequestBuilder.create(endpoint.httpMethod().toString()).setUri(target.toString());

```

dialogue-hc4-client/src/main/java/com/palantir/dialogue/hc4/ApacheHttpClientChannel.java
```diff
@@ -0,0 +1,217 @@
...
+        // Fill headers
+        for (Map.Entry<String, String> header : request.headerParams().entrySet()) {
+            builder.addHeader(header.getKey(), header.getValue());
+        }
+
+        if (request.body().isPresent()) {
+            if (endpoint.httpMethod() == HttpMethod.GET) {
+                throw new SafeIllegalArgumentException("GET endpoints must not have a request body");
+            }
+            if (endpoint.httpMethod() == HttpMethod.DELETE) {
```
ferozco: is this a limitation of the client or are we just enforcing this?

dialogue-hc4-client/src/main/java/com/palantir/dialogue/hc4/ApacheHttpClientChannel.java
```diff
@@ -0,0 +1,217 @@
...
+                            .add(header.getValue());
+                }
+                headers = tmpHeaders;
+            }
+            return headers;
+        }
+
+        @Override
+        public Optional<String> getFirstHeader(String header) {
+            Header first = response.getFirstHeader(header);
```
ferozco:
```suggestion
            return Optional.ofNullable(response.getFirstHeader(header)).map(Header::getValue);

```

dialogue-httpurlconnection-client/src/test/java/com/palantir/dialogue/httpurlconnection/HttpUrlConnectionChannelsTest.java
carterkozak: AbstractChannelTest needs it, I borrowed this from the other implementations but it looks like annotating the abstract class is sufficient.

dialogue-httpurlconnection-client/src/test/java/com/palantir/dialogue/httpurlconnection/HttpUrlConnectionChannelsTest.java
carterkozak: This matches the other AbstractChannelTest implementations, I'd prefer to keep them in sync until we move everything over.

dialogue-httpurlconnection-client/src/main/java/com/palantir/dialogue/httpurlconnection/HttpUrlConnectionBlockingChannel.java
carterkozak: I don't think HttpURLConnection has this limitation, I added validation to avoid behavior differences between clients. Should we allow clients to send deletes with bodies?

dialogue-hc4-client/src/main/java/com/palantir/dialogue/hc4/ApacheHttpClientChannel.java
carterkozak: good call

dialogue-httpurlconnection-client/src/main/java/com/palantir/dialogue/httpurlconnection/HttpUrlConnectionBlockingChannel.java
ferozco: I think if possible we should support DELETE with bodies given that valid conjure definitions may include them

carterkozak: The jdk9 client can support them if we ask it, I'll include that in this change.

dialogue-hc4-client/src/main/java/com/palantir/dialogue/hc4/ApacheHttpClientChannel.java
carterkozak: Was just enforcing this for consistency, but I've fixed all clients to support delete with a body.

dialogue-httpurlconnection-client/src/main/java/com/palantir/dialogue/httpurlconnection/HttpUrlConnectionBlockingChannel.java
ferozco: I had been seeing runtime failures when trying to get it to make delete requests with bodies

carterkozak: The test makes a delete request with a body and passes :-)

dialogue-hc4-client/src/main/java/com/palantir/dialogue/hc4/ApacheHttpClientBlockingChannel.java
```diff
@@ -0,0 +1,199 @@
...
+import org.apache.http.client.methods.RequestBuilder;
+import org.apache.http.impl.client.CloseableHttpClient;
+import org.apache.http.message.BasicHeader;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+final class ApacheHttpClientBlockingChannel implements BlockingChannel {
+    private static final Logger log = LoggerFactory.getLogger(ApacheHttpClientBlockingChannel.class);
+
+    private final CloseableHttpClient client;
```
iamdanfox: Ok so at the moment we're potentially leaving a connection manager open by not closing this anywhere, which is why you wrote up https://github.com/palantir/dialogue/issues/339#issuecomment-587248596 ?

carterkozak: That's correct.

dialogue-hc4-client/src/main/java/com/palantir/dialogue/hc4/ApacheHttpClientBlockingChannel.java
```diff
@@ -0,0 +1,199 @@
...
+                for (Header header : response.getAllHeaders()) {
+                    String value = header.getValue();
+                    if (value != null) {
+                        tmpHeaders
+                                .computeIfAbsent(header.getName(), _name -> new ArrayList<>(1))
+                                .add(header.getValue());
+                    }
+                }
+                headers = tmpHeaders;
+            }
```
iamdanfox: seeing the lazy initialization of this makes me wonder if we should document the thread-safety expectations of Response somewhere?

dialogue-hc4-client/src/main/java/com/palantir/dialogue/hc4/ApacheHttpClientBlockingChannel.java
```diff
@@ -0,0 +1,199 @@
...
+
+        @Override
+        public Map<String, List<String>> headers() {
+            if (headers == null) {
+                Map<String, List<String>> tmpHeaders = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
+                for (Header header : response.getAllHeaders()) {
+                    String value = header.getValue();
+                    if (value != null) {
+                        tmpHeaders
+                                .computeIfAbsent(header.getName(), _name -> new ArrayList<>(1))
```
iamdanfox: you sure you don't want to pull this out to a static method reference? ;) it's gonna allocate one whole lambda every time we call headers()

carterkozak: Oh you know I want to. Keeping simple for now :-)

dialogue-hc4-client/src/main/java/com/palantir/dialogue/hc4/ApacheHttpClientBlockingChannel.java
```diff
@@ -0,0 +1,199 @@
...
+            // Applies to responses.
+            return false;
+        }
+
+        @Override
+        public void consumeContent() {}
+
+        @Override
+        public String toString() {
+            return "RequestBodyEntity{requestBody=" + requestBody + '}';
```
iamdanfox: I'd love to get your immutable RequestBody PR merged, because I think as we have no control over the requestBody implementation, there's currently a danger that this toString value could include sensitive headers (incl. bearertokens).

carterkozak: We did! This is the body object (tuple of `Consumer<OutputStream>` and contentType)

dialogue-hc4-client/src/main/java/com/palantir/dialogue/hc4/ApacheHttpClientBlockingChannel.java
```diff
@@ -0,0 +1,199 @@
...
+        }
+
+        @Override
+        public boolean isRepeatable() {
+            return false;
+        }
+
+        @Override
+        public boolean isChunked() {
+            return true;
```
iamdanfox: Out of curiosity, are there ever worlds where we explicitly wouldn't want chunked? Do people just do it when you know the request size and want to save the bytes in between each chunk?

dialogue-hc4-client/src/main/java/com/palantir/dialogue/hc4/ApacheHttpClientBlockingChannel.java
iamdanfox: ah cool sorry misread as request 🤦‍♂

dialogue-hc4-client/src/main/java/com/palantir/dialogue/hc4/ApacheHttpClientBlockingChannel.java
carterkozak: That's correct. There's an optimization where we could internally buffer data, and if we hit the end of content before something like 8k or 16k bytes we could send a content-length request.
We already have relatively large buffers configured at the OS level so we're fine writing to the socket, we just can't walk back to the headers and add a content-length.

dialogue-httpurlconnection-client/src/main/java/com/palantir/dialogue/httpurlconnection/HttpUrlConnectionChannels.java
```diff
@@ -0,0 +1,49 @@
...
+import com.palantir.tritium.metrics.registry.TaggedMetricRegistry;
+import java.net.MalformedURLException;
+import java.net.URL;
+
+/** Simple Channel factory which produces channels based on HttpURLConnection. */
+public final class HttpUrlConnectionChannels {
+
+    private HttpUrlConnectionChannels() {}
+
+    public static Channel create(ClientConfiguration conf, UserAgent baseAgent, TaggedMetricRegistry metrics) {
```
iamdanfox, line 34: Low importance as I think this will be used for benchmarking over prod usage, but might be worth doing the same ClientConf preconditions here?

Kinda spicy that we literally can't even control cipher suites...

settings.gradle
```diff
@@ -1,8 +1,11 @@
 rootProject.name = 'dialogue'
 
+include 'dialogue-blocking-channels'
 include 'dialogue-client-test-lib'
 include 'dialogue-core'
 include 'dialogue-example'
+include 'dialogue-httpurlconnection-client'
+include 'dialogue-hc4-client'
 include 'dialogue-java-client'
```
iamdanfox, line 9: I know it's not this PR, but could we rename this to `dialogue-java9-client` as the `-java` suffix is a bit overloaded in conjure-land

settings.gradle
```diff
@@ -1,8 +1,11 @@
 rootProject.name = 'dialogue'
 
+include 'dialogue-blocking-channels'
 include 'dialogue-client-test-lib'
 include 'dialogue-core'
 include 'dialogue-example'
+include 'dialogue-httpurlconnection-client'
+include 'dialogue-hc4-client'
```
iamdanfox: dialogue-apache-hc4? seems a tad over concise for anyone who doesn't know what HttpComponents4 are

settings.gradle
carterkozak: We can do that separately. I think email-service will need to be updated.

iamdanfox (approved):

settings.gradle
carterkozak: I thought hc stood for HttpClient, though that name is overloaded with java these days ;-)

Feign calls it `httpclient` https://github.com/OpenFeign/feign/tree/master/httpclient
As does dropwizard-metrics: https://metrics.dropwizard.io/3.1.0/manual/httpclient/

+1 for `apache-hc4` to disambiguate from other clients.

dialogue-httpurlconnection-client/src/main/java/com/palantir/dialogue/httpurlconnection/HttpUrlConnectionChannels.java
carterkozak: Perhaps in a follow-up, I have a WIP to move some of the validation to Channels.create we pick the node selection strategy, qos, etc.

I'll open a PR to update the readme with a description of each of our current clients and describe our confidence (e.g. don't use java9 or httpurlconnection)

PR #337, opened 2020-02-16
[break][improvement] RequestBody.length uses OptionalLong

carterkozak:
## Before this PR
`Optional<Long>`

## After this PR
`OptionalLong` is a more specific type.


carterkozak:
I think this will have merge conflicts with the httpurlclient/hc4 channel impls.

ferozco (approved):

PR #338, opened 2020-02-16
Use default port if one is not provided

ferozco:
## Before this PR
We would fail to execute a request if the provided uri did not include an explicit port number

## After this PR
We fallback to the protocol's default port number if one is not provided

### DIFF
--- a/dialogue-java-client/src/main/java/com/palantir/dialogue/HttpChannel.java
+++ b/dialogue-java-client/src/main/java/com/palantir/dialogue/HttpChannel.java
@@ -58,7 +58,7 @@ public final class HttpChannel implements Channel {
                 null == Strings.emptyToNull(baseUrl.getUserInfo()), "baseUrl user info must be empty");
         this.baseUrl = UrlBuilder.withProtocol(baseUrl.getProtocol())
                 .host(baseUrl.getHost())
-                .port(baseUrl.getPort());
+                .port(getPortOrDefault(baseUrl));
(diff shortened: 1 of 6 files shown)


dialogue-java-client/src/main/java/com/palantir/dialogue/HttpChannel.java
```diff
@@ -58,7 +58,7 @@ private HttpChannel(HttpClient client, URL baseUrl, Duration requestTimeout) {
                 null == Strings.emptyToNull(baseUrl.getUserInfo()), "baseUrl user info must be empty");
         this.baseUrl = UrlBuilder.withProtocol(baseUrl.getProtocol())
                 .host(baseUrl.getHost())
-                .port(baseUrl.getPort());
+                .port(getPortOrDefault(baseUrl));
```
carterkozak: Can we build this into the urlbuilder?

ferozco: sure, at that point we might as well put all of the logic for parsing a URL in URLBuilder

carterkozak: That would be fantastic. I think the api is a bit dangerous currently because it’s mutable and we must remember to call newBuilder before mutating the instance for each request.

Perhaps initialize from a url, and render given a request and endpoint? That should simplify all our clients.

carterkozak:
Thanks @ferozco :+1:

PR #342, opened 2020-02-16
convert client metrics to metric schema

ferozco:
## Before this PR
We had client metrics defined in code under `client` namespace. 

## After this PR
==COMMIT_MSG==
Migrate client metrics to metric schema and move the metrics to `dialogue.client` namespace to avoid conflict with existing clients
==COMMIT_MSG==

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/InstrumentedChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/InstrumentedChannel.java
@@ -25,8 +25,6 @@ import com.palantir.dialogue.Channel;
 import com.palantir.dialogue.Endpoint;
 import com.palantir.dialogue.Request;
 import com.palantir.dialogue.Response;
-import com.palantir.tritium.metrics.registry.MetricName;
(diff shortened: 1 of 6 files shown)


dialogue-core/src/main/java/com/palantir/dialogue/core/InstrumentedChannel.java
```diff
@@ -66,18 +60,11 @@ public void onFailure(Throwable _throwable) {
 
                     private void record(Endpoint endpoint) {
                         long micros = stopwatch.elapsed(TimeUnit.MICROSECONDS);
-                        registry.timer(name(endpoint.serviceName())).update(micros, TimeUnit.MICROSECONDS);
+                        metrics.response(endpoint.serviceName()).update(micros, TimeUnit.MICROSECONDS);
```
carterkozak, line 63: rambling: I like the way conjure-undertow allows us to build this kind of data into the Endpoint, I wonder if there's a clean way we could do the same thing in dialogue.

carterkozak (approved):
thanks!

Issue #340, opened 2020-02-16
Provide a channel factory API to transparently test different client implementations


carterkozak:
## What happened?

Our different client implementations require different pieces of data to be used.

## What did you want to happen?

We should provide a channel factory (Client?) API to create channels from a ClientConfiguration + metric registry so we can transparently drop in alternative implementations.

carterkozak:
Channel factory could be named `DialogueClient`.

ferozco:
Closed by #492

PR #344, opened 2020-02-17
Fix dialogue channel tracing

carterkozak:
## Before this PR

Previously if no active trace existed, outgoing requests did not
contain trace IDs.
Normalized all tests to use Channels.create to validate all
client implementations handle tracing correctly. Note that
the jdk9 client was tested both with and without the Channels
wrapper, now it is only tested with.
We never expect channels to be used without the wrapper, so it's
more important to test the wrapped functionality than unwrapped.

## After this PR
==COMMIT_MSG==
Fix dialogue channel tracing. Previously if no active trace existed, outgoing requests did not contain trace IDs.
==COMMIT_MSG==


dialogue-okhttp-client/src/test/java/com/palantir/dialogue/OkHttpChannelTest.java
```diff
@@ -32,6 +36,9 @@ Channel createChannel(URL baseUrl) {
                 .newBuilder()
                 .dispatcher(new Dispatcher(Executors.newSingleThreadExecutor()))
                 .build();
-        return OkHttpChannel.of(client, baseUrl);
+        return Channels.create(
```
iamdanfox, line 39: At some point I'd love to try and condense the various possible public API methods to make a single idiot-proof entry point for dialogue

carterkozak: Agreed, I filed #340 yesterday which should cover this :-)

iamdanfox (approved):

carterkozak:
dozer?

PR #345, opened 2020-02-17
Delete unused classes from dialogue-target

carterkozak:
==COMMIT_MSG==
Delete unused classes from dialogue-target
==COMMIT_MSG==

## Possible downsides?
could be used somewhere?


carterkozak:
Not calling this a break because nothing touches these classes. Any use would be unrelated to dialogue.

iamdanfox (approved):

PR #346, opened 2020-02-17
Utility to safely build URLs

carterkozak:
Deduplicate client code, and isolate unsafety around UrlBuilder
mutability.

## After this PR
==COMMIT_MSG==
Utility to safely build URLs
==COMMIT_MSG==


dialogue-core/src/main/java/com/palantir/dialogue/core/BaseUrl.java
```diff
@@ -0,0 +1,48 @@
...
+ */
+
+package com.palantir.dialogue.core;
+
+import com.palantir.dialogue.Endpoint;
+import com.palantir.dialogue.Request;
+import com.palantir.dialogue.UrlBuilder;
+import java.net.URL;
+
+/** Convenience utility around {@link UrlBuilder}. */
```
markelliot, line 40: Why not add the render method to UrlBuilder?

carterkozak: I’m worried about introducing bugs that mutate the urlBuilder state

carterkozak: Method on the builder is probably fine but it adds logic to the generator target rather than the runtime, and we still risk mutation

markelliot: Sure...it just seems strange to fully encapsulate one utility class inside another. After this change UrlBuilder wouldn't need to be public, right?

carterkozak: The generated Endpoint objects have a method that consumes a UrlBuilder. I think that object should become an interface to further reduce logic in the generator target API, but that can occur separately.

ferozco (commented):
:+1:  Changing UrlBuilder from a class to interface is an ABI break but it shouldn't be an issue. Now might be the right time to make any breaks that require changes to the codegen

carterkozak:
:+1: policy bot?

carterkozak:
:robot:

ferozco:
👍

PR #348, opened 2020-02-17
[test-only] Simulate dialogue clients against a DeterministicScheduler

iamdanfox:
## Before this PR

I want to conclusively be able to demonstrate all the bits of our RPC stack is expected to behave under various kinds of load, but at the moment we just have to eyeball code and try to construct the expected behaviour in our heads.

## After this PR
==COMMIT_MSG==
A new `:simulation` project runs dialogue calls in a DeterministicScheduler and records the results.
==COMMIT_MSG==

The tests run _instantly_ and are fully deterministic and reproducible :D

Example chart:

sample

Sorry this is such a big PR - I'd recommend just cloning it and running SimulationTest to get a sense of what happens, you can also fiddle with the numbers.


dialogue-core/src/main/java/com/palantir/dialogue/core/BlacklistingChannel.java
```diff
@@ -25,20 +25,32 @@
...
 /**
  * Delegates execution to the given {@link LimitedChannel} until a failure is seen, at which point the channel will
  * return {@link Optional#empty()} until the given {@link Duration} has elapsed.
+ *
+ * When the blacklisting {@link Duration} has elapsed, the channel enters a 'probation' period, where a few requests
+ * are allowed to be sent off - all of which must be successful for probation to be passed and the channel fully
+ * unblacklisted. Without this functionality, hundreds of requests could be sent to a still-broken
+ * server before the first of them returns and tells us it's still broken.
  */
 final class BlacklistingChannel implements LimitedChannel {
```
iamdanfox: moving these changes to a dedicated PR: https://github.com/palantir/dialogue/pull/349

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -26,12 +26,14 @@
...
 import java.util.concurrent.Executor;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.Function;
 import java.util.function.Supplier;
 
 /**
- * Retries calls to the underlying channel upon failure.
+ * Immediately retries calls to the underlying channel upon failure.
  */
 final class RetryingChannel implements Channel {
```
iamdanfox: Dedicated PR for these changes https://github.com/palantir/dialogue/pull/350

dialogue-core/src/main/java/com/palantir/dialogue/core/ConcurrencyLimitedChannel.java
```diff
@@ -42,22 +42,24 @@
...
     private static final Void NO_CONTEXT = null;
+    private static final Supplier<Long> SYSTEM_NANOTIME = System::nanoTime;
 
     private final Channel delegate;
-    private final LoadingCache<Endpoint, Limiter<Void>> limiters;
+    private final LoadingCache<Endpoint, SimpleLimiter<Void>> limiters;
 
     @VisibleForTesting
-    ConcurrencyLimitedChannel(Channel delegate, Supplier<Limiter<Void>> limiterSupplier) {
+    ConcurrencyLimitedChannel(Channel delegate, Supplier<SimpleLimiter<Void>> limiterSupplier) {
```
markelliot: maybe split these changes to a separate PR?

dialogue-core/src/main/java/com/palantir/dialogue/core/GenericRefreshingChannel.java
```diff
@@ -0,0 +1,65 @@
...
+import com.palantir.conjure.java.api.config.service.ServicesConfigBlock;
+import com.palantir.dialogue.Channel;
+import com.palantir.dialogue.Endpoint;
+import com.palantir.dialogue.Request;
+import com.palantir.dialogue.Response;
+import com.palantir.logsafe.exceptions.SafeIllegalStateException;
+import java.util.Optional;
+import java.util.function.Supplier;
+
+final class GenericRefreshingChannel implements Channel {
```
markelliot: split this change to a separate PR?

dialogue-core/src/main/java/com/palantir/dialogue/core/ConcurrencyLimitedChannel.java
iamdanfox: This was just to allow me to log the permits - have reverted as I've deleted the corresponding logging

simulation/src/test/resources/live_reloading[CONCURRENCY_LIMITER].txt
```diff
@@ -0,0 +1 @@
+success=58.3%	client_mean=PT0.7228S      	server_cpu=PT4M49.12S     	received=400/400	codes={200=233, 500=167}
```
Comment (iamdanfox) [inline simulation/src/test/resources/live_reloading[CONCURRENCY_LIMITER].txt:1]: Seems like github hides the pngs by default - gotta click the little "Display the rich diff" thing:

image

dialogue-core/src/main/java/com/palantir/dialogue/core/ConcurrencyLimitedChannel.java
```diff
@@ -42,22 +42,24 @@
  */
 final class ConcurrencyLimitedChannel implements LimitedChannel {
     private static final Void NO_CONTEXT = null;
+    private static final Supplier<Long> SYSTEM_NANOTIME = System::nanoTime;
```
carterkozak: LongSupplier or a custom interface

dialogue-core/src/main/java/com/palantir/dialogue/core/GenericRefreshingChannel.java
iamdanfox: Pulled out here: https://github.com/palantir/dialogue/pull/352

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
```diff
@@ -114,8 +117,12 @@ private void onCompletion() {
      * Try to schedule as many tasks as possible. Called when requests are submitted and when they complete.
      */
     private void schedule() {
+        int numScheduled = 0;
         while (scheduleNextTask()) {
-            // Do nothing
+            numScheduled++;
+        }
+        if (numScheduled > 1) {
```
carterkozak:
```suggestion
        if (log.isDebugEnabled()) {
```
zero and one are also helpful data points imo.

simulation/src/main/java/com/palantir/dialogue/core/Benchmark.java
```diff
@@ -0,0 +1,251 @@
...
+                        .successPercentage(
+                                Math.round(statusCodes.getOrDefault("200", 0) * 1000d / requestsStarted[0]) / 10d)
+                        .numSent(requestsStarted[0])
+                        .numReceived(responsesReceived[0])
+                        .build(),
+                MoreExecutors.directExecutor());
+    }
+
+    private Stream<ScheduledRequest> infiniteRequests(Duration interval) {
+        return Stream.iterate(0, current -> current + 1).map(number -> {
```
carterkozak: nit: IntStream. 2m requests isn't a ton, may be worth using longs here?

simulation/src/main/java/com/palantir/dialogue/core/Benchmark.java
```diff
@@ -0,0 +1,251 @@
...
+
+        @Value.Default
+        default Endpoint endpoint() {
+            return ENDPOINT;
+        }
+    }
+
+    private static Request constructRequest(int number) {
+        return Request.builder()
+                .putHeaderParams("X-B3-TraceId", "req-" + number)
```
carterkozak: `X-B3-TraceId` overloading is a bit funky, perhaps we should either use the existing tracing channels with random IDs, or use a simulation-specific header?

simulation/src/main/java/com/palantir/dialogue/core/HistogramChannel.java
```diff
@@ -0,0 +1,51 @@
...
+import com.codahale.metrics.SlidingTimeWindowArrayReservoir;
+import com.google.common.util.concurrent.ListenableFuture;
+import com.google.common.util.concurrent.MoreExecutors;
+import com.palantir.dialogue.Channel;
+import com.palantir.dialogue.Endpoint;
+import com.palantir.dialogue.Request;
+import com.palantir.dialogue.Response;
+import java.util.concurrent.TimeUnit;
+
+final class HistogramChannel implements Channel {
```
carterkozak, line 29: Could we use our standard instrumented channel with an AbstractTaggedMetricRegistry implementation that overrides `timerSupplier` and uses a custom reservoir supplier? Would be nice to reinforce that our metrics are good enough, but may not be worthwhile here.

simulation/src/test/java/com/palantir/dialogue/core/Strategy.java
```diff
@@ -0,0 +1,133 @@
...
+    static Channel dontTolerateLimits(LimitedChannel limitedChannel) {
+        return new Channel() {
+            @Override
+            public ListenableFuture<Response> execute(Endpoint endpoint, Request request) {
+                Optional<ListenableFuture<Response>> future = limitedChannel.maybeExecute(endpoint, request);
+                if (future.isPresent()) {
+                    return future.get();
+                }
+
+                return Futures.immediateFailedFuture(new SafeRuntimeException("limited channel says no :("));
```
carterkozak: I think queued channel should do this #351

simulation/src/test/java/com/palantir/dialogue/core/Strategy.java
```diff
@@ -0,0 +1,133 @@
...
+
+            @Override
+            public String toString() {
+                return delegate.toString();
+            }
+        };
+    }
+
+    /** This is an alternative to the {@link com.palantir.dialogue.core.QueuedChannel}. */
+    static Channel dontTolerateLimits(LimitedChannel limitedChannel) {
```
carterkozak: Is this used?

simulation/src/main/java/com/palantir/dialogue/core/HistogramChannel.java
iamdanfox: I will happily do that - the only reason I didn't initially was there are a lot of methods on TaggedMetricRegistry and it was gonna inflate the line count of this PR quite severely!

simulation/src/test/java/com/palantir/dialogue/core/Strategy.java
iamdanfox: Yeah I was using this as a substitute for the queued channel, but it's not being used right now, so will delete and bring it back when needed.

simulation/src/main/java/com/palantir/dialogue/core/Benchmark.java
iamdanfox: good point - I'll probably want to visualize traces for these at some point, have switched to a different constant.

dialogue-core/src/main/java/com/palantir/dialogue/core/ConcurrencyLimitedChannel.java
iamdanfox: Turns out the netflix builder requires a `Supplier<Long>`, so I've used the same caffeine Ticker interface we use everywhere else.

carterkozak (approved):
I haven't reviewed this as closely as changes to prod code, but the output is insightful and the implementation looks reasonable.

iamdanfox:
Thanks carter, I'm gonna update my other PRs with this code and hopefully have some pretty graphs to show!

PR #349, opened 2020-02-17
BlacklistingChannel probation and per-endpoint info

iamdanfox:
## Before this PR

- Blacklisting was an all or nothing thing (so on endpoint could stop traffic to an entire host)
- After blacklisting, we'd send a deluge of requests to the newly unblacklisted host, which might still be broken, resulting in lots of failures.
- Also 500s wouldn't result in blacklisting

(It's way easier to review this when I have the `:simulation` graphs to demonstrate what's broken)

## After this PR
==COMMIT_MSG==
BlacklistingChannel keeps track of per-endpoint information and conservatively ramps up.
==COMMIT_MSG==


dialogue-core/src/main/java/com/palantir/dialogue/core/BlacklistingChannel.java
```diff
@@ -47,35 +59,119 @@
...
     @Override
     public Optional<ListenableFuture<Response>> maybeExecute(Endpoint endpoint, Request request) {
-        if (isBlacklisted.getIfPresent(KEY) != null) {
-            return Optional.empty();
-        } else {
-            return delegate.maybeExecute(endpoint, request)
-                    .map(future -> DialogueFutures.addDirectCallback(future, new BlacklistingCallback()));
+        BlacklistState state = perEndpointBlacklistState.getIfPresent(endpoint);
+
+        if (state != null) {
```
carterkozak: nit: unnecessary null check

dialogue-core/src/main/java/com/palantir/dialogue/core/BlacklistingChannel.java
```diff
@@ -47,35 +59,119 @@
...
         this.delegate = delegate;
-        this.isBlacklisted = Caffeine.newBuilder()
-                .maximumSize(1)
-                .expireAfterWrite(duration)
-                .ticker(ticker)
-                .build();
+        this.duration = duration;
+        this.ticker = ticker;
+        this.perEndpointBlacklistState =
+                Caffeine.newBuilder().maximumSize(1000).ticker(ticker).build();
```
carterkozak, line 65: per-endpoint state feels a bit odd here. At the channel level I think we'd want to track health of the channel as a whole. I wonder if we could invert the relationship between channel and Endpoint to allow endpoints to be wrapped similarly to conjure-undertow?

dialogue-core/src/main/java/com/palantir/dialogue/core/BlacklistingChannel.java
iamdanfox: the getIfPresent call on L70 returns null if a channel is not blacklisted, so this check is important to let normal calls go through! (Esp because this isn't a loadingcache, it's just a regular one).

carterkozak: Null is neither an instance of BlacklistUntil nor Probation, so it should be fine, right? I may be misreading.

dialogue-core/src/main/java/com/palantir/dialogue/core/BlacklistingChannel.java
```diff
@@ -47,35 +59,119 @@
...
+        }
+    }
+
+    @Override
+    public String toString() {
+        return "BlacklistingChannel{" + delegate + '}';
+    }
+
+    // I wish java had union types
+    interface BlacklistState {}
```
ferozco, line 146: derive4j is your friend here

dialogue-core/src/main/java/com/palantir/dialogue/core/BlacklistingChannel.java
iamdanfox: Is this something that you want to nail down before moving forwards with this PR? (We've already got a `Cache<Endpoint, Limiter>` in our concurrency limiters...)

This BlacklistingChannel is a key ingredient for an alternative node selection strategy I've been working on, but I'd really like to land this PR as it's not super usable as is.

dialogue-core/src/main/java/com/palantir/dialogue/core/BlacklistingChannel.java
iamdanfox: So I actually really don't want to have the extra faff and tooling setup of derive4j - can you forgive these two purely implementation-detail classes?

dialogue-core/src/main/java/com/palantir/dialogue/core/BlacklistingChannel.java
iamdanfox: ah yep for some reason I thought null would have fallen through those instanceof checks

ferozco:
👍

iamdanfox:
FWIW Channels#create doesn't currently use BlacklistingChannel - it was a piece of floating dead code... but maybe not for long ;)

carterkozak:
If we want to use blacklisting channel we’ll need to wire in eventing to attempt to flush the queue when channels are no longer blacklisted.

-ck

iamdanfox:
Ah right fair enough, as the concurrency limiter's behaviour only changes when a request goes in or comes back (which triggers the queued thing to schedule), whereas since blacklisting is time based, we could have a spike of requests that all get queued and if no further requests come in then they'd just sit on the queue forever.

PR #350, opened 2020-02-17
RetryingChannel retries 429/503s

iamdanfox:
~_Would love to merge simulation charts before this: https://github.com/palantir/dialogue/pull/348, as we'll be able to see the difference in the graphs!_~

# [GRAPHS](https://github.com/palantir/dialogue/blob/dfox/retrying/simulation/src/test/resources/report.md)

## Before this PR

A single blip of brokenness would be passed on to clients, resulting in user-facing impact.

## After this PR
==COMMIT_MSG==
RetryingChannel retries 429/503s
==COMMIT_MSG==

## Possible downsides?
~this might have weird effects if a request is not idempotent (e.g. "sendEmail") and the server breaks halfway though servicing the request (i.e. multiple emails could get sent).~

In a future PR, I think we should consider retrying 500s too.


dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -54,34 +56,55 @@
...
-        Supplier<ListenableFuture<Response>> callSupplier = () -> delegate.execute(endpoint, request);
-        FutureCallback<Response> retryer = new RetryingCallback<>(callSupplier, future);
-        Futures.addCallback(callSupplier.get(), retryer, DIRECT_EXECUTOR);
+        Function<Integer, ListenableFuture<Response>> callSupplier = attempt -> {
+            return delegate.execute(
+                    endpoint,
+                    Request.builder()
+                            .from(request)
+                            .putHeaderParams(
+                                    "X-B3-TraceId", request.headerParams().get("X-B3-TraceId") + "-attempt-" + attempt)
```
carterkozak: What’s going on here?

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -54,34 +56,55 @@
...
         }
 
         @Override
-        public void onSuccess(T result) {
+        public void onSuccess(Response result) {
+            // this condition should really match the BlacklistingChannel so that we don't hit the same host twice in
+            // a row
+            if (result.code() == 503 || result.code() == 500) {
+                retryOrFail(() -> new SafeRuntimeException("Retries exhausted"));
+                return;
```
carterkozak, line 91: Response body leak, we need to close it

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -54,34 +56,50 @@
...
+            }
+
             delegate.set(result);
```
carterkozak: Another leak if this returns false (I think I filed an issue for this one)

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -54,34 +56,50 @@
...
         }
 
         @Override
         public void onFailure(Throwable throwable) {
-            if (failures.incrementAndGet() < maxRetries) {
-                Futures.addCallback(runnable.get(), this, DIRECT_EXECUTOR);
+            retryOrFail(() -> throwable);
+        }
+
+        private void retryOrFail(Supplier<Throwable> throwable) {
```
carterkozak: I think these failures should always be logged at info or warn level, so we can do away with the supplier

markelliot:
for composition reasons, these failure types probably ought to be encoded by a different channel

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -54,34 +56,50 @@
...
 
-        Supplier<ListenableFuture<Response>> callSupplier = () -> delegate.execute(endpoint, request);
-        FutureCallback<Response> retryer = new RetryingCallback<>(callSupplier, future);
-        Futures.addCallback(callSupplier.get(), retryer, DIRECT_EXECUTOR);
+        Function<Integer, ListenableFuture<Response>> callSupplier = attempt -> {
+            // TODO(dfox): include retry number in the request somehow
+            return delegate.execute(endpoint, request);
+        };
+        FutureCallback<Response> retryer = new RetryingCallback(callSupplier, future);
+        Futures.addCallback(callSupplier.apply(0), retryer, DIRECT_EXECUTOR);
```
ferozco, line 68:
```suggestion
        return DialogueFutures.addDirectCallback(callSupplier.apply(0), retryer);
```

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -54,34 +56,50 @@
...
             this.runnable = runnable;
             this.delegate = delegate;
         }
 
         @Override
-        public void onSuccess(T result) {
+        public void onSuccess(Response result) {
+            // this condition should really match the BlacklistingChannel so that we don't hit the same host twice in
+            // a row
+            if (result.code() == 503 || result.code() == 500) {
```
ferozco: why not retry 429's as well?

iamdanfox: I think I'd like to land that in lock-step PR with a change to the BlacklistingChannel, as otherwise you might retry on the same host immediately!

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
iamdanfox: Ah this was some leftover from debugging odd behaviour - removed!

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
markelliot: this is why, per my comment, the location of this check is likely wrong

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
iamdanfox: If it's OK with you I think I'd actually prefer to just stick with the vanilla guava - I find it kinda reassuring that there's no magic going on under the hood

simulation/src/test/java/com/palantir/dialogue/core/SimulationTest.java
```diff
@@ -418,8 +420,17 @@ public void after() throws IOException {
         Path txt = Paths.get("src/test/resources/" + testName.getMethodName() + ".txt");
         String pngPath = "src/test/resources/" + testName.getMethodName() + ".png";
         String onDisk = Files.exists(txt) ? new String(Files.readAllBytes(txt), StandardCharsets.UTF_8) : "";
+
         boolean txtChanged = !longSummary.equals(onDisk);
 
+        if (System.getenv().containsKey("CI")) { // only strict on CI, locally we just overwrite
+            assertThat(onDisk)
+                    .describedAs("Run tests locally to update checked-in file: %s", txt)
+                    .isEqualTo(longSummary);
```
iamdanfox: I realized we need this otherwise PRs might forget to update the checked-in txt & png files. Got the failure I wanted here: https://circleci.com/gh/palantir/dialogue/3833

iamdanfox:
Omg I love graphs https://github.com/palantir/dialogue/blob/dfox/retrying/simulation/src/test/resources/report.md#slow_503s_then_revertconcurrency_limiterpng

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -54,34 +57,69 @@
...
             this.delegate = delegate;
         }
 
         @Override
-        public void onSuccess(T result) {
-            delegate.set(result);
+        public void onSuccess(Response response) {
+            // this condition should really match the BlacklistingChannel so that we don't hit the same host twice in
+            // a row
+            if (response.code() == 503 || response.code() == 500) {
```
iamdanfox: Thinking a bit more carefully here, I think we should possibly just match c-j-r's [old behaviour](https://github.com/palantir/conjure-java-runtime/blob/47e8727462b1f8261eaac2660038ded48d491958/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/RemotingOkHttpCall.java#L319) to minimize disruption on the rollout:

- 1xx and 2xx are considered successful
- QoShandler catches 308, 429, 503 which are eligible for retry.
- everything else goes straight into the error pipe.

We can debate the 'retry 500s' thing separately.

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
ferozco: It just seems like extra boilerplate 🤷‍♂ feels off to have a utility exactly for this and then not use it

ferozco:
👍

PR #352, opened 2020-02-17
[pure refactor] RefreshingChannelFactory exposes generic version

iamdanfox:
No behaviour changes, I just needed this for my tests.

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/RefreshingChannelFactory.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/RefreshingChannelFactory.java
@@ -95,6 +95,8 @@ public final class RefreshingChannelFactory {
          * provided by the {@code confSupplier} is. Avoids invoking the channelFactory if the config hasn't changed.
          */
         static <T> Channel create(Supplier<T> confSupplier, Function<T, Channel> channelFactory) {
+            // Beware: the memoizing composing supplier does updates in a synchronized block, which can block all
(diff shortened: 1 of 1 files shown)


dialogue-core/src/main/java/com/palantir/dialogue/core/RefreshingChannelFactory.java
```diff
@@ -73,7 +82,26 @@ private AlwaysThrowingChannel(String serviceName) {
...
+        private RefreshingChannel(Supplier<Channel> channelSupplier) {
+            this.channelSupplier = channelSupplier;
+        }
+
+        /**
+         * Returns a Channel which will be built by applying the {@code channelFactory} to whatever the latest config
+         * provided by the {@code confSupplier} is. Avoids invoking the channelFactory if the config hasn't changed.
+         */
+        static <T> Channel create(Supplier<T> confSupplier, Function<T, Channel> channelFactory) {
+            Supplier<Channel> channelSupplier = new MemoizingComposingSupplier<>(confSupplier, channelFactory);
```
iamdanfox, line 100: FWIW I don't think this is the ideal long-term solution, because we've it a bug with a similar memoizing supplier that contained a `synchronized` block internally in sls-logging.  Just making the minimum changes necessary to unblock my simulations.

carterkozak: Mind adding a comment along these lines?

carterkozak (approved):

PR #354, opened 2020-02-17
JavaChannels respects the client configuration `enableHttp2` flag

carterkozak:
This changes the default to disabled as the client fails to make requests
to a standard witchcraft server otherwise.

==COMMIT_MSG==
JavaChannels respects the client configuration `enableHttp2` flag
==COMMIT_MSG==


carterkozak:
> This changes the default to disabled as the client fails to make requests
to a standard witchcraft server otherwise.

~~It's also timing out using http/1.1, but less frequently...~~
Disregard, that was still using http/2 with http/2 enabled on the server.

iamdanfox (approved):

PR #355, opened 2020-02-17
Implement an OkHttpChannels factory class

carterkozak:
## Before this PR
Difficult to make an okhttp based channel.

## After this PR
==COMMIT_MSG==
Implement an OkHttpChannels factory class based on CJR without duplicated functionality.
==COMMIT_MSG==

## Possible downsides?
We don't really want to use okhttp, but it's helpful to have a class to compare with.

### DIFF
--- a/dialogue-okhttp-client/src/main/java/com/palantir/dialogue/OkHttpChannels.java
+++ b/dialogue-okhttp-client/src/main/java/com/palantir/dialogue/OkHttpChannels.java
@@ -23,8 +23,10 @@ import com.palantir.conjure.java.api.config.service.BasicCredentials;
 import com.palantir.conjure.java.api.config.service.UserAgent;
 import com.palantir.conjure.java.client.config.CipherSuites;
 import com.palantir.conjure.java.client.config.ClientConfiguration;
+import com.palantir.conjure.java.client.config.NodeSelectionStrategy;
 import com.palantir.dialogue.core.Channels;
 import com.palantir.logsafe.Preconditions;
+import com.palantir.logsafe.SafeArg;
 import com.palantir.logsafe.exceptions.SafeIllegalArgumentException;
 import com.palantir.tritium.metrics.registry.TaggedMetricRegistry;
 import java.net.MalformedURLException;
@@ -104,6 +106,11 @@ public final class OkHttpChannels {
         Preconditions.checkArgument(
                 config.serverQoS() == ClientConfiguration.ServerQoS.AUTOMATIC_RETRY,
                 "Propagating QoS exceptions is not supported");
+        if (config.nodeSelectionStrategy() != NodeSelectionStrategy.ROUND_ROBIN) {
+            log.warn(
+                    "Dialogue currently only supports ROUND_ROBIN node selection strategy. {} will be ignored",
+                    SafeArg.of("requestedStrategy", config.nodeSelectionStrategy()));
+        }
         OkHttpClient.Builder builder = new OkHttpClient()
                 .newBuilder()
                 .dispatcher(dispatcher)
--- a/dialogue-okhttp-client/src/main/java/com/palantir/dialogue/OkHttpChannels.java
+++ b/dialogue-okhttp-client/src/main/java/com/palantir/dialogue/OkHttpChannels.java
@@ -24,6 +24,7 @@ import com.palantir.conjure.java.api.config.service.UserAgent;
 import com.palantir.conjure.java.client.config.CipherSuites;
 import com.palantir.conjure.java.client.config.ClientConfiguration;
 import com.palantir.dialogue.core.Channels;
+import com.palantir.logsafe.Preconditions;
 import com.palantir.logsafe.exceptions.SafeIllegalArgumentException;
 import com.palantir.tritium.metrics.registry.TaggedMetricRegistry;
 import java.net.MalformedURLException;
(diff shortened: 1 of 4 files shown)


dialogue-okhttp-client/src/main/java/com/palantir/dialogue/OkHttpChannels.java
```diff
@@ -0,0 +1,161 @@
...
+
+    public static Channel create(ClientConfiguration config, UserAgent baseAgent, TaggedMetricRegistry metrics) {
+        OkHttpClient.Builder builder = new OkHttpClient()
+                .newBuilder()
+                .dispatcher(dispatcher)
+                .connectionPool(connectionPool)
+                .followRedirects(false) // We implement our own redirect logic.
+                .sslSocketFactory(config.sslSocketFactory(), config.trustManager())
+                // timeouts
+                // Note that Feign overrides OkHttp timeouts with the timeouts given in FeignBuilder#Options if given,
```
iamdanfox: probably don't need the feign mention here?

dialogue-okhttp-client/src/main/java/com/palantir/dialogue/OkHttpChannels.java
```diff
@@ -0,0 +1,161 @@
...
+import okhttp3.ConnectionSpec;
+import okhttp3.Credentials;
+import okhttp3.Dispatcher;
+import okhttp3.OkHttpClient;
+import okhttp3.Protocol;
+import okhttp3.TlsVersion;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public final class OkHttpChannels {
```
iamdanfox, line 48: Thoughts on the naming convention `DialogueOkhttpClient`, as then the main method would be `DialogueOkhttpClient#create` ?

dialogue-okhttp-client/src/main/java/com/palantir/dialogue/OkHttpChannels.java
```diff
@@ -0,0 +1,161 @@
...
+
+    private OkHttpChannels() {}
+
+    public static Channel create(ClientConfiguration config, UserAgent baseAgent, TaggedMetricRegistry metrics) {
+        OkHttpClient.Builder builder = new OkHttpClient()
+                .newBuilder()
+                .dispatcher(dispatcher)
+                .connectionPool(connectionPool)
+                .followRedirects(false) // We implement our own redirect logic.
+                .sslSocketFactory(config.sslSocketFactory(), config.trustManager())
```
iamdanfox, line 119: have we lost the KeepAliveSslSocketFactory here? Maybe we should just make it public in c-j-r?

carterkozak: That's correct, this doesn't record client tls handshake metrics or set keepalive. Agree that we should be consistent across all clients.

dialogue-okhttp-client/src/main/java/com/palantir/dialogue/OkHttpChannels.java
```diff
@@ -0,0 +1,161 @@
...
+            1000,
+            // Most servers use a one minute keepalive for idle connections, by using a shorter keepalive on
+            // clients we can avoid race conditions where the attempts to reuse a connection as the server
+            // closes it, resulting in unnecessary I/O exceptions and retrial.
+            55,
+            TimeUnit.SECONDS);
+
+    private OkHttpChannels() {}
+
+    public static Channel create(ClientConfiguration config, UserAgent baseAgent, TaggedMetricRegistry metrics) {
```
iamdanfox, line 100: so I know there a crap ton of options in a ClientConfiguration, but we dont' support all of them in this method. Could we do a Preconditions.checkState to verify the ones we don't support are empty / set to our expected defaults? e.g. `Optional<HostAndPort> meshProxy();`

iamdanfox: All shall we just start with a builder straight away here? The overloads in c-j-r break my heart.

carterkozak: I think the uniformity between our `XChannels.create` methods is worth something, I'd prefer to define args in dialogue-core and update them all in lock-step.

carterkozak: > Could we do a Preconditions.checkState to verify the ones we don't support are empty / set to our expected defaults?

Do we want to throw if nodeSelectionStrategy is not ROUND_ROBIN? I can't imagine anyone has that set.

dialogue-okhttp-client/src/main/java/com/palantir/dialogue/OkHttpChannels.java
carterkozak: I would prefer to update all of the clients together, we should provide a standard interface for these.

dialogue-okhttp-client/src/main/java/com/palantir/dialogue/OkHttpChannels.java
iamdanfox: There's a few users of it actually: https://internal-github/search?q=org%3Afoundry+ROUND_ROBIN&type=Code

Could do a log.warn otherwise?

carterkozak: Updated.

iamdanfox (approved):
LGTM - don't think the KeepAliveSocketFactory thing should be a blocker

carterkozak:
I've enabled keep-alive on the apache hc implementation in #335 :-)

PR #358, opened 2020-02-18
[test-only] simulations use tritium metric registry

iamdanfox:
## Before this PR

My simulation was keeping track of a map of metrics like an animal.

## After this PR
==COMMIT_MSG==
Simulation metrics are now tracked in a tritium TaggedMetricRegistry implementation, which is hooked up to the simulation clock.
==COMMIT_MSG==

This PR also uses this new metricregistry to better record the global number of responses returned and also verify that all of them were closed. Graphs will also display tags nicely, where previously I was doing string concatenation!

As a bonus, gradle compilation is now much snappier.


build.gradle
```diff
@@ -44,10 +44,13 @@ plugins {
 apply plugin: 'com.palantir.baseline'
 apply plugin: 'com.palantir.git-version'
 
+version System.env.CIRCLE_TAG ?: gitVersion()
```
iamdanfox: I had a 25second gradle configuration time without this: https://scans.gradle.com/s/6j5nbkrgpj54u/performance/configuration#summary-total

simulation/src/main/java/com/palantir/dialogue/core/TaggedMetrics.java
```diff
@@ -0,0 +1,58 @@
...
+import com.codahale.metrics.Counter;
+import com.codahale.metrics.Meter;
+import com.codahale.metrics.SlidingTimeWindowArrayReservoir;
+import com.codahale.metrics.Timer;
+import com.palantir.tritium.metrics.registry.AbstractTaggedMetricRegistry;
+import com.palantir.tritium.metrics.registry.MetricName;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Supplier;
+
+public final class TaggedMetrics extends AbstractTaggedMetricRegistry {
```
iamdanfox, line 29: @carterkozak here's that FLUP from the earlier PR - turned out much easier than I was expecting!

simulation/src/test/resources/report.md
```diff
@@ -1,24 +1,24 @@
...
+             one_endpoint_dies_on_each_server[CONCURRENCY_LIMITER].txt:	success=67.6%	client_mean=PT0.6S         	server_cpu=PT5M6S         	client_received=510/510	server_resps=510	codes={200=345, 500=165}
+                     one_endpoint_dies_on_each_server[ROUND_ROBIN].txt:	success=67.6%	client_mean=PT0.6S         	server_cpu=PT5M6S         	client_received=510/510	server_resps=510	codes={200=345, 500=165}
+                       simplest_possible_case[CONCURRENCY_LIMITER].txt:	success=100.0%	client_mean=PT0.7998S      	server_cpu=PT13M19.8S     	client_received=1000/1000	server_resps=1000	codes={200=1000}
+                               simplest_possible_case[ROUND_ROBIN].txt:	success=100.0%	client_mean=PT0.7998S      	server_cpu=PT13M19.8S     	client_received=1000/1000	server_resps=1000	codes={200=1000}
+                        slow_503s_then_revert[CONCURRENCY_LIMITER].txt:	success=100.0%	client_mean=PT0.129949S    	server_cpu=PT6M29.847333275S	client_received=3000/3000	server_resps=3130	codes={200=3000}
```
iamdanfox: this is an example where our RetryingChannel kicked in and resulted in servers producing responding 3130 responses, but clients just perceived 3000 green responses (as the retries were eventually successful).

simulation/src/main/java/com/palantir/dialogue/core/SimulationMetricsReporter.java
```diff
@@ -226,6 +139,18 @@ public XYChart chart(Pattern metricNameRegex) {
         return chart;
     }
 
+    private static double toDouble(Number number) {
```
carterkozak: this function can be replaced with `number.doubleValue()`

simulation/src/main/java/com/palantir/dialogue/core/TaggedMetrics.java
carterkozak: Ideally I'd prefer to use metrics from our production code, otherwise I'm worried we won't have the data to understand real use. We should be able to update existing timers like this in order to work properly with simulation: #359

build.gradle
carterkozak: Can we make these improvements in separate PRs? Unix philosophy :-)

iamdanfox: Pulled out to https://github.com/palantir/dialogue/pull/365

simulation/src/main/java/com/palantir/dialogue/core/HistogramChannel.java
```diff
@@ -46,7 +46,13 @@ public Histogram getHistogram() {
     public ListenableFuture<Response> execute(Endpoint endpoint, Request request) {
         long start = simulation.clock().read();
         ListenableFuture<Response> future = channel.execute(endpoint, request);
-        future.addListener(() -> histogram.update(simulation.clock().read() - start), MoreExecutors.directExecutor());
+        future.addListener(
+                () -> {
+                    long micros =
+                            TimeUnit.MICROSECONDS.convert(simulation.clock().read() - start, TimeUnit.NANOSECONDS);
+                    histogram.update(micros);
```
carterkozak: can we use a timer instead?

iamdanfox: Have updated to use the vanilla InstrumentedChannel!

carterkozak: Thanks!

iamdanfox:
Ok I think this is ready to go :)

carterkozak (approved):

PR #359, opened 2020-02-18
Simplify InstrumentedChannel

carterkozak:
Instead of a Guava stopwatch we use a codahale Timer.Context. This
allows the InstrumentedChannel to use the timer provided by
simulation tests.

Because the same code is executed for both success and failure
we can use the simple ListenableFuture.addListener function.

==COMMIT_MSG==
Simplify InstrumentedChannel and provide flexibility for custom clocks.
==COMMIT_MSG==

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/InstrumentedChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/InstrumentedChannel.java
@@ -16,17 +16,13 @@
 
 package com.palantir.dialogue.core;
 
-import com.google.common.base.Stopwatch;
-import com.google.common.util.concurrent.FutureCallback;
-import com.google.common.util.concurrent.Futures;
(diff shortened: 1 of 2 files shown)


dialogue-core/src/main/java/com/palantir/dialogue/core/InstrumentedChannel.java
```diff
@@ -43,28 +39,9 @@
...
-                    }
-
-                    private void record(Endpoint endpoint) {
-                        long micros = stopwatch.elapsed(TimeUnit.MICROSECONDS);
-                        metrics.response(endpoint.serviceName()).update(micros, TimeUnit.MICROSECONDS);
-                    }
-                },
-                MoreExecutors.directExecutor());
-
+        response.addListener(context::stop, MoreExecutors.directExecutor());
```
ferozco, line 44: nice!

ferozco (approved):

PR #360, opened 2020-02-18
fix #353: Response is closeable

### LINKED ISSUE #353: Response should be closable
## What happened?

We've had several bugs where response streams aren't closed.

 We can make this more obvious and easier to remediate by making the Response object closable.

carterkozak:
Implementations are expected not to throw in order to simplify
response handling. By making it easy to close responses it's more
likely consumers will actually do it.

## Before this PR
Easy to forget to close Responses, and difficult to close them.

## After this PR
The Response object is Closeable, and doesn't require special IOException handling to do.

==COMMIT_MSG==
Response is closeable
==COMMIT_MSG==

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/ContentDecodingChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/ContentDecodingChannel.java
@@ -33,6 +33,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.zip.GZIPInputStream;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Adds support for transparently requesting and decoding <code>Content-Encoding: gzip</code> responses
@@ -45,6 +47,8 @@ import java.util.zip.GZIPInputStream;
  * https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.11
  */
 final class ContentDecodingChannel implements Channel {
+
+    private static final Logger log = LoggerFactory.getLogger(ContentDecodingChannel.class);
 
     private static final String ACCEPT_ENCODING = "accept-encoding";
     private static final String CONTENT_ENCODING = "content-encoding";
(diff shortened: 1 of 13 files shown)


dialogue-core/src/main/java/com/palantir/dialogue/core/ContentDecodingChannel.java
```diff
@@ -98,9 +102,6 @@ private static Response decompress(Response input) {
 
         @Override
         public InputStream body() {
-            if (body == null) {
-                delegate.body();
-            }
```
carterkozak, line 103: :man_facepalming: I must have had a bad re-stack on the content-decoding change, sorry!

dialogue-okhttp-client/src/main/java/com/palantir/dialogue/OkHttpResponse.java
```diff
@@ -48,4 +48,9 @@ public int code() {
     public Map<String, List<String>> headers() {
         return delegate.headers().toMultimap();
     }
+
+    @Override
+    public void close() {
+        delegate.close();
```
iamdanfox, line 54: :+1: nice - this would have been a painful discovery otherwise!

dialogue-target/src/main/java/com/palantir/dialogue/Response.java
```diff
@@ -17,12 +17,13 @@
...
 
 import com.google.common.collect.ImmutableList;
+import java.io.Closeable;
 import java.io.InputStream;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 
-public interface Response {
+public interface Response extends Closeable {
```
iamdanfox, line 26: Would putting a `@MustBeClosed` annotation on this help us catch any other cases?

iamdanfox (approved):

dialogue-target/src/main/java/com/palantir/dialogue/Response.java
carterkozak: Unfortunately `MustBeClosed` is only valid for methods to require return values to be closed.

PR #364, opened 2020-02-18
Make the pin until error strategy available

iamdanfox:
# [GRAPHS](https://github.com/palantir/dialogue/blob/dfox/pin-until-error/simulation/src/test/resources/report.md)

## Before this PR

Migrating all clients in foundry to dialogue will involve changing the vast 99% of clients from pin-until-error to round robin, which is a little scary as some people may rely on warm caches to deliver adequate performance (especially on larger deployments where there are more nodes).

To avoid the switch to dialogue causing a performance regression, I think it's worth having this code available, even if we end up using something better.

## After this PR
==COMMIT_MSG==
A new `PinUntilErrorChannel` provides the same functionality as conjure-java-runtime's default NodeSelectionStrategy.
==COMMIT_MSG==

Note - the original [UrlSelectorImpl](https://github.com/palantir/conjure-java-runtime/blob/develop/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/UrlSelectorImpl.java) _also_ could mark nodes as failed and then test them before resuming. In dialogue-land, that functionality is provided by the BlacklistingChannel, so doesn't appear here.

## Possible downsides?

- this doens't handle 308 responses where a server specifies a specific node


dialogue-core/src/main/java/com/palantir/dialogue/core/PinUntilErrorChannel.java
```diff
@@ -0,0 +1,199 @@
...
+        LimitedChannel channel = channels.get(currentIndex % channels.size());
+
+        Optional<ListenableFuture<Response>> maybeFuture = channel.maybeExecute(endpoint, request);
+        if (!maybeFuture.isPresent()) {
+            OptionalInt next = incrementCurrentHost(currentIndex);
+            log.debug(
+                    "Current channel rejected request, switching to next channel",
+                    SafeArg.of("currentIndex", currentIndex),
+                    UnsafeArg.of("current", channel),
+                    SafeArg.of("nextIndex", next));
```
carterkozak: nit: if log.isDebugEnabled

I prefer guarding trace and debug statements with logsafe args to prevent allocation in the default case.

carterkozak: Do we also want to check that next is present?

iamdanfox: ~Eh, I think it's fine to get a safearg with an optionalint.empty in it because it emphasises that we got an out of order response.~
NVM i see what you mean the log line will be misleading will change.

dialogue-core/src/main/java/com/palantir/dialogue/core/PinUntilErrorChannel.java
```diff
@@ -0,0 +1,245 @@
...
+ * Upside: ensures clients will always hit warm caches
+ * Downsides:
+ * <ul>
+ *     <li>a single high volume client can result in unbalanced utilization of server nodes.
+ *     <li>when a node is restarted (and clients see 500s/connect exceptions), they will _all_ pin to a different node
+ * </ul>
+ *
+ * To alleviate the second downside, we reshuffle all nodes every 10 minutes.
+ */
+public final class PinUntilErrorChannel implements LimitedChannel {
```
ferozco: do you want to wire this up?

dialogue-core/src/main/java/com/palantir/dialogue/core/PinUntilErrorChannel.java
```diff
@@ -0,0 +1,245 @@
...
+     */
+    private OptionalInt incrementCurrentHost(int currentIndex) {
+        int next = currentIndex + 1;
+        int nextIndex = (next >= 0) ? next % channels.size() : 0; // we don't want negative indexes!
+
+        boolean saved = currentHost.compareAndSet(currentIndex, nextIndex);
+        return saved ? OptionalInt.of(nextIndex) : OptionalInt.empty(); // we've moved on already
+    }
+
+    private final class Reshuffle {
```
ferozco: would it be cleaner to provide a view over the list of channels that handles reshuffling on access thus isolating the reshuffle logic

dialogue-core/src/main/java/com/palantir/dialogue/core/PinUntilErrorChannel.java
carterkozak: This should make it easier to do: #366

dialogue-core/src/main/java/com/palantir/dialogue/core/PinUntilErrorChannel.java
iamdanfox: Just pushed a version where it's hidden behind a single abstraction, what do you think?

dialogue-core/src/main/java/com/palantir/dialogue/core/PinUntilErrorChannel.java
```diff
@@ -0,0 +1,260 @@
...
+                new ReshufflingNodeList(channels, ThreadLocalRandom.current(), System::nanoTime);
+        return new PinUntilErrorChannel(shufflingNodeList);
+    }
+
+    /**
+     * This strategy results in unbalanced server utilization, especially after an upstream restarts.
+     * @deprecated prefer {@link #pinUntilError}
+     */
+    @Deprecated
+    public static PinUntilErrorChannel pinUntilErrorWithoutReshuffle(List<LimitedChannel> channels) {
```
ferozco: why do these need to be public?

dialogue-core/src/main/java/com/palantir/dialogue/core/PinUntilErrorChannel.java
```diff
@@ -0,0 +1,260 @@
...
+    // we also add some jitter to ensure that there isn't a big spike of reshuffling every 10 minutes.
+    private static final Duration RESHUFFLE_EVERY = Duration.ofMinutes(10);
+
+    // TODO(dfox): could we make this endpoint-specific?
+    private final AtomicInteger currentHost = new AtomicInteger(0); // always positive
+    private final NodeList maybeShufflingNodeList;
+
+    @VisibleForTesting
+    PinUntilErrorChannel(NodeList nodeList) {
+        this.maybeShufflingNodeList = nodeList;
```
ferozco: s/maybeShufflingNodeList/nodeList/

dialogue-core/src/main/java/com/palantir/dialogue/core/PinUntilErrorChannel.java
```diff
@@ -0,0 +1,260 @@
...
+            }
+        }));
+    }
+
+    /**
+     * If we have some reason to think the currentIndex is bad, we want to move to the next host. This is done with a
+     * compareAndSet to ensure that out of order responses which signal information about a previous host don't kick
+     * us off a good one.
+     */
+    private OptionalInt incrementCurrentHost(int currentIndex) {
```
ferozco: maybeIncrementAndGet ? kinda wish you had parens on that i.e. maybe(IncrementAndGet)

dialogue-core/src/main/java/com/palantir/dialogue/core/PinUntilErrorChannel.java
```diff
@@ -0,0 +1,260 @@
...
+         */
+        LimitedChannel get(int index);
+    }
+
+    @VisibleForTesting
+    static final class ConstantNodeList implements NodeList {
+        private final List<LimitedChannel> channels;
+
+        ConstantNodeList(List<LimitedChannel> channels, Random random) {
+            this.channels = shuffleImmutableList(channels, random);
```
ferozco, line 147: why do we need to shuffle here?

dialogue-core/src/main/java/com/palantir/dialogue/core/PinUntilErrorChannel.java
```diff
@@ -0,0 +1,260 @@
...
+        }
+
+        @Override
+        public LimitedChannel get(int index) {
+            return channels.get(index % channels.size());
+        }
+    }
+
+    @VisibleForTesting
+    static final class ReshufflingNodeList implements NodeList {
```
ferozco, line 162: this is nice :)

dialogue-core/src/main/java/com/palantir/dialogue/core/PinUntilErrorChannel.java
```diff
@@ -0,0 +1,260 @@
...
+    }
+
+    /** Returns a new shuffled list, without mutating the input list (which may be immutable). */
+    private static <T> ImmutableList<T> shuffleImmutableList(List<T> sourceList, Random random) {
+        List<T> mutableList = new ArrayList<>(sourceList);
+        Collections.shuffle(mutableList, random);
+        return ImmutableList.copyOf(mutableList);
+    }
+
+    private void debugLogCurrentChannelRejected(int currentIndex, LimitedChannel channel, OptionalInt next) {
```
ferozco, line 219: timber would be nice for things like this

dialogue-core/src/test/java/com/palantir/dialogue/core/PinUntilErrorChannelTest.java
```diff
@@ -0,0 +1,156 @@
...
+import java.util.Random;
+import java.util.concurrent.TimeUnit;
+import java.util.stream.IntStream;
+import org.junit.Before;
+import org.junit.Test;
+import org.mockito.Mockito;
+
+public class PinUntilErrorChannelTest {
+
+    private LimitedChannel channel1 = mock(LimitedChannel.class);
```
ferozco: nit: `@Mock` for consistency and added strictness with mockitoExtension

dialogue-core/src/test/java/com/palantir/dialogue/core/PinUntilErrorChannelTest.java
```diff
@@ -0,0 +1,156 @@
...
+    public void reshuffle_happens_roughly_every_10_mins() {
+        setResponse(channel1, 100);
+        setResponse(channel2, 204);
+
+        assertThat(IntStream.range(0, 6).map(number -> getCode(pinUntilError)))
+                .describedAs("First batch on channel1")
+                .contains(100, 100, 100, 100, 100, 100);
+
+        when(clock.read()).thenReturn(Duration.ofMinutes(11).toNanos());
+        assertThat(IntStream.range(0, 6).map(number -> getCode(pinUntilError)))
```
ferozco, line 107: this might be flaky since reshuffle could result in the same order as before no?

dialogue-core/src/main/java/com/palantir/dialogue/core/PinUntilErrorChannel.java
iamdanfox: This is actually pretty important - if there are 100 independent servers all talking to a few catalog nodes and they all start off with the nodelist in the same order, then when catalog node1 restarts, all that nodes traffic will shift onto node2 and stay there. Node2 will then have double traffic, and when it restarts, they'll all shift to node3.

Making sure that the order is shuffled at startup time means that when a single upstream goes down, the traffic should be distributed across upstreams.

I can add a comment to this effect if you like?

ferozco: ok that totally makes sense. Ya a comment would be helpful thanks!

dialogue-core/src/test/java/com/palantir/dialogue/core/PinUntilErrorChannelTest.java
iamdanfox: I seed the Random instance at the beginning of the test, so it's 100% deterministic.

dialogue-core/src/main/java/com/palantir/dialogue/core/PinUntilErrorChannel.java
iamdanfox: What about if I just called it `markCurrentHostAsBad` ?

dialogue-core/src/main/java/com/palantir/dialogue/core/PinUntilErrorChannel.java
```diff
@@ -0,0 +1,260 @@
...
+
+    @VisibleForTesting
+    PinUntilErrorChannel(NodeList nodeList) {
+        this.maybeShufflingNodeList = nodeList;
+    }
+
+    public static PinUntilErrorChannel pinUntilError(List<LimitedChannel> channels) {
+        Preconditions.checkArgument(!channels.isEmpty(), "List of channels must not be empty");
+        ReshufflingNodeList shufflingNodeList =
+                new ReshufflingNodeList(channels, ThreadLocalRandom.current(), System::nanoTime);
```
carterkozak: bug: ThreadLocalRandom.current() returns the caller threads ThreadLocalRandom, the node list may be used by another thread, potentially causing contention. I wrote a simple `SafeThreadLocalRandom` random at one point that internally delegates to the current threads instance for all calls but never pushed it out broadly.
For now we may want a `Supplier<Random>`

dialogue-core/src/main/java/com/palantir/dialogue/core/PinUntilErrorChannel.java
ferozco: It doesn't quite capture the return type though. `maybeIncrementCurrentHost`?

dialogue-core/src/main/java/com/palantir/dialogue/core/PinUntilErrorChannel.java
```diff
@@ -0,0 +1,260 @@
...
+                    .plus(Duration.ofSeconds(random.nextInt(60) - 30))
+                    .toNanos();
+            this.nextReshuffle = new AtomicLong(clock.read() + intervalWithJitter);
+            this.clock = clock;
+        }
+
+        @Override
+        public LimitedChannel get(int index) {
+            reshuffleChannelsIfNecessary();
+            return channels.get(index % channels.size());
```
carterkozak: concurrency bug: index out of bounds when another thread reshuffles. We need a single volatile read, and then use get+size on the local reference.

dialogue-core/src/main/java/com/palantir/dialogue/core/PinUntilErrorChannel.java
```diff
@@ -0,0 +1,260 @@
...
+        }
+
+        private void reshuffleChannelsIfNecessary() {
```
carterkozak, line 193: Worth an early return when `channels.size <= 1` to avoid nanotime syscalls

dialogue-core/src/main/java/com/palantir/dialogue/core/PinUntilErrorChannel.java
```diff
@@ -0,0 +1,260 @@
...
+            long reshuffleTime = nextReshuffle.get();
+            if (clock.read() < reshuffleTime) {
+                return;
+            }
+
+            if (nextReshuffle.compareAndSet(reshuffleTime, reshuffleTime + intervalWithJitter)) {
```
carterkozak:
```suggestion
            if (nextReshuffle.compareAndSet(reshuffleTime, clock.read() + intervalWithJitter)) {
```

We don't want to be caught in a state where each request reshuffles the list to catch up. This can occur if a vm is paused or laptop is closed :-)

dialogue-core/src/main/java/com/palantir/dialogue/core/PinUntilErrorChannel.java
```diff
@@ -0,0 +1,269 @@
...
+        }
+
+        @Override
+        public LimitedChannel get(int index) {
+            reshuffleChannelsIfNecessary();
+            return channels.get(index % channels.size());
+        }
+
+        private void reshuffleChannelsIfNecessary() {
+            if (channels.size() < 1) {
```
carterkozak: If there's only one channel there's no need to shuffle either
```suggestion
            if (channels.size() <= 1) {
```

iamdanfox: 🤦‍♂

dialogue-core/src/main/java/com/palantir/dialogue/core/PinUntilErrorChannel.java
iamdanfox: How could we get out of bounds if the channel size is always constant? Also we're not shuffling in place, but re-assigning the entire channels list?

dialogue-core/src/main/java/com/palantir/dialogue/core/PinUntilErrorChannel.java
```diff
@@ -0,0 +1,270 @@
...
+        }));
+    }
+
+    /**
+     * If we have some reason to think the currentIndex is bad, we want to move to the next host. This is done with a
+     * compareAndSet to ensure that out of order responses which signal information about a previous host don't kick
+     * us off a good one.
+     */
+    private OptionalInt markCurrentHostAsBad(int currentIndex) {
+        int nextIndex = Math.max(currentIndex + 1, 0); // we want Integer.MAX_VALUE to wrap around to zero
```
iamdanfox: ~interestingly I think the round robin channel has this same bug, because if a list has 10 nodes, and we end up doing:~

```
(Integer.MAX_VALUE + 1) % 10 = -8
```

~And I'm _pretty_ sure arraylists don't like negative indexes.~

Actually nvm it always calls toIndex before saving

dialogue-core/src/main/java/com/palantir/dialogue/core/PinUntilErrorChannel.java
carterkozak: Ah right, I suppose since we don't have the ability to update targets the size is a constant.

It smells a off that we may be calling channels.size on a different instance than channels.get if a shuffle occurs while we're executing. I think the code would be safer/more obvious with an atomic read (or fully extracting the size) so we don't shoot ourselves in the foot modifying this later!

dialogue-core/src/main/java/com/palantir/dialogue/core/PinUntilErrorChannel.java
carterkozak: Can we write a test to expose it?

dialogue-core/src/main/java/com/palantir/dialogue/core/PinUntilErrorChannel.java
```diff
@@ -0,0 +1,270 @@
...
+ * Upside: ensures clients will always hit warm caches
+ * Downsides:
+ * <ul>
+ *     <li>a single high volume client can result in unbalanced utilization of server nodes.
+ *     <li>when a node is restarted (and clients see 500s/connect exceptions), they will _all_ pin to a different node
+ * </ul>
+ *
+ * To alleviate the second downside, we reshuffle all nodes every 10 minutes.
+ */
+final class PinUntilErrorChannel implements LimitedChannel {
```
iamdanfox, line 53: I wonder if we should make a specialized version of this for the case when there's exactly one upstream cos all the logic is pointless in that case (0 +1 % 1 is still 0)? Or am I over-optimizing?

carterkozak: 1 target is a very common case, I think it's worth optimizing for.
It's difficult to say _where_ we should do this optimization without implementing config reloading, if we continued with the 'recreate-everything' approach I'd prefer to simply bypass the load balancing channel entirely and wire the next LimitedChannel in the chain.

iamdanfox:
Ok so I think I've optimized this as much as it deserves for a first stab 😂can someone pull the trigger for me?

carterkozak:
:+1: Thanks Dan!

ferozco:
👍

ferozco:
dozer?

iamdanfox:
Thanks for the review! This one was more subtle than I was expecting

dialogue-core/src/main/java/com/palantir/dialogue/core/PinUntilErrorChannel.java
```diff
@@ -0,0 +1,275 @@
...
+            OptionalInt next = incrementHostIfNecessary(currentIndex);
+            debugLogCurrentChannelRejected(currentIndex, channel, next);
+            return Optional.empty(); // if the caller retries immediately, we'll get the next host
+        }
+
+        ListenableFuture<Response> future = maybeFuture.get();
+        return Optional.of(DialogueFutures.addDirectCallback(future, new FutureCallback<Response>() {
+            @Override
+            public void onSuccess(Response response) {
+                if (response.code() >= 300) {
```
markelliot, line 104: putting this sort of logic in onSuccess handlers throughout the code is going to continue to make this framework much harder to understand than necessary. 

please consider an approach where success means _success_ always and that there should instead be a channel which coerces failures into failed futures.

why are failures here defined as >= 300, but elsewhere in code as >= 500?

carterkozak: Looks like the CJR implementation only cycles when an I/O exception is thrown making a request, not based on status (aside from QoS). We wouldn't want 4xx responses to trigger rotation.

PR #365, opened 2020-02-18
Faster gradle

iamdanfox:
## Before this PR

Was getting 20 second configuration times from gitVersion.

## After this PR
==COMMIT_MSG==
Local gradle development goes fast.
==COMMIT_MSG==


carterkozak (approved):
thanks!

PR #367, opened 2020-02-18
Fix off-by-one in RetryingChannel, use the configured number of retries

carterkozak:
==COMMIT_MSG==
Fix off-by-one in RetryingChannel, use the configured number of retries
==COMMIT_MSG==

## Possible downsides?
Doesn't support the configured backoff interval yet.

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
@@ -16,22 +16,24 @@
 
 package com.palantir.dialogue.core;
 
-import com.google.common.util.concurrent.FutureCallback;
+import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
-import com.google.common.util.concurrent.SettableFuture;
+import com.google.common.util.concurrent.MoreExecutors;
 import com.palantir.dialogue.Channel;
 import com.palantir.dialogue.Endpoint;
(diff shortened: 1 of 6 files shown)


dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -59,23 +48,21 @@
...
         };
         FutureCallback<Response> retryer = new RetryingCallback(callSupplier, future);
-        Futures.addCallback(callSupplier.apply(0), retryer, DIRECT_EXECUTOR);
-
+        DialogueFutures.addDirectCallback(callSupplier.apply(0), retryer);
         return future;
     }
 
     private final class RetryingCallback implements FutureCallback<Response> {
         private final AtomicInteger failures = new AtomicInteger(0);
```
carterkozak: This should be an `int` given that retries are sequenced we can rely on the happens-before relationship.

ferozco: do you want to make that change as part of this PR?

ferozco:
👍

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
carterkozak: I'll add it to a an improvement on top of this PR, might be best if you'd like to review that in isolation :-]

PR #368, opened 2020-02-18
fix #311: RetryingChannel propagates cancelation

### LINKED ISSUE #311: RetryingChannel does not propagate cancelation
## What happened?

When futures are canceled, the underlying request is not.

Cancellation should make its way to the actual http client.

carterkozak:
Generally safer by using existing utility functionality to chain
futures instead of reimplementing async-transformation via
callbacks.

## After this PR
==COMMIT_MSG==
RetryingChannel propagates cancelation
==COMMIT_MSG==

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
@@ -16,17 +16,15 @@
 
 package com.palantir.dialogue.core;
 
-import com.google.common.util.concurrent.FutureCallback;
+import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
-import com.google.common.util.concurrent.SettableFuture;
+import com.google.common.util.concurrent.MoreExecutors;
 import com.palantir.dialogue.Channel;
 import com.palantir.dialogue.Endpoint;
 import com.palantir.dialogue.Request;
 import com.palantir.dialogue.Response;
 import com.palantir.logsafe.exceptions.SafeRuntimeException;
 import java.util.Optional;
-import java.util.concurrent.atomic.AtomicInteger;
-import java.util.function.IntFunction;
 
 /**
  * Immediately retries calls to the underlying channel upon failure.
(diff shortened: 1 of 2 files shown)


dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -46,61 +44,59 @@
 
     @Override
     public ListenableFuture<Response> execute(Endpoint endpoint, Request request) {
-        SettableFuture<Response> future = SettableFuture.create();
-
-        IntFunction<ListenableFuture<Response>> callSupplier = attempt -> {
```
carterkozak, line 51: Replaced this abstraction by passing the channel, endpoint, and request into the retryer. We still have the attempt  index available for instrumentation (RetryingCallback.failures), but I find it's easier to spot bugs without the indirection.

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -46,61 +44,59 @@
...
-        return future;
+        return new RetryingCallback(delegate, endpoint, request, maxRetries).execute();
     }
 
-    private final class RetryingCallback implements FutureCallback<Response> {
-        private final AtomicInteger failures = new AtomicInteger(0);
-        private final IntFunction<ListenableFuture<Response>> runnable;
-        private final SettableFuture<Response> delegate;
+    private static final class RetryingCallback {
+        private final Channel channel;
```
ferozco: s/channel/delegate/ for consistency

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -46,61 +44,59 @@
...
-                return;
+                return retryOrFail(Optional.empty());
             }
 
             // TODO(dfox): if people are using 308, we probably need to support it too
 
-            boolean setSuccessfully = delegate.set(response);
-            if (!setSuccessfully) {
-                response.close();
-            }
```
carterkozak, line 85: No longer have to worry about this case because cancel is propagated through the future chain.

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -46,61 +44,59 @@
...
-            } else {
-                if (throwable.isPresent()) {
-                    delegate.setException(throwable.get());
-                } else {
-                    delegate.setException(new SafeRuntimeException("Retries exhausted"));
-                }
+                return execute();
             }
+            return Futures.immediateFailedFuture(
+                    throwable.orElseGet(() -> new SafeRuntimeException("Retries exhausted")));
```
ferozco: It might be nice to know what caused us to retry

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -46,61 +44,59 @@
...
+        }
+
+        private ListenableFuture<Response> wrap(ListenableFuture<Response> input) {
+            return Futures.catchingAsync(
```
ferozco, line 115: this is really neat!

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
carterkozak: good call, might as well add some helpful info logging prior to each retry while I'm here.

PR #369, opened 2020-02-18
Migrate dialogue-core to junit-jupiter

ferozco:
## Before this PR
We used junit4 in dialogue-core, preventing us from using the tracing extension

## After this PR
We use junit-jupiter.

I only now found out about `lenient().when` from mockito which lets us get some of the advantages of strict mocking without as much boilerplate.

I'll probably migrate the rest of repo...

## Possible downsides?
N/A


iamdanfox:
I think we may need to open-source our internal junit5-extensions in order to get a junit5 parameterized class for my simulations...

ferozco:
we can make some progress with tracing without the simulations, but you're totally right it would be sweet to see the simulations and the traces at once

carterkozak:
:+1:

PR #372, opened 2020-02-18
add retry tracing

ferozco:
## Before this PR
We did not produce separate spans for each retry attempt

## After this PR
We produce a standalone span for each retry attempt that are all siblings

## Possible downsides?
I'm not sure why we see the following trace, in particular the variance in request time and the huge gabs between them

image


iamdanfox:
> I'm not sure why we see the following trace, in particular the variance in request time and the huge gabs between them

Isn't this because java will be running your code in interpreter mode initially, i.e. nothing has been jit'd yet? also all your futures return immediately?

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -113,10 +114,16 @@ private void logRetry(Throwable throwable) {
...
             return Futures.catchingAsync(
-                    Futures.transformAsync(input, this::success, MoreExecutors.directExecutor()),
+                    Futures.transformAsync(wrapWithTracing(input), this::success, MoreExecutors.directExecutor()),
                     Throwable.class,
                     this::failure,
                     MoreExecutors.directExecutor());
         }
+
+        private ListenableFuture<Response> wrapWithTracing(ListenableFuture<Response> input) {
+            DetachedSpan detachedSpan = DetachedSpan.start(String.format("request-attempt-%d", failures));
```
carterkozak:
```suggestion
            DetachedSpan detachedSpan = DetachedSpan.start("request-attempt-" + failures);
```

String.format is slow. compiling for jdk11 presizes string concatenation (even better than java8 automatic string builders)
https://openjdk.java.net/jeps/280

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -113,10 +114,16 @@ private void logRetry(Throwable throwable) {
 
         private ListenableFuture<Response> wrap(ListenableFuture<Response> input) {
             return Futures.catchingAsync(
-                    Futures.transformAsync(input, this::success, MoreExecutors.directExecutor()),
+                    Futures.transformAsync(wrapWithTracing(input), this::success, MoreExecutors.directExecutor()),
```
carterkozak, line 117: We don't capture it in our tests, but all our real channels are wrapped with code to produce tracing spans. It doesn't have a retry index, but the order should make it clear.
What if we wrapped the outermost channel returned from Channels.create with another traced channel instead?

ferozco: sure

PR #373, opened 2020-02-19
RoundRobin uses a new heuristic: Balanced

iamdanfox:
## Before this PR

- **Bad blue/green upgrades wreck users.** Right now, if one out of two servers get upgraded and starts blasting 500s, our current concurrencylimiter+roundrobin setup will result in the user experiencing a large proportion of brokenness even if there were other perfectly healthy nodes.  **[Here's a graph showing the brokenness](https://github.com/palantir/dialogue/blob/96ebe45024b1961b11cf212f105f5940fcf1f3c2/simulation/src/test/resources/fast_500s_then_revert%5BCONCURRENCY_LIMITER%5D.png)**. (clients perceive 77% success). pin_until_error alleviates this (**[see graph](https://github.com/palantir/dialogue/blob/96ebe45024b1961b11cf212f105f5940fcf1f3c2/simulation/src/test/resources/fast_500s_then_revert%5BPIN_UNTIL_ERROR%5D.png)**), but it results in one node getting absolutely hammered, even after the bad node recovers.

- **We're leaving latency on the table.** if one node suddenly gets slow for some reason (but continues 200ing), we might not hit our concurrency limiter ceiling so round robin will just keep sending 50% traffic to this server, while another fast node receives it's 50% share of traffic but returns it instantly.  pin_until_error isn't really any better because it's all luck whether you get pinned to the slow one or the fast one. If they're both 200ing, then it's equal chance either way. See this in [drastic_slowdown graphs](https://github.com/palantir/dialogue/blob/dfox/lowest-utilization/simulation/src/test/resources/report.md#drastic_slowdownconcurrency_limiterpng).

- **Bugs are often localized to endpoints.** We're getting better at shipping small chunks of code to production. This often means that if a bug goes out to prod, it often only affects one or two endpoints. To achieve some theoretical maximum performance our clients need to understand endpoint granularity and be able to keep using the still working endpoints even if one is broken.

_PR reopened on 24th Apr after the internal P0 PDS-117808 which revealed how bad the plain round robin strategy is._

## After this PR
==COMMIT_MSG==
When users ask for `ROUND_ROBIN`, a new `Balanced` heuristic picks the node with the fewest active requests in flight.
==COMMIT_MSG==

# [Check out the graphs](https://github.com/palantir/dialogue/blob/dfox/lowest-utilization/simulation/src/test/resources/report.md#fast_500s_then_revertprefer_lowest)

## Possible downsides?
- ~the current impl spends a ton of cycles allocating, sorting and generally faffing about.  If we thing this idea has potential, I'd like to try and make a fast approximation of this heuristic.~
- not appropriate for transactional workloads (e.g. internal ski product) as successive requests will often get sent to different nodes
- this heuristic can't match the cache-hit likelihood of PIN_UNTIL_ERROR (although it's no worse than ROUND_ROBIN), so when we have vast server capacity and never error then I think PIN_UNTIL_ERROR could out-perform it.


simulation/src/test/java/com/palantir/dialogue/core/Strategy.java
```diff
@@ -81,15 +84,29 @@ private static Channel pinUntilError(Simulation sim, Supplier<List<SimulationSer
...
             return new RetryingChannel(channel, 4 /* ClientConfigurations.DEFAULT_MAX_NUM_RETRIES */);
         });
     }
 
+    private static Channel preferLowest(Simulation sim, Supplier<List<SimulationServer>> channelSupplier) {
+        Random pseudoRandom = new Random(21735712L);
+        return RefreshingChannelFactory.RefreshingChannel.create(channelSupplier, channels -> {
+            ImmutableList<LimitedChannel> limitedChannels = channels.stream()
+                    .map(c1 -> new ConcurrencyLimitedChannel(
+                            c1, () -> ConcurrencyLimitedChannel.createLimiter(sim.clock())))
```
iamdanfox: You can turn off ConcurrencyLimiters here (and use `UnlimitedChannel` instead) and you get marginally faster performance in most conditions - the only case that's worse if all nodes are 500ing, we end up just blasting one node with a massive spike of traffic. Here's the [comparison](https://github.com/palantir/dialogue/blob/dfox/lowest-utilization/simulation/src/test/resources/report.md#all_nodes_500prefer_lowestpng)

all_nodes_500 PREFER_LOWEST

carterkozak:
At a glance I think the lowestUtilization simulation is also using the blacklist channel, which I'd expect to similarly improve round robin.

Initial thoughts:
I prefer to avoid smart clients, but this logic is fairly complicated. Are we confident the upside is significantly better than round robin?
With heavy concurrent utilization I'd expect the law of averages to produce similar results using round robin.
The servers view of its own utilization is going to be more accurate than the clients. There are often many more clients than servers, so balancing active requests across hosts may not actually balance server load.
Not all endpoints have the same cost, we can track client-side data per endpoint describing how long requests are expected to take, but some endpoints wait for events which is nearly free but time consuming, while others may use gigabytes of memory and spin several cores.

iamdanfox:
So I accept that we should probably get rollout of vanilla dialogue first, but I do still want to revisit this because the graphs seem to show it producing better client-perceived performance.

dialogue-core/src/main/java/com/palantir/dialogue/core/PreferLowestUtilization.java
```diff
@@ -0,0 +1,118 @@
...
+ */
+final class PreferLowestUtilization implements LimitedChannel {
+    private static final Logger log = LoggerFactory.getLogger(PreferLowestUtilization.class);
+
+    private final Random random;
+    private final ImmutableList<LimitedChannel> channels;
+
+    /** Each AtomicInteger stores the number of active requests that a channel is currently serving. */
+    private final LoadingCache<LimitedChannel, AtomicInteger> activeRequestsPerChannel =
+            Caffeine.newBuilder().maximumSize(1000).build(upstream -> new AtomicInteger());
```
carterkozak: I don't think we need a cache here, can we transform the List<LimitedChannel> into objects that are composed of the LimitedChannel and AtomicInteger?

iamdanfox: Yep I'm gonna reimplement this to be non-bonkers don't worry, just wanted to get the simulations checked in so I can see whether my re-implementation does what expected!

dialogue-core/src/main/java/com/palantir/dialogue/core/PreferLowestUtilization.java
```diff
@@ -0,0 +1,118 @@
...
+        for (LimitedChannel channel : shuffleImmutableList(channels)) {
+            int activeRequests = activeRequestsPerChannel.get(channel).get();
+            channelsByUtilization.put(activeRequests, channel);
+        }
+
+        return channelsByUtilization;
+    }
+
+    /** Returns a new shuffled list, without mutating the input list (which may be immutable). */
+    private <T> List<T> shuffleImmutableList(ImmutableList<T> sourceList) {
```
carterkozak: Not sure we need to go so far as sorting lists here, but there's a quick optimization in which we sort the result of the inner for lop in `maybeExecute` where we can no-op the simple case that sourceList has fewer than two elements.

iamdanfox:
So I think the black_hole case is most similar to the recent P0 https://github.com/palantir/dialogue/blob/dfox/lowest-utilization/simulation/src/test/resources/report.md#black_holeconcurrency_limiter_round_robin although as requests never return, they're not counted in the client_mean=600ms.

dialogue-core/src/main/java/com/palantir/dialogue/core/PreferLowestUtilization.java
carterkozak: Longer term this object knows when each channel receives and completes a request, so it can use that to maintain an internal data structure that doesn't require building b-trees for each request.

carterkozak:
We've seen several cases lately where we discover URIs that don't point to a real service and result in a quick connect failure. Similarly we're rolling out a change to immediately 503 when requests are received if readiness checks aren't green yet. In these cases, won't we effectively pin to the fast-failing channel?

iamdanfox:
I don't think we'll end up pinned to that channel, because 503s should take down the concurrency limiter. So perhaps in our sorting that channel will appear first in the list, but we'll always end up making a request to one that has capacity.

The fast_500s_then_revert simulation case should cover this.

iamdanfox:
So @carterkozak I've reimplemented this and it no longer constructs any maps on the hot codepath. What do you think?

carterkozak:
I think the issue is that 503 responses and connection errors will generally be super fast (sub milliseconds) where successful requests will take much longer. Thoughts on adding a simulation with three servers where one responds 503 within ~150 microseconds and the other two process requests resulting in 200 status responses taking >100 milliseconds?

My fear is that a concurrency limit of one permit still allows 6k reqs/sec to flow through in serial where the working channels will only be able to handle 10 reqs/sec at the same level of saturation.

iamdanfox:
ok yep so https://github.com/palantir/dialogue/blob/dfox/lowest-utilization/simulation/src/test/resources/report.md#fast_503s_then_revertunlimited_round_robin very much does demonstrate how this strategy leans into a node that is failing fast. I think it still would have averted an outage in PDS-117808 though, because that was _not_ a fail fast scenario (the node was unresponsive, so more of a black hole).

I miss my blacklisting channel 😢

carterkozak:
I like the idea of taking a both a sliding window of success/failure rates and utilization into account. I'm fearful that the blacklisting channel will wreck us if we get intermittent connection failures every 100k requests while successfully handling 99.9% of requests, which the adaptive model could solve.

iamdanfox:
~I'm gonna try a really trivial memory first before the sliding window thing, one sec~ lol this does not help at all.

dialogue-core/src/main/java/com/palantir/dialogue/core/Balanced.java
```diff
@@ -0,0 +1,172 @@
...
+            public void onSuccess(Response result) {
+                inflight.decrementAndGet();
+                if (Responses.isQosStatus(result) || Responses.isServerError(result)) {
+                    recentFailures.update(1);
+                }
+            }
+
+            @Override
+            public void onFailure(Throwable _throwable) {
+                recentFailures.update(1);
```
carterkozak, line 126:
```suggestion
                inflight.decrementAndGet();
                recentFailures.update(1);
```

dialogue-core/src/main/java/com/palantir/dialogue/core/Balanced.java
```diff
@@ -0,0 +1,172 @@
...
+                // we rely heavily on streams being lazy here
+                .findFirst();
+    }
+
+    /** Returns a new shuffled list, without mutating the input list (which may be immutable). */
+    private static <T> List<T> shuffleImmutableList(ImmutableList<T> sourceList, Random random) {
+        List<T> mutableList = new ArrayList<>(sourceList);
+        Collections.shuffle(mutableList, random);
+        return mutableList;
+    }
```
carterkozak, line 97: I don't think the shuffle buys us anything because utilization is immediately reflected. We shouldn't need to break ties because we have enough data that either they're uncommon, or it doesn't matter which channel we select because we know it works just as well.

dialogue-core/src/main/java/com/palantir/dialogue/core/Balanced.java
```diff
@@ -0,0 +1,172 @@
...
+        @VisibleForTesting
+        final AtomicInteger inflight = new AtomicInteger(0);
+
+        /**
+         * We keep track of failures within a time window to do well in scenaios where an unhealthy server returns
+         * errors much faster than healthy nodes can serve good responses. See
+         * {@link SimulationTest#fast_503s_then_revert}.
+         */
+        @VisibleForTesting
+        final SlidingTimeWindowArrayReservoir recentFailures;
```
carterkozak: Have we validated the implementation of reservoir for read performance? They're generally written to be performant with more reads than writes, but our usage pattern is the opposite.

iamdanfox: Not yet - I think this one would end up calling `ChunkedAssociativeLongArray#size`, which doesn't look _too_ horrible...

```java
   synchronized int size() {
        int result = 0;
        for (Chunk chunk : chunks) {
            result += chunk.cursor - chunk.startIndex;
        }
        return result;
    }
```

We could probably implement something with a single array and a circular buffer though if we wanted to micro-optimize.

sfackler: Since full accuracy isn't really important here, you can use something simpler but a bit more fuzzy. Pseudocode:

```
class FuzzyReservoir {
    long count;
    long[] windows;
    int cursor;
    long rolloverNanos;
    long nextRollover;

    FuzzerReservoir(long sizeNanos, int windows) {
        windows = new long[windows];
        rolloverNanos = sizeNanos / windows;
        nextRollover = System.nanoTime() + sizeNanos;
    }


    void increment() {
        maybeRoll();
        count += 1;
        windows[cursor] += 1;
    }

    long get() {
        return count;
    }

    void maybeRoll() {
        if (System.nanoTime() < nextRollover) {
            return;
        }

        cursor = (cursor + 1) % windows.length;
        count -= windows[cursor];
        windows[cursor] = 0;
        nextRollover += rolloverNanos;
    }
}
```

If you make count and nextRollover atomics, you can have (mostly) lock free reads, and just synchronize writes and rollovers.

carterkozak: Generally in favor of the idea, we would need to check for rollover on read otherwise a high-cost channel will never be selected, and have no way to decay cost over time.

dialogue-core/src/main/java/com/palantir/dialogue/core/Balanced.java
```diff
@@ -0,0 +1,173 @@
...
+        }
+
+        // pre-shuffling is pretty important here, otherwise when there are no requests in flight, we'd
+        // *always* prefer the first channel of the list, leading to a higher overall load.
+        List<ChannelWithStats> preShuffled = shuffleImmutableList(channels, random);
+
+        // TODO(dfox): P2C optimization when we have high number of nodes to save CPU?
+        // http://www.eecs.harvard.edu/~michaelm/NEWWORK/postscripts/twosurvey.pdf
+        return preShuffled.stream()
+                .sorted(BY_SCORE)
```
carterkozak: I think this sort can throw because score will be modified by concurrent requests, list[a] > list[b] when first validated, but not the second time.

dialogue-core/src/main/java/com/palantir/dialogue/core/Balanced.java
iamdanfox: Are you sure? if we have hundreds of clients and we took out this shuffle, then every single time there's a client has a blip of zero active requests, they will always pick the first channel to hit first when they make their next request. The simulations show a pretty noticeable effect where server utilization is just not balanced anymore.

dialogue-core/src/main/java/com/palantir/dialogue/core/Balanced.java
iamdanfox: Hmmm, I guess we could copy the mutable things into an immutable list and then sort that?

dialogue-core/src/main/java/com/palantir/dialogue/core/Balanced.java
carterkozak: That’s a good point, I suppose we could do something similar to the random selection implementation

iamdanfox: Given that the number of nodes is pretty much always <10, I think it might be fine to leave it as is for now tbh

dialogue-core/src/main/java/com/palantir/dialogue/core/Balanced.java
carterkozak: Yep, that should do it

dialogue-core/src/main/java/com/palantir/dialogue/core/Balanced.java
```diff
@@ -0,0 +1,173 @@
...
+
+        ChannelWithStats(LimitedChannel delegate, CodahaleClock clock) {
+            this.delegate = delegate;
+            this.recentFailures =
+                    new SlidingTimeWindowArrayReservoir(FAILURE_MEMORY.toNanos(), TimeUnit.NANOSECONDS, clock);
+        }
+
+        /** Low = good. */
+        int score() {
+            return inflight.get() + recentFailures.size();
```
carterkozak: Perhaps we should scale failures by a factor of ~ten, that way slow successes don't require so many failures to balance out?

carterkozak, line 154: Perhaps we should scale failures by a factor of ~ten, that way slow successes don't require so many failures to balance out?

dialogue-core/src/main/java/com/palantir/dialogue/core/Balanced.java
```diff
@@ -0,0 +1,173 @@
...
+ * {@link PinUntilErrorChannel} remains the best choice for these.
+ */
+final class Balanced implements LimitedChannel {
+    private static final Logger log = LoggerFactory.getLogger(Balanced.class);
+
+    /**
+     * We'd like to remember failures for a long time, but this would increase CPU on a hot codepath as we compute
+     * the number of failures for each channel in order to sort. See {@link ChannelWithStats#score}.
+     */
+    private static final Duration FAILURE_MEMORY = Duration.ofSeconds(5);
```
carterkozak: We could use a simple single-bucket exponential decay implementation instead of sliding window. That way we can remember failures for a long time and simply cut the value in half every N seconds.

dialogue-core/src/main/java/com/palantir/dialogue/core/Balanced.java
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+    static final class ChannelWithStats implements LimitedChannel {
+        private final LimitedChannel delegate;
+
+        @VisibleForTesting
+        final AtomicInteger inflight = new AtomicInteger(0);
+
+        /**
+         * We keep track of failures within a time window to do well in scenaios where an unhealthy server returns
```
carterkozak:
```suggestion
         * We keep track of failures within a time window to do well in scenarios where an unhealthy server returns
```

dialogue-core/src/main/java/com/palantir/dialogue/core/Balanced.java
```diff
@@ -0,0 +1,173 @@
...
+         * errors much faster than healthy nodes can serve good responses. See
+         * {@link SimulationTest#fast_503s_then_revert}.
```
carterkozak: Invalid link
```suggestion
         * <code>SimulationTest.fast_503s_then_revert</code>.
```

ferozco:
I think it would be helpful to have a discussion about the algorithm separate from the code so that we can all agree on the things we want capture in the alg. and hopefully make it easier for us to explain to others what it is doing.

It might also be cool to expose a way to tune the hyperparameters to the algorithm so that ideally machines (but maybe initially humans) can turn knobs to eek out improved performance/reliability

dialogue-core/src/main/java/com/palantir/dialogue/core/Balanced.java
iamdanfox: Honestly given how well the sliding window performs in the simulations, I think I find it a little easier to reason about, at least as an MVP.  Maybe we can open a PR to swap it out so we can compare the simulation diff later?

dialogue-core/src/main/java/com/palantir/dialogue/core/Balanced.java
iamdanfox: Have implemented, see LockFreeTimeWindowReservoir.java

dialogue-core/src/main/java/com/palantir/dialogue/core/LockFreeTimeWindowReservoir.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    void mark() {
+        maybeRoll();
+        buckets[cursor.get()] += 1;
+        count.incrementAndGet();
+    }
+
+    long size() {
+        return count.get();
```
sfackler: You'll need to maybeRoll here as well or the error rate will never decay if errors stop ocurring.

dialogue-core/src/main/java/com/palantir/dialogue/core/LockFreeTimeWindowReservoir.java
```diff
@@ -0,0 +1,118 @@
...
+    LockFreeTimeWindowReservoir(Duration memory, int granularity, Ticker clock) {
+        this.buckets = new long[granularity];
+        this.clock = clock;
+        this.bucketSizeNanos = memory.toNanos() / granularity;
+        this.nextRollover = new AtomicLong(clock.read() + bucketSizeNanos);
+    }
+
+    void mark() {
+        maybeRoll();
+        buckets[cursor.get()] += 1;
```
sfackler: buckets probably needs to be a `AtomicLong[]` so this behaves under concurrency, right?

iamdanfox:
In terms of the actual algorithm in use here, happy to discuss/document whatever you think would be helpful. To be honest I didn't really have a priori intentions of what I wanted to stick into the algorithm, I was really just trying stuff until the simulations looked good.

Presently, the algorithm is:
- track the number of inflight reqs to each host
- track the number of recentFailures seen in the last 5 seconds for each host
- route each request to the host with the lowest `inflight + recentFailures`

This delivers a big improvement in two key simulations, the first of which would have helped with PDS-117808 I think.
- [drastic_slowdown](https://github.com/palantir/dialogue/blob/dfox/lowest-utilization/simulation/src/test/resources/report.md#drastic_slowdownconcurrency_limiter_round_robin)
- [slow_503s_then_revert](https://github.com/palantir/dialogue/blob/dfox/lowest-utilization/simulation/src/test/resources/report.md#slow_503s_then_revertconcurrency_limiter_round_robin)
- [fast_503s_then_revert](https://github.com/palantir/dialogue/blob/dfox/lowest-utilization/simulation/src/test/resources/report.md#fast_503s_then_revertconcurrency_limiter_round_robin)

carterkozak:
I put together a simple exponential decay implementation optimized toward reads: https://github.com/palantir/dialogue/compare/develop...ckozak/simple_coarse_exponential_decay
I think this should give us a bit more historical context with a simpler implementation.

iamdanfox:
Mmm the does have a lot less concurrency in it (which seems v. desirable). Also kinda nice that it would ramp traffic after failures I guess... do we see this in the simulations?

carterkozak:
I'll try it out in a branch based on this one.

dialogue-core/src/main/java/com/palantir/dialogue/core/Balanced.java
iamdanfox: Have now implemented an immutable snapshot.

simulation/src/test/resources/report.md
```diff
@@ -2,36 +2,35 @@
...
+                           drastic_slowdown[UNLIMITED_ROUND_ROBIN].txt:	success=100.0%	client_mean=PT0.275952999S 	server_cpu=PT18M23.81199998S	client_received=4000/4000	server_resps=4000	codes={200=4000}
         fast_503s_then_revert[CONCURRENCY_LIMITER_PIN_UNTIL_ERROR].txt:	success=100.0%	client_mean=PT0.120012009S 	server_cpu=PT1H30M0.00000004S	client_received=45000/45000	server_resps=45004	codes={200=45000}
-            fast_503s_then_revert[CONCURRENCY_LIMITER_ROUND_ROBIN].txt:	success=98.1%	client_mean=PT0.277457972S 	server_cpu=PT1H28M17.7602765S	client_received=45000/45000	server_resps=71798	codes={200=44148, 503=852}
-                      fast_503s_then_revert[UNLIMITED_ROUND_ROBIN].txt:	success=98.1%	client_mean=PT0.277457972S 	server_cpu=PT1H28M17.7602765S	client_received=45000/45000	server_resps=71798	codes={200=44148, 503=852}
+            fast_503s_then_revert[CONCURRENCY_LIMITER_ROUND_ROBIN].txt:	success=100.0%	client_mean=PT0.120105435S 	server_cpu=PT1H30M0.0000004S	client_received=45000/45000	server_resps=45040	codes={200=45000}
```
iamdanfox: really nice to get to success=100% here

carterkozak (approved):
I'm happy with this, thanks @iamdanfox! I think there's some room for micro-optimization in the details, but we shouldn't block on it.

iamdanfox:
Ok I'm gonna merge this as-is, but not gonna release until I've added the same kind of instrumentation as we have in the pin until error channel so I can make graphs.

:+1:

iamdanfox:
Thanks for the careful reviews all!

PR #376, opened 2020-02-19
Add simple cost based routing scheme

ferozco:
## Before this PR
Full disclaimer this was largely insipired by @iamdanfox awesome work and @markelliot's ideas. I thought it would be interesting to quickly whip up a cost based approach for routing where we attempt to make requests to channels that historically have been the cheapest. In this case we attribute cost for successful requests as the microseconds spent executing the request and for failures as the median cost of all known channels (very much a WIP of a heuristic).

As expected this heuristic doesn't behave incredibly well in some of the simulator cases since servers have constant response times. I expect that real world variance will lead to very interesting emergent behavior especially during upgrade windows.

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/CostBasedRouting.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/CostBasedRouting.java
@@ -1,0 +1,106 @@
+/*
+ * (c) Copyright 2020 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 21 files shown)


dialogue-core/src/main/java/com/palantir/dialogue/core/CostBasedRouting.java
```diff
@@ -0,0 +1,103 @@
...
+import java.util.concurrent.TimeUnit;
+import java.util.function.LongConsumer;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+final class CostBasedRouting implements LimitedChannel {
+    private final ImmutableList<LimitedChannel> channels;
+    // Each Reservoir stores the perceived cost of making a request to a channel. We may want to consider bookkeeping at
+    // the endpoint level in the future
+    private final LoadingCache<LimitedChannel, Reservoir> costPerChannel =
+            Caffeine.newBuilder().maximumSize(1000).build(upstream -> new ExponentiallyDecayingReservoir());
```
iamdanfox: you gotta wire clock into these - you can get one from simulation.codahaleClock()

ferozco:
We can revisit this later if we want

PR #378, opened 2020-02-19
Respect ClientConfiguration.serverQos

carterkozak:
## Before this PR
pref was ignored.

==COMMIT_MSG==
Respect ClientConfiguration.serverQos
==COMMIT_MSG==


dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -40,29 +40,34 @@
...
         private final Channel delegate;
         private final Endpoint endpoint;
         private final Request request;
         private final int maxRetries;
+        private final boolean propagateQoS;
         private int failures = 0;
 
-        private RetryingCallback(Channel delegate, Endpoint endpoint, Request request, int maxRetries) {
+        private RetryingCallback(
+                Channel delegate, Endpoint endpoint, Request request, int maxRetries, boolean disableQosRetries) {
```
iamdanfox: could we actually just plumb this enum into the RetryingChannel?

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -81,9 +85,7 @@ private RetryingCallback(Channel delegate, Endpoint endpoint, Request request, i
                     logRetry(failure);
                     return execute();
                 }
-                // TODO(ckozak): It's out of scope for this class to map responses to exceptions. After retries
-                // are exhausted the open response should be returned.
-                return Futures.immediateFailedFuture(new SafeRuntimeException("Retries exhausted", failure));
```
iamdanfox, line 86: :+1: makes sense! users want to see something from the server, not deep in here. Maybe debug log instead?

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -112,11 +114,12 @@ private void logRetry(Throwable throwable) {
...
-            return Futures.catchingAsync(
-                    Futures.transformAsync(input, this::success, MoreExecutors.directExecutor()),
-                    Throwable.class,
-                    this::failure,
-                    MoreExecutors.directExecutor());
+            ListenableFuture<Response> result = input;
+            if (!propagateQoS) {
+                result = Futures.transformAsync(result, this::success, MoreExecutors.directExecutor());
+            }
+            result = Futures.catchingAsync(result, Throwable.class, this::failure, MoreExecutors.directExecutor());
```
iamdanfox, line 132: this is where I think it wold be clearer to put that switch statement

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
carterkozak: I went back and forth on this. Feels a little odd to pipe the ClientConfiguration bits this deep, but an enum is cleaner. I'm okay with reusing the enum since RetryingChannel isn't public.

iamdanfox (approved):

iamdanfox:
As we are starting from a clean state with dialogue, I think it's also worth being judicious about what we end up exposing to users as the API surface of dialogue. 

Like maybe we need to replicate c-j-r's options for testing and evaluation purposes, but if we settle on something that we thing is 'just better' I would love to enforce a slightly more homogenous usage pattern if possible.

PR #383, opened 2020-02-19
Apache channel supports proxy credentials

carterkozak:
==COMMIT_MSG==
Apache channel supports proxy credentials
==COMMIT_MSG==

mesh proxy is still a todo.


dialogue-apache-hc4-client/src/main/java/com/palantir/dialogue/hc4/ApacheHttpClientChannels.java
```diff
@@ -97,4 +128,64 @@ private static URL url(String uri) {
             throw new SafeIllegalArgumentException("Failed to parse URL", e);
         }
     }
+
+    private enum NullCredentialsProvider implements CredentialsProvider {
```
ferozco, line 133: oh my boilerplate. Why do we need to provide null implementations?

dialogue-apache-hc4-client/src/main/java/com/palantir/dialogue/hc4/ApacheHttpClientChannels.java
```diff
@@ -82,7 +102,18 @@ public static Channel create(ClientConfiguration conf, UserAgent baseAgent) {
                                         ? CipherSuites.allCipherSuites()
                                         : CipherSuites.fastCipherSuites(),
                                 new DefaultHostnameVerifier()))
-                .build();
+                .setDefaultCredentialsProvider(NullCredentialsProvider.INSTANCE)
+                .setTargetAuthenticationStrategy(NullAuthenticationStrategy.INSTANCE)
+                .setProxyAuthenticationStrategy(NullAuthenticationStrategy.INSTANCE)
+                .setDefaultAuthSchemeRegistry(
+                        RegistryBuilder.<AuthSchemeProvider>create().build());
+        conf.proxyCredentials().ifPresent(credentials -> builder.setDefaultCredentialsProvider(
```
ferozco: could you add `{}` I think it makes it much more readable

dialogue-client-test-lib/src/main/java/com/palantir/dialogue/AbstractProxyConfigTest.java
```diff
@@ -0,0 +1,177 @@
...
+import okhttp3.mockwebserver.MockWebServer;
+import okhttp3.mockwebserver.RecordedRequest;
+import org.junit.Rule;
+import org.junit.Test;
+
+public abstract class AbstractProxyConfigTest {
+
+    protected static final UserAgent AGENT = UserAgent.of(UserAgent.Agent.of("test", "0.0.1"));
+    private static final SslConfiguration SSL_CONFIG = SslConfiguration.of(
+            Paths.get("../dialogue-client-test-lib/src/main/resources/trustStore.jks"),
```
ferozco, line 49: these paths seem a little off

dialogue-apache-hc4-client/src/main/java/com/palantir/dialogue/hc4/ApacheHttpClientChannels.java
carterkozak: I want to avoid dealing with credentials in all possible cases so we don't have to deal with folks getting used to client-specific behavior.
When proxy credentials are provided I want to make sure they can only be used by proxies by disabling the target auth impl, and all schemes except basic to match okhttp.

dialogue-client-test-lib/src/main/java/com/palantir/dialogue/AbstractProxyConfigTest.java
carterkozak: I'd like to reuse this config in other tests too, but avoided modifying them here. It's a bit odd, but it would allow us to check in a single ssl config with our abstract tests and reuse it in all client implementations. We have a lot of boilerplate to implement a test.

dialogue-apache-hc4-client/src/main/java/com/palantir/dialogue/hc4/ApacheHttpClientChannels.java
carterkozak: ack

ferozco:
👍

PR #385, opened 2020-02-19
Trace entire reqeuest

ferozco:
## Before this PR
We traced individual attempts but not the entire request.

## After this PR
==COMMIT_MSG==

==COMMIT_MSG==
Trace an entire request across attempts:
image


dialogue-core/src/main/java/com/palantir/dialogue/core/Channels.java
```diff
@@ -46,16 +46,17 @@ public static Channel create(
...
                 // TracedChannel must wrap TracedRequestChannel to ensure requests have tracing headers.
                 .map(TracedRequestChannel::new)
-                .map(channel -> new TracedChannel(channel, "Concurrency-Limited Dialogue Request"))
                 .map(ContentDecodingChannel::new)
                 .map(concurrencyLimiter(config))
                 .collect(ImmutableList.toImmutableList());
 
         LimitedChannel limited = nodeSelectionStrategy(config, limitedChannels);
         Channel channel = new QueuedChannel(limited, DispatcherMetrics.of(config.taggedMetricRegistry()));
+        channel = new TracedChannel(channel, "Dialogue-request-attempt");
```
carterkozak, line 56: It’s a little odd that we provide spans for attempts which may or may not actually make it to an http client (blacklisting, concurrency limiting, etc).
Perhaps we should wrap the client channel as well to show real http requests?

carterkozak: I'm not sure that wrapping the client channel would show up in our tests, I think we need to add a bit of tracing-aware code to queued channel (imo this would be totally reasonable for a separate change)

dialogue-core/src/test/java/com/palantir/dialogue/core/ChannelsTest.java
```diff
@@ -138,4 +140,20 @@ public void bad_channel_throwing_an_error_still_returns_a_future() {
         // only when we access things do we allow exceptions
         assertThatThrownBy(() -> Futures.getUnchecked(future)).hasCauseInstanceOf(NoClassDefFoundError.class);
     }
+
+    @Test
+    @TestTracing
+    @SuppressWarnings("FutureReturnValueIgnored")
+    public void traces_on_retries() {
+        when(response.code()).thenReturn(429);
+        channel.execute(endpoint, request);
```
carterkozak: Feels a little less brittle instead of `@SuppressWarnings("FutureReturnValueIgnored")`
```suggestion
        channel.execute(endpoint, request).get();
```

dialogue-core/src/test/java/com/palantir/dialogue/core/ChannelsTest.java
```diff
@@ -138,4 +140,20 @@ public void bad_channel_throwing_an_error_still_returns_a_future() {
...
+    }
+
+    @Test
+    @TestTracing(snapshot = true)
+    @SuppressWarnings("FutureReturnValueIgnored")
+    public void traces_on_succes() {
+        when(response.code()).thenReturn(200);
+        channel.execute(endpoint, request);
```
carterkozak: here as well

dialogue-core/src/main/java/com/palantir/dialogue/core/Channels.java
ferozco: Ya totally agree, I decided to go down this path since it seemed nice from a users perspective to always be able to see N request attempts regardless oh what happened during the attempt. I can bring back https://github.com/palantir/dialogue/pull/385/files/a82380f80f878ed1e33394317d7b640f79f3cf11#diff-3e072c17c066c68f3570b3b5f19ed29dL49 to show when "real" requests are made

carterkozak (approved):
As is I think this is an improvement.

dialogue-core/src/main/java/com/palantir/dialogue/core/Channels.java
carterkozak: 👍

PR #387, opened 2020-02-20
fix #301: QueuedChannel futures don't wait forever when delegates throw

### LINKED ISSUE #301: QueuedChannel result listeners can block forever when exceptions are thrown from delegates
## What happened?
Channels aren't supposed to throw. We have a backstop against bugs when they do, however the QueuedChannel needs to wrap delegates with a similar backstop, otherwise failures are not propagated to callers.

All futures should be completed eventually.

carterkozak:
## Before this PR
futures could never be completed.

## After this PR
==COMMIT_MSG==
QueuedChannel futures don't wait forever when delegates throw
==COMMIT_MSG==

### DIFF
--- a/dialogue-core/src/test/java/com/palantir/dialogue/core/QueuedChannelTest.java
+++ b/dialogue-core/src/test/java/com/palantir/dialogue/core/QueuedChannelTest.java
@@ -37,6 +37,7 @@ import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
 import org.mockito.Mock;
+import org.mockito.Mockito;
 import org.mockito.junit.jupiter.MockitoExtension;
 import org.mockito.stubbing.OngoingStubbing;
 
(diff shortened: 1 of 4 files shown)


dialogue-core/src/test/java/com/palantir/dialogue/core/QueuedChannelTest.java
```diff
@@ -114,6 +115,26 @@ public void testQueuedRequestExecutedOnNextSubmission() {
...
 
+    @Test
+    public void testQueuedRequestExecutedOnNextSubmission_throws() throws ExecutionException, InterruptedException {
+        Request queuedRequest = Mockito.mock(Request.class);
+        when(delegate.maybeExecute(endpoint, queuedRequest)).thenReturn(Optional.empty());
+        ListenableFuture<Response> queuedFuture = queuedChannel.execute(endpoint, queuedRequest);
+        verify(delegate, times(2)).maybeExecute(endpoint, queuedRequest);
+        assertThat(queuedFuture).isNotDone();
+
+        futureResponse.set(mockResponse);
```
ferozco, line 128: I would find it helpful to leave little comments explaining what you expect to be happening since just reading the test case requires tracing through the queuedChannel impl to understand why this is testing what it says it is

carterkozak: good call, updated.

ferozco:
👍

PR #388, opened 2020-02-20
client tolerates unexpected body

ferozco:
## Before this PR
If a client received a response but was not expecting one it would throw

## After this PR
==COMMIT_MSG==
Clients are tolerant of unexpected response body
==COMMIT_MSG==

## Possible downsides?
N/A

### DIFF
--- a/dialogue-client-test-lib/src/main/java/com/palantir/dialogue/AbstractSampleServiceClientTest.java
+++ b/dialogue-client-test-lib/src/main/java/com/palantir/dialogue/AbstractSampleServiceClientTest.java
@@ -142,6 +142,18 @@ public abstract class AbstractSampleServiceClientTest {
     }
 
     @Test
+    public void testBlocking_voidToVoid_doesNotThrowsWhenResponseBodyIsNonEmpty() {
+        server.enqueue(new MockResponse().setBody("Unexpected response"));
+        blockingClient.voidToVoid();
+    }
+
+    @Test
+    public void testAsync_voidToVoid_doesNotThrowWhenResponseBodyIsNonEmpty() throws Exception {
+        server.enqueue(new MockResponse().setBody("Unexpected response"));
+        assertThat(asyncClient.voidToVoid().get()).isNull();
+    }
+
+    @Test
     public void testAsync_stringToString_throwsWhenResponseBodyIsEmpty() {
         server.enqueue(new MockResponse().addHeader(Headers.CONTENT_TYPE, "application/json"));
         assertThatThrownBy(() ->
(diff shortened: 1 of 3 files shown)


dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ConjureBodySerDe.java
```diff
@@ -78,13 +78,7 @@
...
     public Deserializer<Void> emptyBodyDeserializer() {
         return input -> {
-            try (Response response = input) {
-                if (response.body().read() != -1) {
-                    throw new SafeRuntimeException("Expected empty response body");
-                }
-            } catch (IOException e) {
-                throw new SafeRuntimeException("Failed to read from response body", e);
-            }
+            input.close();
```
carterkozak, line 82: Let's leave a comment describing that `void`->`type` is allowed. Is there a quick test we can add?

ferozco: 👍 I'll add a comment, I already added tests :)

iamdanfox (approved):
Oof yep this would be a super annoying constraint from back-compat perspective!

dialogue-client-test-lib/src/main/java/com/palantir/dialogue/AbstractSampleServiceClientTest.java
```diff
@@ -141,6 +141,18 @@ public void testBlocking_stringToString_throwsWhenResponseBodyIsEmpty() {
                 .hasMessageContaining("Failed to deserialize response stream. Syntax error?");
     }
 
+    @Test
+    public void testBlocking_voidToVoid_doesNotThrowsWhenResponseBodyIsNonEmpty() {
```
uschi2000, line 145: nit: doesNotThrowWhen

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ConjureBodySerDe.java
```diff
@@ -78,13 +78,8 @@
...
     public Deserializer<Void> emptyBodyDeserializer() {
         return input -> {
-            try (Response response = input) {
-                if (response.body().read() != -1) {
-                    throw new SafeRuntimeException("Expected empty response body");
-                }
-            } catch (IOException e) {
-                throw new SafeRuntimeException("Failed to read from response body", e);
-            }
+            // We should not fail if a server that previously return nothing starts returning a response
```
uschi2000, line 81: nit: grammar is a little funky here

uschi2000: was there a bug here previously w.r.t. the missing `input.close()`? if so, can you add a test?

ferozco: No there was not a bug, it was previously closed by the try-with-resource. It was simply more concise to call close directly

PR #390, opened 2020-02-20
QueuedChannel.scheduleNextTask applies components.span

carterkozak:
Screen Shot 2020-02-20 at 11 02 27 AM

This propagates tracing information to the delegate request, and
avoids potential tracing state from the thread that invoked
scheduleNextTask.

## Before this PR
incorrect trace data associated with queued events

## After this PR
==COMMIT_MSG==
QueuedChannel.scheduleNextTask applies correct trace data
==COMMIT_MSG==


ferozco (commented):
👍 I think that this is better although I'm not a huge fan of polluting the trace with a bunch of wrapper spans to ensure that trace is propagated correctly. We might want to consider a typed way of passing detached spans between channels

PR #393, opened 2020-02-20
fix #392: QueuedChannel no longer leaks responses on cancel

### LINKED ISSUE #392: QueuedChannel leaks responses when cancel is invoked first
## What happened?

resource leak

Should not leak resources

carterkozak:
==COMMIT_MSG==
QueuedChannel no longer leaks responses on cancel
==COMMIT_MSG==

Still a lot of bugs here...

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
@@ -165,13 +165,17 @@ final class QueuedChannel implements Channel {
 
         @Override
         public void onSuccess(Response result) {
-            response.set(result);
+            if (!response.set(result)) {
+                result.close();
+            }
             schedule();
         }
 
         @Override
         public void onFailure(Throwable throwable) {
-            response.setException(throwable);
+            if (!response.setException(throwable)) {
+                log.info("Call failed after the future completed", throwable);
+            }
             schedule();
         }
     }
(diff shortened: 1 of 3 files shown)


dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
```diff
@@ -165,13 +165,17 @@ private boolean scheduleNextTask() {
 
         @Override
         public void onSuccess(Response result) {
-            response.set(result);
+            if (!response.set(result)) {
```
iamdanfox, line 168: time for an error-prone rule to flag this kind of thing to us?

carterkozak: Ya, I think so. It's a super common mistake, so rollout will be difficult more broadly.

I think we can suggest a fix of either:
`Preconditions.checkState(settableFuture.set[Exception](value), "failed to complete")`
or
```java
if (!settableFuture.set[Exception](value)) {
    log.warn("Failed to complete future", value); // requires a logger to be defined, otherwise the check becomes much more complicated.
}
```

iamdanfox: even if it's hard to roll out everywhere, I think I'd be happy to have it just in this repo to avoid our own foot guns!

iamdanfox (approved):

PR #394, opened 2020-02-20
fix #391: QueuedChannel correctly propagates cancellation

### LINKED ISSUE #391: QueuedChannel queued requests fail to propagate cancellation
## What happened?

the returned future may be cancelled but it has no impact on execution of the queued work.

Cancellation should be respected

carterkozak:
==COMMIT_MSG==
QueuedChannel correctly propagates cancellation
==COMMIT_MSG==

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
@@ -152,8 +152,8 @@ final class QueuedChannel implements Channel {
             return true;
         }
         try (CloseableSpan ignored = components.span().childSpan("Dialogue-request-scheduled")) {
-            Optional<ListenableFuture<Response>> maybeResponse =
-                    delegate.maybeExecute(components.endpoint(), components.request());
+            Endpoint endpoint = components.endpoint();
+            Optional<ListenableFuture<Response>> maybeResponse = delegate.maybeExecute(endpoint, components.request());
 
             if (maybeResponse.isPresent()) {
                 ListenableFuture<Response> response = maybeResponse.get();
@@ -167,7 +167,13 @@ final class QueuedChannel implements Channel {
                                 // TODO(ckozak): Consider capturing the argument value provided to cancel to propagate
                                 // here.
                                 // Currently cancel(false) will be converted to cancel(true)
-                                response.cancel(true);
+                                if (!response.cancel(true) && log.isDebugEnabled()) {
+                                    log.debug(
+                                            "Failed to cancel delegate response, it should be reported by"
+                                                    + " ForwardAndSchedule logging",
+                                            SafeArg.of("service", endpoint.serviceName()),
+                                            SafeArg.of("endpoint", endpoint.endpointName()));
+                                }
                             }
                         },
                         DIRECT);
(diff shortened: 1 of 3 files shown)


dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
```diff
@@ -144,16 +144,33 @@ private boolean scheduleNextTask() {
...
-            if (response.isPresent()) {
+            if (maybeResponse.isPresent()) {
+                ListenableFuture<Response> response = maybeResponse.get();
                 components.span().complete();
                 numRunningRequests.incrementAndGet();
-                response.get().addListener(numRunningRequests::decrementAndGet, DIRECT);
-                Futures.addCallback(response.get(), new ForwardAndSchedule(components.response()), DIRECT);
+                response.addListener(numRunningRequests::decrementAndGet, DIRECT);
+                Futures.addCallback(response, new ForwardAndSchedule(queuedResponse), DIRECT);
+                queuedResponse.addListener(
```
iamdanfox, line 164: so this is just saying if a request comes off the queue and gets sent down the delegate pipe, but the user decides to cancel the settablefuture they were holding, then we try and cancel the thing we sent down the delegate chain?

How will our lowest-level layer clients perceive this? does it just set the interrupted flag which we'd then hope something in apache would check?

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
```diff
@@ -144,16 +144,33 @@ private boolean scheduleNextTask() {
...
+                        () -> {
+                            if (queuedResponse.isCancelled()) {
+                                // TODO(ckozak): Consider capturing the argument value provided to cancel to propagate
+                                // here.
+                                // Currently cancel(false) will be converted to cancel(true)
+                                response.cancel(true);
```
iamdanfox: the javadoc on this says `@return {@code false} if the task could not be cancelled, typically because it has already completed normally;`... i wonder if it's worth putting a debug log line in that case? I feel like the timelock folks had so much time debugging cancellations that they'd want every bit of signal they could get

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
carterkozak: Since we use the blocking channel adapter, the thread running the request is interrupted which may result in the request stopping depending on the point in the lifecycle.

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
carterkozak: good call!

iamdanfox (approved):

dialogue-core/src/test/java/com/palantir/dialogue/core/QueuedChannelTest.java
```diff
@@ -212,18 +215,62 @@ public void emitsQueueCapacityMetrics_whenChannelHasCapacity() {
...
         queuedChannel.execute(endpoint, request);
-        verify(delegate, times(3)).maybeExecute(endpoint, request);
-        futureResponse.set(mockResponse);
+        assertThat(result).isCancelled();
+        verify(delegate, times(1)).maybeExecute(endpoint, request);
         verify(mockResponse, times(1)).close();
     }
 
+    @Test
+    @SuppressWarnings("FutureReturnValueIgnored")
```
uschi2000, line 240: nit: can probably add this at class-level?

PR #395, opened 2020-02-20
fix #386: Limit connections-per-host using a LimitedChannel

### LINKED ISSUE #386: Replace client implementation per-host connection limits with a LimitedChannel
## What happened?

Each client configures its own connection pool. These can be unlimited to avoid unexpected internal blocking/queueing in favor of a limited channel that applies to all channels.

The default per host should be 256 connections to match cjr-okhttp.

carterkozak:
## After this PR
==COMMIT_MSG==
Limit connections-per-host using a LimitedChannel
==COMMIT_MSG==

## Possible downsides?
Unlike the remoting-okhttp implementation there's no static global state, so two unique channels created for the same target will each get 256 requests.

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/Channels.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/Channels.java
@@ -27,11 +27,10 @@ import com.palantir.logsafe.exceptions.SafeRuntimeException;
 import java.util.Collection;
 import java.util.List;
 import java.util.function.Function;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 public final class Channels {
-    private static final Logger log = LoggerFactory.getLogger(Channels.class);
+
+    private static final int MAX_REQUESTS_PER_CHANNEL = 256;
 
     private Channels() {}
 
@@ -49,6 +48,7 @@ public final class Channels {
                 .map(channel -> new TracedChannel(channel, "Dialogue-http-request"))
                 .map(ContentDecodingChannel::new)
                 .map(concurrencyLimiter(config))
+                .map(channel -> new FixedLimitedChannel(channel, MAX_REQUESTS_PER_CHANNEL))
                 .collect(ImmutableList.toImmutableList());
 
         LimitedChannel limited = nodeSelectionStrategy(config, limitedChannels);
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/FixedLimitedChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/FixedLimitedChannel.java
@@ -42,7 +42,7 @@ final class FixedLimitedChannel implements LimitedChannel {
 
     @Override
     public Optional<ListenableFuture<Response>> maybeExecute(Endpoint endpoint, Request request) {
-        // Doesn't check for integer underflow, we don't have enough threads for that to be possible
+        // Doesn't check for integer overflow, we don't have enough threads for that to occur.
         if (availablePermits.decrementAndGet() < 0) {
             availablePermits.incrementAndGet();
             if (log.isDebugEnabled()) {
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/FixedLimitedChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/FixedLimitedChannel.java
@@ -1,0 +1,70 @@
+/*
+ * (c) Copyright 2020 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.palantir.dialogue.core;
+
+import com.google.common.util.concurrent.ListenableFuture;
+import com.google.common.util.concurrent.MoreExecutors;
+import com.palantir.dialogue.Endpoint;
+import com.palantir.dialogue.Request;
+import com.palantir.dialogue.Response;
+import com.palantir.logsafe.SafeArg;
+import java.util.Optional;
+import java.util.concurrent.atomic.AtomicInteger;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Channel with a fixed number of available permits. This can be used to enforce a per-route request limit.
+ */
+final class FixedLimitedChannel implements LimitedChannel {
+    private static final Logger log = LoggerFactory.getLogger(FixedLimitedChannel.class);
+
+    private final LimitedChannel delegate;
+    private final AtomicInteger availablePermits;
+
+    FixedLimitedChannel(LimitedChannel delegate, int permits) {
+        this.delegate = delegate;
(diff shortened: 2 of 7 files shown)


dialogue-core/src/main/java/com/palantir/dialogue/core/FixedLimitedChannel.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    @Override
+    public Optional<ListenableFuture<Response>> maybeExecute(Endpoint endpoint, Request request) {
+        // Doesn't check for integer underflow, we don't have enough threads for that to be possible
+        if (availablePermits.decrementAndGet() < 0) {
+            availablePermits.incrementAndGet();
+            if (log.isDebugEnabled()) {
+                log.debug(
+                        "Permits have been exhausted",
```
iamdanfox: permits as a safearg too?

dialogue-core/src/main/java/com/palantir/dialogue/core/FixedLimitedChannel.java
```diff
@@ -0,0 +1,70 @@
...
+
+    FixedLimitedChannel(LimitedChannel delegate, int permits) {
+        this.delegate = delegate;
+        this.availablePermits = new AtomicInteger(permits);
+    }
+
+    @Override
+    public Optional<ListenableFuture<Response>> maybeExecute(Endpoint endpoint, Request request) {
+        // Doesn't check for integer overflow, we don't have enough threads for that to occur.
+        if (availablePermits.decrementAndGet() < 0) {
```
iamdanfox: could we hoist this to a variable just to emphasise that it's stateful and affects both the if body and the fallthrough?

dialogue-core/src/main/java/com/palantir/dialogue/core/FixedLimitedChannel.java
```diff
@@ -0,0 +1,70 @@
...
+import org.slf4j.LoggerFactory;
+
+/**
+ * Channel with a fixed number of available permits. This can be used to enforce a per-route request limit.
+ */
+final class FixedLimitedChannel implements LimitedChannel {
+    private static final Logger log = LoggerFactory.getLogger(FixedLimitedChannel.class);
+
+    private final LimitedChannel delegate;
+    private final AtomicInteger availablePermits;
```
iamdanfox: Somehow my brain prefers the inversion of this, where this number starts at zero and counts the number of active requests?

dialogue-core/src/main/java/com/palantir/dialogue/core/FixedLimitedChannel.java
```diff
@@ -0,0 +1,70 @@
...
+            availablePermits.incrementAndGet();
+            if (log.isDebugEnabled()) {
+                log.debug(
+                        "Permits have been exhausted",
+                        SafeArg.of("service", endpoint.serviceName()),
+                        SafeArg.of("endpoint", endpoint.endpointName()));
+            }
+            return Optional.empty();
+        }
+        boolean resetPermit = true;
```
iamdanfox: what about flipping this round so that it's named `delegateRejectedRequest = false`, and then after we get an optional.empty we set it to true?

dialogue-core/src/main/java/com/palantir/dialogue/core/FixedLimitedChannel.java
iamdanfox: Just a thought - do you think a LongAdder would reduce contention and increase throughput here? I'm just thinking if all the witchcraft worker threads in a server are constantly blasting though these FixedLimitedChannels and there are two atomicinteger writes per call to maybeExecute, perhaps making the single read a bit more expensive by calling `LongAdder#sum()` would be worth it?

Quite possibly over-optimizing, just wanted to throw the idea out there

dialogue-core/src/main/java/com/palantir/dialogue/core/FixedLimitedChannel.java
```diff
@@ -0,0 +1,70 @@
...
+        try {
+            Optional<ListenableFuture<Response>> result = delegate.maybeExecute(endpoint, request);
+            if (result.isPresent()) {
+                result.get().addListener(availablePermits::incrementAndGet, MoreExecutors.directExecutor());
```
iamdanfox: do method references like this allocate each time?

iamdanfox:
> so two unique channels created for the same target will each get 256 requests.

We can always stick a cache in WC to prevent people from creating endless clients pointing to the same thing if we have to!

dialogue-core/src/main/java/com/palantir/dialogue/core/Channels.java
```diff
@@ -49,6 +48,7 @@ public static Channel create(
                 .map(channel -> new TracedChannel(channel, "Dialogue-http-request"))
                 .map(ContentDecodingChannel::new)
                 .map(concurrencyLimiter(config))
+                .map(channel -> new FixedLimitedChannel(channel, MAX_REQUESTS_PER_CHANNEL))
```
iamdanfox, line 51: would you mind dropping this into the simulator Strategy enum too please?

dialogue-core/src/main/java/com/palantir/dialogue/core/FixedLimitedChannel.java
carterkozak: LongAdder doesn't provide any concurrency guarantees, I believe reading the value is more expensive though writing is cheap. It works well for things like counters/meters because they have orders of magnitude more writes than reads, and the value doesn't need to be precise.

There's always AtomicIntegerFieldUpdater, but it performs similarly to AtomicInteger with lower allocation overhead for creation, we could use it if we wanted to optimize around fitting the entire channel chain into a cache line... but that's likely a bit too far.

dialogue-core/src/main/java/com/palantir/dialogue/core/Channels.java
carterkozak: good call. Perhaps we should overload channels with a method that takes clocks so we can keep them in sync?

dialogue-core/src/main/java/com/palantir/dialogue/core/FixedLimitedChannel.java
carterkozak: Yep (mostly, I think it's possible to get it inlined, but that takes a while and is fiddly). I'll update :-)

dialogue-core/src/main/java/com/palantir/dialogue/core/FixedLimitedChannel.java
carterkozak: That fails if the delegate channel throws. I should add a test for that case ;-)

dialogue-core/src/main/java/com/palantir/dialogue/core/FixedLimitedChannel.java
carterkozak: good call

dialogue-core/src/main/java/com/palantir/dialogue/core/Channels.java
iamdanfox: Yep I think my facade is gonna include a clock

dialogue-core/src/main/java/com/palantir/dialogue/core/FixedLimitedChannel.java
iamdanfox: On first read I was trying to figure out what fancy logic the ‘resetPermits’ thing was doing, like was it a backstop or something.

What if we called it “resetOptimisticallyConsumedPermit” or something?

carterkozak: 👍

iamdanfox (approved):
Out of curiosity, does this change any of the simulations? Or did nothing get to 256 requests/host?

carterkozak:
I don't think any of them hit concurrent 256 requests/host

PR #396, opened 2020-02-20
Clean up channel chain ergonomics

carterkozak:
ConcurrencyLimitedChannel is no longer a bridge between limited channels and standard channels, allowing us to more easily optimize order moving forward.
Shuffles channel order to avoid re-evaluating higher level functionality for each retry.
Fixes a potential leak in concurrency limited channel if the
next channel in the chain happens to throw.


iamdanfox (approved):

ferozco:
👍

PR #398, opened 2020-02-20
New entry-point: Dialogue.newClientPool()

iamdanfox:
## Before this PR

- We need to give people an easy way to build multiple clients that all use the same connection pool under the hood.
- There's quite a few of different little bits of API surface in dialogue, which makes it a little confusing for users to know what they're supposed to interact with
- We don't currently cleanup or dispose of client resources when we're done
- We also rely on the old c-j-r ClientConfiguration type everywhere

## After this PR
==COMMIT_MSG==
A new `Dialogue.newClientPool()` method gives people a single clear entrypoint to the library.
==COMMIT_MSG==

It people build clients in a declarative way that people are familiar with from feign:

```js
AsyncFooService foo = clients.get(AsyncFooService.class, listenableConfig);
AsyncOtherService other = clients.get(AsyncOtherService.class, listenableConfig);

// ... later, wc can call:
clients.close()
```

By getting conjure to generate _enums_, we can actually do this without using any sketchy reflection (no InvocationTargetExceptions anywhere!)

_Obviously this doesn't have enough tests to merge yet, just wanted to see what you guys think of the idea before I sink too much time into it!_

## Possible downsides?
- yet another refreshable implementation


dialogue-core/src/main/java/com/palantir/dialogue/core/ClientPool.java
```diff
@@ -0,0 +1,51 @@
...
+
+import com.palantir.dialogue.Channel;
+import java.io.Closeable;
+import java.io.IOException;
+
+/**
+ * Facilitates creating many clients which all share the same connection pool. Should only create one of these per
+ * server. Close it when your server shuts down to release resources.
+ */
+public interface ClientPool extends Closeable {
```
iamdanfox, line 27: This is probably a reasonable place to start reading this PR if you want

dialogue-core/src/main/java/com/palantir/dialogue/core/Dialogue.java
```diff
@@ -0,0 +1,35 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.dialogue.core;
+
+import com.google.errorprone.annotations.MustBeClosed;
+import com.palantir.dialogue.ConjureRuntime;
+
+public final class Dialogue {
```
iamdanfox, line 22: My dream is if we can make all the things people care about accessible from one top-level starting point, so they can just remember "the method is on Dialogue somewhere".

markelliot:
> We need to give people an easy way to build multiple clients that all use the same connection pool under the hood.

Is this actually desirable?

dialogue-core/src/main/java/com/palantir/dialogue/core/ClientPool.java
```diff
@@ -0,0 +1,51 @@
...
+     * appropriately to the various available nodes. Live-reloaded every time the urls change.
+     */
+    Channel smartChannel(Listenable<DialogueConfig> config);
+
+    /**
+     * Gets a direct channel to a single host within the specified Config. Live-reloads under the hood. The channel
+     * will always fail if the specified uri is not listed in the latest version of the config. Somewhat dangerous
+     * because this has no limits / failover.
+     */
+    Channel rawHttpChannel(String uri, Listenable<DialogueConfig> config);
```
markelliot, line 43: is this actually something that should be exposed?

dialogue-core/src/main/java/com/palantir/dialogue/core/ClientPool.java
```diff
@@ -0,0 +1,51 @@
...
+public interface ClientPool extends Closeable {
+
+    /** Returns an implementation of the given dialogueInterface, hooked up to a fresh smart channel underneath. */
+    <T> T get(Class<T> dialogueInterface, Listenable<DialogueConfig> config);
+
+    /**
+     * Returns a channel for interacting with the given abstract upstream service, which routes traffic
+     * appropriately to the various available nodes. Live-reloaded every time the urls change.
+     */
+    Channel smartChannel(Listenable<DialogueConfig> config);
```
markelliot, line 36: when does one use this instead of `get`?

iamdanfox:
I think so, because the average WC server has 38 clients according to datadog, but I think some have much fewer (like maybe 2) and some have even more.  This means any limits are gonna seem to apply quite differently.  It also tries to provide similar-ish behaviour to our old c-j-r okhttp world, where there was just one static instance.

> From the metric summary page in datadog (which shows you tag cardinality), we've got 26.7k witchcraft servers (reporting the `jvm.safepoint.time` metric) and 1,016,011 series of `client.response.count`. 1016011 / 26700 = 38

dialogue-core/src/main/java/com/palantir/dialogue/core/ClientPool.java
iamdanfox: Probably not to be honest - I started with this 3-member interface just to hash out in my head what the key building blocks were. Happy to push it down to package private.

dialogue-core/src/main/java/com/palantir/dialogue/core/Dialogue.java
markelliot: I think if that's the case it'd be better to return a ClientFactory that uses the provided ClientPool so that no user of Dialogue has to know anything about something called a ClientPool.

dialogue-apache-hc4-client/src/main/java/com/palantir/dialogue/hc4/DialogueApacheHttpClient.java
```diff
@@ -0,0 +1,334 @@
...
+    private static final String STORE = "DialogueApacheHttpClient";
+
+    @Override
+    public Channel construct(String uri, Listenable<DialogueConfig> config, SharedResources sharedResources) {
+        CloseableHttpClient client = sharedResources
+                .getStore(STORE)
+                .getOrComputeIfAbsent(
+                        "one-off-client-construction",
+                        unused -> {
+                            return constructLiveReloadingClient(config); // we'll re-use this instance every time
```
carterkozak, line 85: I think we need a unique CloseableHttpClient instance for each ssl configuration -- each service can have different trust data. A few other prefs from the configuration are applied directly to the CloseableHttpClient so they will require unique instances

dialogue-core/src/main/java/com/palantir/dialogue/core/ClientPool.java
carterkozak, line 36: I think we should have some API separation between channel creation and using a channel to create a generated client (which would delegate to the first component)

dialogue-core/src/main/java/com/palantir/dialogue/core/ClientPoolImpl.java
```diff
@@ -0,0 +1,115 @@
...
+                "@ConstructUsing annotation must be present on interface",
+                SafeArg.of("interface", dialogueInterface.getName()));
+
+        Class<?> factoryClass = annotation.value();
+        // this is safe because of the ConstructUsing annotation's type parameters
+        Factory<T> factory = (Factory<T>) getOnlyEnumConstant(factoryClass);
+        return factory.construct(smartChannel, runtime);
+    }
+
+    private static <F> F getOnlyEnumConstant(Class<F> factoryClass) {
```
carterkozak, line 105:
```suggestion
    private static <F extends Enum<F>> F getOnlyEnumConstant(Class<F> factoryClass) {
```

dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueConfig.java
```diff
@@ -0,0 +1,107 @@
...
+
+        private final String className;
+
+        HttpClientType(String className) {
+            this.className = className;
+        }
+
+        Class<? extends HttpChannelFactory> getFactoryClass() {
+            try {
+                return (Class<? extends HttpChannelFactory>) Class.forName(className);
```
uschi2000, line 97: why do we need reflection here?

uschi2000: i.e., why not `Supplier<Channel>`?

dialogue-core/src/main/java/com/palantir/dialogue/core/Dialogue.java
uschi2000: it's not so clear to me that this code should live in dialogue land rather than in wc?

iamdanfox: @uschi2000 are you referring to just this one entrypoint or the whole PR?

dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueConfig.java
iamdanfox: I didn't want people who depend on dialogue-core to transitively get dialogue-apache, dialogue-java, dialogue-okhttp etc.

Might be able to doing this with a compileOnly dependency if you'd prefer?

dialogue-core/src/main/java/com/palantir/dialogue/core/Dialogue.java
uschi2000: I think the idea of a "singleton client factory", i.e., something that holds on to resources for the lifetime of a process, is something I'd expect in a server library, not a client library.

dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueConfig.java
uschi2000: I'm not a huge fan of runtime dependencies like this. might indicate that the decomp is a little off, and maybe also that this should live in the server repo?

dialogue-core/src/main/java/com/palantir/dialogue/core/ClientPoolImpl.java
iamdanfox: The only way I can get this to work is if I then make the argument `Class<?>`, which seems a bit less nice?

image

dialogue-core/src/main/java/com/palantir/dialogue/core/Dialogue.java
iamdanfox: I think if we try to make all our libraries good enough to use on their own, then when we pull them into WC they'll still be good. I actually think a client library that doesn't give you a convenient way to release resources when you want to seems broken/incomplete.

To be clear this API doesn't know anything about the lifetime of a process, and there are no static singletons here.  it would still be up to witchcraft to make sure there's one client pool and witchcrafts responsibility to close it at shutdown time.

uschi2000: `Dialogue` is designed to be a singleton, no?
Yes, it's a bug today that we don't close channels, but then why isn't it sufficient for this library to add `Channel#close`? I prefer hierarchical ownership (with delegating `.close()`) over complex "inside-out" resource management; is that not an option here?

iamdanfox:
Closing out as I'm resuscitating my PR to do this in WC

PR #401, opened 2020-02-21
Blacklisting channel notifies the queue when it can handle events

carterkozak:
## Before this PR
Blacklisting channel left us in a state where requests could be queued forever.

## After this PR
==COMMIT_MSG==
Blacklisting channel notifies the queue when it can handle events
==COMMIT_MSG==

## Possible downsides?
This implementation is a bit dirty, the circular references aren't ideal.

Alternatives considered:
- subscriber model, though there's no reason to subscribe multiple queues. Potential for memory leaks upon later code changes if unsubscribe isn't called.
- scheduler owned by the queue running at a fixed rate. This can cause hiccups+ooms on high-throughput applications in the time between unblacklisting and requests beginning to flow again, only the channel which limits requests can tell us when it's ready to try again.
- LimitedChannel return value union of success/limit with a callback. This feels cleaner, however the queue must track a limited response listener for every request it has attempted, exploding complexity.


simulation/src/test/resources/report.md
```diff
@@ -1,7 +1,7 @@
 # Report
 <!-- Run SimulationTest to regenerate this report. -->
 ```
-          all_nodes_500[CONCURRENCY_LIMITER_BLACKLIST_ROUND_ROBIN].txt:	success=78.1%	client_mean=PT2.54955S     	server_cpu=PT20M          	client_received=2000/2000	server_resps=2000	codes={200=1562, 500=438}
+          all_nodes_500[CONCURRENCY_LIMITER_BLACKLIST_ROUND_ROBIN].txt:	success=70.6%	client_mean=PT1.7802S      	server_cpu=PT20M          	client_received=2000/2000	server_resps=2000	codes={200=1411, 500=589}
```
iamdanfox, line 4: this looks kinda better, although success has gone down due to stuff being executed more promptly https://github.com/palantir/dialogue/blob/ckozak/blacklist_signal/simulation/src/test/resources/report.md#all_nodes_500concurrency_limiter_blacklist_round_robin

dialogue-core/src/main/java/com/palantir/dialogue/core/Channels.java
```diff
@@ -39,6 +39,9 @@ public static Channel create(
         Preconditions.checkState(!channels.isEmpty(), "channels must not be empty");
 
         DialogueClientMetrics clientMetrics = DialogueClientMetrics.of(config.taggedMetricRegistry());
+        // n.b. This becomes cleaner once we support reloadable channels, the queue can be created first, and
```
ferozco, line 42: I think it would be helpful to actually wire up the blacklisting channel so that it is used outside of tests

ferozco:
👍 lets try it out

dialogue-core/src/main/java/com/palantir/dialogue/core/Channels.java
carterkozak: It's funky that a single failure can derail a blacklisting channel, perhaps we could use a sliding window if >20% of requests over that period resulted in failures?

ferozco: Sure we could add more complex heuristics, but I think that the right long term solution is to go back to per endpoint blacklisting once we address the risk of immediately retrying on the same node

simulation/src/main/java/com/palantir/dialogue/core/Simulation.java
```diff
@@ -80,6 +81,10 @@ public Ticker clock() {
         return ticker; // read only!
     }
 
+    public ScheduledExecutorService scheduler() {
+        return deterministicExecutor;
+    }
```
iamdanfox, line 86: Hang on exposing this directly might interact weirdly with the simulator, because the above `schedule` methods ensure that the clock is set correctly, whereas direct access means that a scheduled runnable could call clock.read() and get an old value!!

carterkozak: ack

PR #402, opened 2020-02-21
Centralize status code checks in a utility class

carterkozak:
Functional changes:
* ConcurrencyLimitedChannel uses both 503 and 429 (was only 429)
* Blacklisting channel reacts to 429 and all 5xx response codes
  (was 503 and 500)
* PinUntilErrorChannel rotates on 429 and all 500 response codes.
  (was everything >= 300)

## After this PR
==COMMIT_MSG==
Centralize status code checks in a utility class
==COMMIT_MSG==

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/Responses.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/Responses.java
@@ -1,0 +1,43 @@
+/*
+ * (c) Copyright 2020 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.palantir.dialogue.core;
+
+import com.palantir.dialogue.Response;
+
+/** Utility functionality for {@link Response} handling. */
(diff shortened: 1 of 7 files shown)


dialogue-core/src/main/java/com/palantir/dialogue/core/Responses.java
```diff
@@ -0,0 +1,43 @@
...
+final class Responses {
+
+    private static final int TOO_MANY_REQUESTS_429 = 429;
+    private static final int UNAVAILABLE_503 = 503;
+
+    static boolean isTooManyRequests(Response response) {
+        return response.code() == TOO_MANY_REQUESTS_429;
+    }
+
+    static boolean isUnavailable(Response response) {
```
iamdanfox, line 30: could be private?

dialogue-core/src/main/java/com/palantir/dialogue/core/Responses.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+package com.palantir.dialogue.core;
+
+import com.palantir.dialogue.Response;
+
+/** Utility functionality for {@link Response} handling. */
+final class Responses {
+
+    private static final int TOO_MANY_REQUESTS_429 = 429;
+    private static final int UNAVAILABLE_503 = 503;
```
iamdanfox, line 24: I think I'd be fine with just inlining these if you want - the method name provides a nice english description of what they mean, so I think this is quite readable:
```java
    static boolean isTooManyRequests(Response response) {
        return response.code() == 429;
    }
```

iamdanfox (approved):

dialogue-core/src/main/java/com/palantir/dialogue/core/Responses.java
carterkozak: :+1:

dialogue-core/src/main/java/com/palantir/dialogue/core/Responses.java
carterkozak: :+1:

Issue #407, opened 2020-02-23
BlacklistingChannel and QueuedChannel interact poorly


carterkozak:
## What happened?
QueuedChannel assumes all requests (channel+endpoint combinations) have equal likelyhood to succeed.
BlacklistingChannel may blacklist a single endpoint without blacklisting the rest of the channel, so the head of the queued channel blocks all requests to a given channel.

## What did you want to happen?

QueuedChannel and BlacklistingChannel should agree.

carterkozak:
Applies to ConcurrencyLimitedChannel as well.

carterkozak:
fixed via #410

PR #409, opened 2020-02-23
Record metrics any time a channel limits an outgoing request

carterkozak:
## Before this PR
Difficult to tell which LimitedChannel resulted in a delayed request.

## After this PR
==COMMIT_MSG==
Record metrics any time a channel limits an outgoing request
==COMMIT_MSG==

## Possible downsides?
This only adds support for ConcurrencyLimitedChannel and
FixedLimitChannel to avoid conflicts with work that's currently
in progress.

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/FixedLimitedChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/FixedLimitedChannel.java
@@ -43,7 +43,7 @@ final class FixedLimitedChannel implements LimitedChannel {
     FixedLimitedChannel(LimitedChannel delegate, int totalPermits, DialogueClientMetrics metrics) {
         this.delegate = delegate;
         this.totalPermits = totalPermits;
-        this.limitedMeter = metrics.limited("FixedLimitedChannel");
+        this.limitedMeter = metrics.limited(getClass().getSimpleName());
         // Doesn't check for integer overflow, we don't have enough threads for that to occur.
         Preconditions.checkArgument(totalPermits <= 1_000_000, "total permits must not exceed one million");
         this.returnPermit = usedPermits::decrementAndGet;
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/FixedLimitedChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/FixedLimitedChannel.java
@@ -15,6 +15,7 @@
  */
 package com.palantir.dialogue.core;
 
+import com.codahale.metrics.Meter;
 import com.google.common.util.concurrent.ListenableFuture;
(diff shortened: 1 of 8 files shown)


dialogue-core/src/main/metrics/dialogue-core-metrics.yml
```diff
@@ -23,3 +23,7 @@ namespaces:
         type: meter
         tags: [service-name]
         docs: Rate of deprecated endpoints being invoked.
+      limited:
+        type: meter
+        tags: [reason]
+        docs: Rate of client-side request rejection.
```
iamdanfox: We might need to tweak this a little as it sounds quite scary, especially if people don't know about the retries or queue.

dialogue-core/src/main/java/com/palantir/dialogue/core/FixedLimitedChannel.java
```diff
@@ -51,6 +54,7 @@
         boolean optimisticallyAcquiredPermit = usedPermits.incrementAndGet() > totalPermits;
         if (optimisticallyAcquiredPermit) {
             returnPermit.run();
+            limitedMeter.mark();
```
iamdanfox, line 57: Should these be tagged by endpoint / service ? I know it would be expensive, but I'm imagining looking at a datadog graph and seeing a big chunk of limits, but not knowing whether they're evenly spread across all outgoing requests or just a few of them?

dialogue-core/src/main/java/com/palantir/dialogue/core/FixedLimitedChannel.java
```diff
@@ -35,12 +36,14 @@
...
     private final AtomicInteger usedPermits = new AtomicInteger(0);
+    private final Meter limitedMeter;
     private final int totalPermits;
     private final Runnable returnPermit;
 
-    FixedLimitedChannel(LimitedChannel delegate, int totalPermits) {
+    FixedLimitedChannel(LimitedChannel delegate, int totalPermits, DialogueClientMetrics metrics) {
         this.delegate = delegate;
         this.totalPermits = totalPermits;
+        this.limitedMeter = metrics.limited("FixedLimitedChannel");
```
iamdanfox: `FixedLimitedChannel.class.getSimpleName()` ? Then it will get auto-refactored nicely for us?

iamdanfox (approved):
Big fan of having more metrics :)

dialogue-core/src/main/java/com/palantir/dialogue/core/FixedLimitedChannel.java
carterkozak: I'm not sure we should add endpoint data per #407. Service might be alright, for now I'd like to record some data so we can figure out where things are getting flagged.

ferozco (approved):

PR #410, opened 2020-02-23
LimitedChannels operate on channels, ignoring Endpoint and Request

carterkozak:
Related to #407

==COMMIT_MSG==
Update LimitedChannel documentation
==COMMIT_MSG==


dialogue-core/src/main/java/com/palantir/dialogue/core/LimitedChannel.java
```diff
@@ -26,6 +26,9 @@
  * Similar to {@link com.palantir.dialogue.Channel}, but may not actually execute the call (eg: when the channel is
  * overloaded). Semantics match {@link com.palantir.dialogue.Channel} aside from returning an
  * {@link Optional optional response}.
+ * Limited channels must limit exclusively based on the state of the {@link com.palantir.dialogue.Channel}, not
+ * the {@link Endpoint} or {@link Request} arguments, otherwise the caller (generally a {@link QueuedChannel})
```
ferozco, line 30: Given our conversation this morning, shall we also update the blacklisting channel to adhere to this spec change?

carterkozak: Might as well :-)

iamdanfox: Wait but are you really gonna rip out my per endpoint blacklisting but leave per-endpoint concurrency limiters?

carterkozak: We can get it back, but we'll need a separate queueing + limiting mechanism. For what it's worth, I think we want the ability to both limit entire channels as well as per-endpoint separately using similar mechanisms triggered with different configurations (rate of failure, response codes, duration of the blacklist etc)

carterkozak: I figured we'd pull both for now, we can do something else if you like.

dialogue-core/src/main/java/com/palantir/dialogue/core/BlacklistingChannel.java
```diff
@@ -50,7 +49,7 @@
     private final LimitedChannel delegate;
     private final Duration duration;
     private final Ticker ticker;
-    private final Cache<Endpoint, BlacklistState> perEndpointBlacklistState;
+    private final AtomicReference<BlacklistState> channelBlacklistState;
```
carterkozak, line 52: This could be volatile since nothing takes advantage of CAS operations. I think we should be using CAS to prevent multiple failures clobbering state, but that could be done in a follow up (current semantics match the cache, just more obvious where we can improve now)

dialogue-core/src/main/java/com/palantir/dialogue/core/ConcurrencyLimitedChannel.java
```diff
@@ -46,15 +43,14 @@
...
+    private final Limiter<Void> limiter;
 
     @VisibleForTesting
     ConcurrencyLimitedChannel(
             LimitedChannel delegate, Supplier<Limiter<Void>> limiterSupplier, DialogueClientMetrics metrics) {
         this.delegate = new NeverThrowLimitedChannel(delegate);
         this.limitedMeter = metrics.limited(getClass().getSimpleName());
-        this.limiters =
-                Caffeine.newBuilder().expireAfterAccess(Duration.ofMinutes(5)).build(key -> limiterSupplier.get());
+        this.limiter = limiterSupplier.get();
```
carterkozak: Remove supplier, we can pass the Limiter object.

ferozco:
👍

PR #416, opened 2020-02-24
fix #406: QueuedChannel does not leak responses in edge cases

### LINKED ISSUE #406: QueuedChannel leaks queued requests and throws IllegalStateException when full
## What happened?

When the queue is filled while the first element is attempted, and the first element cannot be executed (limiter is still limiting calls, the most likely case), it cannot be pushed back to the queue and an IllegalStateException is thrown.

See
https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/BlockingDeque.html#addFirst-E-

At minimum nothing should throw and nothing should leak.

We may want to consider using racily updated size estimate to limit the queue size, and always allow elements from the front of the queue to be re-added.

carterkozak:
Updated to use the more scalable ConcurrentLinkedQueue with a
separate variable to track the estimated size. This allows us to
avoid decrementing after attempting to send a request until we
know the request has not been rejected. This also allows us
to make a lock-free read when events are enqueued to avoid the
fast path when the queue is not empty.

## After this PR
==COMMIT_MSG==
QueuedChannel does not leak responses in edge cases
==COMMIT_MSG==

## Possible downsides?
We must be careful not to leak queue permits, the atomic integer and deque state are closely coupled. We could build a simple data structure that encapsulates both, but we could lose fringe benefits (it's nice to pull the first element from the deque without modifying the size).


dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
```diff
@@ -78,9 +81,11 @@
     @SuppressWarnings("FutureReturnValueIgnored")
     QueuedChannel(LimitedChannel delegate, int maxQueueSize, DispatcherMetrics metrics) {
         this.delegate = new NeverThrowLimitedChannel(delegate);
-        this.queuedCalls = new LinkedBlockingDeque<>(maxQueueSize);
+        // Do _not_ call size on a ConcurrentLinkedDeque. Unlike other collections, size is an O(n) operation.
+        this.queuedCalls = new ConcurrentLinkedDeque<>();
```
carterkozak: I considered subclassing this to throw if `size()` is called. Thoughts?

Longer term we may want to consider something along the lines of https://github.com/undertow-io/undertow/blob/21035b2ee38e703d82dc74292853acfbdbffc46b/core/src/main/java/io/undertow/util/FastConcurrentDirectDeque.java to remove items from the queue immediately (O(1)) when `future.cancel` is invoked. That's a journey for another day though.

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
```diff
@@ -91,12 +95,20 @@
...
-            return result;
+        if (queueSizeEstimate.get() <= 0) {
+            Optional<ListenableFuture<Response>> maybeResult = delegate.maybeExecute(endpoint, request);
+            if (maybeResult.isPresent()) {
+                ListenableFuture<Response> result = maybeResult.get();
+                numRunningRequests.incrementAndGet();
+                result.addListener(this::onCompletion, DIRECT);
+                return result;
+            }
+        }
```
iamdanfox, line 106: this is a behaviour change right? where previously requests could jump the queue if the channel happened to open up, but they're now more FIFO?

carterkozak: That's correct, it's why the tests changed. I avoided checking the queue size using the old implementation because LinkedBlockingQueue acquires locks to check the size, but this is a quick volatile read.

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
```diff
@@ -149,13 +167,15 @@ private boolean scheduleNextTask() {
         // There's a race where cancel may be invoked between this check and execution, but the scheduled
         // request will be quickly cancelled in that case.
         if (queuedResponse.isDone()) {
+            queueSizeEstimate.decrementAndGet();
             return true;
         }
         try (CloseableSpan ignored = components.span().childSpan("Dialogue-request-scheduled")) {
             Endpoint endpoint = components.endpoint();
```
iamdanfox: could we rename `components` to `headOfTheQueue` or something? components seems unnecessarily ambiguous

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
```diff
@@ -179,7 +199,24 @@ private boolean scheduleNextTask() {
...
+                                    "Failed to req-queue request",
+                                    SafeArg.of("service", endpoint.serviceName()),
+                                    SafeArg.of("endpoint", endpoint.endpointName())))) {
+                        log.debug(
+                                "Queued response has already been completed",
+                                SafeArg.of("service", endpoint.serviceName()),
+                                SafeArg.of("endpoint", endpoint.endpointName()));
+                    }
+                }
                 return false;
```
iamdanfox, line 220: you know the thing we talked about in today's zoom, where if we can't make progress on the head request due to endpoint-level blacklisting, then the entire queue can't make progress...

if we just iterated through the whole queue here rather than just stopping on the first one what was limited, wouldn't this problem go away? (Could use the weird iterator#remove method if it succeeds)

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
iamdanfox: +1 to subclassing, and if possible could your subclass actually just keep track of the queueSizeEstimate? It would clear out a few incrementAndGet/decrementAndGet calls from this file

carterkozak: It could, but as mentioned in the PR description we would lose the ability to pull the head element and attempt to execute it without modifying the estimated size, so the queue would slowly grow when the queue is supposed to be full

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
carterkozak: 👍

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
carterkozak: We attempt to schedule a _lot_. In normal operation, the queue is empty so there's nothing to worry about. When channels are limiting most requests, turning this from an O(1) operation to an O(n) operation would be problematic.

iamdanfox (approved):

PR #418, opened 2020-02-24
Simulate low latency with 1% server error rate

carterkozak:
==COMMIT_MSG==
Simulate low latency with 1% server error rate
==COMMIT_MSG==


simulation/src/test/java/com/palantir/dialogue/core/SimulationTest.java
```diff
@@ -367,6 +370,43 @@ public void live_reloading() {
...
+                .sendUntil(Duration.ofSeconds(10))
+                .clients(10, i -> strategy.getChannel(simulation, servers))
+                .simulation(simulation)
+                .abortAfter(Duration.ofSeconds(10))
+                .run();
+    }
+
+    private Function<SimulationServer, Response> respond500Every(int reqs) {
+        return new Function<SimulationServer, Response>() {
+            private final AtomicInteger count = new AtomicInteger();
```
iamdanfox: I've been using an int array with a single element to emphasise this entire thing is single threaded

simulation/src/test/java/com/palantir/dialogue/core/SimulationTest.java
```diff
@@ -367,6 +370,43 @@ public void live_reloading() {
...
+        servers = servers(
+                SimulationServer.builder()
+                        .serverName("fast0")
+                        .simulation(simulation)
+                        .handler(h -> h.response(respond500Every(100)).responseTime(Duration.ofNanos(1000)))
+                        .build(),
+                SimulationServer.builder()
+                        .serverName("fast1")
+                        .simulation(simulation)
+                        .handler(h -> h.response(respond500Every(100)).responseTime(Duration.ofNanos(1000)))
```
iamdanfox: i've found anything that is exactly regular like this can have weird effects, e.g. if you have two nodes and pure round robin, then all of these 500s will exactly land on one of the nodes. Would recommend using a pseudorandom instance to put a bit of jitter into these (e.g. perhaps varying between 90 and 110?)

simulation/src/test/java/com/palantir/dialogue/core/SimulationTest.java
```diff
@@ -367,6 +370,43 @@ public void live_reloading() {
...
+        result = Benchmark.builder()
+                .requestsPerSecond(1000)
+                .sendUntil(Duration.ofSeconds(10))
+                .clients(10, i -> strategy.getChannel(simulation, servers))
+                .simulation(simulation)
+                .abortAfter(Duration.ofSeconds(10))
+                .run();
+    }
+
+    private Function<SimulationServer, Response> respond500Every(int reqs) {
```
iamdanfox: I think it's fine to drop this as a default method into the HandlerBuilder0 interface or whatever it was

carterkozak: 500 is an odd case, I'd like to test what happens when there's a connect or socket exception, but that doesn't appear to be implemented yet. I don't think most of our limiters actually care about 500.

simulation/src/test/java/com/palantir/dialogue/core/SimulationTest.java
carterkozak: :+1:

simulation/src/test/resources/report.md
```diff
@@ -37,6 +37,10 @@ slowdown_and_error_thresholds[CONCURRENCY_LIMITER_BLACKLIST_ROUND_ROBIN].txt:	su
...
               slowdown_and_error_thresholds[UNLIMITED_ROUND_ROBIN].txt:	success=1.2%	client_mean=PT3.974119999S 	server_cpu=PT11H2M21.19999998S	client_received=10000/10000	server_resps=10000	codes={200=120, 500=9880}
+        uncommon_flakes[CONCURRENCY_LIMITER_BLACKLIST_ROUND_ROBIN].txt:	success=97.6%	client_mean=PT0.125642485S 	server_cpu=PT0.009853S    	client_received=9853/10000	server_resps=9853	codes={200=9755, 500=98}
+              uncommon_flakes[CONCURRENCY_LIMITER_PIN_UNTIL_ERROR].txt:	success=99.0%	client_mean=PT0.000001S    	server_cpu=PT0.01S        	client_received=10000/10000	server_resps=10000	codes={200=9901, 500=99}
+                  uncommon_flakes[CONCURRENCY_LIMITER_ROUND_ROBIN].txt:	success=99.0%	client_mean=PT0.000001S    	server_cpu=PT0.01S        	client_received=10000/10000	server_resps=10000	codes={200=9901, 500=99}
+                            uncommon_flakes[UNLIMITED_ROUND_ROBIN].txt:	success=99.0%	client_mean=PT0.000001S    	server_cpu=PT0.01S        	client_received=10000/10000	server_resps=10000	codes={200=9901, 500=99}
```
iamdanfox: graphs link: https://github.com/palantir/dialogue/blob/ckozak/rapid_fire/simulation/src/test/resources/report.md#uncommon_flakesconcurrency_limiter_blacklist_round_robin

simulation/src/test/java/com/palantir/dialogue/core/SimulationTest.java
iamdanfox: Feel free to just return a response containing that exception! No need to stick to status codes

ferozco:
Graphs for anyone interested: https://github.com/palantir/dialogue/blob/ckozak/rapid_fire/simulation/src/test/resources/report.md

simulation/src/test/java/com/palantir/dialogue/core/SimulationTest.java
carterkozak: It doesn't look like most of our limiters care about exceptions yet, I'll leave this as is for the time being

iamdanfox (approved):

PR #420, opened 2020-02-24
Simplify BlacklistingChannel

carterkozak:
More thread safety, use compare-and-swap where possible, reuse the instance. This change prevents us from mutating state based on old permits.

==COMMIT_MSG==
Simplify BlacklistingChannel
==COMMIT_MSG==

## Possible downsides?
Could introduce bugs.


dialogue-core/src/main/java/com/palantir/dialogue/core/BlacklistingChannel.java
```diff
@@ -166,43 +136,59 @@ public String toString() {
 
     private final class BlacklistState {
```
ferozco, line 113: implements `LimitedChannel`

dialogue-core/src/main/java/com/palantir/dialogue/core/BlacklistingChannel.java
```diff
@@ -70,6 +70,30 @@
...
+
+        @Override
+        public void onSuccess(Response response) {
+            // TODO(jellis): use the Retry-After header (if present) to determine how long to blacklist the channel
+            if (Responses.isQosStatus(response) || Responses.isServerError(response)) {
+                log.debug(
+                        "Blacklisting {} due to status code {}",
+                        UnsafeArg.of("delegate", delegate),
+                        SafeArg.of("code", response.code()));
+                channelBlacklistState.set(new BlacklistState(duration, NUM_PROBATION_REQUESTS));
```
ferozco: In cases where there are multiple requests in flight and they all fail, we will rest the blacklist multiple times. Should we CAS to start blacklisting on the first failure?

carterkozak: I went back and forth on this, I think you're right though and we should avoid resetting the timeout when requests come in.

dialogue-core/src/main/java/com/palantir/dialogue/core/BlacklistingChannel.java
carterkozak: I'm not sure that buys us anything, does it?

dialogue-core/src/main/java/com/palantir/dialogue/core/BlacklistingChannel.java
```diff
@@ -166,43 +136,59 @@ public String toString() {
...
+                return delegate.maybeExecute(endpoint, request)
+                        .map(future -> DialogueFutures.addDirectCallback(future, blacklistingCallback));
+            } else {
+                log.debug("Probation channel request not allowed");
+                return Optional.empty();
+            }
         }
 
-        BlacklistStage maybeProgressAndGet() {
+        Optional<Probation> maybeProgressAndGet() {
```
iamdanfox: rename to `isBlacklistingComplete` ? or `maybeBeginProbation`?

dialogue-core/src/main/java/com/palantir/dialogue/core/BlacklistingChannel.java
```diff
@@ -166,43 +136,59 @@ public String toString() {
...
             if (inProbation.get()) {
-                return probation;
+                return Optional.of(probation);
             }
-            if (ticker.read() >= blacklistUntil.untilNanos) {
+            if (ticker.read() >= blacklistUntilNanos) {
                 if (inProbation.compareAndSet(false, true)) {
```
iamdanfox: elsewhere we have debug logs for interesting transitions - can we include one here for beginning probation?

dialogue-core/src/main/java/com/palantir/dialogue/core/BlacklistingChannel.java
```diff
@@ -166,43 +136,59 @@ public String toString() {
 
     private final class BlacklistState {
         private final AtomicBoolean inProbation = new AtomicBoolean();
```
iamdanfox: why do we need a boolean and the `Probation` field - wouldn't an AtomicReference<Probation> or maybe volatile Probation do the trick using null or Optional.empty to signify that we're not in probation?

dialogue-core/src/main/java/com/palantir/dialogue/core/BlacklistingChannel.java
carterkozak: good call

dialogue-core/src/main/java/com/palantir/dialogue/core/BlacklistingChannel.java
carterkozak: Updated to use an AtomicReference for probation

dialogue-core/src/main/java/com/palantir/dialogue/core/BlacklistingChannel.java
carterkozak: 👍

dialogue-core/src/main/java/com/palantir/dialogue/core/BlacklistingChannel.java
carterkozak: Updated, it makes some of the simulations better :-)

iamdanfox (approved):

PR #422, opened 2020-02-24
Differentiate between limited and blacklisted requests

ferozco:
## Before this PR
With blacklisting enabled we would could fail to the over all request without actually ever making a remote call if we hit a blacklisted channel on each retry

## After this PR
==COMMIT_MSG==
Enable channel blacklisting by default and do not count requests to blacklisted hosts against retry limit
==COMMIT_MSG==


dialogue-core/src/main/java/com/palantir/dialogue/core/Channels.java
```diff
@@ -42,14 +42,15 @@ public static Channel create(
...
+        List<CompositeLimitedChannel> limitedChannels = channels.stream()
                 // Instrument inner-most channel with metrics so that we measure only the over-the-wire-time
                 .map(channel -> new InstrumentedChannel(channel, clientMetrics))
                 // TracedChannel must wrap TracedRequestChannel to ensure requests have tracing headers.
                 .map(TracedRequestChannel::new)
                 .map(channel -> new TracedChannel(channel, "Dialogue-http-request"))
-                .map(LimitedChannelAdapter::new)
+                .map(CompositeLimitedChannelAdapter::new)
                 .map(concurrencyLimiter(config, clientMetrics))
+                .map(channel -> new BlacklistingChannel(channel, config.failedUrlCooldown(), queueListener))
```
carterkozak, line 53: after any failure, this will put the channel into probation mode by default (failedUrlCooldown == 0 using the default configuration).
This could be problematic for timelock, perhaps we should check if clientQos is disabled for this?

dialogue-core/src/main/java/com/palantir/dialogue/core/RoundRobinChannel.java
```diff
@@ -46,11 +50,14 @@
...
-            LimitedChannel channel = delegates.get(toIndex(host + i));
-            Optional<ListenableFuture<Response>> maybeCall = channel.maybeExecute(endpoint, request);
-            if (maybeCall.isPresent()) {
-                return maybeCall;
+            CompositeLimitedChannel channel = delegates.get(toIndex(host + i));
+            LimitedResponse maybeCall = channel.maybeExecute(endpoint, request);
+            if (maybeCall.matches(isBlacklisted)) {
+                continue;
+            } else if (maybeCall.matches(isLimited)) {
+                return Optional.empty();
```
carterkozak, line 58: If a response is limited by one channel (too many concurrent requests?) we don't attempt to rotate to the next channel?

dialogue-core/src/main/java/com/palantir/dialogue/core/Channels.java
ferozco: ya good catch

dialogue-core/src/main/java/com/palantir/dialogue/core/CompositeLimitedChannel.java
```diff
@@ -0,0 +1,24 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.dialogue.core;
+
+import com.palantir.dialogue.Endpoint;
+import com.palantir.dialogue.Request;
+
+public interface CompositeLimitedChannel {
```
carterkozak, line 22: package private

dialogue-core/src/main/java/com/palantir/dialogue/core/LimitedResponse.java
```diff
@@ -0,0 +1,34 @@
...
+package com.palantir.dialogue.core;
+
+import com.google.common.util.concurrent.ListenableFuture;
+import com.palantir.dialogue.Response;
+import org.derive4j.Data;
+
+@Data
+public interface LimitedResponse {
+    interface Cases<T> {
+        T blacklisted();
```
carterkozak, line 26: Is this only set by blacklisting channel? As a consumer how do I handle this differently from a limited response? The blacklisted channel will become un-blacklisted eventually, much like a limited channel will open up.

I think docs would help me a lot.

dialogue-core/src/main/java/com/palantir/dialogue/core/CompositeLimitedChannel.java
carterkozak: Is there any reason we should have both LimitedChannel and CompositeLimitedChannel?

dialogue-core/src/main/java/com/palantir/dialogue/core/LimitedResponse.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+package com.palantir.dialogue.core;
+
+import com.google.common.util.concurrent.ListenableFuture;
+import com.palantir.dialogue.Response;
+import org.derive4j.Data;
+
+@Data
+public interface LimitedResponse {
```
carterkozak, line 24: This appears to only be used by RoundRobinChannel, is that correct?

dialogue-core/src/main/java/com/palantir/dialogue/core/CompositeLimitedChannel.java
ferozco: It was to better separate concerns. Node selection strategies should know about blacklisting and limiting while queueing/retrying only needs to know about whether a request occurred or not

dialogue-core/src/main/java/com/palantir/dialogue/core/LimitedResponse.java
ferozco: I'll add some docs

dialogue-core/src/main/java/com/palantir/dialogue/core/LimitedResponse.java
ferozco: No, all selection strategies convert from LimitedReponses to `Optional<Response>`. The idea is that selection strategies need more information to properly select which channel to make a request to

dialogue-core/src/main/java/com/palantir/dialogue/core/RoundRobinChannel.java
ferozco: ya it seems correct to skip over the the limited channel

carterkozak: But this stops iterating and doesn't attempt any more channels unless I'm reading it incorrectly.

ferozco: Yes, I think its right to stop iterating if you hit a limited channel to give time for the limit to be released. Otherwise you'll end up just spinning until a limit is released

carterkozak: Each delegate has its own limit, the next channel may not be limited in the same way

carterkozak: For example when a new host is discovered we want to allow most requests to rotate to that channel even if the rest are some combination of saturated/blacklisted.

simulation/src/test/resources/report.md
```diff
@@ -1,41 +1,41 @@
...
-         live_reloading[CONCURRENCY_LIMITER_BLACKLIST_ROUND_ROBIN].txt:	success=79.2%	client_mean=PT4.5097744S   	server_cpu=PT1H53M27.37S  	client_received=2500/2500	server_resps=2500	codes={200=1979, 500=521}
-               live_reloading[CONCURRENCY_LIMITER_PIN_UNTIL_ERROR].txt:	success=58.9%	client_mean=PT3.5763136S   	server_cpu=PT1H58M42.9S   	client_received=2500/2500	server_resps=2500	codes={200=1473, 500=1027}
-                   live_reloading[CONCURRENCY_LIMITER_ROUND_ROBIN].txt:	success=58.6%	client_mean=PT3.5376608S   	server_cpu=PT1H58M19S     	client_received=2500/2500	server_resps=2500	codes={200=1466, 500=1034}
+         live_reloading[CONCURRENCY_LIMITER_BLACKLIST_ROUND_ROBIN].txt:	success=79.6%	client_mean=PT4.5579616S   	server_cpu=PT1H54M0.29S   	client_received=2500/2500	server_resps=2500	codes={200=1990, 500=510}
+               live_reloading[CONCURRENCY_LIMITER_PIN_UNTIL_ERROR].txt:	success=93.0%	client_mean=PT7.1876512S   	server_cpu=PT2H48.6S      	client_received=2500/2500	server_resps=2500	codes={200=2326, 500=174}
```
iamdanfox, line 21: is it expected that live pin_until_error now seems to be not touching the newly live-reloaded server? https://github.com/palantir/dialogue/blob/fo/differentiate-limits-blacklist/simulation/src/test/resources/report.md#live_reloadingconcurrency_limiter_pin_until_error

simulation/src/test/resources/report.md
```diff
@@ -1,41 +1,41 @@
...
-slowdown_and_error_thresholds[CONCURRENCY_LIMITER_BLACKLIST_ROUND_ROBIN].txt:	success=1.8%	client_mean=PT1M8.847548999S	server_cpu=PT10H40M58.53999998S	client_received=10000/10000	server_resps=10000	codes={200=183, 500=9817}
-slowdown_and_error_thresholds[CONCURRENCY_LIMITER_PIN_UNTIL_ERROR].txt:	success=1.8%	client_mean=PT20.068609533S	server_cpu=PT10H35M7.200666646S	client_received=10000/10000	server_resps=10000	codes={200=176, 500=9824}
-    slowdown_and_error_thresholds[CONCURRENCY_LIMITER_ROUND_ROBIN].txt:	success=1.2%	client_mean=PT16.859225466S	server_cpu=PT10H30M49.207333306S	client_received=10000/10000	server_resps=10000	codes={200=120, 500=9880}
+slowdown_and_error_thresholds[CONCURRENCY_LIMITER_BLACKLIST_ROUND_ROBIN].txt:	success=1.8%	client_mean=PT1M8.848058466S	server_cpu=PT10H40M56.53999998S	client_received=10000/10000	server_resps=10000	codes={200=183, 500=9817}
+slowdown_and_error_thresholds[CONCURRENCY_LIMITER_PIN_UNTIL_ERROR].txt:	success=9.4%	client_mean=PT1M4.588908199S	server_cpu=PT10H49M0.733999957S	client_received=10000/10000	server_resps=10000	codes={200=936, 500=9064}
```
iamdanfox, line 37: as expected, by not consuming some retries in our limits, this means requests can hang on for longer (so take longer from a user-perspective), but ultimately more succeed. 1.8% -> 9.4%

dialogue-core/src/main/java/com/palantir/dialogue/core/BlacklistingChannel.java
```diff
@@ -49,7 +50,7 @@
  * unblacklisted. Without this functionality, hundreds of requests could be sent to a still-broken
  * server before the first of them returns and tells us it's still broken.
  */
-final class BlacklistingChannel implements LimitedChannel {
+final class BlacklistingChannel implements CompositeLimitedChannel {
```
iamdanfox, line 53: can we tweak the names so that your CompositeLimitedChannel just becomes LimitedChannel and then rename LimitedChannel -> NodeSelectionChannel (because that's our PinUntilError, RoundRobin implementations)?

dialogue-core/src/main/java/com/palantir/dialogue/core/BlacklistingChannel.java
```diff
@@ -96,29 +98,27 @@
...
 
     @Override
-    public Optional<ListenableFuture<Response>> maybeExecute(Endpoint endpoint, Request request) {
+    public LimitedResponse maybeExecute(Endpoint endpoint, Request request) {
         BlacklistState state = channelBlacklistState.get();
         if (state != null) {
             BlacklistStage stage = state.maybeProgressAndGet();
             if (stage instanceof BlacklistUntil) {
-                return Optional.empty();
+                return LimitedResponses.blacklisted();
```
iamdanfox, line 106: Rather than `LimitedResponses.blacklisted`, what do you think of `LimitedResponses.backOff` ?

dialogue-core/src/main/java/com/palantir/dialogue/core/ConcurrencyLimitedChannel.java
```diff
@@ -74,20 +75,18 @@ static ConcurrencyLimitedChannel create(LimitedChannel delegate, DialogueClientM
...
-    public Optional<ListenableFuture<Response>> maybeExecute(Endpoint endpoint, Request request) {
+    public LimitedResponse maybeExecute(Endpoint endpoint, Request request) {
         Optional<Limiter.Listener> maybeListener = limiter.acquire(NO_CONTEXT);
         if (maybeListener.isPresent()) {
             Limiter.Listener listener = maybeListener.get();
-            Optional<ListenableFuture<Response>> result = delegate.maybeExecute(endpoint, request);
-            if (result.isPresent()) {
-                DialogueFutures.addDirectCallback(result.get(), new LimiterCallback(listener));
-            } else {
-                listener.onIgnore();
```
iamdanfox, line 85: why are we no longer calling `listener.onIgnore` in the other two cases? I think using a visitor on the limitedResponse would be more reassuring to me here

dialogue-core/src/main/java/com/palantir/dialogue/core/PinUntilErrorChannel.java
```diff
@@ -80,40 +80,52 @@ static LimitedChannel pinUntilError(List<LimitedChannel> channels) {
...
             @Override
-            public void onSuccess(Response response) {
-                if (Responses.isQosStatus(response) || Responses.isServerError(response)) {
-                    OptionalInt next = incrementHostIfNecessary(currentIndex);
-                    debugLogReceivedErrorStatus(currentIndex, channel, response, next);
-                    // TODO(dfox): handle 308 See Other somehow, as we currently don't have a host -> channel mapping
-                }
+            public Optional<ListenableFuture<Response>> limited() {
+                debugLogLimitedRequest(currentIndex, channel);
+                return Optional.empty();
```
iamdanfox, line 105: what does it semantically mean for the PinUntilErrorChannel to return Optional.empty? If an inner concurrency limiter decided this host was overloaded, why are we not selecting another host here?

iamdanfox: I think this also explains why https://github.com/palantir/dialogue/blob/fo/differentiate-limits-blacklist/simulation/src/test/resources/report.md#drastic_slowdownconcurrency_limiter_pin_until_error doesn't look good after the slow node was reverted.

ferozco:
Closing in favour of #432

PR #423, opened 2020-02-24
NullAway

iamdanfox:
## Before this PR

We generally run NullAway on most infra libraries, just to hold ourselves to the highest standard we can, but it wasn't yet enabled on dialogue yet.

https://github.com/uber/NullAway

## After this PR
==COMMIT_MSG==
Dialogue now passes Uber's NullAway static analysis
==COMMIT_MSG==

## Possible downsides?
- if it gets annoying for local iteration, people might need to start using `./gradlew test -Dcom.palantir.baseline-error-prone.disable`

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,3 @@
 *.png filter=lfs diff=lfs merge=lfs -text
+versions.props merge=union
+versions.lock merge=union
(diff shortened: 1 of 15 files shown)


carterkozak (approved):
thanks!

PR #426, opened 2020-02-25
Track png in git-lfs

ferozco:
Some PNG somehow managed to make it onto develop without being in git lfs


carterkozak:
I think this was my fault, committed from another machine without lfs. Sorry!

:+1:

-ck

ferozco:
🤖

PR #427, opened 2020-02-25
Downgrade hc4 to 4.5.10

ferozco:
Downgrade to 4.5.10 after @mglazer flagged that 4.5.11 was broken due to https://issues.apache.org/jira/browse/HTTPCLIENT-2047.


carterkozak:
We use 4.5.11 in our large internal product, and this implementation is not impacted by HTTPCLIENT-2047 because we use `new DefaultHostnameVerifier()` without providing a `PublicSuffixMatcher`.

versions.props
```diff
@@ -13,7 +13,7 @@ com.palantir.tokens:auth-tokens = 3.6.1
 com.palantir.tracing:* = 4.4.0
 com.uber.nullaway:nullaway = 0.7.9
 io.dropwizard.metrics:metrics-core = 3.2.6
-org.apache.httpcomponents:httpclient = 4.5.11
+org.apache.httpcomponents:httpclient = 4.5.10
```
carterkozak: I think we need to pull this into a do-not-upgrade block

carterkozak (approved):

PR #430, opened 2020-02-25
ApacheHttpChannels exposes a Closeable instance

iamdanfox:
## Before this PR

Previously, our `ApacheHttpChannels#create` method would create a client pool internally, but gave us no way to shut this down.

## After this PR
==COMMIT_MSG==
ApacheHttpChannels exposes a Closeable instance, allowing callers to release resources when they determine the client is no longer needed.
==COMMIT_MSG==

## Possible downsides?
- This approach means we have more API surface, as we have to expose quite primitive building blocks and trust that the WC facade will abstract them appropriately.  This is in contrast to my earlier [`clientPool` approach](https://github.com/palantir/dialogue/pull/398) which hid all these details.


dialogue-apache-hc4-client/src/main/java/com/palantir/dialogue/hc4/ApacheHttpClientChannels.java
```diff
@@ -75,15 +75,26 @@
...
+                conf.uris().stream().map(uri -> createSingleUri(uri, client)).collect(Collectors.toList());
+
+        return Channels.create(channels, agent, conf);
+    }
+
+    public static Channel createSingleUri(String uri, CloseableHttpClient client) {
+        return BlockingChannelAdapter.of(new ApacheHttpClientBlockingChannel(client, url(uri)));
+    }
+
+    public static CloseableHttpClient createCloseableHttpClient(ClientConfiguration conf) {
```
carterkozak: A little bit wonky in public api since it's not clear what functionality the returned client provides compared to our channel wrappers. Will we cause bugs if we move functionality from the client into a channel?
At the same time, this should be an implementation dependency of components which simply provide channels, so it's probably fine.

dialogue-apache-hc4-client/src/main/java/com/palantir/dialogue/hc4/ApacheHttpClientChannels.java
```diff
@@ -75,15 +75,26 @@
...
+    public static Channel create(ClientConfiguration conf, UserAgent agent) {
+        CloseableHttpClient client = createCloseableHttpClient(conf);
+        List<Channel> channels =
+                conf.uris().stream().map(uri -> createSingleUri(uri, client)).collect(Collectors.toList());
+
+        return Channels.create(channels, agent, conf);
+    }
+
+    public static Channel createSingleUri(String uri, CloseableHttpClient client) {
+        return BlockingChannelAdapter.of(new ApacheHttpClientBlockingChannel(client, url(uri)));
```
carterkozak: Another possibility is that we create a new client for each URI for additional isolation and avoiding contention in the connection pool.

dialogue-apache-hc4-client/src/main/java/com/palantir/dialogue/hc4/ApacheHttpClientChannels.java
```diff
@@ -75,15 +75,26 @@
 
     private ApacheHttpClientChannels() {}
 
-    public static Channel create(ClientConfiguration conf, UserAgent baseAgent) {
+    public static Channel create(ClientConfiguration conf, UserAgent agent) {
```
carterkozak: Should all channel implementations be closeable?

dialogue-apache-hc4-client/src/main/java/com/palantir/dialogue/hc4/ApacheHttpClientChannels.java
iamdanfox: This API gives us the freedom to do that from our WC facade, because we can choose to pass in the same closeablehttpclient or not.

dialogue-apache-hc4-client/src/main/java/com/palantir/dialogue/hc4/ApacheHttpClientChannels.java
iamdanfox: so I wanted to just convey the idea of some closeable shared resource - would you prefer if I wrap it in some opaque type so that people can't use it directly?

dialogue-apache-hc4-client/src/main/java/com/palantir/dialogue/hc4/ApacheHttpClientChannels.java
iamdanfox: I actually started with this, but I think it will get hella confusing in the scenario where our NodeSelectionStrategy channels contain a bunch of internal channels and possibly want to live reload a new uri... I think it's clearer to only make things closeable that actually directly hold resources, and if they don't hold resources then they can just be thrown away.

carterkozak: I think queuedchannel is the only other piece that holds resources

carterkozak:
:+1:

dialogue-apache-hc4-client/src/main/java/com/palantir/dialogue/hc4/ApacheHttpClientChannels.java
```diff
@@ -75,14 +77,25 @@
...
+                conf.uris().stream().map(uri -> createSingleUri(uri, client)).collect(Collectors.toList());
+
+        return Channels.create(channels, conf);
+    }
+
+    public static Channel createSingleUri(String uri, CloseableClient client) {
+        return BlockingChannelAdapter.of(new ApacheHttpClientBlockingChannel(client.client, url(uri)));
+    }
+
+    public static CloseableClient createCloseableHttpClient(ClientConfiguration conf) {
```
uschi2000, line 91: ClientCloseable? :) https://github.com/palantir/conjure-java/pull/790

uschi2000, line 91: ClientCloseable? :) https://github.com/palantir/conjure-java/pull/790

dialogue-apache-hc4-client/src/main/java/com/palantir/dialogue/hc4/ApacheHttpClientChannels.java
```diff
@@ -127,12 +140,27 @@ public static Channel create(ClientConfiguration conf) {
...
+        }
+
+        @Override
+        public void close() throws IOException {
+            client.close();
+        }
+
+        @Override
+        public String toString() {
+            return "SharedResource{client=" + client + '}';
```
uschi2000, line 162: SharedResource?

dialogue-apache-hc4-client/src/main/java/com/palantir/dialogue/hc4/ApacheHttpClientChannels.java
```diff
@@ -127,12 +140,27 @@ public static Channel create(ClientConfiguration conf) {
...
-        ImmutableList<Channel> channels = conf.uris().stream()
-                .map(uri -> BlockingChannelAdapter.of(new ApacheHttpClientBlockingChannel(client, url(uri))))
-                .collect(ImmutableList.toImmutableList());
 
-        return Channels.create(channels, conf);
+        return new CloseableClient(builder.build());
+    }
+
+    /** Intentionally opaque wrapper type - we don't want people using the inner Apache client directly. */
+    public static final class CloseableClient implements Closeable {
```
uschi2000, line 148: why is this public if it's never used outside this class?

iamdanfox: It needs to be public because we create some of these in Witchcraft and then re-use them to avoid re-creating clients on every url live-reload.  

See https://internal-github/foundry/witchcraft/pull/2233/files#diff-2819734323a961448d10003f575f1b4bR217

dialogue-apache-hc4-client/src/main/java/com/palantir/dialogue/hc4/ApacheHttpClientChannels.java
iamdanfox: 🤦‍♂ https://github.com/palantir/dialogue/pull/449

PR #431, opened 2020-02-25
Simplify (ClientConfig, UserAgent) signatures to just (ClientConfig)

iamdanfox:
## Before this PR

Same motivation as https://github.com/palantir/conjure-java-runtime/pull/1501

## After this PR
==COMMIT_MSG==
A bunch of our method signatures now just take a `ClientConfiguration` parameter instead of `ClientConfiguration` and `UserAgent`.
==COMMIT_MSG==

## Possible downsides?
- this is a break


carterkozak (approved):

iamdanfox:
Also thoughts on enabling revapi again? I'm going to need to go and fix WC to accommodate this break, and I think it would be nice to make sure PR authors get a preview of their PR impact (which they can always choose to just accept and move on)

PR #432, opened 2020-02-25
fix #351 fix #302 fix #312 Replace QueuedChannel with a backoff based retryer

### LINKED ISSUE #302: QueuedChannel metrics only report data for the most recently constructed channel
## What happened?

QueuedChannel metrics use gauges and don't differentiate between instances. Only the most recently created QueuedChannel is instrumented.

These should use Counter and increment/decrement a global value.

### LINKED ISSUE #312: RoundRobinChannel should fail when no targets exist

### LINKED ISSUE #351: QueuedChannel should fail when all delegates are limited and the queue is full
## What happened?

QueuedChannel returns a fake 429 response instead of a failure when all delegates are limited and the queue is full. Responses mean that a request was made and a server provided feedback.

When no request is made we should not return a response.

carterkozak:
==COMMIT_MSG==
Replace QueuedChannel with a backoff based retryer
==COMMIT_MSG==


dialogue-core/src/main/java/com/palantir/dialogue/core/ExponentialBackoffStrategy.java
```diff
@@ -0,0 +1,40 @@
...
+import java.util.function.DoubleSupplier;
+
+final class ExponentialBackoffStrategy implements BackoffStrategy {
+
+    private final Duration backoffSlotSize;
+    private final DoubleSupplier random;
+
+    ExponentialBackoffStrategy(Duration backoffSlotSize, DoubleSupplier random) {
+        this.backoffSlotSize = backoffSlotSize;
+        this.random = random;
```
iamdanfox: s/random/jitter/ - it's fine for this to always return 1 in testing, and be a random implementation in prod?

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -37,46 +48,87 @@
...
+     * Shared single thread executor is reused between all retrying channels. If it becomes oversaturated
+     * we may wait longer than expected before resuming requests, but this is an
+     * edge case where services are already operating in a degraded state and we should not
+     * spam servers.
+     */
+    private static final Supplier<ListeningScheduledExecutorService> sharedScheduler = Suppliers.memoize(
+            () -> MoreExecutors.listeningDecorator(Tracers.wrap(
+                    "dialogue-BlacklistingChannel-scheduler",
+                    Executors.newSingleThreadScheduledExecutor(new ThreadFactoryBuilder()
+                            .setNameFormat("dialogue-BlacklistingChannel-scheduler-%d")
```
iamdanfox: s/Blacklisting/Retrying/ ?

carterkozak: 🤦‍♂

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -37,46 +48,87 @@
...
+                    () -> delegate.execute(endpoint, request), backoffNanoseconds, TimeUnit.NANOSECONDS);
+            return wrap(Futures.transformAsync(scheduled, input -> input, MoreExecutors.directExecutor()));
+        }
+
+        private long getBackoffNanoseconds() {
+            if (failures == 0) {
+                return 0L;
+            }
+            int upperBound = (int) Math.pow(2, failures - 1);
+            return Math.round(backoffSlotSize.toNanos() * random.getAsDouble() * upperBound);
```
iamdanfox: We're gonna plumb the BackoffStrategy thing in here right?

dialogue-core/src/main/java/com/palantir/dialogue/core/ExponentialBackoffStrategy.java
carterkozak: It's important to get jitter in our simulations as well, otherwise the we may cause waves of requests. In tests using a constant is fine, but ideally we don't measure wall clock time in tests and watch the request rate/etc.

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
carterkozak: We could, but I don't like it as written in CJR, I find it confusing that getting the duration mutates state and increments failures. I had an implementation that took a backoffstrategy as a function from failure count to backoff duration that worked fine, but without needing alternative implementations it didn't add anything.
Thoughts?

iamdanfox: Yeah definitely would like to not do the mutation as part of the getter. Fair enough that if we're always going to use the exponential strategy, then maybe factoring out is unnecessary!

iamdanfox:
Since this is quite a big change in behaviour, I'd like to try and articulate the pros and cons of this switch.

- an upside of the old 'queued channel & instant retry' thing was if there's at least one channel which is not currently limited, then you should be able to get a request out the door instantly. Now however, you might get the "Failed to make a request" exception which causes a backoff if the one node you were pinned to limited you

carterkozak:
Upsides:
* This allows us to implement per-endpoint limiting because we don’t need to worry about a queue. (We could make limitedChannel extend channel and try to find a non-limited channel before attempting for force our way through, trying a channel+endpoint combination that is likely broken)
* Easier to reason about because a single limited request doesn’t cause jitter for other requests triggered from separate threads slightly later

carterkozak:
(sorry to split these, poking at this between cleaning/dishes/etc)
Downsides:
 * Potentially lower throughput due to backoff rather than edge-triggering. In practice the BlacklistingChannel resulted in an approximation of backoff without the upsides, and it was too easy to introduce regressions.

dialogue-core/src/main/java/com/palantir/dialogue/core/Channels.java
```diff
@@ -39,26 +38,24 @@ public static Channel create(Collection<? extends Channel> channels, ClientConfi
...
-        QueuedChannel queuedChannel = new QueuedChannel(limited, DispatcherMetrics.of(config.taggedMetricRegistry()));
-        queueListener.delegate = queuedChannel::schedule;
-        Channel channel = queuedChannel;
+        Channel channel = new LimitedChannelToChannelAdapter(limited);
         channel = new TracedChannel(channel, "Dialogue-request-attempt");
-        channel = new RetryingChannel(channel, config.maxNumRetries(), config.serverQoS());
+        if (config.maxNumRetries() > 0) {
+            channel =
+                    new RetryingChannel(channel, config.maxNumRetries(), config.backoffSlotSize(), config.serverQoS());
+        }
```
iamdanfox, line 58: cute little optimization :)

dialogue-core/src/main/java/com/palantir/dialogue/core/ExponentialBackoffStrategy.java
```diff
@@ -0,0 +1,40 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.dialogue.core;
+
+import java.time.Duration;
+import java.util.function.DoubleSupplier;
+
+final class ExponentialBackoffStrategy implements BackoffStrategy {
```
iamdanfox: this class seems unused - delete? (also the interface)

carterkozak: doh.

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
```diff
@@ -1,299 +0,0 @@
...
-                        log.debug(
-                                "Queued response has already been completed",
-                                SafeArg.of("service", endpoint.serviceName()),
-                                SafeArg.of("endpoint", endpoint.endpointName()));
-                    }
-                }
-                return false;
-            }
-        }
-    }
```
iamdanfox, line 223: so much spiciness into the bin 🌶

carterkozak: 🥛:-)

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -103,18 +154,18 @@ private RetryingCallback(
...
         }
 
-        private void logRetry(Throwable throwable) {
+        private void logRetry(long backoffNanoseconds, Throwable throwable) {
             if (log.isInfoEnabled()) {
                 log.info(
                         "Retrying call after failure",
                         SafeArg.of("failures", failures),
                         SafeArg.of("maxRetries", maxRetries),
+                        SafeArg.of("backoffNanoseconds", backoffNanoseconds),
```
iamdanfox: could we do a TimeUnit.Milliseconds.convert here and present backoffMillis instead? It's really hard to eyeball such big numbers when they appear in the logs

dialogue-core/src/test/resources/tracing/ChannelsTest/traces_on_retries.log
```diff
@@ -1,22 +1,26 @@
...
-{"traceId":"3af31cbecda2b46c","parentSpanId":"28686cc7c2695fb2","spanId":"749bb99b3951da3f","type":"LOCAL","operation":"Dialogue-request-attempt initial","startTimeMicroSeconds":1582562924534927,"durationNanoSeconds":544058,"metadata":{}}
-{"traceId":"3af31cbecda2b46c","parentSpanId":"697e878d5bfc5628","spanId":"28686cc7c2695fb2","type":"LOCAL","operation":"Dialogue-request-attempt","startTimeMicroSeconds":1582562924534905,"durationNanoSeconds":577068,"metadata":{}}
-{"traceId":"3af31cbecda2b46c","parentSpanId":"efaf41a21cc7c6a9","spanId":"697e878d5bfc5628","type":"LOCAL","operation":"Dialogue-request initial","startTimeMicroSeconds":1582562924525312,"durationNanoSeconds":13461016,"metadata":{}}
-{"traceId":"3af31cbecda2b46c","parentSpanId":null,"spanId":"efaf41a21cc7c6a9","type":"LOCAL","operation":"Dialogue-request","startTimeMicroSeconds":1582562924525264,"durationNanoSeconds":13530208,"metadata":{}}
+{"traceId":"09b26a1be498af5c","parentSpanId":"49c88b0d4a7b609a","spanId":"802d8f9c0d0333a7","type":"LOCAL","operation":"Dialogue-http-request initial","startTimeMicroSeconds":1582681721129000,"durationNanoSeconds":7754255,"metadata":{}}
```
iamdanfox, line 1: https://5436-164943450-gh.circle-artifacts.com/0/~/artifacts/ChannelsTest/traces_on_retries/actual.html

iamdanfox: There is a lot of empty space in these - can we get a span containing the backoff duration in there somehow? (Could be as a FLUP)

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
carterkozak: good call

simulation/src/test/java/com/palantir/dialogue/core/Strategy.java
```diff
@@ -115,18 +113,14 @@ private static Channel roundRobin(Simulation sim, Supplier<List<SimulationServer
     }
 
     private static Channel queuedChannelAndRetrying(Simulation sim, LimitedChannel limited) {
```
iamdanfox: worth renaming this as it no longer includes a queue

dialogue-core/src/test/resources/tracing/ChannelsTest/traces_on_retries.log
carterkozak: Ya, I'd prefer to do this in a follow-up

simulation/src/test/resources/report.md
```diff
@@ -1,43 +1,43 @@
...
                       fast_500s_then_revert[UNLIMITED_ROUND_ROBIN].txt:	success=76.7%	client_mean=PT0.055463644S 	server_cpu=PT3M27.988666346S	client_received=3750/3750	server_resps=3750	codes={200=2876, 500=874}
-         live_reloading[CONCURRENCY_LIMITER_BLACKLIST_ROUND_ROBIN].txt:	success=82.4%	client_mean=PT4.5664688S   	server_cpu=PT1H52M14.26S  	client_received=2500/2500	server_resps=2500	codes={200=2061, 500=439}
-               live_reloading[CONCURRENCY_LIMITER_PIN_UNTIL_ERROR].txt:	success=59.0%	client_mean=PT3.5693656S   	server_cpu=PT1H58M42.9S   	client_received=2500/2500	server_resps=2500	codes={200=1476, 500=1024}
-                   live_reloading[CONCURRENCY_LIMITER_ROUND_ROBIN].txt:	success=58.6%	client_mean=PT3.5376608S   	server_cpu=PT1H58M19S     	client_received=2500/2500	server_resps=2500	codes={200=1466, 500=1034}
+         live_reloading[CONCURRENCY_LIMITER_BLACKLIST_ROUND_ROBIN].txt:	success=59.3%	client_mean=PT2.717022131S 	server_cpu=PT1H21M18.049029538S	client_received=2500/2500	server_resps=1865	codes={200=1483, 500=382, Failed to make a request=635}
```
iamdanfox, line 20: I think the 82% -> 59% success makes sense because a bunch of requests are no longer hanging out on the queue, and instead we get a chunk of them come back as "failed to make a request"
https://github.com/palantir/dialogue/blob/ckozak/no_queue/simulation/src/test/resources/report.md#live_reloadingconcurrency_limiter_blacklist_round_robin

dialogue-core/src/main/java/com/palantir/dialogue/core/LimitedChannelToChannelAdapter.java
```diff
@@ -0,0 +1,51 @@
...
+import com.palantir.logsafe.exceptions.SafeRuntimeException;
+import java.util.function.Supplier;
+
+/** Adapter from {@link LimitedChannel} to {@link Channel} which produces a failed future when results are limited. */
+final class LimitedChannelToChannelAdapter implements Channel {
+
+    // Avoid method reference allocations
+    @SuppressWarnings("UnnecessaryLambda")
+    private static final Supplier<ListenableFuture<Response>> limitedResultSupplier =
+            () -> Futures.immediateFailedFuture(new SafeRuntimeException("Failed to make a request"));
```
iamdanfox, line 34: I think we're going to need to give people more information in this exception message (or possibly a comment here) otherwise we're going to be fielding questions in `#dev-foundry-infra` from people asking "why dialogue is breaking them" and telling them they can't make requests ;)

When people hit this, we probably don't want them to do their own retrying right?  Perhaps we should even put some instrumentation on this, so we know exactly how many times our client refused to send a request out the door?

iamdanfox (approved):
This is clearly a lot closer to the client-behaviour people are used to, so should be less scary to roll out. Excited that it gives us the possibility of bringing back per-endpoint smartness.

I think we might want to come back to the idea of immediate failover again at some point, but given that there's so much c-j-r pain right now, I'd be happy to get this out first.

dialogue-core/src/main/java/com/palantir/dialogue/core/LimitedChannelToChannelAdapter.java
carterkozak: Agreed, I like the idea of replacing the limited-channel result of `Optional<future<response>>` with a union of `limited-reason` or `future<response>`. Also like the idea of forcefully attempting a request through the limiter if all nodes have limited our result to avoid purely client-side badness.

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -37,46 +48,87 @@
...
+     * edge case where services are already operating in a degraded state and we should not
+     * spam servers.
+     */
+    private static final Supplier<ListeningScheduledExecutorService> sharedScheduler = Suppliers.memoize(
+            () -> MoreExecutors.listeningDecorator(Tracers.wrap(
+                    "dialogue-RetryingChannel-scheduler",
+                    Executors.newSingleThreadScheduledExecutor(new ThreadFactoryBuilder()
+                            .setNameFormat("dialogue-RetryingChannel-scheduler-%d")
+                            .setDaemon(false)
+                            .build()))));
```
iamdanfox, line 63: one more flup please- can we have instrumentation on this? e.g. how many things are currently on it? or will there be a way that I can pass a WC scheduled executor to this? Also possibly want an uncaught exception handler given the recent spiciness we discovered

carterkozak: good idea, would be nice to get metrics on this and identical tracing to other components.

PR #433, opened 2020-02-25
Fix the simulation scheduler to avoid fast-forward in external use

carterkozak:
Simulations failed when I tried testing the backoff retryer change because each time we schedule a future task, we fast-forward.


iamdanfox:
I don't fully understand why there's now a big spike at the end of some of the graphs (e.g. https://github.com/palantir/dialogue/blob/ckozak/fix_scheduler/simulation/src/test/resources/report.md#slowdown_and_error_thresholdsconcurrency_limiter_round_robin)

also 

https://github.com/palantir/dialogue/blob/ckozak/fix_scheduler/simulation/src/test/resources/report.md#live_reloadingconcurrency_limiter_blacklist_round_robin

carterkozak:
I wonder if we’re rendering what happens between the end of measurement and fast-forward-to-infinity?

iamdanfox:
Ah right yep the x axis on [this one](https://github.com/palantir/dialogue/blob/ckozak/fix_scheduler/simulation/src/test/resources/report.md#live_reloadingconcurrency_limiter_blacklist_round_robin) went up to 13 seconds, but this time was truncated to 10... that seems a bit wrong

iamdanfox:
So conceptually, I still want 'external users' to be able to schedule tasks in the future on our executor, and as long as the executor starts them in order, then shouldn't we be able to fast-forward the ticker monotonically? I don't quite get why this problem has started happening (esp since the Benchmark and SimulationServer classes all use scheduled futures already)

carterkozak:
Without this change the backoff retryer commit complains about rewinding time too far, fails requests, and causes some simulations to hang. Not confident this is the correct solution, but it does allow the tests to run, heh.

PR #436, opened 2020-02-26
All metrics include dialogueVersion

iamdanfox:
## Before this PR

As we roll out dialogue, we could get into the situation where people show us graphs on datadog and ask "why did dialogue do that" - the first step will probably be to figure out what version of dialogue was actually in use at the time to figure out if they're experiencing a bug in latest or if they're on an old version.

## After this PR
==COMMIT_MSG==
All tagged metrics produced by dialogue will now have an extra `dialogueVersion` tag.
==COMMIT_MSG==

Might want to move this into tritium/metric-schema if we decide we like the pattern.

## Possible downsides?
- literally can't think of a downside here. This should not increase the cardinality at all, because I'm not expecting there to be multiple versions of dialogue inside a single service!

### DIFF
--- a/changelog/@unreleased/pr-436.v2.yml
+++ b/changelog/@unreleased/pr-436.v2.yml
@@ -1,0 +1,6 @@
+type: improvement
+improvement:
+  description: All tagged metrics produced by dialogue will now have an extra `dialogueVersion`
+    tag.
+  links:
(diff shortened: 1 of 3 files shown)


carterkozak (approved):

ferozco:
👍

PR #437, opened 2020-02-26
Metrics for PinUntilErrorChannel

iamdanfox:
## Before this PR
The vast majority of services will use this channel. I'd like to be able to look at graphs to see how it behaves in production.

## After this PR
==COMMIT_MSG==
Users of the PinUntilError channel now have metrics available capturing when the list reshuffles, when node switches (including a reason) and how many successful requests we see on each channel.
==COMMIT_MSG==

## Possible downsides?
- more metrics = more money. I _think_ we could build a flag to switch some of these metrics on/off later, but I think it's OK to keep em for now as we do a gradual rollout of dialogue.

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/PinUntilErrorChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/PinUntilErrorChannel.java
@@ -37,6 +37,8 @@ import java.util.OptionalInt;
 import java.util.Random;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
+import java.util.stream.IntStream;
+import javax.annotation.Nullable;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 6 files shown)


dialogue-core/src/main/java/com/palantir/dialogue/core/PinUntilErrorChannel.java
```diff
@@ -216,59 +220,102 @@ private void reshuffleChannelsIfNecessary() {
...
+                            UnsafeArg.of("current", channel),
+                            throwable);
+                }
+            }
+        }
+
+        private void instrumentSuccessfulResponse(int currentIndex, NodeList nodeList) {
+            if (nodeList.size() < 10) {
+                // hard limit ensures we don't create unbounded tags
+                metrics.success(Integer.toString(currentIndex));
```
carterkozak: I think we're missing a `@CheckReturnValue` on the generated utility class, I imagine we want to mark this meter.
Can we create an array of meters and run `successfulResponseMeters.get(currentIndex).mark()`?

iamdanfox: Done

carterkozak (approved):

PR #440, opened 2020-02-26
Implement missing retry configuration options

carterkozak:
==COMMIT_MSG==
Implement missing retry configuration options, RetryOnTimeout and RetryOnSocketException
==COMMIT_MSG==


dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -155,11 +175,38 @@ private long getBackoffNanoseconds() {
...
+            if (retryOnTimeout == ClientConfiguration.RetryOnTimeout.DISABLED) {
+                if (throwable instanceof SocketTimeoutException) {
+                    // non-connect timeouts should not be retried
+                    SocketTimeoutException socketTimeout = (SocketTimeoutException) throwable;
+                    return socketTimeout.getMessage() != null
+                            // String matches CJR RemotingOkHttpCall.shouldRetry
+                            && socketTimeout.getMessage().contains("connect timed out");
+                }
+            }
+            return true;
```
iamdanfox, line 205: The RetryOnSocketException=DANGEROUS_DISABLED thing was implemented purely as a hack for alta, and if possible I'd kinda like to just not support it from the beginning to minimize complexity (as I think they can make single-uri channels to achieve their use-case).

https://internal-github/search?q=org%3Afoundry+RetryOnSocketException&type=Code

carterkozak: I've pulled it from this PR and added a check so we throw eagerly if the pref is enabled.

iamdanfox: Do you still want to keep the socketexception normalizing thing?

carterkozak: Might as well get rid of it since ConnectTimeoutException isn't a socket exception

dialogue-apache-hc4-client/build.gradle
```diff
@@ -1,5 +1,4 @@
 apply from: "$rootDir/gradle/publish-jar.gradle"
-apply plugin: 'com.palantir.revapi'
```
iamdanfox: why the deletion here? i'd prefer to just use `./gradlew revapiAcceptAllBreaks` and start building up a revapi.yml of accepted breaks?

carterkozak: I picked the wrong number of commits in `git checkout HEAD~4 -- dialogue-apache-hc4-client` :-( fixed.

iamdanfox (approved):

PR #441, opened 2020-02-26
Tracing spans for RetryingChannel backoff

carterkozak:
## After this PR
==COMMIT_MSG==
Tracing spans for RetryingChannel backoff
==COMMIT_MSG==

## Possible downsides?
Doesn't produce the prettiest graphs, but it fills space. The results
can be a touch confusing because backoff spans use the retrying channel
as a parent, not the request that produced a failure.

Screen Shot 2020-02-26 at 2 30 25 PM


dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -98,6 +99,7 @@
     private final class RetryingCallback {
         private final Endpoint endpoint;
         private final Request request;
+        private final DetachedSpan span = DetachedSpan.start("Dialogue-RetryingChannel");
```
iamdanfox, line 102: ooh nice so this actually won't appear if a request succeeds first time?

carterkozak: Yep :-)

carterkozak: The successful request tracing test is not impacted by this change

iamdanfox (approved):
Looks like a strict improvement to me! I'd still like to try and prune down all the `foo` and `foo initial` spans, but don't want to block this PR on that.

PR #443, opened 2020-02-26
Instrument ConcurrencyLimitedChannel

iamdanfox:
## Before this PR

'Concurrency Limiters' is a scary word to most people, and people fear what they don't understand. I'd like to have graphs to show when concurrency limiters are actually affecting people.

## After this PR
==COMMIT_MSG==
Two new gauges, `dialogue.concurrencylimiter.max` and `dialogue.concurrencylimiter.utilization`, describe the state of each host's concurrency limiter.
==COMMIT_MSG==

~`do not merge` until we've decided how to plumb in the service-name (otherwise if a service creates multiple dialogue clients each with 3 hosts, these gauges will flicker horribly).~

## Possible downsides?
- more metrics = more DD $
- Witchcraft's metric sampling is roughly every 30 seconds, so even if a DataDog graph shows low utilization, there's a possibility that it was spiking in between samples, which is kinda misleading.


dialogue-core/src/main/java/com/palantir/dialogue/core/Channels.java
```diff
@@ -23,48 +23,41 @@
...
-                .map(channel -> new InstrumentedChannel(channel, clientMetrics))
-                // TracedChannel must wrap TracedRequestChannel to ensure requests have tracing headers.
-                .map(TracedRequestChannel::new)
-                .map(channel -> new TracedChannel(channel, "Dialogue-http-request"))
-                .map(ChannelToLimitedChannelAdapter::new)
-                .map(concurrencyLimiter(config, clientMetrics))
-                .map(channel -> new FixedLimitedChannel(channel, MAX_REQUESTS_PER_CHANNEL, clientMetrics))
-                .collect(ImmutableList.toImmutableList());
+        ImmutableList.Builder<LimitedChannel> limitedChannels = ImmutableList.builder();
+        for (int hostIndex = 0; hostIndex < channels.size(); hostIndex++) {
```
carterkozak: Can we sort the input uris and index based on sorted order? That way if the input list changes order we don't shuffle metrics, and I think it's cleaner to consume a collection/iterable here.

iamdanfox: At the moment we just get 'Channel' as the input type - it's not comparable and we can't correlate between any particular instances.

Thoughts on asking people to provide a ChannelWithId so we can do that sorting?

dialogue-core/src/main/java/com/palantir/dialogue/core/ConcurrencyLimitedChannel.java
```diff
@@ -45,22 +47,35 @@
...
         this.delegate = new NeverThrowLimitedChannel(delegate);
-        this.limitedMeter = metrics.limited(getClass().getSimpleName());
+        this.limitedMeter =
+                DialogueClientMetrics.of(taggedMetrics).limited(getClass().getSimpleName());
         this.limiter = limiter;
+
+        hostIndex.ifPresent(index -> {
+            DialogueConcurrencylimiterMetrics metrics = DialogueConcurrencylimiterMetrics.of(taggedMetrics);
+            // TODO(dfox): hook up the service-name somehow? also when nodes reshuffle these metrics will look odd.
+            metrics.utilization().hostIndex(Integer.toString(index)).build(this::getUtilization);
```
carterkozak: I'm not confident this works when we have multiple discovered services, each combined channel (alta both gatekeeper, for example) will have separate concurrency limited channels using the same gauge for index zero. We don't know which one we're measuring.

dialogue-core/src/main/java/com/palantir/dialogue/core/Channels.java
carterkozak: Ah right. Not sure channel IDs are helpful outside of this use case. We can sort by identity hash code and index from that list ¯\_(ツ)_/¯

iamdanfox: I think they'd also be useful for staying pinned on the same host after uris live-reload. Let me try something with identifyhashcode.

dialogue-core/src/main/java/com/palantir/dialogue/core/ConcurrencyLimitedChannel.java
iamdanfox: Yep that's why I added the `do not merge`. I think the thing we're really interested in here is some stable identifier per host.

I don't think we can actually use hashcode because then a ton of uri reloads will result in an unbounded number of tags, even for the uris that don't change.

iamdanfox:
@ellisjoe I think you were keen to avoid plumbing uris/channel ids deep into the internals of dialogue, so I'm trying to think of a way to produce these graphs (especially for the initial rollout), so that we can really clearly explain exactly what dialogue is doing at any point in time.

What if at Channels#create time, in addition to a List<Channel>, we asked users to pass in an opaque List<ChannelMetricSink> or something, and then just called methods on that interface to report stuff about a specific channel?

(Also I think it would be really helpful to articulate the dangerous you see in plumbing uris / channeld identifiers in some kind of architecture decision record).

iamdanfox:
Two open questions: 
- if people create clients with a high number of uris then this will create a high tag DD tag cardinality.
- if the same URI appears in multiple service-discovery-blocks e.g. one for multipass and one for gatekeeper, then the metrics will likely be confusing.

dialogue-core/src/main/java/com/palantir/dialogue/core/ConcurrencyLimitedChannel.java
```diff
@@ -41,26 +43,30 @@
...
+        this.instrumentation = instrumentation;
         this.limiter = limiter;
+        instrumentation.registerMaxGauge(limiter::getLimit);
+        instrumentation.registerUtilizationGauge(this::getUtilization);
     }
 
-    static ConcurrencyLimitedChannel create(LimitedChannel delegate, DialogueClientMetrics metrics) {
-        return new ConcurrencyLimitedChannel(
-                delegate, ConcurrencyLimitedChannel.createLimiter(SYSTEM_NANOTIME), metrics);
+    interface Instrumentation {
```
ferozco: why do we need this extra interface? can we just pass a `DialogueConcurrencyLimiterMetrics`?

dialogue-core/src/main/java/com/palantir/dialogue/core/ConcurrencyLimitedChannel.java
```diff
@@ -125,4 +132,33 @@ public void onFailure(Throwable _throwable) {
...
+            @Override
+            public void markLimited() {
+                limited.mark();
+            }
+
+            @Override
+            public void registerUtilizationGauge(DoubleSupplier gauge) {
+                DialogueConcurrencylimiterMetrics.of(registry)
+                        .utilization()
+                        .host(uri)
```
ferozco: I don't think we consider uri to always be safe @carterkozak

dialogue-core/src/main/java/com/palantir/dialogue/core/ConcurrencyLimitedChannel.java
iamdanfox: I had this originally, but it required passing in the string uri which was only used for instrumentation purposes (here's the commit where you can see this: https://github.com/palantir/dialogue/pull/443/commits/8a8139d09704d80974bd388df29df5b7e5df28ed).

This felt a bit gross, as the string uri could in theory be re-purposed for other stuff. Also felt odd to always be passing in the only possible tag value rather than just pre-combining them.

dialogue-core/src/main/java/com/palantir/dialogue/core/ConcurrencyLimitedChannel.java
iamdanfox: URIs of palantir-controlled hosts are very much already present in DD, not sure if we can say the same for outgoing rquests.

dialogue-core/src/main/java/com/palantir/dialogue/core/ConcurrencyLimitedChannel.java
```diff
@@ -41,26 +43,30 @@
...
+        /** Called exactly once to register a gauge. */
+        void registerUtilizationGauge(DoubleSupplier gauge);
+
+        /** Called exactly once to register a gauge. */
+        void registerMaxGauge(IntSupplier gauge);
     }
 
-    @VisibleForTesting
-    static Limiter<Void> createLimiter(Ticker nanoTimeClock) {
+    static SimpleLimiter<Void> createLimiter(Ticker nanoTimeClock) {
```
ferozco, line 77: could we wrap this in instrumentation instead?

dialogue-core/src/main/java/com/palantir/dialogue/core/ConcurrencyLimitedChannel.java
carterkozak: front-door proxy names aren't always safe, when requests span enclaves

dialogue-core/src/main/java/com/palantir/dialogue/core/ConcurrencyLimitedChannel.java
iamdanfox: I've just tried this (see snippet), but we'd end up jumping through so many hoops to try and avoid gauges (counters only support integers, whereas utilization is a ratio), they also don't allow us to just set a new value.

```java
    private static class InstrumentedSimpleLimiter implements Limiter<Void> {
        private final SimpleLimiter<Void> simpleLimiter;

        private InstrumentedSimpleLimiter(
                SimpleLimiter<Void> simpleLimiter, DialogueConcurrencylimiterMetrics metrics, int hostIndex) {
            this.simpleLimiter = simpleLimiter;
        }

        private void updateMetrics(Instrumentation instrumentation, SimpleLimiter<Void> simpleLimiter) {
            instrumentation.updateMax(simpleLimiter.getLimit());
            instrumentation.updateUtilization(((double) simpleLimiter.getInflight()) / simpleLimiter.getLimit());
        }

        @Override
        public Optional<Listener> acquire(Void context) {
            Optional<Listener> maybeListener = simpleLimiter.acquire(context);
            if (!maybeListener.isPresent()) {
                return Optional.empty();
            }
            return Optional.of(new Listener() {
                @Override
                public void onSuccess() {
                    maybeListener.get().onSuccess();
                    updateMetrics(instrumentation, simpleLimiter);
                }

                @Override
                public void onDropped() {
                    maybeListener.get().onDropped();
                    updateMetrics(instrumentation, simpleLimiter);
                }

                @Override
                public void onIgnore() {
                    maybeListener.get().onIgnore();
                }
            });
        }
    }
```

dialogue-core/src/main/java/com/palantir/dialogue/core/ConcurrencyLimitedChannel.java
iamdanfox: Have switched to a uriIndex to resolve this.

dialogue-core/src/main/java/com/palantir/dialogue/core/ConcurrencyLimitedChannel.java
ferozco: ok if its not coming out cleanly we can try the gauges

dialogue-core/src/main/java/com/palantir/dialogue/core/ConcurrencyLimitedChannel.java
```diff
@@ -41,26 +45,36 @@
...
         this.limiter = limiter;
-    }
 
-    static ConcurrencyLimitedChannel create(LimitedChannel delegate, DialogueClientMetrics metrics) {
-        return new ConcurrencyLimitedChannel(
-                delegate, ConcurrencyLimitedChannel.createLimiter(SYSTEM_NANOTIME), metrics);
+        weakGauge(
+                taggedMetrics,
+                MetricName.builder()
+                        .safeName("dialogue.concurrencylimiter.utilization")
```
ferozco, line 62: kinda sad that we can't use metric schema :(

iamdanfox: Agreed, but I think it's ok to experiment with new APIs by just manually writing them. If we find this weak reference pattern is actually widespread and worth recommending, then we could bake it into tritium/metric-schema.

dialogue-core/src/main/java/com/palantir/dialogue/core/ConcurrencyLimitedChannel.java
```diff
@@ -41,26 +45,36 @@
     @Nullable
     private static final Void NO_CONTEXT = null;
 
-    private static final Ticker SYSTEM_NANOTIME = System::nanoTime;
-
     private final Meter limitedMeter;
     private final LimitedChannel delegate;
-    private final Limiter<Void> limiter;
+    private final SimpleLimiter<Void> limiter;
```
ferozco, line 50: do we have to narrow the type?

dialogue-core/src/main/java/com/palantir/dialogue/core/ConcurrencyLimitedChannel.java
iamdanfox: I think we could almost recommend _everyone_ express gauges in terms of a `T` and a `Function<T,Number>` (and make this a first-class tritium) because in the current world I bet the majority of gauges are actually preventing things being GC'd!

dialogue-core/src/main/java/com/palantir/dialogue/core/ConcurrencyLimitedChannel.java
iamdanfox: yes

image

dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueChannel.java
```diff
@@ -97,13 +98,17 @@ public void updateUris(Collection<String> uris) {
         Sets.SetView<String> newUris = Sets.difference(uniqueUris, limitedChannelByUri.keySet());
 
         staleUris.forEach(limitedChannelByUri::remove);
-        newUris.forEach(uri -> limitedChannelByUri.put(uri, createLimitedChannel(uri)));
+        ImmutableList<String> allUris = ImmutableList.<String>builder()
```
ferozco, line 101: These indices aren't stable across updates so it will probably pretty confusing to look at individual metrics. Not sure of a better solution, but worth flagging

ferozco:
👍

dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueChannel.java
iamdanfox: Yep. Given that it's not recommended to just hash things, I think this is the best we can do.  In practise, I hope it'll be reasonably easy to see n coloured lines go into a live-reload event and then either n+1 or n-1 lines come out.  People should be able to visually connect em up.

PR #444, opened 2020-02-26
Allow routing strategies to bypass channel limits

carterkozak:
This allows us to avoid failing hard when channels are limited
and we may be able to make a successful request. In most cases
the failure from a bad server provides more signal than a failure
from blocking requests client-side on a limited channel.

==COMMIT_MSG==
Allow routing strategies to bypass channel limits
==COMMIT_MSG==

## Possible downsides?
Possible that bugs can allow a single client to DoS a server.


carterkozak:
draft to collect feedback

dialogue-core/src/test/java/com/palantir/dialogue/core/PinUntilErrorChannelTest.java
```diff
@@ -71,7 +70,8 @@ public void before() {
         pinUntilError = new PinUntilErrorChannel(shufflingList, metrics);
     }
 
-    @Test
+    // TODO(ckozak): fix tests
+    // @Test
```
carterkozak, line 74: merge blocking.

dialogue-core/src/main/java/com/palantir/dialogue/core/LimitedChannel.java
```diff
@@ -27,6 +28,6 @@
  * overloaded). Semantics match {@link com.palantir.dialogue.Channel} aside from returning an
  * {@link Optional optional response}.
  */
-interface LimitedChannel {
+interface LimitedChannel extends Channel {
```
markelliot, line 31: this seems really strange as a semantic; the idea of having a LimitedChannel and a Channel are sort of core to ensure separation of concerns; would love for a discussion around this that includes @ellisjoe

ellisjoe:
So I think this sort of thing is probably related to Dan's desire to plumb channel ids down with the channels themselves, but the intent behind the way this was decomped was to completely decouple a request from the way it's handled by the underlying channels (eg: why should a retried request1 be routed using different logic than request2 being tried for the first time). I think I'd like to better understand why we think it's better for higher level logic to be responsible for plumbing requests through things like the LimitedChannel rather than having the limited channel be a bit smarter (for example, every so often it could ignore the limit, attempt the send the request, and then possible increase the limit).

I feel like blurring these lines is going to reintroduce a bunch of the complexity that exists in cjr today.

carterkozak:
This is a hack to approximate the desired end state of weighted channels based on our discussion this morning. The limited channel only allows us to attempt requests immediately, or fail to something else, there’s a necessary coupling between the node selection strategy and limiters that we are missing.

-ck

iamdanfox:
I think the outcome this achieves is something we want - people will _definitely_ rage at us if they have 5 uris configured but our client-side heuristic throws them an exception of "Failed to make a request" due to everything being limited (especially if our limiters turn out to be over-conservative).

In this scenario, it seems to me like the node selection strategy (i.e. the thing that understands that all channels are returning optional.empty) should have a way of getting a request out the door anyway.

markelliot:
Not sure I agree — the way we break limited channel boundaries today says either the limiting is too aggressive (because we hit this case and apparently it’s okay to break through it) or that limiting is entirely bogus (because we hit this case often enough that the _normal_ case is that we break through it). Further, the way we delay and retry even when limits are reached _increases_ load in circumstances where the limiting should’ve just immediately shed load without even making a request, protecting the system as a whole at the expense of the limited request. 

But, all that said, supposing we want to break through the limits we still don’t need to say all LimitedChannel implementations must be breakable. As an alternative structure, suppose the LimitedChannel implmentation were given two channels, one which is Limited and another which is just a Channel; then the implementation owns the full contract without leaking that _all_ LimitedChannel implementations are breakable.

PR #445, opened 2020-02-27
Pin until error strategy attempts to find an active channel

carterkozak:
PinUntilErrorChannel behavior matches round robin more closely
when many channels are blacklisted. Clients are no longer forced
to use N retries to find the next available channel.

==COMMIT_MSG==
Pin until error strategy attempts to find an active channel rather than bailing on the first limited channel. This matches round robin more closely when many channels are blacklisted. Clients are no longer forced to use N retries to find the next available channel.
==COMMIT_MSG==


iamdanfox (approved):
Yep makes sense. No need to burn retries hopping slowly through a list of blacklisted channels before we get to one that's ready!

PR #448, opened 2020-02-27
Reloading preserves pinned url

iamdanfox:
## Before this PR

When URLs live-reload, the list gets get reshuffled so we'll end up pinned to a new node. I think this is bad because url live-reloads happen at pretty much the same time across an entire stack, which means nodes in the middle of a blue/green will experience a sudden spike of unfamiliar traffic and cache misses.

## After this PR
==COMMIT_MSG==
The `DialogueChannel.builder().from()` method can be used to construct a new client (with a new set of urls) that will start off pinned to the same URL that it was before.
==COMMIT_MSG==

This ensures that clients will all keep refreshing on their existing 10 minute cadence so a new node will pick up traffic more gracefully.

Note that this PR is _not_ suggesting making fiddly mutations the whole way through our stack (as I think that will be impossible to maintain and super buggy). We still recreate the entire stack, it just gives us the opportunity to continue where we left off if possible.

As a nice side-effect, as the new builder API lets us plug in a clock, we can use _exactly_ the same codepath in prod and in our simulations! This should reduce the maintenance burden of the simulations.

## Possible downsides?
- a bunch of our channels now have intellij-generated `equals` methods. If people add more state to these channels they may forget to update these methods.


dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueChannel.java
```diff
@@ -0,0 +1,234 @@
...
+                case PIN_UNTIL_ERROR:
+                case PIN_UNTIL_ERROR_WITHOUT_RESHUFFLE:
+                    // create a fresh instance (ignoring the idea of live reloading initially) - we might discard this
+                    PinUntilErrorChannel goalInstance = PinUntilErrorChannel.of(
+                            config.nodeSelectionStrategy(), channels, DialoguePinuntilerrorMetrics.of(metrics), random);
+
+                    // find out whether we could have live-reloaded to get to the goal state
+                    Optional<PinUntilErrorChannel> maybeLiveReloaded = maybeExisting
+                            .flatMap(existing -> goalInstance.reloadableFrom(existing.nodeSelectionStrategy))
+                            .map(existing -> existing.liveReloadNewInstance(channels));
```
iamdanfox, line 213: this is where the magic happens

dialogue-core/src/test/java/com/palantir/dialogue/core/PinUntilErrorChannelTest.java
```diff
@@ -150,14 +152,71 @@ public void out_of_order_responses_dont_cause_us_to_switch_channel() throws Exce
...
 
         assertThat(getCode(pinUntilError))
-                .describedAs("We're still on channel2, as an older 500 doesn't mark this host as bad")
-                .isEqualTo(101);
+                .describedAs("We're still on channel0, as an older 500 doesn't mark this host as bad")
+                .isEqualTo(100);
+    }
+
+    @Test
+    void reloading_the_shuffling_version_stays_pinned_on_the_same_host_where_possible() {
```
iamdanfox, line 166: and here is the test showing it working :D

simulation/src/test/java/com/palantir/dialogue/core/Strategy.java
```diff
@@ -54,114 +48,69 @@ public Channel getChannel(Simulation simulation, Supplier<List<SimulationServer>
...
+                            .from(stubConfig())
+                            .taggedMetricRegistry(sim.taggedMetrics())
+                            .nodeSelectionStrategy(NodeSelectionStrategy.ROUND_ROBIN)
+                            .failedUrlCooldown(Duration.ofMillis(200))
+                            .clientQoS(ClientConfiguration.ClientQoS.DANGEROUS_DISABLE_SYMPATHETIC_CLIENT_QOS)
+                            .build())
+                    .clock(sim.clock())
+                    .random(random)
+                    .scheduler(sim.scheduler())
+                    .build();
```
iamdanfox, line 100: I think using the same ClientConfiguration in our simulation makes these a bit more believable. It also forces us to build new channels/functionality in a way that allows us to inject clocks/schedulers/randomness.

dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueChannel.java
```diff
@@ -0,0 +1,234 @@
...
+
+        private static LimitedChannel buildSingleUri(
+                Channel channel, DialogueClientMetrics clientMetrics, ClientConfiguration conf, Ticker clock) {
+            // Instrument inner-most channel with metrics so that we measure only the over-the-wire-time
+            Channel chan = new InstrumentedChannel(channel, clientMetrics);
+            // TracedChannel must wrap TracedRequestChannel to ensure requests have tracing headers.
+            chan = new TracedRequestChannel(chan);
+            chan = new TracedChannel(chan, "Dialogue-http-request");
+            LimitedChannel limitedChan = new ChannelToLimitedChannelAdapter(chan);
+            limitedChan = concurrencyLimiter(limitedChan, conf, clientMetrics, clock);
```
iamdanfox, line 155: a future piece of work would be to copy the state of the concurrencylimiter too. didn't want to go too crazy in this PR though.

dialogue-core/src/main/java/com/palantir/dialogue/core/PinUntilErrorChannel.java
```diff
@@ -144,7 +160,49 @@ private OptionalInt incrementHostIfNecessary(int currentIndex) {
         return saved ? OptionalInt.of(nextIndex) : OptionalInt.empty(); // we've moved on already
     }
 
-    interface NodeList {
+    Optional<PinUntilErrorChannel> reloadableFrom(Object other) {
```
carterkozak, line 163: The object casting logic is a bit difficult to follow, perhaps we can provide a `NodeSelector` interface with the functionality we need, and implement it for round-robin as well? Round robin could effectively do what we're already doing, but using a volatile delegate instance that's recreated when nodes change.

dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueChannel.java
carterkozak: If we used a strategy where we take a `List<uri>` and `function<uri, channel>` we can hold references to fully-wrapped/limited channels, solving the concurrency limiter state problem, and making the code easier to follow without reliance on equals/hashCode in all implementations. For backcompat from Channels.create I think we could pass a random uuid as the uri to retain the old behavior of always-recreate-everything. Thoughts?

dialogue-core/src/main/java/com/palantir/dialogue/core/Channels.java
```diff
@@ -16,93 +16,18 @@
...
 import com.palantir.dialogue.Channel;
-import com.palantir.logsafe.Preconditions;
-import com.palantir.logsafe.SafeArg;
-import com.palantir.logsafe.exceptions.SafeIllegalStateException;
-import com.palantir.logsafe.exceptions.SafeRuntimeException;
 import java.util.Collection;
-import java.util.List;
-import java.util.function.Function;
 
 public final class Channels {
```
carterkozak, line 23: Should we deprecate this? I think the builder is strictly better.

simulation/src/test/java/com/palantir/dialogue/core/Strategy.java
carterkozak: This is fantastic :-)

simulation/src/test/java/com/palantir/dialogue/core/Strategy.java
```diff
@@ -54,114 +48,69 @@ public Channel getChannel(Simulation simulation, Supplier<List<SimulationServer>
...
-                    .collect(Collectors.toList());
-            LimitedChannel limited1 = new RoundRobinChannel(limitedChannels);
-            return retryingChannel(sim, limited1);
+            return DialogueChannel.builder()
+                    .channels(channels)
+                    .clientConfiguration(ClientConfiguration.builder()
+                            .from(stubConfig())
+                            .taggedMetricRegistry(sim.taggedMetrics())
+                            .nodeSelectionStrategy(NodeSelectionStrategy.ROUND_ROBIN)
+                            .failedUrlCooldown(Duration.ofMillis(200))
```
carterkozak, line 59: Do we consume `failedUrlCooldown` anywhere?

iamdanfox: Don't think so, but you can't construct a ClientConfiguration instance without specifying it!!

carterkozak: Ah I see, thanks! Looks like the default in `ClientConfigurations.java` is `Duration.ZERO`, we may want to emulate that.

carterkozak: In a follow-up we could use positive values to wrap delegates with a BlacklistingChannel inside of RoundRobinChannel.

dialogue-core/src/main/java/com/palantir/dialogue/core/PinUntilErrorChannel.java
iamdanfox: I think the "reloadableFrom" then "liveReloadNewInstance" dance is really a key bit of the PR, so would like to try and make it as clear as possible. The reason I've got the instanceof check is because in the case where there's only a single uri, we have the clever short-circuiting thing:

```
            if (channels.size() == 1) {
                return channels.get(0); // no fancy node selection heuristic can save us if our one node goes down
            }
```

Also I want to make it clear that if someone was on ROUND_ROBIN, then there's no way to live-reload our way to PIN_UNTIL_ERROR, we just have to recreate.  That's where the `reloadableFrom` method would just throw up its hands and say I can't work worth with this, just use a fresh instance.

simulation/src/test/java/com/palantir/dialogue/core/Strategy.java
iamdanfox: Mmm looks like we can't get away with zero:
```
SafeIllegalArgumentException: If nodeSelectionStrategy is ROUND_ROBIN then failedUrlCooldown must be positive
	at com.palantir.logsafe.Preconditions.checkArgument(Preconditions.java:53)
	at com.palantir.conjure.java.client.config.ClientConfiguration.check(ClientConfiguration.java:132)
```

simulation/src/test/java/com/palantir/dialogue/core/Strategy.java
iamdanfox: Ok I've factored out all the builder stuff into a (hopefully) non-controversial PR: https://github.com/palantir/dialogue/pull/452

ferozco:
Closing in favour of #492

PR #451, opened 2020-02-27
simulate one big spike of requests

iamdanfox:
## Before this PR
From the `#dev-rpc` slack channel, this is intended to be somewhat representative of alta clients.

## After this PR
==COMMIT_MSG==
New simulation case 'one_big_spike' demonstrates the workflow where a single client may kick off a ton of requests in a hot loop and just wait for them all to come back before being able to return to the user.
==COMMIT_MSG==

# [GRAPHS](https://github.com/palantir/dialogue/blob/dfox/simulate-one-big-spike/simulation/src/test/resources/report.md#one_big_spikeconcurrency_limiter_blacklist_round_robin)

show 79% success rate

### DIFF
--- a/simulation/src/test/java/com/palantir/dialogue/core/SimulationTest.java
+++ b/simulation/src/test/java/com/palantir/dialogue/core/SimulationTest.java
@@ -393,6 +393,35 @@ public class SimulationTest {
                 .run();
     }
 
+    /**
+     * This simulates an alta client, which might load up some keys and then lookup each key in order to build a big
+     * response for the user. The goal is 100% client-perceived success here, because building up half the response
+     * is no good.
+     */
+    @Test
+    public void one_big_spike() {
+        int capacity = 100;
+        servers = servers(
+                SimulationServer.builder()
+                        .serverName("node1")
+                        .simulation(simulation)
+                        .handler(h -> h.respond200UntilCapacity(429, capacity).responseTime(Duration.ofMillis(150)))
+                        .build(),
+                SimulationServer.builder()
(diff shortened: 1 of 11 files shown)


simulation/src/test/java/com/palantir/dialogue/core/SimulationTest.java
```diff
@@ -393,6 +393,35 @@ public void uncommon_flakes() {
...
+     * is no good.
+     */
+    @Test
+    public void one_big_spike() {
+        int capacity = 100;
+        servers = servers(
+                SimulationServer.builder()
+                        .serverName("node1")
+                        .simulation(simulation)
+                        .handler(h -> h.respond200UntilCapacity(429, capacity).responseTime(Duration.ofMillis(150)))
```
carterkozak, line 408: Is 429 representative here? I think we use 503 in most cases, unless alta does something special.

iamdanfox: Looks like alta uses `QosException.throttle()` (see QosAltaTransaction.java#L186), so I think it is representative.

carterkozak: 👍 thanks. I'm not sure if it's intentional that ConcurrencyLimitedChannel only checks 429 and not 503.

iamdanfox: Yeah that just seemed weird to me too!

carterkozak (approved):

PR #452, opened 2020-02-27
Use same Channels#builder API in prod & simulation code

iamdanfox:
## Before this PR

I spent all this time building these elaborate simulations, but they constructed all their channels in an entirely different way than we normally do in prod (they didn't use `Channels#create`). This made me a little suspicious of them, and made me trust them a tiny bit less.

## After this PR
==COMMIT_MSG==
Simulations use the same codepath to construct channels that we used in prod.
==COMMIT_MSG==


dialogue-core/src/main/java/com/palantir/dialogue/core/ConcurrencyLimitedChannel.java
```diff
@@ -94,6 +94,10 @@ static ConcurrencyLimitedChannel create(LimitedChannel delegate, DialogueClientM
         }
     }
 
+    @Override
+    public String toString() {
+        return "ConcurrencyLimitedChannel{" + delegate + '}';
```
carterkozak, line 99: Nice, thanks for adding these!

dialogue-core/src/main/java/com/palantir/dialogue/core/FixedLimitedChannel.java
```diff
@@ -82,4 +82,16 @@ private void logExhaustion(Endpoint endpoint) {
...
         }
     }
+
+    @Override
+    public String toString() {
+        return "FixedLimitedChannel{"
+                + "delegate="
+                + delegate
+                + ", usedPermits="
+                + usedPermits
```
carterkozak, line 92: Heh, this could be confusing because toString isn't evaluated when limits are, but I'm all for more data.

carterkozak (approved):

PR #454, opened 2020-02-28
UnknownRemoteException

iamdanfox:
## Before this PR

A lot of people compile against conjure-java-runtime-api's UnknownRemoteException to cases outside the conjure ecosystem (e.g. nginx gateway timeout), but dialogue currently throws a made-up SafeRuntimeException instead.

https://internal-github/search?l=Java&q=org%3Afoundry+UnknownRemoteException&type=Code

## After this PR
==COMMIT_MSG==
DefaultErrorDecoder throws UnknownRemoteExceptions when we can't deserialize json into our expected `SerializableError` type.
==COMMIT_MSG==

### DIFF
--- a/dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/DefaultErrorDecoder.java
+++ b/dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/DefaultErrorDecoder.java
@@ -24,8 +24,6 @@ import com.palantir.conjure.java.api.errors.UnknownRemoteException;
 import com.palantir.conjure.java.serialization.ObjectMappers;
 import com.palantir.dialogue.ErrorDecoder;
 import com.palantir.dialogue.Response;
-import com.palantir.logsafe.SafeArg;
-import com.palantir.logsafe.exceptions.SafeRuntimeException;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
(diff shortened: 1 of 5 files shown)


dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/DefaultErrorDecoder.java
```diff
@@ -44,29 +42,26 @@ public RemoteException decode(Response response) {
...
+
         Optional<String> contentType = response.getFirstHeader(HttpHeaders.CONTENT_TYPE);
         if (contentType.isPresent() && contentType.get().equals("application/json")) {
-            final String body;
-            try {
-                body = toString(response.body());
-            } catch (NullPointerException | IOException e) {
-                throw new SafeRuntimeException(
-                        "Failed to read response body, could not deserialize SerializableError", e);
-            }
```
iamdanfox, line 55: @uschi2000 any concerns here?

carterkozak:
This would be a good time to spec out a similar exception type for client failures that don’t make it to a server (no status code). Feign uses FeignException, I recall that being handled specifically in some code, do we want a similar type for conjure clients?

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/DefaultErrorDecoder.java
```diff
@@ -44,29 +42,26 @@ public RemoteException decode(Response response) {
         // TODO(rfink): What about HTTP/101 switching protocols?
         // TODO(rfink): What about HEAD requests?
 
+        String body;
+        try {
+            body = toString(response.body());
+        } catch (NullPointerException | IOException e) {
+            UnknownRemoteException exception = new UnknownRemoteException(response.code(), "<unparseable>");
```
carterkozak, line 49: Do we need special handling in ConjureExceptions.blocking?

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/DefaultErrorDecoder.java
uschi2000: yeah, I think this is ok.

carterkozak (approved):

PR #457, opened 2020-02-28
Apache connection pool instrumentation

iamdanfox:
## Before this PR

We have hypotheses that connection re-use has a big impact on our most performance-sensitive services like timelock. Currently, we just use the tls.handshake graphs to 

## After this PR
==COMMIT_MSG==
Each closeable apache client now exposes methods which reveal statistics about the connection pool.
==COMMIT_MSG==

After merging, we'll have to carefully wire these up to gauges in witchcraft, ensuring that we don't prevent objects from being GC'd just by having a reference in a tritium registry somewhere.

## Possible downsides?
- if the apache defaults change then by building our connection manager ourself we might miss out on improvements

### DIFF
--- a/dialogue-apache-hc4-client/src/main/java/com/palantir/dialogue/hc4/ApacheHttpClientChannels.java
+++ b/dialogue-apache-hc4-client/src/main/java/com/palantir/dialogue/hc4/ApacheHttpClientChannels.java
@@ -15,6 +15,9 @@
  */
 package com.palantir.dialogue.hc4;
 
+import com.codahale.metrics.Gauge;
+import com.codahale.metrics.Metric;
+import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.primitives.Ints;
 import com.palantir.conjure.java.api.config.service.BasicCredentials;
@@ -27,6 +30,8 @@ import com.palantir.logsafe.Preconditions;
 import com.palantir.logsafe.SafeArg;
 import com.palantir.logsafe.exceptions.SafeIllegalArgumentException;
 import com.palantir.logsafe.exceptions.SafeRuntimeException;
+import com.palantir.tritium.metrics.registry.MetricName;
+import com.palantir.tritium.metrics.registry.TaggedMetricSet;
 import java.io.Closeable;
 import java.io.IOException;
 import java.net.MalformedURLException;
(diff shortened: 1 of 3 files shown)


dialogue-apache-hc4-client/src/test/java/com/palantir/dialogue/hc4/ApacheHttpClientChannelsTest.java
```diff
@@ -58,6 +59,26 @@ public void close_works() throws Exception {
...
+
+            Channel channel = ApacheHttpClientChannels.createSingleUri("http://neverssl.com", client);
+            ListenableFuture<Response> response =
+                    channel.execute(new TestEndpoint(), Request.builder().build());
+            assertThat(Futures.getUnchecked(response).code()).isEqualTo(200);
+
+            assertThat(client.getNumRoutes()).describedAs("routes").isEqualTo(1);
+            assertThat(client.getPoolStats().getAvailable())
+                    .describedAs("available")
+                    .isEqualTo(0);
```
iamdanfox, line 93: I think graphing the number `available` will be kinda interesting, as this should tell us how quickly we could handle bursts of traffic 

```java
 /**
     * Gets the number idle persistent connections.
     * <p>
     * The total number of connections in the pool is equal to {@code available} plus {@code leased}.
     * </p>
     *
     * @return number idle persistent connections.
     */
    public int getAvailable() {
```

dialogue-apache-hc4-client/src/main/java/com/palantir/dialogue/hc4/ApacheHttpClientChannels.java
```diff
@@ -96,6 +100,26 @@ public static CloseableClient createCloseableHttpClient(ClientConfiguration conf
...
+                new String[] {"TLSv1.2"},
+                conf.enableGcmCipherSuites()
+                        ? jvmSupportedCipherSuites(CipherSuites.allCipherSuites())
+                        : jvmSupportedCipherSuites(CipherSuites.fastCipherSuites()),
+                new DefaultHostnameVerifier());
+
+        PoolingHttpClientConnectionManager connectionManager = new PoolingHttpClientConnectionManager(
+                RegistryBuilder.<ConnectionSocketFactory>create()
+                        .register("http", PlainConnectionSocketFactory.getSocketFactory())
+                        .register("https", sslSocketFactory)
```
carterkozak, line 116: I wonder if we can wrap these to add a meter  of connection rate, and wrap the pool manager to capture the rate connections requested?

iamdanfox: Mmm that would be cool, so dividing one by the other tells us how effectively we're re-using... Do we not already get that connection rate from the tritium tls.handshakes?

carterkozak: Tritium handshake instrumentation would work assuming https, but I don’t think we wire it up yet.

ferozco:
Closing in favour of #498

PR #461, opened 2020-02-28
Use SafeThreadLocalRandom from separate project

carterkozak:
https://github.com/palantir/safe-thread-local-random

==COMMIT_MSG==
Use SafeThreadLocalRandom from separate project
==COMMIT_MSG==


ferozco:
👍

PR #476, opened 2020-03-03
use generated clients in test

ferozco:
## Before this PR
We used hand rolled dialogue clients in our tests

## After this PR
Use conjure defined and generated clients

## Possible downsides?
It makes breaks to the generator target harder


dialogue-example/dialogue-example-dialogue/build/reports/checkstyle/main.html
```diff
@@ -0,0 +1,119 @@
+<html>
```
carterkozak: I think we're missing `build/` in the .gitignore

ferozco: 🤦‍♂

ferozco:
We'll probably want to wait on #475 merging and conjure-java adopting the change before merging this in

.gitignore
```diff
@@ -8,6 +8,7 @@
 .node
 /build/
 */build/
+*/*/build
 examples/*/build/
 extras/*/build/
```
carterkozak: I think these two can be deleted since we have `*/*/build`

ferozco:
Looks like my modules cache somehow got corrupted and was causing --write-locks to emit the wrong result 😨 Luckily `rm -rf ~/.gradle/caches/modules-2/` seems to have fixed it up.

carterkozak:
Oh wow, next time I wonder if it's worth backing up the directory to root cause the corruption when we have cycles to dig into it

carterkozak:
Thanks! :+1:

PR #479, opened 2020-03-04
Remove unused methods from PlainSerDe

carterkozak:
List and Set serializers:
Generated code uses loops in order to handle nonstandard types,
aliases and external type imports. A subset of cases could be
updated to use these specific methods, but it would increase
complexity of the generator, and require more specific test coverage.
The loop-based approach has the fringe benefit of avoiding
intermediate collection allocation, relying on the builders internal
state instead.

Optional serializers:
The request object does not support optional values, so writes
to the request builder always involve checking optional.isPresent
before calling standard serializers on the result of optional.get.
The optional serializer methods aren't currently used, nor would
they reduce the complexity of the generated code if they were.

## After this PR
==COMMIT_MSG==
Remove unused methods from PlainSerDe
==COMMIT_MSG==

## Possible downsides?
break

### DIFF
--- a/.palantir/revapi.yml
+++ b/.palantir/revapi.yml
@@ -26,3 +26,120 @@ acceptedBreaks:
       old: "class com.palantir.dialogue.RemoteExceptions"
       new: null
       justification: "alpha"
+  "0.12.3":
+    com.palantir.dialogue:dialogue-apache-hc4-client: []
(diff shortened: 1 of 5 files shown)


ferozco:
👍 Awesome! happy to get rid of these

PR #480, opened 2020-03-04
Remove unnecessary header constants from the target api

carterkozak:
==COMMIT_MSG==
Remove unnecessary header constants from the target api
==COMMIT_MSG==

## Possible downsides?
break


ferozco (commented):
👍 we're already pretty tied to guava so might as well use their definitions

ferozco:
👍

PR #482, opened 2020-03-04
fixes #468: Dialogue clients send Accept headers

### LINKED ISSUE #468: Dialogue requests are missing an Accept header
## What happened?

No accept header is added to dialogue requests

requests should `Accept` content types based on their Deserializer instance.

carterkozak:
## Before this PR
1. Clients violate the conjure specification by excluding Accept headers.
2. Endpoints which return `binary` failed to parse errors, instead producing an illegal argument exception "Unsupported Content-Type" (with no response status check).

## After this PR
==COMMIT_MSG==
Dialogue clients send Accept headers
==COMMIT_MSG==

## Possible downsides?
api churn

### DIFF
--- a/dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ConjureBodySerDe.java
+++ b/dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ConjureBodySerDe.java
@@ -34,6 +34,7 @@ import java.io.InputStream;
 import java.io.OutputStream;
 import java.util.List;
 import java.util.Optional;
+import java.util.stream.Collectors;
 import javax.ws.rs.core.HttpHeaders;
 
 /** Package private internal API. */
@@ -74,13 +75,8 @@ final class ConjureBodySerDe implements BodySerDe {
     }
 
     @Override
-    @SuppressWarnings("NullAway") // empty body is a special case
     public Deserializer<Void> emptyBodyDeserializer() {
-        return input -> {
-            // We should not fail if a server that previously returned nothing starts returning a response
-            input.close();
-            return null;
-        };
+        return EmptyBodyDeserializer.INSTANCE;
     }
 
     @Override
@@ -154,12 +150,16 @@ final class ConjureBodySerDe implements BodySerDe {
 
         private final ImmutableList<EncodingDeserializerContainer<T>> encodings;
         private final ErrorDecoder errorDecoder;
+        private final Optional<String> acceptValue;
 
         EncodingDeserializerRegistry(List<Encoding> encodings, ErrorDecoder errorDecoder, TypeMarker<T> token) {
             this.encodings = encodings.stream()
                     .map(encoding -> new EncodingDeserializerContainer<>(encoding, token))
                     .collect(ImmutableList.toImmutableList());
             this.errorDecoder = errorDecoder;
+            // Encodings are applied to the accept header in the order of preference based on the provided list.
+            this.acceptValue =
+                    Optional.of(encodings.stream().map(Encoding::getContentType).collect(Collectors.joining(", ")));
         }
 
         @Override
@@ -175,6 +175,11 @@ final class ConjureBodySerDe implements BodySerDe {
             } finally {
                 response.close();
             }
+        }
+
+        @Override
+        public Optional<String> accepts() {
+            return acceptValue;
         }
 
         /** Returns the {@link EncodingDeserializerContainer} to use to deserialize the request body. */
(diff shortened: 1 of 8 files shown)


dialogue-target/src/main/java/com/palantir/dialogue/BodySerDe.java
```diff
@@ -40,6 +48,11 @@
      * <p>
      * This method is named <pre>deserializeInputStream</pre> not <pre>deserializeBinary</pre>
      * to support future streaming binary bindings without conflicting method signatures.
+     *
+     * @deprecated Prefer {@link #inputStreamDeserializer()}
      */
-    InputStream deserializeInputStream(Response response);
+    @Deprecated
+    default InputStream deserializeInputStream(Response response) {
```
ferozco, line 55: Why not just get rid of this

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/DefaultClients.java
```diff
@@ -120,4 +126,20 @@ public void onFailure(Throwable throwable) {
             log.info("Canceled call failed", throwable);
         }
     }
+
+    private static Request accepting(Request original, String acceptValue) {
+        Preconditions.checkNotNull(acceptValue, "Accept value is required");
+        Preconditions.checkState(!acceptValue.isEmpty(), "Accept value must not be empty");
+        if (original.headerParams().containsKey(HttpHeaders.ACCEPT)) {
```
ferozco, line 133: should we append the acceptValue?

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ConjureBodySerDe.java
```diff
@@ -177,6 +172,11 @@ public T deserialize(Response response) {
             }
         }
 
+        @Override
+        public Optional<String> accepts() {
```
ferozco, line 176: Instead of joining the separate accept headers into a single string should we expose a `List<String>`

dialogue-target/src/main/java/com/palantir/dialogue/BodySerDe.java
carterkozak: our circular dependency on generated code would break

dialogue-serde/src/test/java/com/palantir/conjure/java/dialogue/serde/DefaultClientsTest.java
```diff
@@ -60,4 +70,19 @@ public void testCall() throws ExecutionException, InterruptedException {
...
+        when(deserializer.deserialize(eq(response))).thenReturn("value");
+        String expectedAccept = "application/json";
+        when(deserializer.accepts()).thenReturn(Optional.of(expectedAccept));
+        when(channel.execute(eq(endpoint), any())).thenReturn(Futures.immediateFuture(response));
+        ListenableFuture<String> result = DefaultClients.INSTANCE.call(channel, endpoint, request, deserializer);
+        assertThat(result).isDone();
+        assertThat(result.get()).isEqualTo("value");
+        verify(channel).execute(eq(endpoint), requestCaptor.capture());
+        assertThat(requestCaptor.getValue().headerParams().get(HttpHeaders.ACCEPT))
+                .containsExactly(expectedAccept);
```
ferozco: nit: I find the separating different parts of the test with whitespace makes it much more readable
```suggestion
        String expectedAccept = "application/json";
        Request request = Request.builder().build();

        when(deserializer.deserialize(eq(response))).thenReturn("value");
        when(deserializer.accepts()).thenReturn(Optional.of(expectedAccept));
        when(channel.execute(eq(endpoint), any())).thenReturn(Futures.immediateFuture(response));

        ListenableFuture<String> result = DefaultClients.INSTANCE.call(channel, endpoint, request, deserializer);

        assertThat(result).isDone();
        assertThat(result.get()).isEqualTo("value");
        verify(channel).execute(eq(endpoint), requestCaptor.capture());
        assertThat(requestCaptor.getValue().headerParams().get(HttpHeaders.ACCEPT))
                .containsExactly(expectedAccept);
```

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/DefaultClients.java
carterkozak: I don't think so, we'd need to deduplicate repeated accept values, and the existing values will be preferred since we're using a list, which isn't what we want here. I considered throwing instead of logging but it's difficult to trace where the value was added (unlike the null/empty inputs) so we would want to return a future.

dialogue-serde/src/test/java/com/palantir/conjure/java/dialogue/serde/DefaultClientsTest.java
carterkozak: :+1: good call

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ConjureBodySerDe.java
carterkozak: I tried this initially, but found that it maps to adding multiple `Accept` headers each with a single value, but it's most common to use `Accept` header with multiple values. The optional approach allows us to pre-calculate the value.

ferozco: It seems surprising to encode multiple accept values as a header value since in a multimap set up  I would expect each value in the map to be a single header value.

carterkozak: I completely agree that multiple accept header key/value pairs would be clean, but I don't think I've seen it used that way in practice. I think we'll reduce odd edge case failures by following the principle of least astonishment.

ferozco:
👍

PR #483, opened 2020-03-04
Dialogue Response headers are represented using a multimap

carterkozak:
## After this PR
==COMMIT_MSG==
Dialogue Response headers are represented using a multimap for symmetry with request headers.
==COMMIT_MSG==

## Possible downsides?
break


versions.lock
```diff
@@ -44,6 +44,7 @@ net.bytebuddy:byte-buddy-agent:1.10.5 (1 constraints: 410b37de)
 org.apache.httpcomponents:httpclient:4.5.10 (1 constraints: 3c054b3b)
 org.apache.httpcomponents:httpcore:4.4.12 (1 constraints: 581015a2)
 org.assertj:assertj-core:3.15.0 (3 constraints: 4f25f54b)
+org.assertj:assertj-guava:3.3.0 (1 constraints: 08050336)
```
ferozco, line 47: this probably should be in the test dependencies section. Perhaps we can mark the test project as such: https://github.com/palantir/gradle-consistent-versions/#common-workflow-internal-test-utility-projects

ferozco (approved):

versions.lock
carterkozak: Thanks for the link! Added to my backlog to revisit.

PR #484, opened 2020-03-04
Remove the arbitrary FixedLimitedChannel

carterkozak:
## Before this PR
The limit provided by FixedLimitedChannel by definition does
not scale, which is the antithesis of what dialogue aims to provide.

## After this PR
==COMMIT_MSG==
Remove the arbitrary FixedLimitedChannel
==COMMIT_MSG==


ferozco:
👍

PR #485, opened 2020-03-04
instrument retrying channels scheduled executor service

ferozco:
## Before this PR
We had no metrics on number of pending requests

## After this PR
==COMMIT_MSG==
We produce metrics about requests that are going to be retried
==COMMIT_MSG==

## Possible downsides?
This uses tritium to instrument the scheduledExecutor so the metric names won't be the best


dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -80,6 +83,7 @@
             ClientConfiguration.RetryOnTimeout retryOnTimeout) {
         this(
                 delegate,
+                new DefaultTaggedMetricRegistry(),
```
carterkozak, line 88: I think we need to add this to the constructor and pass the registry from our ClientConfiguration

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -245,4 +250,10 @@ private static boolean shouldPropagateQos(ClientConfiguration.ServerQoS serverQo
         throw new SafeIllegalStateException(
                 "Encountered unknown propagate QoS configuration", SafeArg.of("serverQoS", serverQoS));
     }
+
+    private static ListeningScheduledExecutorService instrument(
+            ScheduledExecutorService delegate, TaggedMetricRegistry metrics) {
+        return MoreExecutors.listeningDecorator(
+                Tracers.wrap(SCHEDULER_NAME, MetricRegistries.instrument(metrics, delegate, SCHEDULER_NAME)));
```
carterkozak, line 259: nice

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
ferozco: Yes we do do that, this constructor is only visible for testing. Let me  make that more clear

carterkozak: oh doh, I misread, sorry!

carterkozak (approved):
thanks!

ferozco:
🤖

PR #486, opened 2020-03-04
Report metrics for in-flight requests

carterkozak:
## Before this PR
Difficult to tell how many requests were actively in flight.

## After this PR
==COMMIT_MSG==
Report metrics for in-flight requests
==COMMIT_MSG==

## Possible downsides?
metrics ain't free?

### DIFF
--- a/dialogue-core/src/main/metrics/dialogue-core-metrics.yml
+++ b/dialogue-core/src/main/metrics/dialogue-core-metrics.yml
@@ -9,6 +9,10 @@ namespaces:
         type: timer
         tags: [service-name]
         docs: Request time, note that this does not include time spent reading the response body.
+      requests.active:
+        type: counter
+        tags: [stage]
+        docs: Number of requests that are actively running. The stage may refer to 'in-flight' requests actively executing over the wire, or 'processing' which may be awaiting a client or backing off for a retry.
       deprecations:
         type: meter
         tags: [service-name]
--- a/dialogue-core/src/main/metrics/dialogue-core-metrics.yml
+++ b/dialogue-core/src/main/metrics/dialogue-core-metrics.yml
@@ -9,10 +9,13 @@ namespaces:
         type: timer
         tags: [service-name]
         docs: Request time, note that this does not include time spent reading the response body.
-      requests.active:
+      request.active:
         type: counter
-        tags: [stage]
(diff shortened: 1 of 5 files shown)


carterkozak:
Draft, still needs test coverage.

A couple open questions before I clean it up, will add them inline.

dialogue-core/src/main/metrics/dialogue-core-metrics.yml
```diff
@@ -9,6 +9,10 @@ namespaces:
         type: timer
         tags: [service-name]
         docs: Request time, note that this does not include time spent reading the response body.
+      requests.active:
+        type: counter
+        tags: [stage]
```
carterkozak: I considered adding `service-name` here as well, that way we have more insight into precisely where our active requests are going. Thoughts?

dialogue-core/src/main/metrics/dialogue-core-metrics.yml
```diff
@@ -9,6 +9,10 @@ namespaces:
...
+        docs: Number of requests that are actively running. The stage may refer to 'in-flight' requests actively executing over the wire, or 'processing' which may be awaiting a client or backing off for a retry.
```
carterkozak: Looking for feedback on naming `dialogue.client.requests.active` should probably use singular `request`.
stage `in-flight`: maybe rename to `wire`.
`processing` is fairly vague, but as such it's not going to be incorrect.

ferozco: `request.active` seems alright.  Agree `in-flight` is a bit confusing since from a users perspective the request is `in-flight` as soon as it hits our code.  Perhaps `running` and `queued` for consistency with CJR?

dialogue-core/src/main/metrics/dialogue-core-metrics.yml
ferozco: Yes I think adding service name would be good

dialogue-core/src/main/metrics/dialogue-core-metrics.yml
carterkozak: I like it, thanks!

carterkozak: oh hmm, queued isn't quite right because it includes everything, both queued and running.

ferozco:
👍

PR #489, opened 2020-03-05
Remove stopgap components from circular generator dependency upgrade

carterkozak:
## Before this PR
unnecessary cruft in the api

## After this PR
==COMMIT_MSG==
Remove legacy deserializer apis left over to allow upgrading both the generator and target.
==COMMIT_MSG==

## Possible downsides?
break


ferozco:
👍

PR #491, opened 2020-03-05
Regenerate simulation data with current metrics

carterkozak:
Images aren't regenerated unless the report txt is updated, which
includes a subset of the data.

## After this PR
==COMMIT_MSG==
Regenerate simulation data with current metrics
==COMMIT_MSG==


ferozco:
👍

PR #492, opened 2020-03-05
Support live reloading uris

ferozco:
Still a WIP wanted to get some thoughts on the approach before commiting to fixing up the simulator

## Before this PR
We would force consumers to throw away all previous channel state when any part of the client configuration changed.

## After this PR
==COMMIT_MSG==
Persist all channel state across updates to uris
==COMMIT_MSG==


dialogue-core/src/main/java/com/palantir/dialogue/core/Channels.java
```diff
@@ -1,184 +0,0 @@
...
-
-    public static Channel create(Collection<? extends Channel> channels, ClientConfiguration config) {
-        return builder().channels(channels).clientConfiguration(config).build();
-    }
-
-    public static Builder builder() {
-        return new Builder();
-    }
-
-    public static final class Builder {
```
ferozco, line 50: mostly moved around in `DialogueChannel` but git didn't get it

dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueChannel.java
```diff
@@ -0,0 +1,276 @@
...
+        this.delegate = wrap(
+                new SupplierChannel(nodeSelectionStrategy::get), clientConfiguration, scheduler, random, clientMetrics);
+    }
+
+    @Override
+    public ListenableFuture<Response> execute(Endpoint endpoint, Request request) {
+        return delegate.execute(endpoint, request);
+    }
+
+    public void updateUris(List<String> uris) {
```
ferozco: Considered exposing an update method that would recreate the entire client if the client configuration wasn't identical (except for uris) but it seemed like more faff than it was worth

dialogue-core/src/main/java/com/palantir/dialogue/core/PinUntilErrorChannel.java
```diff
@@ -82,27 +82,41 @@ static PinUntilErrorChannel of(
...
+            DialoguePinuntilerrorMetrics metrics,
+            Random random) {
         /**
          * The *initial* list is shuffled to ensure that clients across the fleet don't all traverse the in the
          * same order.  If they did, then restarting one upstream node n would shift all its traffic (from all
          * servers) to upstream n+1. When n+1 restarts, it would all shift to n+2. This results in the disastrous
          * situation where there might be many nodes but all clients have decided to hammer one of them.
          */
         ImmutableList<LimitedChannel> initialShuffle = shuffleImmutableList(channels, random);
+        int initialHost = initialShuffle.indexOf(initialChannel);
```
ferozco, line 102: We only rely on reference equality since in LimitedChannels are reused across updates

dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueChannel.java
carterkozak: Makes sense, we can handle that externally

carterkozak: I think this method should be synchronized. May not be 100% necessary, but concurrent calls to updateUris could leave us in an inconsistent state.

dialogue-core/src/main/java/com/palantir/dialogue/core/PinUntilErrorChannel.java
carterkozak: I think that's reasonable, future us might benefit from a comment along those lines!

dialogue-apache-hc4-client/src/main/java/com/palantir/dialogue/hc4/ApacheHttpClientChannels.java
```diff
@@ -78,10 +77,10 @@ private ApacheHttpClientChannels() {}
 
     public static Channel create(ClientConfiguration conf) {
         CloseableClient client = createCloseableHttpClient(conf);
-        List<Channel> channels =
-                conf.uris().stream().map(uri -> createSingleUri(uri, client)).collect(Collectors.toList());
-
-        return Channels.create(channels, conf);
+        return DialogueChannel.builder()
+                .clientConfiguration(conf)
+                .channelFactory(uri -> createSingleUri(uri, client))
```
carterkozak, line 82: nice

dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueChannel.java
ferozco: I thought about that. Currently concurrent requests to updateUris will result in a channel being created with a limited channel corresponding the the intersection of all `uris` which I think it probably ok

uschi2000:
@ferozco Do we have evidence that this solves an actual problem? E.g., how often do URIs get reloaded/changed, and when they do change, what bad thing happens if we start with clean state?

ferozco:
> @ferozco Do we have evidence that this solves an actual problem? E.g., how often do URIs get reloaded/changed, and when they do change, what bad thing happens if we start with clean state?

URI changes can happen whenever a service upgrades or goes through the blue/green process. For small stacks with few replicas per service this would not be incredibly frequent, but for larger stacks with many replicas this could be quite frequent.

By starting from a clean slate on URI updates you lose all of the information the client had slowly built up about the services it was communicating with leading to decreased throughput and increased failure rates. For example:
- Resetting blacklisting will result in requests being made to a host that is in a degraded state
- Resetting concurrency limiters would artificial limit the number of concurrent requests while the limit is slowly re-learned
- Resetting node selection strategy (specifically pinUntilError) could result in slower responses time due to a loss of request locality

carterkozak (approved):

PR #493, opened 2020-03-05
Support weighted encodings

carterkozak:
==COMMIT_MSG==
Support weighted encodings for `Accept` type preference.
==COMMIT_MSG==

### DIFF
--- a/dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ConjureBodySerDe.java
+++ b/dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ConjureBodySerDe.java
@@ -18,6 +18,7 @@ package com.palantir.conjure.java.dialogue.serde;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Lists;
 import com.palantir.dialogue.BinaryRequestBody;
 import com.palantir.dialogue.BodySerDe;
 import com.palantir.dialogue.Deserializer;
@@ -32,6 +33,8 @@ import com.palantir.logsafe.exceptions.SafeRuntimeException;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.util.ArrayList;
+import java.util.Comparator;
 import java.util.List;
 import java.util.Optional;
 import java.util.stream.Collectors;
@@ -40,7 +43,7 @@ import javax.ws.rs.core.HttpHeaders;
 /** Package private internal API. */
 final class ConjureBodySerDe implements BodySerDe {
 
-    private final List<Encoding> encodings;
+    private final List<Encoding> encodingsSortedByWeight;
     private final ErrorDecoder errorDecoder;
     private final Encoding defaultEncoding;
     private final Deserializer<InputStream> binaryInputStreamDeserializer;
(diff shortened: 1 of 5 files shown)


dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/WeightedEncoding.java
```diff
@@ -0,0 +1,75 @@
...
+
+import com.palantir.logsafe.Preconditions;
+import java.util.Objects;
+
+/**
+ * Represents an encoding with a weight for <code>Accept</code> types.
+ * Note that the weight may not be applied to the Accept header, rather
+ * used to order values.
+ */
+final class WeightedEncoding {
```
carterkozak, line 27: initially this was public, but the API feels a lot cleaner exposing a builder method overload with weight.

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ConjureBodySerDe.java
```diff
@@ -50,30 +53,36 @@
      * {@link Encoding#supportsContentType supports} the serialization format {@link HttpHeaders#ACCEPT accepted}
      * by a given request, or the first serializer if no such serializer can be found.
      */
-    ConjureBodySerDe(List<Encoding> encodings) {
-        // TODO(jellis): consider supporting cbor encoded errors
```
carterkozak, line 54: Removed this todo, a while ago we discussed content-type negotiation and agreed that failures should always use the most basic encoding, that way if bugs are introduced we have the best shot at understanding what is going wrong.

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/WeightedEncoding.java
ferozco, line 27: might be simpler to use an immutable?

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/WeightedEncoding.java
```diff
@@ -0,0 +1,75 @@
...
+ */
+
+package com.palantir.conjure.java.dialogue.serde;
+
+import com.palantir.logsafe.Preconditions;
+import java.util.Objects;
+
+/**
+ * Represents an encoding with a weight for <code>Accept</code> types.
+ * Note that the weight may not be applied to the Accept header, rather
```
ferozco, line 24: Why lock ourselves out from applying q-factor to accept headers?

carterkozak: My goal was to make it clear that it doesn't necessarily use q-factor values, but could in the future.

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/WeightedEncoding.java
carterkozak: I could go either way, we don't use annotation processors in this module yet. I tend to avoid immutables for fewer than three parameters ¯\_(ツ)_/¯

ferozco: fair, I lean towards immutables for consistency with other data classes and to reduce boilerplate

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/DefaultConjureRuntime.java
```diff
@@ -62,13 +65,23 @@ public Clients clients() {
...
 
-        private final List<Encoding> encodings = new ArrayList<>();
+        private final List<WeightedEncoding> encodings = new ArrayList<>();
 
         private Builder() {}
 
         @CanIgnoreReturnValue
         public Builder encodings(Encoding value) {
-            encodings.add(Preconditions.checkNotNull(value, "Value is required"));
+            encodings.add(WeightedEncoding.of(Preconditions.checkNotNull(value, "Value is required")));
```
ferozco: unnecessary since the weight encoding does the same checks

ferozco:
👍

PR #494, opened 2020-03-06
Replace slf4j-simple with log4j2 for tests

carterkozak:
This gives us more flexibility to collect data while running
tests.

==COMMIT_MSG==
Replace slf4j-simple with log4j2 for tests
==COMMIT_MSG==


ferozco:
👍

PR #498, opened 2020-03-06
Apache client connection pool instrumentation

carterkozak:
See #457

## After this PR
==COMMIT_MSG==
Apache client connection pool instrumentation
==COMMIT_MSG==


dialogue-apache-hc4-client/src/main/java/com/palantir/dialogue/hc4/ApacheClientGauges.java
```diff
@@ -0,0 +1,82 @@
...
+        poolMetrics.size().state("idle").build(idle);
+        poolMetrics.size().state("leased").build(leased);
+        poolMetrics.size().state("pending").build(pending);
+    }
+
+    static <T> T collect(
+            Function<PoolingHttpClientConnectionManager, T> extractor, BiFunction<T, T, T> combiner, T initial) {
+        T current = initial;
+        Iterator<WeakReference<PoolingHttpClientConnectionManager>> iterator = pools.iterator();
+        while (iterator.hasNext()) {
```
ferozco, line 62: for loop?

carterkozak: For loop doesn't allow the iterator.remove operation we use when a garbage collected item is encountered.

dialogue-apache-hc4-client/src/main/java/com/palantir/dialogue/hc4/ApacheHttpClientChannels.java
```diff
@@ -92,9 +96,37 @@ public static CloseableClient createCloseableHttpClient(ClientConfiguration conf
...
+
+        ApacheClientGauges.register(connectionManager);
+
+        connectionManager.setDefaultSocketConfig(socketConfig);
+        connectionManager.setMaxTotal(Integer.MAX_VALUE);
+        connectionManager.setDefaultMaxPerRoute(Integer.MAX_VALUE);
+        // Increased from two seconds to twenty-five seconds because we have strong support for retries
+        // and can optimistically avoid expensive connection checks.
+        connectionManager.setValidateAfterInactivity(
+                Ints.checkedCast(Duration.ofSeconds(25).toMillis()));
```
carterkozak, line 128: I snuck in a behavior change here, I can pull this into a separate PR if you prefer. I dropped it here because this PR adds direct access to the PoolingHttpClientConnectionManager, which is the only to configure `setValidateAfterInactivity`

dialogue-apache-hc4-client/src/main/java/com/palantir/dialogue/hc4/ApacheHttpClientChannels.java
```diff
@@ -105,12 +137,10 @@ public static CloseableClient createCloseableHttpClient(ClientConfiguration conf
                         .setRedirectsEnabled(false)
                         .setRelativeRedirectsAllowed(false)
                         .build())
-                .setDefaultSocketConfig(
-                        SocketConfig.custom().setSoKeepAlive(true).build())
+                .setDefaultSocketConfig(socketConfig)
                 .evictIdleConnections(55, TimeUnit.SECONDS)
-                .setMaxConnPerRoute(Integer.MAX_VALUE)
-                .setMaxConnTotal(Integer.MAX_VALUE)
-                // TODO(ckozak): proxy credentials
```
carterkozak, line 113: removed this stale TODO, it's TODONE

dialogue-apache-hc4-client/src/main/metrics/dialogue-apache-client-metrics.yml
```diff
@@ -0,0 +1,11 @@
+options:
+  javaPackage: com.palantir.dialogue.hc4
+  javaVisibility: packagePrivate
+namespaces:
+  dialogue.client.pool:
```
carterkozak, line 5: The metric name doesn't reference the apache client, I tried to keep it as general as possible so we could implement it for other clients in the future if applicable. Thoughts?

ferozco:
👍

dialogue-apache-hc4-client/src/main/metrics/dialogue-apache-client-metrics.yml
ferozco: I like the idea of having consistent metric regardless of underlying client implementation. We can update the docs if we decide to work on another client

PR #499, opened 2020-03-06
Properly support optional empty response

ferozco:
## Before this PR
We would throw if a server responded with a 204

## After this PR
==COMMIT_MSG==
Handle 204s when the expected response type is an optional
==COMMIT_MSG==

We do not properly support cases where a server responds with a 204 but the response type is an alias or any other collection. In practice, it does not seem that any server follows the recommendation of the conjure spec

### DIFF
--- a/dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ConjureBodySerDe.java
+++ b/dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ConjureBodySerDe.java
@@ -154,6 +154,8 @@ final class ConjureBodySerDe implements BodySerDe {
 
         private final ImmutableList<EncodingDeserializerContainer<T>> encodings;
         private final ErrorDecoder errorDecoder;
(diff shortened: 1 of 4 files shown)


dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ConjureBodySerDe.java
```diff
@@ -171,6 +175,8 @@ public T deserialize(Response response) {
             try {
                 if (errorDecoder.isError(response)) {
                     throw errorDecoder.decode(response);
+                } else if (isOptionalType && response.code() == 204) {
+                    return TypeMarkers.getEmptyOptional(token);
```
carterkozak, line 179: should we just use the 204 check, then throw if `isOptionalType` is false?

carterkozak (approved):
thanks!

ferozco:
the bots are too fast

carterkozak:
ha, first time in a while!

PR #500, opened 2020-03-06
improve error on supported 204 type

ferozco:
Follow up to #499


dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ConjureBodySerDe.java
```diff
@@ -175,7 +175,11 @@ public T deserialize(Response response) {
             try {
                 if (errorDecoder.isError(response)) {
                     throw errorDecoder.decode(response);
-                } else if (isOptionalType && response.code() == 204) {
+                } else if (response.code() == 204) {
+                    Preconditions.checkState(
+                            isOptionalType,
+                            "Unable to deserialize non-optional response type from 204",
+                            SafeArg.of("type", token));
```
carterkozak: nit: invert to an `if` expression to avoid SafeArg allocation in the common case.

carterkozak (approved):
thanks!

PR #501, opened 2020-03-06
Dialogue supports optional binary responses

carterkozak:
==COMMIT_MSG==
Dialogue supports optional binary responses using a new method on ConjureBodySerDe which must be used by generated code.
==COMMIT_MSG==

### DIFF
--- a/dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/BinaryEncoding.java
+++ b/dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/BinaryEncoding.java
@@ -17,9 +17,10 @@
 package com.palantir.conjure.java.dialogue.serde;
 
 import com.palantir.dialogue.TypeMarker;
-import com.palantir.logsafe.Preconditions;
 import com.palantir.logsafe.SafeArg;
+import com.palantir.logsafe.exceptions.SafeIllegalStateException;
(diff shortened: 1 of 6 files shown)


ferozco:
How do you disambiguate between an absent inputstream and an empty inputstream?

carterkozak:
204 response code is an absent stream, 200 with zero bytes is an empty stream.

carterkozak:
We have validation to disallow optional binary request bodies because they can't differentiate between the two :-)

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/BinaryEncoding.java
```diff
@@ -41,11 +43,13 @@
     @Override
     @SuppressWarnings("unchecked")
     public <T> Deserializer<T> deserializer(TypeMarker<T> type) {
-        Preconditions.checkArgument(
-                InputStream.class.equals(type.getType()),
-                "BinaryEncoding only supports InputStream",
-                SafeArg.of("requested", type));
-        return input -> (T) input;
+        if (MARKER.equals(type)) {
+            return input -> (T) input;
```
ferozco, line 47:
```suggestion
            return Function.identity();
```

ferozco (approved):

PR #506, opened 2020-03-06
Setup verifier

ferozco:
We actually are more spec compliant than CJR. The big thing that hits us is our handling of aliases of collections


dialogue-client-test-lib/build.gradle
```diff
@@ -1,3 +1,7 @@
+versionsLock {
+    testProject()
```
carterkozak, line 2: thanks!

carterkozak (approved):

PR #507, opened 2020-03-06
Upgrade the generator to fix binary alias handling

carterkozak:
==COMMIT_MSG==
Upgrade the generator to fix binary alias handling
==COMMIT_MSG==


ferozco:
Nice! 👍

PR #511, opened 2020-03-09
Pin until error only switches pin on error

ferozco:
## Before this PR
Pin until error would actually switch the pinned channel if it was limited.

## After this PR
==COMMIT_MSG==
Pin until error only switches the pinned channel if to an error response is received. Requests that are made while the the pinned channel is limited will be redirected to the first available channel.
==COMMIT_MSG==

## Possible downsides?
Its unclear whether the previous behaviour is/was every actually desirable. In the previous model spikes in request volume could trigger concurrency limits, pinning to the next available channel. Depending on the impact host locality has on the outgoing requests, this switch could result in a loss of performance. 

Interestingly, switching the pin could also have a cascade effect since the newly selected channel may have few available limits (due to being previously unused) and therefore quickly result in switching the pin again because of a limited request. In effect, you would round robin through the available channels "warming" the concurrency limits until you reached a point where there sufficient available limits on a given channel to handle request volume. 

Introducing periodic errors into the mix makes things even more complex. The errors would result both in the pin changed and the number of limits being multiplicatively decreased which in turn would cause you to lose short term and medium term locality since you'd be immediately switching hosts and when you eventually did circle back around to the previously pinned channel you would be more likely to hit the concurrency limit again and switch pins again.

---

Think about this a bit more, the rerouting (load shedding) behaviour introduced in this PR starts to make blacklisting more desirable since it could help you make a better decision about where to reroute a request to. It may not be necessary, but something to think about...


dialogue-core/src/main/java/com/palantir/dialogue/core/PinUntilErrorChannel.java
```diff
@@ -125,21 +125,22 @@ LimitedChannel getCurrentChannel() {
...
-    private Optional<ListenableFuture<Response>> executeInternal(Endpoint endpoint, Request request, int depth) {
-        int currentIndex = currentHost.get();
+    private Optional<ListenableFuture<Response>> executeInternal(
+            Endpoint endpoint, Request request, int currentIndex, int depth) {
         LimitedChannel channel = nodeList.get(currentIndex);
 
         Optional<ListenableFuture<Response>> maybeFuture = channel.maybeExecute(endpoint, request);
+        // Simply reroute the request to the fist available channel rather than waiting for the pinned channel to
+        // become available. We attempt to route the request at most nodeList.size() times, but due to reshuffling we
+        // are not guaranteed to attempt every distinct channel.
```
carterkozak, line 138: In this case shouldn’t we fail and let the retryer do it’s thing?

ferozco: Further simplifying the logic and simply returning the pinned channels response is totally an option. I find it difficult to determine which approach would be better, since doing so would require making certain assumptions about the retry behaviour and encode a trade off between locality and latency which may not be universally true.

As mentioned on slack, I merely opened this PR to present an alternative and start a discussion about how node selection and limited channels should behave

carterkozak: As written I’m worried the proposed chang would further complicate the algorithm because it differentiates changing the pinned channel from attempting another channel without changing the pin. This could result in undesirable behavior for subsequent calls.
I think we need to decide whether or not a limited result is considered a failure in the context a pin until error. If so, the existing implementation is reasonable, but if not we should return the limited (but not failed) result. In the future we may implement a blocking limit to avoid burning retries unnecessarily and failing when retries aren’t allowed.
Pin until error with overflow is difficult for me to reason about, I’m not sure what the benefits are over rotating the pinned channel.

ferozco:
Closing in favour of #515

PR #513, opened 2020-03-09
Concurrency Limits decrease in response to 429, 503 and 500s

ferozco:
## Before this PR
We would only reduce the number of limits if the server returned an explicit 429.

## After this PR
==COMMIT_MSG==
Reduce the number of available concurrency limits in response to 429's and 500's
==COMMIT_MSG==

I could go either way on whether general 500's should result in reduced concurrency limits.


carterkozak:
```diff
- Reduce the number of available concurrency limits in response to 429's and 500's
+ Reduce the number of available concurrency limits in response to 429's and 5xx's
```

carterkozak (approved):
Thanks!

PR #514, opened 2020-03-09
fix #510: Use random selection over round robin

### LINKED ISSUE #510: Consider replacing round robin with random selection
## What happened?

The combination of round robin and blacklisting results in hot spotting on nodes following those that are blacklisted.
We shuffle the list periodically, but this is a coarse operation which attempts to approximate random selection.

Randomly choose a node, if it is limited, randomly select among the remaining channels instead of iterating in order.
This strategy is stateless and incurs no synchronization penalty.

carterkozak:
## Before this PR
See #510

## After this PR
==COMMIT_MSG==
Use random selection over round robin to avoid hot spotting
==COMMIT_MSG==

## Possible downsides?
This strategy does not match the name in ClientConfiguration, however it will provide better results and the guarantees are identical for consumers with more than one application thread.


dialogue-core/src/main/java/com/palantir/dialogue/core/RandomSelectionChannel.java
```diff
@@ -0,0 +1,89 @@
...
+    private final Random random;
+
+    RandomSelectionChannel(List<LimitedChannel> delegates, Random random) {
+        this.delegates = ImmutableList.copyOf(delegates);
+        this.random = random;
+    }
+
+    @Override
+    public Optional<ListenableFuture<Response>> maybeExecute(Endpoint endpoint, Request request) {
+        if (delegates.isEmpty()) {
```
ferozco: add a precondition that fails to construct the channel if list of delegates is empty instead

dialogue-core/src/main/java/com/palantir/dialogue/core/RandomSelectionChannel.java
```diff
@@ -0,0 +1,89 @@
...
+            }
+            visitedChannels.set(host);
+            --elements;
+        }
+        return Optional.empty();
+    }
+
+    /** Transforms an index in remaining channels into an actual index in the delegates list. */
+    @VisibleForTesting
+    static int toIndex(@Nullable BitSet visited, int index) {
```
ferozco, line 75: I found this rather hard to follow. Perhaps a comment explaining the algo?

dialogue-core/src/main/java/com/palantir/dialogue/core/RandomSelectionChannel.java
carterkozak: good call, this is an artifact from the round robin channel

ferozco:
👍

PR #516, opened 2020-03-09
Ressurect QueuedChannel

carterkozak:
## Before this PR
The retryer acted as  a queue of sorts.

## After this PR
==COMMIT_MSG==
The QueuedChannel is back with the removal of BlacklistingChannel, which could previously invalidate the queue. This means we cannot limit based on endpoints, but must only take the channel into consideration, and queued requests are scheduled when requests are queued and complete.
==COMMIT_MSG==

## Possible downsides?
The queue is complex, it must be taken into account by all new LimitedChannel implementations.


dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
```diff
@@ -0,0 +1,264 @@
...
+    QueuedChannel(LimitedChannel channel) {
+        this(channel, 1_000);
+    }
+
+    @VisibleForTesting
+    QueuedChannel(LimitedChannel delegate, int maxQueueSize) {
+        this.delegate = new NeverThrowLimitedChannel(delegate);
+        // Do _not_ call size on a ConcurrentLinkedDeque. Unlike other collections, size is an O(n) operation.
+        this.queuedCalls = new ProtectedConcurrentLinkedDeque<>();
+        this.maxQueueSize = maxQueueSize;
```
ferozco, line 80: shall we instrument the queue?

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
```diff
@@ -0,0 +1,264 @@
...
+ *     wait until the next scheduling period</li>
+ *     <li>Schedule in a spin loop: this would allow us to schedule without delay, but requires a thread constantly
+ *     doing work, much of which will be wasted</li>
+ * </ul>
+ *
+ * TODO(jellis): record metrics for queue sizes, num requests in flight, time spent in queue, etc.
+ */
+final class QueuedChannel implements LimitedChannel {
+    private static final Logger log = LoggerFactory.getLogger(QueuedChannel.class);
+    private static final Executor DIRECT = MoreExecutors.directExecutor();
```
ferozco, line 62: nit: we don't really need this constant

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
```diff
@@ -0,0 +1,264 @@
...
+                                                    + " ForwardAndSchedule logging",
+                                            SafeArg.of("service", endpoint.serviceName()),
+                                            SafeArg.of("endpoint", endpoint.endpointName()));
+                                }
+                            }
+                        },
+                        DIRECT);
+                return true;
+            } else {
+                if (!queuedCalls.offerFirst(queueHead)) {
```
ferozco, line 190: could we `addFirst` instead?

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
```diff
@@ -0,0 +1,264 @@
...
+            Endpoint endpoint = queueHead.endpoint();
+            Optional<ListenableFuture<Response>> maybeResponse = delegate.maybeExecute(endpoint, queueHead.request());
+
+            if (maybeResponse.isPresent()) {
+                queueSizeEstimate.decrementAndGet();
+                ListenableFuture<Response> response = maybeResponse.get();
+                queueHead.span().complete();
+                numRunningRequests.incrementAndGet();
+                response.addListener(numRunningRequests::decrementAndGet, DIRECT);
+                Futures.addCallback(response, new ForwardAndSchedule(queuedResponse), DIRECT);
```
ferozco, line 171:
```suggestion
                DialogueFutures.addDirectCallback(response, new ForwardAndSchedule(queuedResponse));
```

dialogue-core/src/test/java/com/palantir/dialogue/core/QueuedChannelTest.java
```diff
@@ -0,0 +1,250 @@
...
+        mockHasCapacity();
+        queuedChannel.maybeExecute(endpoint, request);
+        verify(delegate, times(4)).maybeExecute(endpoint, request);
+    }
+
+    @Test
+    public void testQueuedRequestExecutedOnNextSubmission_throws() throws ExecutionException, InterruptedException {
+        // First request is limited by the channel and queued
+        Request queuedRequest = Mockito.mock(Request.class);
+        when(delegate.maybeExecute(endpoint, queuedRequest)).thenReturn(Optional.empty());
```
ferozco, line 115:
```suggestion
        mockNoCapacity();
```

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
carterkozak: This reintroduces the old queue, I’d prefer to make improvements separately

ferozco: Oh it didn't seem like this was just a raw revert

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
carterkozak: addFirst can throw on failure, I’d prefer not to introduce a scenario that forces us to throw and catch. The current implementation is expected not to throw because we don’t use a queue with a maximum size, but I’d like to keep the queue safe to allow future changes (around removing elements upon cancelation) without risking reintroducing bugs.

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
carterkozak: The only diff is that this doesn’t include the two broken gauges

ferozco:
👍

PR #517, opened 2020-03-09
Implement a non-configurable ten minute timeout

carterkozak:
==COMMIT_MSG==
Implement a non-configurable ten minute timeout
==COMMIT_MSG==

Should this be configurable?

### DIFF
--- a/changelog/@unreleased/pr-517.v2.yml
+++ b/changelog/@unreleased/pr-517.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Implement a non-configurable ten minute timeout
+  links:
+  - https://github.com/palantir/dialogue/pull/517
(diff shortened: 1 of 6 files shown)


ferozco:
@ellisjoe do you have any thoughts on introducing such a time out? I'm hesitant since there could be services out in the wild where this value is too low

markelliot:
Probably risky to add this unless CJR already has an enforcement. AFAIU the CJR timeouts affect idle time but not request run-time, and it's just fine to stream bytes for >10mins if in fact the connection is active.

carterkozak:
Happy to punt on this for now. Long term I think a time-based queue limit will scale better than a size-based limit, though ideal may lie in the middle.

markelliot:
This appears to limit request duration, not just time-to-schedule. I'd be favor in never starting requests past a fixed deadline, but think more work needs to be put in to understanding durations of active requests if the limit is going to include full request duration.

ellisjoe:
Yeah, the other thing that always comes up with deadlines here is whether or not requests should be retried when the deadline is hit. We don't currently have support server side to shutdown all the work associated with a request when it disconnects, so you can end up pegging upstream services with duplicative work that's never expected to complete within the deadline.

PR #518, opened 2020-03-09
Upgrade to fixed version of apache httpclient, resume auto-upgrades

carterkozak:
## Before this PR
Pinned to an older release due to https://issues.apache.org/jira/browse/HTTPCLIENT-2047

## After this PR
==COMMIT_MSG==
Upgrade to fixed version of apache httpclient, resume auto-upgrades
==COMMIT_MSG==


ferozco:
👍

PR #522, opened 2020-03-09
ConcurrencyLimitedChannel has no maximum limit value

carterkozak:
The client and server may negotiate a reasonable value based on
load.
Limits apply per channel (host) rather than per-endpoint, so
200 was incredibly restrictive.

## Before this PR
Limit of 200 requests per host.

## After this PR
==COMMIT_MSG==
ConcurrencyLimitedChannel has no maximum limit value
==COMMIT_MSG==

## Possible downsides?
Possible to grow load when servers fail to respond with 429 or 5xx. Our primary server does not allow requests to queue, responding 503 when incoming requests cannot be processed immediately. It's possible although unlikely and already suboptimal that some systems do not behave in this way.

### DIFF
--- a/changelog/@unreleased/pr-522.v2.yml
+++ b/changelog/@unreleased/pr-522.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: ConcurrencyLimitedChannel has no maximum limit value
+  links:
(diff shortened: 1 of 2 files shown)


ferozco:
👍

carterkozak:
:robot:

PR #524, opened 2020-03-10
Instrument channels with channel name

ferozco:
## Before this PR
We would produce metrics where the `service-name` tag was endpoints service name which corresponds the the service interface name. This could a bit confusing since multiple logical services (`FooService`, `BarService`) could be implemented by a single server and single channel is re-used for all clients to a given serer.

## After this PR
We add a required `channelName` field and tag all output metrics with it.


dialogue-core/src/main/metrics/dialogue-core-metrics.yml
```diff
@@ -16,6 +16,10 @@ namespaces:
           Number of requests that are actively running. The `stage` may refer to `running` requests actively
           executing over the wire or `processing` which may be awaiting a client or backing off for a retry.
           Note that running requests are also counted as processing.
+      requests.queued:
+        type: gauge
```
carterkozak: Can we use a counter, and wrap both the atomic int + counter with a simple abstraction to expose `incrementAndGet` `decrementAndGet` and `get`? This way we don't have to worry about gauge weirdness.

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
```diff
@@ -59,25 +56,27 @@
  */
 final class QueuedChannel implements LimitedChannel {
     private static final Logger log = LoggerFactory.getLogger(QueuedChannel.class);
-    private static final Executor DIRECT = MoreExecutors.directExecutor();
 
     private final Deque<DeferredCall> queuedCalls;
     private final LimitedChannel delegate;
+    private final String serviceName;
     // Tracks requests that are current executing in delegate and are not tracked in queuedCalls
     private final AtomicInteger numRunningRequests = new AtomicInteger(0);
```
carterkozak: hmm, I don't think this is actually used, our ActiveRequestInstrumentationChannel provides the same data.

dialogue-core/src/main/java/com/palantir/dialogue/core/DeprecationWarningChannel.java
```diff
@@ -56,7 +57,7 @@
                 delegate.execute(endpoint, request), DialogueFutures.onSuccess(response -> {
                     if (response != null) {
                         response.getFirstHeader("deprecation").ifPresent(deprecated -> {
-                            metrics.deprecations(endpoint.serviceName()).mark();
```
carterkozak: I think `endpoint.serviceName()` makes the most sense here since deprecation is defined in conjure with the endpoint itself.

dialogue-core/src/main/java/com/palantir/dialogue/core/ActiveRequestInstrumentationChannel.java
```diff
@@ -29,27 +28,22 @@
 
     private final String stage;
     private final Channel delegate;
-    private final DialogueClientMetrics metrics;
+    private final Counter active;
 
     ActiveRequestInstrumentationChannel(
-            Channel delegate, @CompileTimeConstant String stage, DialogueClientMetrics metrics) {
+            Channel delegate, String serviceName, @CompileTimeConstant String stage, DialogueClientMetrics metrics) {
```
carterkozak: There's an argument for both the endpoint service name and target name depending on the use case. We could do both, and come back to remove the one that we find less helpful.

dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueFutures.java
```diff
@@ -33,6 +33,12 @@ private DialogueFutures() {}
         return future;
     }
 
+    @CanIgnoreReturnValue
+    static <T> ListenableFuture<T> addDirectListener(ListenableFuture<T> future, Runnable listener) {
+        future.addListener(listener, MoreExecutors.directExecutor());
+        return future;
```
carterkozak, line 39: :+1:

dialogue-core/src/main/metrics/dialogue-core-metrics.yml
ferozco: any reason to not just use the counter as a source of truth instead of atomic int?

ferozco:
Need to fix up the benchmarks because they relied on metrics being tagged in a particular way

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
carterkozak: I think `numRunningRequests` can be deleted now. I don't think it's ever read.

carterkozak:
:+1:

PR #531, opened 2020-03-13
Tag connection pool metrics by channel name

carterkozak:
==COMMIT_MSG==
Tag connection pool metrics by channel name
==COMMIT_MSG==


dialogue-apache-hc4-client/src/main/java/com/palantir/dialogue/hc4/ApacheHttpClientChannels.java
```diff
@@ -92,7 +93,17 @@ public static Channel createSingleUri(String uri, CloseableClient client) {
         return BlockingChannelAdapter.of(new ApacheHttpClientBlockingChannel(client.client, url(uri)));
     }
 
+    /**
+     * Prefer {@link #createCloseableHttpClient(ClientConfiguration, String)}.
+     *
+     * @deprecated Use the overload with a channel name.
+     */
+    @Deprecated
     public static CloseableClient createCloseableHttpClient(ClientConfiguration conf) {
```
iamdanfox, line 104: 😢my nice clean signatures.

What if as an alternative, the ClientConfiguration passed in by WC contained a TaggedMetricRegistry which happened to add the channel-name tag to everything (akin to the VersionedTaggedMetricRegistry)?

carterkozak: Our list of all connection pools would be added to every channel name, right?

dialogue-apache-hc4-client/src/main/java/com/palantir/dialogue/hc4/ApacheClientGauges.java
```diff
@@ -35,28 +43,60 @@
 @SuppressWarnings("UnnecessaryLambda")
 final class ApacheClientGauges {
 
-    private static final List<WeakReference<PoolingHttpClientConnectionManager>> pools = new CopyOnWriteArrayList<>();
```
iamdanfox, line 38: maybe I don't fully understand the context about why we had a static field here in the first place, but what would happen if we used the same `weakGauge` thing we just added in ConcurrencyLimitedChannel? https://github.com/palantir/dialogue/blob/7dc6e7cfbc508231220ab3b2f108445ac690c24b/dialogue-core/src/main/java/com/palantir/dialogue/core/ConcurrencyLimitedChannel.java#L153-L174

dialogue-core/src/main/java/com/palantir/dialogue/core/VersionedTaggedMetricRegistry.java
```diff
@@ -114,9 +114,9 @@ public Counter counter(MetricName metricName, Supplier<Counter> counterSupplier)
     }
 
     @Override
-    public void addMetrics(String _safeTagName, String _safeTagValue, TaggedMetricSet _metrics) {
-        throw new UnsupportedOperationException(
-                "Operations involving transforming metricsets are not implemented as we don't need them in dialogue");
+    public void addMetrics(String safeTagName, String safeTagValue, TaggedMetricSet metrics) {
+        // Passed through without adding version tags
```
iamdanfox: is this just to minimize the LoC?

dialogue-apache-hc4-client/src/main/java/com/palantir/dialogue/hc4/ApacheClientGauges.java
carterkozak: This utility is built to capture the sum of values when we roll over. I think those statistics are important for tracking connection pools, where we don't want to replace the existing value outright.

dialogue-core/src/main/java/com/palantir/dialogue/core/VersionedTaggedMetricRegistry.java
carterkozak: It didn't seem as important for this metric and the change is already a tad complex, I can add versions here if you like.

dialogue-apache-hc4-client/src/main/java/com/palantir/dialogue/hc4/ApacheClientGauges.java
iamdanfox: got it makes sense - I think I hadn't fully understood the purpose of this class initially, but it makes sense now.

dialogue-apache-hc4-client/src/main/java/com/palantir/dialogue/hc4/ApacheClientGauges.java
```diff
@@ -35,28 +43,60 @@
...
+    private static final TaggedMetricSet metricSet = new TaggedMetricSet() {
+        @Override
+        public Map<MetricName, Metric> getMetrics() {
+            ImmutableMap.Builder<MetricName, Metric> results = ImmutableMap.builder();
+            forEachMetric(results::put);
+            return results.build();
+        }
+
+        @Override
+        public void forEachMetric(BiConsumer<MetricName, Metric> consumer) {
```
iamdanfox: does the specialized implementation of this really buy us anything? esp since we're already adding them all into a `new DefaultTaggedMetricRegistry()`, seems like we could just move this code into `getMetrics()` and then rely on the default impl?

carterkozak: Adding the version when metrics are created allows us to do the work once when the metric is created/accessed as opposed to each time metrics are loaded, in theory. However quite a few places create metric names with service/endpoint information which moves that work to each request... so I think you're right, it's best if we tag on access.
Thoughts on making that change in this PR vs an isolated change?

dialogue-apache-hc4-client/src/main/java/com/palantir/dialogue/hc4/ApacheClientGauges.java
```diff
@@ -35,28 +43,60 @@
...
         pools.add(new WeakReference<>(connectionPool));
         pools.removeIf(item -> item.get() == null);
     }
 
     static void install(TaggedMetricRegistry registry) {
-        DialogueClientPoolMetrics poolMetrics = DialogueClientPoolMetrics.of(registry);
-        poolMetrics.size().state("idle").build(idle);
-        poolMetrics.size().state("leased").build(leased);
-        poolMetrics.size().state("pending").build(pending);
+        registry.addMetrics("set", "apacheClientGauges", metricSet);
```
iamdanfox: any concerns if I just PR tritium to let us just `addMetrics(TaggedMetricSet)` without having to specify the extra tag names?

carterkozak: I think metric sets are identified by a tag, I suppose a concurrent set would be fine if the original is tracked for removal. However it becomes easy to foot-gun two instances of the same type of metric set in a way that causes duplicate keys, which makes getMetrics throw.

dialogue-apache-hc4-client/src/main/java/com/palantir/dialogue/hc4/ApacheClientGauges.java
```diff
@@ -35,28 +43,60 @@
...
     }
 
     static <T> T collect(
```
iamdanfox: private static ?

dialogue-apache-hc4-client/src/main/java/com/palantir/dialogue/hc4/ApacheClientGauges.java
iamdanfox: I think that exact foot-gun (with the current API) was the cause of @jkozlowski's disappearing metrics PDS ticket. I think it's not significantly more dangerous tbh, and I kinda like the ability to compose things without having to add throwaway strings into the mix

carterkozak: I don't have a strong opinion either way, defer to dave

iamdanfox:
So it seems to me like the majority of the complexity comes from the book-keeping to keep track of the careful additions when we're draining an old connection pool and starting to ramp up a new one.  Can I push up an option that encapsulates this into a mutable class? I'd like to avoid the statics if possible!

carterkozak:
Sure, let’s give it a shot

dialogue-apache-hc4-client/src/main/java/com/palantir/dialogue/hc4/ApacheHttpClientChannels.java
iamdanfox: Ok so now that we don't have the statics, what do you think of this WC-controlled channel-name tag idea? Then we wouldn't even need to change the signature?

carterkozak: That’s probably for the best, though consumers outside of witchcraft will need to know to do that as well, which isn’t likely to happen. That case is likely one-channel-per-application and fine without the extra data.

dialogue-core/src/main/java/com/palantir/dialogue/core/VersionedTaggedMetricRegistry.java
iamdanfox: should be fine to revert this now right?

iamdanfox (approved):

dialogue-core/src/main/java/com/palantir/dialogue/core/VersionedTaggedMetricRegistry.java
carterkozak: Yep, will do

dialogue-apache-hc4-client/src/main/java/com/palantir/dialogue/hc4/ApacheHttpClientChannels.java
```diff
@@ -80,9 +82,10 @@
     private ApacheHttpClientChannels() {}
 
     public static Channel create(ClientConfiguration conf) {
```
iamdanfox, line 84: thoughts on deprecating/deleting this one at the same time? it leaks resources by never closing the CloseableClient, and we don't use it from WC...

iamdanfox:
Looks like my approval isn't counted as I've ocntributed some commits, but if you also `+1` I think it'll go.

dialogue-apache-hc4-client/src/main/java/com/palantir/dialogue/hc4/ApacheHttpClientChannels.java
carterkozak: I think this should move to a test utility rather than main sources, in favor of deprecating this method but I'd prefer to do it in isolation -- looks like ~7 uses in test code that will need to be either updated or suppress the deprecation.

carterkozak:
👍 thanks @iamdanfox

PR #534, opened 2020-03-16
Record metrics describing time spent queued

carterkozak:
This fixes a bug where tracing spans were not recorded when queued
requests were canceled.

## Before this PR
Only tracing provided insight into the time requests spent queued.

## After this PR
==COMMIT_MSG==
Record metrics describing time spent queued
==COMMIT_MSG==

## Possible downsides?
more complexity surrounding the queue.

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
@@ -17,6 +17,7 @@
 package com.palantir.dialogue.core;
 
 import com.codahale.metrics.Counter;
+import com.codahale.metrics.Timer;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.util.concurrent.FutureCallback;
 import com.google.common.util.concurrent.ListenableFuture;
@@ -32,6 +33,7 @@ import com.palantir.tracing.DetachedSpan;
 import java.util.Deque;
 import java.util.Optional;
 import java.util.concurrent.ConcurrentLinkedDeque;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 import org.immutables.value.Value;
 import org.slf4j.Logger;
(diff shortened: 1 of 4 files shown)


dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
```diff
@@ -92,6 +96,8 @@
             if (maybeResult.isPresent()) {
                 ListenableFuture<Response> result = maybeResult.get();
                 DialogueFutures.addDirectListener(result, this::onCompletion);
+                // While the queue was avoid, this is equivalent to spending zero time on the queue.
+                queuedTime.update(0, TimeUnit.NANOSECONDS);
```
carterkozak, line 98: I could go either way on this, but I opted toward updating with zero time to avoid biasing our statistics toward problematic regions. This code path is an optimization to avoid unnecessary work in the queue unless requests are already queued.

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
```diff
@@ -65,6 +67,7 @@
     private final AtomicInteger queueSizeEstimate = new AtomicInteger(0);
     private final int maxQueueSize;
     private Counter queueSizeCounter;
+    private Timer queuedTime;
```
iamdanfox: why would this not be final?

carterkozak: doh, this and the counter should be final

iamdanfox (approved):
LGTM, can probably also amend the TODO:
```
TODO(jellis): record metrics for queue sizes, num requests in flight, time spent in queue, etc.
```

ferozco:
👍

carterkozak:
good call, I've removed the todo because we've implemented the relevant queue metrics, in-flight is out of scope for the queue and already recorded elsewhere.

PR #535, opened 2020-03-16
[stacked] Keep track of apache connection pool metrics using WeakSummingGauge

iamdanfox:
## Before this PR

We're doing some careful book-keeping to try and make sure that outgoing connection pools are still represented in our metrics when we live-reload and create new ones, but it involved some central static static, which kinda felt odd in such a nice clean dialogue world.

## After this PR
==COMMIT_MSG==
a mutable `WeakSummingGauge` class allows us to keep track of however many connection pools are still being actively used, so that the metrics don't miss anything. No statics.
==COMMIT_MSG==

Stacked on top of https://github.com/palantir/dialogue/pull/531


dialogue-apache-hc4-client/src/main/java/com/palantir/dialogue/hc4/ApacheHttpClientChannels.java
```diff
@@ -178,6 +178,39 @@ public static CloseableClient createCloseableHttpClient(ClientConfiguration conf
...
+    private static void setupConnectionPoolMetrics(
+            TaggedMetricRegistry taggedMetrics,
+            String channelName,
+            PoolingHttpClientConnectionManager connectionManager) {
+        WeakSummingGauge.getOrCreate(
+                pool -> pool.getTotalStats().getAvailable(),
+                connectionManager,
+                taggedMetrics,
+                MetricName.builder()
+                        .safeName("dialogue.client.pool.size")
```
carterkozak, line 190: shame we can't use metric-schema here, this is one of the most important sets of metrics (for me anyhow, service authors shouldn't need to worry about these unless something has gone south)

dialogue-apache-hc4-client/src/main/java/com/palantir/dialogue/hc4/WeakSummingGauge.java
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    /** Register a new source element which will be used to compute the future summary integer. */
+    public void add(T sourceElement) {
+        weakSet.add(sourceElement);
+    }
+
+    @Override
+    public Integer getValue() {
+        return weakSet.stream().mapToInt(gaugeFunction).sum();
```
carterkozak, line 52: This isn't threadsafe: https://docs.oracle.com/javase/8/docs/api/java/util/Collections.html#synchronizedSet-java.util.Set-

dialogue-apache-hc4-client/src/main/java/com/palantir/dialogue/hc4/ApacheHttpClientChannels.java
iamdanfox: I think if metric schema exposed just the MetricName we could use it? Also it's still present in the metrics.yml file, so at least it should show up in the witchcraft generated metrics.md

carterkozak (approved):

carterkozak:
Thanks @iamdanfox!

PR #542, opened 2020-03-19
Instrument client TLS handshakes

carterkozak:
==COMMIT_MSG==
Instrument client TLS handshakes
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-542.v2.yml
+++ b/changelog/@unreleased/pr-542.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Instrument client TLS handshakes
+  links:
+  - https://github.com/palantir/dialogue/pull/542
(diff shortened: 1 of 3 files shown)


carterkozak:
@ferozco I can't seem to add you as a reviewer on this, not sure what's going on!

iamdanfox:
I think all the graphs we have at the moment assume that these connections are only instrumented on the server-side - can we differentiate the client & server side in the metric names somehow?

Also if these just say `tls.handshake` then in our birds-eye graphs we might end up double-counting requests that come from dialogue clients?

carterkozak:
The server produces metrics tagged `context: server`, CJR already provides these from the client side:
https://github.com/palantir/conjure-java-runtime/pull/1446

iamdanfox (approved):
Ah great, ok I see we'd get a `context: gatekeeperservice` tag in DD.

PR #545, opened 2020-03-19
Implement a LeakDetectingChannel to notify when responses are leaked

carterkozak:
## Before this PR
resources probably aren't leaked, but we don't have validation.

## After this PR
==COMMIT_MSG==
Log warnings and record metrics when resource leaks are discovered.
==COMMIT_MSG==

## Possible downsides?
Implementing a finalize adds objects to the finalizer queue which runs on a single thread. It shouldn't be problematic and the cost is effectively zero compared to the cost of a resource leak.

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/LeakDetectingChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/LeakDetectingChannel.java
@@ -140,6 +140,7 @@ final class LeakDetectingChannel implements Channel {
 
         private final Response delegate;
         private final LeakDetector leakDetector;
+        @Nullable
         private InputStream leakDetectingStream;
 
         LeakDetectingResponse(Response delegate, LeakDetector leakDetector) {
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/LeakDetectingChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/LeakDetectingChannel.java
@@ -1,0 +1,167 @@
+/*
+ * (c) Copyright 2020 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.dialogue.core;
+
+import com.google.common.collect.ListMultimap;
+import com.google.common.util.concurrent.Futures;
+import com.google.common.util.concurrent.ListenableFuture;
+import com.google.common.util.concurrent.MoreExecutors;
+import com.palantir.dialogue.Channel;
+import com.palantir.dialogue.Endpoint;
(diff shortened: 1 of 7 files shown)


dialogue-core/src/main/java/com/palantir/dialogue/core/LeakDetectingChannel.java
```diff
@@ -0,0 +1,187 @@
...
+            this.creationTrace = log.isTraceEnabled() ? new SafeRuntimeException("created here") : null;
+        }
+
+        void disarm() {
+            armed = false;
+        }
+
+        @Override
+        @SuppressWarnings("NoFinalizer")
+        protected void finalize() throws Throwable {
```
iamdanfox, line 85: Seems like the finalize method is deprecated in java9 with a recommendation to use `java.lang.ref.Cleaner` or PhantomReference instead - https://bugs.openjdk.java.net/browse/JDK-8165641... seems like it should just be a matter of calling cleaner.register with a nice runnable? https://docs.oracle.com/javase/9/docs/api/java/lang/ref/Cleaner.html

dialogue-core/src/main/java/com/palantir/dialogue/core/LeakDetectingChannel.java
```diff
@@ -0,0 +1,187 @@
...
+
+        LeakDetectingResponse(Response delegate, LeakDetector leakDetector) {
+            this.delegate = delegate;
+            this.leakDetector = leakDetector;
+        }
+
+        @Override
+        public InputStream body() {
+            if (leakDetectingStream == null) {
+                leakDetectingStream = new LeakDetectingInputStream(delegate.body(), leakDetector);
```
iamdanfox, line 156: why the lazy initialization here?

dialogue-core/src/main/java/com/palantir/dialogue/core/LeakDetectingChannel.java
carterkozak: We can’t do that with jdk8 though.

iamdanfox: Ah dang ok

iamdanfox:
I really like the idea of being able to look at graphs in prod and be really confident that there are zero leaks! Kinda sad we have to implement deprecated APIs (I guess if we really cared we could do some kind of crazy service-loading from a multi-release jar to switch between Cleaner & finalize method).  Reckon this is probably good to go as is though ;)

dialogue-core/src/main/java/com/palantir/dialogue/core/LeakDetectingChannel.java
iamdanfox: If this method ever manages to block then do we end up preventing all garbage collection from succeeding?

iamdanfox (approved):

dialogue-core/src/main/java/com/palantir/dialogue/core/LeakDetectingChannel.java
carterkozak: In the case delegate.body throws, I figure this will produce a better stack trace than failing when the response is wrapped.
It’s probably fine either way though.

dialogue-core/src/main/java/com/palantir/dialogue/core/LeakDetectingChannel.java
carterkozak: It can block the finalizer queue from progressing, but that should only occur if we have leaked a resource. I’d prefer to have the data to resolve the situation in that scenario though.

iamdanfox: yep I think that makes sense!

PR #548, opened 2020-03-21
BinaryRequestBody: fix confusing variable names/comments

iamdanfox:
## Before this PR

I was implementing BinaryRequestBody and got confused by the argument name & javadoc.

## After this PR
==COMMIT_MSG==
BinaryRequestBody javadoc mentions 'request' not 'response'.
==COMMIT_MSG==

### DIFF
--- a/dialogue-target/src/main/java/com/palantir/dialogue/BinaryRequestBody.java
+++ b/dialogue-target/src/main/java/com/palantir/dialogue/BinaryRequestBody.java
@@ -26,7 +26,7 @@ import java.io.OutputStream;
 /** Streamed binary response data with Content-Type <code>application/octet-stream</code>. */
 public interface BinaryRequestBody extends Closeable {
 
-    /** Invoked to write data to the request stream. Called exactly once. */
+    /** Invoked to write data to the request stream. */
     void write(OutputStream requestBody) throws IOException;
 
     /** This method may be overridden to return resources. */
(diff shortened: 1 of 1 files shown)


dialogue-target/src/main/java/com/palantir/dialogue/BinaryRequestBody.java
```diff
@@ -23,15 +23,11 @@
 import java.io.InputStream;
 import java.io.OutputStream;
 
-/**
- * Streamed binary response data with Content-Type <code>application/octet-stream</code>.
- */
+/** Streamed binary response data with Content-Type <code>application/octet-stream</code>. */
```
carterkozak:
```suggestion
/** Streamed binary request data with Content-Type <code>application/octet-stream</code>. */
```

dialogue-target/src/main/java/com/palantir/dialogue/BinaryRequestBody.java
```diff
@@ -23,15 +23,11 @@
...
 public interface BinaryRequestBody extends Closeable {
 
-    /**
-     * Invoked to write data to the response stream. Called exactly once.
-     */
-    void write(OutputStream responseBody) throws IOException;
+    /** Invoked to write data to the request stream. Called exactly once. */
```
carterkozak:
```suggestion
    /** Invoked to write data to the request stream. */
```

If we're limited and no space exists on the queue, or the request is cancelled, this may be called at all, but that should be okay since this is closeable, and we do close the BinaryRequestBody in those scenarios.

iamdanfox:
:+1:

carterkozak (approved):

ferozco:
👍

PR #553, opened 2020-03-23
POC for sticky sessioning

j-baker:
Will discuss internally, but we have a lot of garbage at present related
to handling load balancing - we need a 'session' of requests to an
underlying channel.

Wondered if this might be cleaner?

Adds a 'tag' concept to requests and responses, where channels can add
and remove data as relevant (although the present implementation of
'getTag()'.

Then, uses the tag concept to implement a limited channel which can be
used to implement sticky sessioning.

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/StickySessioning.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/StickySessioning.java
@@ -1,0 +1,120 @@
+/*
+ * (c) Copyright 2020 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.dialogue.core;
+
+import com.google.common.collect.ListMultimap;
+import com.google.common.util.concurrent.FluentFuture;
+import com.google.common.util.concurrent.Futures;
+import com.google.common.util.concurrent.ListenableFuture;
+import com.google.common.util.concurrent.MoreExecutors;
+import com.palantir.dialogue.Channel;
+import com.palantir.dialogue.Endpoint;
+import com.palantir.dialogue.Request;
+import com.palantir.dialogue.Response;
+import com.palantir.dialogue.TagKey;
+import java.io.InputStream;
+import java.util.Optional;
+
+public final class StickySessioning {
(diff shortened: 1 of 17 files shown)


dialogue-core/src/main/java/com/palantir/dialogue/core/StickySessioning.java
```diff
@@ -0,0 +1,120 @@
...
+import com.google.common.util.concurrent.MoreExecutors;
+import com.palantir.dialogue.Channel;
+import com.palantir.dialogue.Endpoint;
+import com.palantir.dialogue.Request;
+import com.palantir.dialogue.Response;
+import com.palantir.dialogue.TagKey;
+import java.io.InputStream;
+import java.util.Optional;
+
+public final class StickySessioning {
```
j-baker, line 32: naming is not great...

dialogue-target/src/main/java/com/palantir/dialogue/Response.java
```diff
@@ -46,4 +46,6 @@
      */
     @Override
     void close();
+
+    <T> Optional<T> getTag(TagKey<T> tagKey);
```
j-baker, line 50: this is possibly a non-ideal api, beacuse it can only really be implemented by if/else chains. OTOH, if the if/else chain is more than a couple deep, we have a problem.

dialogue-core/src/main/java/com/palantir/dialogue/core/StickySessioning.java
```diff
@@ -0,0 +1,120 @@
...
+        public Optional<ListenableFuture<Response>> maybeExecute(
+                Endpoint endpoint, Request request) {
+            return request.getTag(EXECUTED_ON).orElse(delegate).maybeExecute(endpoint, request);
+        }
+    }
+
+    private static final class StickySessionChannel implements Channel {
+        private final Channel delegate;
+
+        private FluentFuture<LimitedChannel> stuckChannel;
```
jkozlowski, line 94: I spent 10 minutes just perusing the codebase, and reading through the generated client code too, it seems to me like you give a channel to the generated client and it keeps it.

So I guess we'd have to create this channel at the start of a sticky session, and then from the sticky channel create a generated client. Then we'd get load balancing, but each session would be sticky.

That's how I understand this would be constructed.

jkozlowski: The flow is a bit mysterious for what this is trying to achieve, you let a request execute, then you remember that in the request at a special tag, then you remember that in here.

dialogue-core/src/main/java/com/palantir/dialogue/core/StickySessioning.java
```diff
@@ -0,0 +1,120 @@
...
+
+        private StickySessionChannel(Channel delegate) {
+            this.delegate = delegate;
+        }
+
+        @Override
+        public synchronized ListenableFuture<Response> execute(Endpoint endpoint, Request request) {
```
jkozlowski, line 101: If this is constructed in the way I described, this synchronized is not a problem.

jkozlowski: Would an explicit mapping in the response for "this is the Channel I executed on" be a bit less open?

PR #554, opened 2020-03-24
Response leak detection samples 1% of requests (down from 100%)

carterkozak:
This matches Netty behavior, allowing us to collect signal when
leaks occur without as severe a performance penalty even when
stack traces are collected.

## After this PR
==COMMIT_MSG==
Response leak detection samples 1% of requests (down from 100%)
==COMMIT_MSG==

## Possible downsides?
Only 1% of leaks will be recorded. Goal is to discover when there are leaks while being as non-invasive as possible.

### DIFF
--- a/changelog/@unreleased/pr-554.v2.yml
+++ b/changelog/@unreleased/pr-554.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Response leak detection samples 1% of requests (down from 100%)
+  links:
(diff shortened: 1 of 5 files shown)


carterkozak:
Created as a draft, I don't think we want to merge this until we have more signal/validation from existing consumers.

carterkozak:
lol... simulation is impacted by interactions with the random instance

iamdanfox (approved):

PR #555, opened 2020-03-24
Stop publishing alternative channel impls (in prep for 1.0)

iamdanfox:
## Before this PR

@callumr thought the recommended way of using dialogue was the new java9 HttpClient thing, but we actually discourage this one due to it failing to complete a benchmark (so bugs in the JVM)

## After this PR
==COMMIT_MSG==
`dialogue-apache-hc4-client` is the only published channel implementation, as this is our recommended and supported client.
==COMMIT_MSG==

## Possible downsides?
- when it comes to benchmarking, we may have to copy and paste the other impls.


carterkozak (approved):

iamdanfox:
🍿

PR #556, opened 2020-03-24
Validate ciphers against the provided socket factory

carterkozak:
Previously we only checked against the JVM default context which
isn't representative of all inputs.

==COMMIT_MSG==
Validate ciphers against the provided socket factory, this adds support for Conscrypt.
==COMMIT_MSG==

Downsides:
We don't have real tests for this, TODO to add tests to this repository which use TLS.


iamdanfox:
Good spot!

ferozco:
👍

PR #561, opened 2020-03-25
ApacheHttpClientChannels logs client names with cipher suites

carterkozak:
==COMMIT_MSG==
ApacheHttpClientChannels logs client names with cipher suites
==COMMIT_MSG==


dialogue-apache-hc4-client/src/main/java/com/palantir/dialogue/hc4/ApacheHttpClientChannels.java
```diff
@@ -247,6 +249,7 @@ public String toString() {
         if (!unsupported.isEmpty()) {
             log.info(
```
iamdanfox: were you gonna take this down to debug after callum's suggestion?

carterkozak: I could go either way on this, it's logged each time a client is created, not per-call as he though, but still probably a bit too noisy for info.

iamdanfox (approved):

PR #564, opened 2020-03-26
Responses are no longer leaked when blocking channel futures are interrupted

carterkozak:
## After this PR
==COMMIT_MSG==
Responses are no longer leaked when blocking channel futures are interrupted
==COMMIT_MSG==

### DIFF
--- a/dialogue-apache-hc4-client/src/main/java/com/palantir/dialogue/hc4/ApacheHttpClientBlockingChannel.java
+++ b/dialogue-apache-hc4-client/src/main/java/com/palantir/dialogue/hc4/ApacheHttpClientBlockingChannel.java
@@ -25,6 +25,7 @@ import com.palantir.dialogue.RequestBody;
 import com.palantir.dialogue.Response;
 import com.palantir.dialogue.blocking.BlockingChannel;
 import com.palantir.dialogue.core.BaseUrl;
+import com.palantir.dialogue.core.ResponseLeakDetector;
 import com.palantir.logsafe.Preconditions;
 import com.palantir.logsafe.exceptions.SafeRuntimeException;
 import java.io.IOException;
@@ -48,10 +49,13 @@ final class ApacheHttpClientBlockingChannel implements BlockingChannel {
 
     private final CloseableHttpClient client;
     private final BaseUrl baseUrl;
+    private final ResponseLeakDetector responseLeakDetector;
 
-    ApacheHttpClientBlockingChannel(CloseableHttpClient client, URL baseUrl) {
+    ApacheHttpClientBlockingChannel(
+            CloseableHttpClient client, URL baseUrl, ResponseLeakDetector responseLeakDetector) {
         this.client = client;
         this.baseUrl = BaseUrl.of(baseUrl);
+        this.responseLeakDetector = responseLeakDetector;
     }
 
     @Override
@@ -70,7 +74,7 @@ final class ApacheHttpClientBlockingChannel implements BlockingChannel {
             RequestBody body = request.body().get();
             builder.setEntity(new RequestBodyEntity(body));
         }
-        return new HttpClientResponse(client.execute(builder.build()));
+        return responseLeakDetector.wrap(new HttpClientResponse(client.execute(builder.build())), endpoint);
     }
 
     private static final class HttpClientResponse implements Response {
--- a/dialogue-apache-hc4-client/src/main/java/com/palantir/dialogue/hc4/ApacheHttpClientBlockingChannel.java
+++ b/dialogue-apache-hc4-client/src/main/java/com/palantir/dialogue/hc4/ApacheHttpClientBlockingChannel.java
@@ -73,7 +73,20 @@ final class ApacheHttpClientBlockingChannel implements BlockingChannel {
             RequestBody body = request.body().get();
             builder.setEntity(new RequestBodyEntity(body));
         }
-        return responseLeakDetector.wrap(new HttpClientResponse(client.execute(builder.build())), endpoint);
+        CloseableHttpResponse httpClientResponse = client.execute(builder.build());
+        // Defensively ensure that resources are closed if failures occur within this block,
+        // for example HttpClientResponse allocation may throw an OutOfMemoryError.
+        boolean close = true;
+        try {
+            Response dialogueResponse = new HttpClientResponse(httpClientResponse);
+            Response leakDetectingResponse = responseLeakDetector.wrap(dialogueResponse, endpoint);
+            close = false;
+            return leakDetectingResponse;
+        } finally {
+            if (close) {
+                httpClientResponse.close();
+            }
+        }
     }
 
     private static final class HttpClientResponse implements Response {
--- a/dialogue-apache-hc4-client/src/main/java/com/palantir/dialogue/hc4/ApacheHttpClientBlockingChannel.java
+++ b/dialogue-apache-hc4-client/src/main/java/com/palantir/dialogue/hc4/ApacheHttpClientBlockingChannel.java
@@ -25,7 +25,6 @@ import com.palantir.dialogue.RequestBody;
 import com.palantir.dialogue.Response;
 import com.palantir.dialogue.blocking.BlockingChannel;
 import com.palantir.dialogue.core.BaseUrl;
-import com.palantir.dialogue.core.ResponseLeakDetector;
 import com.palantir.logsafe.Preconditions;
 import com.palantir.logsafe.exceptions.SafeRuntimeException;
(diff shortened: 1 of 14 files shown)


carterkozak:
~~Following up separately to move leak detection closer to the source.~~ did it here.

dialogue-blocking-channels/src/main/java/com/palantir/dialogue/blocking/BlockingChannelAdapter.java
```diff
@@ -52,42 +57,66 @@ private BlockingChannelAdapter() {}
...
 
-        BlockingChannelAdapterChannel(BlockingChannel delegate, ListeningExecutorService executor) {
+        BlockingChannelAdapterChannel(BlockingChannel delegate, ExecutorService executor) {
             this.delegate = delegate;
             this.executor = executor;
         }
 
         @Override
         public ListenableFuture<Response> execute(Endpoint endpoint, Request request) {
-            return executor.submit(new BlockingChannelAdapterTask(delegate, endpoint, request));
```
iamdanfox, line 64: I don't really understand what was deficient here and why we need the SettableFuture indirection - previously, if a user called `.cancel(true)` on the listenable future, wouldn't that set the interrupted flag on the thread that was actively processing the BlockingChannelAdapterTask?

iamdanfox: Ah I think it makes sense now - the apache client might still construct a response even if the interrupted flag is set.  Would an alternative solution be to just invoke `delegate.execute(endpoint, request);` and then check if the current thread is interrupted and then close the response?

carterkozak: It would set the interrupted flag, but there's no guarantee that would stop execution. In the case that the apache hc response was already received and we were wrapping it in the dialogue response shim at the time, nothing would signal that the response would not be used, and any interruption checks we add to that method would be racy.

By using a settable future, we allow the code that sets the result to check the return value of the `set` operation and close the response if the result has already been completed via a cancel invocation.

The issue is that the future returned from the ListeningExecutorService is considered complete as soon as cancel is called, regardless of whether or not code is running, with no way to signal to that code that the result will be ignored.

iamdanfox: So I just tried adding one tiny bit to the BlockingChannelAdapterTask and it passed the `testCancel` unit test:

```java
            @Override
            public Response call() throws Exception {
                Response response = delegate.execute(endpoint, request);
                if (Thread.currentThread().isInterrupted()) {
                    response.close();
                }
                return response;
            }
```

Is there something missing here?

carterkozak: It's hard to test race conditions, but a response will leak in that case if cancel is invoked just after the `isInterrupted` invocation and before the response is returned.

dialogue-blocking-channels/src/main/java/com/palantir/dialogue/blocking/BlockingChannelAdapter.java
```diff
@@ -52,42 +57,66 @@ private BlockingChannelAdapter() {}
...
+            SettableFuture<Response> result = SettableFuture.create();
```
iamdanfox: s/result/settableFuture/? just makes it a tiny bit clearer which direction things are going when we're hooking up the cancel listener

dialogue-blocking-channels/src/main/java/com/palantir/dialogue/blocking/BlockingChannelAdapter.java
```diff
@@ -52,42 +57,66 @@ private BlockingChannelAdapter() {}
...
+            BlockingChannelAdapterTask runnable = new BlockingChannelAdapterTask(delegate, endpoint, request, result);
+            Future<?> future = executor.submit(runnable);
+            // The executor task should be interrupted on termination
+            result.addListener(() -> future.cancel(true), MoreExecutors.directExecutor());
```
iamdanfox: seems a tiny bit odd that this fires even when the result is completed normally... why not use `Futures.addCallback` to only fire in the event that you get a CancellationException?

dialogue-blocking-channels/src/main/java/com/palantir/dialogue/blocking/BlockingChannelAdapter.java
```diff
@@ -52,42 +57,66 @@ private BlockingChannelAdapter() {}
...
                 this.endpoint = endpoint;
                 this.request = request;
             }
 
             @Override
-            public Response call() throws IOException {
-                return delegate.execute(endpoint, request);
+            public void run() {
+                if (result.isDone()) {
+                    return;
```
iamdanfox, line 118: Tiny bit confusing that we're using isDone as a proxy for `isCancelled`... I think a little comment would be helpful here, something along the lines of `if a user cancels the SettableFuture we exposed, then there's no need to call the delegate`.

iamdanfox: actually how could this ever be set to a value already (but not cancelled) here? isn't that an invariant violation?

dialogue-core/src/main/java/com/palantir/dialogue/core/ResponseLeakDetector.java
```diff
@@ -35,40 +32,33 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-final class LeakDetectingChannel implements Channel {
+public final class ResponseLeakDetector {
```
iamdanfox: thoughts on just moving this entire class to the `com.palantir.dialogue.hc4` package? this feels more internal than an intentionally exposed piece of public API

dialogue-blocking-channels/src/main/java/com/palantir/dialogue/blocking/BlockingChannelAdapter.java
carterkozak: it's a bit heavier weight, but probably not something to optimize around at this point, updated.

dialogue-apache-hc4-client/src/main/java/com/palantir/dialogue/hc4/ApacheHttpClientBlockingChannel.java
```diff
@@ -70,7 +74,7 @@ public Response execute(Endpoint endpoint, Request request) throws IOException {
             RequestBody body = request.body().get();
             builder.setEntity(new RequestBodyEntity(body));
         }
-        return new HttpClientResponse(client.execute(builder.build()));
+        return responseLeakDetector.wrap(new HttpClientResponse(client.execute(builder.build())), endpoint);
```
iamdanfox: could we pull some of these out to variables so that it's not a one-liner? it makes stacktraces more immediately obvious and also makes stepping through in a debugger a little easier :)

iamdanfox:
This leaking on cancellation stuff is mega subtle - did you spot this just be eyeballing code or was there a test or use-case that exposed this?

carterkozak:
eyeball :-/

dialogue-blocking-channels/src/main/java/com/palantir/dialogue/blocking/BlockingChannelAdapter.java
carterkozak: `isCancelled` works just as well, I chose isDone because it maps more closely to the final check we make on completion, settableFuture.set returns false if the future is done (which includes isCancelled).

dialogue-apache-hc4-client/src/main/java/com/palantir/dialogue/hc4/ApacheHttpClientBlockingChannel.java
carterkozak: Updated, added another defensive check to close responses if exceptions are thrown between client.execute and returning a the dialogue response.

dialogue-core/src/main/java/com/palantir/dialogue/core/ResponseLeakDetector.java
carterkozak: Updated. Unfortunate we mix junit 4 and 5 here, heh

iamdanfox (approved):

carterkozak:
:robot::popcorn:

Issue #565, opened 2020-03-26
RetryingChannel stacktraces don't mention callsite


iamdanfox:
## What happened?

Some products are seeing a lot of `Retrying call after failure` INFO lines, with a stacktrace like this:
```
com.palantir.logsafe.exceptions.SafeRuntimeException: Received retryable response
	at com.palantir.dialogue.core.RetryingChannel$RetryingCallback.success(RetryingChannel.java:190)
	at com.google.common.util.concurrent.AbstractTransformFuture$AsyncTransformFuture.doTransform(AbstractTransformFuture.java:213)
	at com.google.common.util.concurrent.AbstractTransformFuture$AsyncTransformFuture.doTransform(AbstractTransformFuture.java:202)
	at com.google.common.util.concurrent.AbstractTransformFuture.run(AbstractTransformFuture.java:118)
	at com.google.common.util.concurrent.DirectExecutor.execute(DirectExecutor.java:30)
	at com.google.common.util.concurrent.AbstractFuture.executeListener(AbstractFuture.java:1164)
	at com.google.common.util.concurrent.AbstractFuture.complete(AbstractFuture.java:958)
	at com.google.common.util.concurrent.AbstractFuture.set(AbstractFuture.java:727)
	at com.google.common.util.concurrent.AbstractTransformFuture$TransformFuture.setResult(AbstractTransformFuture.java:247)
	at com.google.common.util.concurrent.AbstractTransformFuture.run(AbstractTransformFuture.java:163)
	at com.google.common.util.concurrent.DirectExecutor.execute(DirectExecutor.java:30)
	at com.google.common.util.concurrent.AbstractFuture.executeListener(AbstractFuture.java:1164)
	at com.google.common.util.concurrent.AbstractFuture.complete(AbstractFuture.java:958)
	at com.google.common.util.concurrent.AbstractFuture.set(AbstractFuture.java:727)
	at com.google.common.util.concurrent.TrustedListenableFutureTask$TrustedFutureInterruptibleTask.afterRanInterruptibly(TrustedListenableFutureTask.java:131)
	at com.google.common.util.concurrent.InterruptibleTask.run(InterruptibleTask.java:133)
	at com.google.common.util.concurrent.TrustedListenableFutureTask.run(TrustedListenableFutureTask.java:78)
	at com.palantir.tracing.Tracers$TracingAwareRunnable.run(Tracers.java:501)
[...]

PR #568, opened 2020-03-26
RetryingChannel records stacktraces for logs (when debug logging is enabled)

iamdanfox:
## Before this PR

Motivation here: https://github.com/palantir/dialogue/issues/565

## After this PR
==COMMIT_MSG==
RetryingChannel records informational stacktraces when debug logging enabled.
==COMMIT_MSG==

## Possible downsides?
- recording stacktraces is expensive, so we don't want to do it all the time
- don't have any tests, could regress

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
@@ -148,10 +148,10 @@ final class RetryingChannel implements Channel {
     @Override
     public ListenableFuture<Response> execute(Endpoint endpoint, Request request) {
         if (isRetryable(request)) {
-            Optional<SafeRuntimeException> exceptionForStacktrace = log.isDebugEnabled()
+            Optional<SafeRuntimeException> debugStacktrace = log.isDebugEnabled()
                     ? Optional.of(new SafeRuntimeException("Exception for stacktrace"))
                     : Optional.empty();
-            return new RetryingCallback(endpoint, request, exceptionForStacktrace).execute();
+            return new RetryingCallback(endpoint, request, debugStacktrace).execute();
         }
         return delegate.execute(endpoint, request);
     }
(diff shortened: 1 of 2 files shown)


carterkozak:
What if we go the other direction and add retried failures as suppressed exceptions to the next failure — if the retry causes a call failure it will include suppressed exceptions from retries.

iamdanfox:
I don't want people's stacktraces to get _too_ long though... I feel like a chain of 4 might be pretty hard to read.

iamdanfox:
Whichever approach we go with, I think it would probably be nice to have _some_ kind of test here... Would you recommend using the SystemOutRule for this??

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -117,12 +120,22 @@
...
     }
 
     @Override
     public ListenableFuture<Response> execute(Endpoint endpoint, Request request) {
         if (isRetryable(request)) {
-            return new RetryingCallback(endpoint, request).execute();
+            Optional<SafeRuntimeException> exceptionForStacktrace =
+                    shouldRecordStacktrace.getIfPresent(endpoint) != null
+                            ? Optional.of(new SafeRuntimeException("Exception for stacktrace"))
+                            : Optional.empty();
```
carterkozak: I'm worried about the implications for high-load services running in environments without the best routing hardware, this could further tank performance. Let's chat through it at tomorrows standup :-)

carterkozak:
Does SystemOutRule work with jupiter?

iamdanfox:
Will ditch the cache and check isDebugEnabled.

carterkozak (approved):

carterkozak:
Might want to rename this PR for cleaner git history

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -143,12 +144,20 @@
                 .channelName(channelName)
                 .reason("qosResponse")
                 .build();
+        this.retryDueToThrowable = DialogueClientMetrics.of(metrics)
+                .requestRetry()
+                .channelName(channelName)
+                .reason("throwable")
+                .build();
```
iamdanfox, line 151: I also tacked on this bit of instrumentation, as I realized we missed a case!

carterkozak: Nice catch, thanks!

PR #570, opened 2020-03-26
Return unclosed InputStreams for binary and optional<binary> endpoints

### LINKED ISSUE #569: ContentDecodingChannel closes all InputStreams, which breaks optional<binary> responses
## What happened?

@jkozlowski's data storage client has a conjure endpoint which returns `optional<binary>`. As of [conjure-java 5.10.3](https://github.com/palantir/conjure-java/pull/828), the codegen uses a nice new `optionalInputStreamDeserializer`, but unfortunately the ContentDecodingChannel wrecks us. (this bug arose with [conjure-java 5.11.2](https://internal-github/foundry/a/blob/1.212.0/versions.props#L6), dialogue 1.0.2)

```
Caused by: java.io.IOException: Attempted read on closed stream.
	at org.apache.http.conn.EofSensorInputStream.isReadAllowed(EofSensorInputStream.java:107)
	at org.apache.http.conn.EofSensorInputStream.read(EofSensorInputStream.java:116)
	at java.base/java.util.zip.CheckedInputStream.read(CheckedInputStream.java:60)
	at java.base/java.util.zip.GZIPInputStream.readUByte(GZIPInputStream.java:267)
	at java.base/java.util.zip.GZIPInputStream.readUShort(GZIPInputStream.java:259)
	at java.base/java.util.zip.GZIPInputStream.readHeader(GZIPInputStream.java:165)
	at java.base/java.util.zip.GZIPInputStream.<init>(GZIPInputStream.java:80)
	at com.palantir.dialogue.core.ContentDecodingChannel$DeferredGzipInputStream.getDelegate(ContentDecodingChannel.java:160)
	at com.palantir.dialogue.core.ContentDecodingChannel$DeferredGzipInputStream.read(ContentDecodingChannel.java:188)
	at java.base/sun.nio.cs.StreamDecoder.readBytes(StreamDecoder.java:284)
	at java.base/sun.nio.cs.StreamDecoder.implRead(StreamDecoder.java:326)
	at java.base/sun.nio.cs.StreamDecoder.read(StreamDecoder.java:178)
	at java.base/java.io.InputStreamReader.read(InputStreamReader.java:185)
	at java.base/java.io.BufferedReader.fill(BufferedReader.java:161)
	at java.base/java.io.BufferedReader.readLine(BufferedReader.java:326)
	at java.base/java.io.BufferedReader.readLine(BufferedReader.java:392)
	at org.assertj.core.internal.Diff.linesFromBufferedReader(Diff.java:103)
	at org.assertj.core.internal.Diff.diff(Diff.java:90)
	at org.assertj.core.internal.Diff.diff(Diff.java:52)
	at org.assertj.core.internal.InputStreams.assertSameContentAs(InputStreams.java:71)
	... 24 more
```

- For the majority of API calls, closing the inputstream is good (i.e. when we're deserializing using jackson)
- When we're returning a raw `InputStream` or `Optional<InputStream>` (and perhaps eventually aliases thereof), we need to hand out _unclosed_ inputstreams, so users can actually get the data out at their leisure.

iamdanfox:
## Before this PR
Dialogue's binary handling is _totally broken_ right now, which obviously blocks usage in internal replacement for atlas. 

fixes https://github.com/palantir/dialogue/issues/569

## After this PR
==COMMIT_MSG==
Conjure endpoints returning `binary` or `optional<binary>` now correctly return unclosed InputStreams. Users must be careful to close these InputStreams, otherwise resources will be leaked.
==COMMIT_MSG==

TODO:
- [x] I think we need to hook up some way to close the _Response_ object automatically for these dudes
- [x] how defensive is this if people define their own encodings

## Possible downsides?
<!-- Please describe any way


dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ConjureBodySerDe.java
```diff
@@ -210,28 +211,31 @@ public void close() {
...
+            Optional<String> contentType = response.getFirstHeader(HttpHeaders.CONTENT_TYPE);
+            if (!contentType.isPresent()) {
                 response.close();
+                throw new SafeIllegalArgumentException(
+                        "Response is missing Content-Type header",
+                        SafeArg.of("received", response.headers().keySet()));
             }
+            Encoding.Deserializer<T> deserializer = getResponseDeserializer(contentType.get());
+            return deserializer.deserialize(response.body(), response);
         }
```
iamdanfox, line 242: this is really the key change in this PR - previously the finally block would sweep up _all_ response instances and call close (which made the binary cases _completely_ unusable), whereas now we rely on each individual deserializer to do the right thing, and this lets us put some special logic in the binary deserializers.

dialogue-client-test-lib/src/main/java/com/palantir/dialogue/CloseRecordingInputStream.java
```diff
@@ -0,0 +1,92 @@
...
+
+package com.palantir.dialogue;
+
+import com.palantir.logsafe.exceptions.SafeRuntimeException;
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Optional;
+import org.assertj.core.api.Assertions;
+
+/** A test-only inputstream which can only be closed once. */
```
carterkozak, line 25: is this true? It looks like close can be called several times, which should work correctly based on the closeable javadoc.

dialogue-client-test-lib/src/main/java/com/palantir/dialogue/CloseRecordingInputStream.java
```diff
@@ -0,0 +1,92 @@
...
+
+    @Override
+    public long skip(long num) throws IOException {
+        assertNotClosed();
+        return delegate.skip(num);
+    }
+
+    @Override
+    public void close() throws IOException {
+        closeCalled = Optional.of(new SafeRuntimeException("close was called here"));
```
carterkozak, line 89: should we record only the first close invocation? subsequent close calls will replace that value.

dialogue-client-test-lib/src/main/java/com/palantir/dialogue/TestResponse.java
```diff
@@ -0,0 +1,84 @@
...
+        }
+    }
+
+    public boolean isClosed() {
+        return closeCalled.isPresent();
+    }
+
+    private void checkNotClosed() {
+        if (closeCalled.isPresent()) {
+            throw new SafeRuntimeException("Please don't close twice", closeCalled.get());
```
carterkozak, line 75: Why should this only be closed exactly once? I think multiple invocations are fine, as long as we don't accidentally close things we shouldn't -- we don't want to leak resources!

dialogue-client-verifier/src/test/java/com/palantir/verification/BinaryReturnTypeTest.java
```diff
@@ -0,0 +1,169 @@
...
+    private static final UserAgent USER_AGENT = UserAgent.of(UserAgent.Agent.of("foo", "1.0.0"));
+    private static final SslConfiguration SSL_CONFIG = SslConfiguration.of(
+            Paths.get("../dialogue-core/src/test/resources/trustStore.jks"),
+            Paths.get("../dialogue-core/src/test" + "/resources/keyStore.jks"),
+            "keystore");
+
+    @Test
+    public void undertow_server() {
+        Undertow undertow = Undertow.builder()
+                .addHttpListener(0, "localhost", new HttpHandler() {
```
carterkozak:
```suggestion
                .addHttpListener(0, "localhost", new BlockingHandler(new HttpHandler() {
```

This way we don't need to call `startBlocking` in the custom handler -- as written we're running blocking i/o on a non-blocking i/o thread, which wouldn't be a good idea in production

dialogue-client-verifier/src/test/java/com/palantir/verification/BinaryReturnTypeTest.java
```diff
@@ -0,0 +1,169 @@
...
+                    @Override
+                    public void handleRequest(HttpServerExchange exchange) throws IOException {
+                        exchange.getResponseHeaders().put(Headers.CONTENT_TYPE, "application/octet-stream");
+                        exchange.getResponseHeaders().put(Headers.CONTENT_ENCODING, "gzip");
```
carterkozak: Does compression have anything to do with the problem, or does gzip expose the issue because otherwise the content fits into a buffer?

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/BinaryEncoding.java
```diff
@@ -81,17 +84,40 @@ public String toString() {
...
         }
     }
+
+    static class ResponseClosingInputStream extends ForwardingInputStream {
+        private final InputStream inputStream;
+        private final Closeable response;
+
+        ResponseClosingInputStream(InputStream inputStream, Closeable response) {
+            this.inputStream = inputStream;
+            this.response = response;
```
carterkozak: Not sure it's necessary to pass along the response object here -- all our implementations only require the body stream to be closed, and allow Response.close as a convenience to close the body.

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ConjureBodySerDe.java
```diff
@@ -210,28 +211,31 @@ public void close() {
 
         @Override
         public T deserialize(Response response) {
-            try {
-                if (errorDecoder.isError(response)) {
+            if (errorDecoder.isError(response)) {
```
carterkozak: we may want to use a defensive pattern here to handle unexpected cases:

```java
boolean closeResponse = true;
try {
  /* work*/
  T result = deserializer.deserialize(response.body(), response);
   closeResponse = false; // deserialier is responsible for closing resources when necessary
   return result;
} finally {
    if (closeResponse) response.close();
}
```

dialogue-client-test-lib/src/main/java/com/palantir/dialogue/CloseRecordingInputStream.java
iamdanfox: I found this useful because while I was refactoring, I often ended up adding close calls to several different layers which quickly got confusing as it was never quite clear which bit was _actually_ responsible for closing everything.

I agree that the contract of Closeable doesn't require us to do this whole exactly-once thing - I just found it helpful to try and condense the closing responsibility into one place and not duplicate it.

dialogue-client-test-lib/src/main/java/com/palantir/dialogue/TestResponse.java
iamdanfox: happy to undo this if you prefer, as I mainly found it useful to guide refactoring.

dialogue-client-verifier/src/test/java/com/palantir/verification/BinaryReturnTypeTest.java
iamdanfox: The original stacktrace I saw made me think it was somehow the ContentDecodingChannel's fault so I tried to be as representative as possible:

```
Caused by: java.io.IOException: Attempted read on closed stream.
	at org.apache.http.conn.EofSensorInputStream.isReadAllowed(EofSensorInputStream.java:107)
	at org.apache.http.conn.EofSensorInputStream.read(EofSensorInputStream.java:116)
	at java.base/java.util.zip.CheckedInputStream.read(CheckedInputStream.java:60)
	at java.base/java.util.zip.GZIPInputStream.readUByte(GZIPInputStream.java:267)
	at java.base/java.util.zip.GZIPInputStream.readUShort(GZIPInputStream.java:259)
	at java.base/java.util.zip.GZIPInputStream.readHeader(GZIPInputStream.java:165)
	at java.base/java.util.zip.GZIPInputStream.<init>(GZIPInputStream.java:80)
	at com.palantir.dialogue.core.ContentDecodingChannel$DeferredGzipInputStream.getDelegate(ContentDecodingChannel.java:160)
	at com.palantir.dialogue.core.ContentDecodingChannel$DeferredGzipInputStream
```
In practise turning off all the content decoding & gzip stuff just gave me a different exception instead because the root cause was in the ConjureBodySerde class

carterkozak: Got it, thanks for the description

dialogue-client-test-lib/src/main/java/com/palantir/dialogue/TestResponse.java
carterkozak: Since it's test code I'm not too worried, there's a chance future changes to fix leaks cause this to fail validation unnecessarily. ¯\_(ツ)_/¯

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/BinaryEncoding.java
iamdanfox: Yeah so this was something I wasn't quite sure about - given that people are free to implement their own `Response`, it seemed like there's a possibility that closing the Response and closing the InputStream could release different resources somehow...

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ConjureBodySerDe.java
iamdanfox: Oh yeah that's much nicer - will update

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/BinaryEncoding.java
carterkozak: Thoughts on updating the Response javadoc to note that at least one of response or response.body must be closed, but both aren't required? That way we force implementors to handle resource release -- all our current implementations only hold resources in the body, so I don't think there's risk, and it keeps the api a touch cleaner.

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/Encoding.java
```diff
@@ -57,11 +58,15 @@
...
+         * corresponding object.
+         *
+         * <p>Implementations must close the {@link InputStream} and the {@code Closeable response}, except in the
+         * case that an InputStream is returned to the user, e.g. for a binary response type.
+         *
+         * <p>Format-related deserialization errors surface as {@link IllegalArgumentException}. Inputs and outputs
          * must never be null.
          */
-        T deserialize(InputStream input);
+        T deserialize(InputStream input, Closeable response);
```
carterkozak: I think we can safely back this out, if we'd like to ensure the response is closed I think we should pass the response object through instead of the stream, but that expands the API surface in a way I'd prefer to avoid.

iamdanfox: Hang on a sec - there's a bunch of interesting behaviour that we've put into the `Response#close` method, e.g. your LeakDetectingResponse disarms the leak detector.  If we say InputStreams can be closed even if we never call `Response#close`, won't this leak detector complain?

carterkozak: I think we have a test for precisely this case :-)
The leak detector shouldn't complain.

iamdanfox: Ok cool.

dialogue-client-verifier/src/test/java/com/palantir/verification/BinaryReturnTypeTest.java
```diff
@@ -0,0 +1,222 @@
...
+                .describedAs("Should receive exactly the number of bytes we sent!")
+                .isEqualTo(limit);
+
+        System.out.printf("%d MB took %d millis%n", megabytes, sw.elapsed(TimeUnit.MILLISECONDS));
+    }
+
+    private void setBinaryGzipResponse(String stringToCompress) {
+        undertowHandler = exchange -> {
+            exchange.getResponseHeaders().put(Headers.CONTENT_TYPE, "application/octet-stream");
+            exchange.getResponseHeaders().put(Headers.CONTENT_ENCODING, "gzip");
```
carterkozak, line 130: We might want to assert that there's a `Accept-Encoding` request header that contains `gzip` to prevent confusion in the future.

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ConjureBodySerDe.java
```diff
@@ -227,10 +229,14 @@ public T deserialize(Response response) {
                             "Response is missing Content-Type header",
                             SafeArg.of("received", response.headers().keySet()));
                 }
-                EncodingDeserializerContainer<T> container = getResponseDeserializer(contentType.get());
-                return container.deserializer.deserialize(response.body());
+                Encoding.Deserializer<T> deserializer = getResponseDeserializer(contentType.get());
+                // deserializer is responsible for closing the response
+                closeResponse = false;
```
carterkozak, line 235: nit: I'd actually prefer to set this after executing the deserializer and before returning it's result, that way if it throws and fails in an unexpected way we don't risk leaking resources.
That may result in multiple close invocations, but it beats the alternative of leaking resources in the case where a deserializer fails prior to a try-with-resources block.

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/BinaryEncoding.java
iamdanfox: Have added a bit of extra wording - does this capture your intentions?

carterkozak: :+1: yep, thanks.

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ConjureBodySerDe.java
iamdanfox: Have also added a test for this

dialogue-client-test-lib/src/main/java/com/palantir/dialogue/CloseRecordingInputStream.java
iamdanfox: Eh, not sure it's worth the faff tbh - people can unwind one at at time :)

carterkozak (approved):
Thanks @iamdanfox!

dialogue-client-verifier/src/test/java/com/palantir/verification/BinaryReturnTypeTest.java
```diff
@@ -0,0 +1,227 @@
...
+        };
+
+        Stopwatch sw = Stopwatch.createStarted();
+
+        InputStream maybeBinary = sampleServiceBlocking().getOptionalBinary().get();
+        assertThat(ByteStreams.exhaust(maybeBinary))
+                .describedAs("Should receive exactly the number of bytes we sent!")
+                .isEqualTo(limit);
+
+        System.out.printf("%d MB took %d millis%n", megabytes, sw.elapsed(TimeUnit.MILLISECONDS));
```
iamdanfox, line 120: maybe it's just my mac, but this takes like 28 seconds for 3GB sometimes

carterkozak: That’s odd considering we’re using plain text http, how long does the build system take?

iamdanfox:
Power merging because I have green CI but Circle is queuing me really hard:

image

https://circleci.com/workflow-run/0a2af8e1-6e2a-49c9-9f48-6b930622bae5

PR #571, opened 2020-03-27
Some helpful toStrings for debugging

iamdanfox:
## Before this PR

I noticed a few classes don't have nice toStrings, which makes it a bit trickier when hooked up to a debugger because you have to expand a bunch of classes.

## After this PR
==COMMIT_MSG==
More encoding-related classes now have helpful toString implementations
==COMMIT_MSG==


dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/BinaryEncoding.java
```diff
@@ -61,4 +61,37 @@ public String getContentType() {
...
+        INSTANCE;
+
+        @Override
+        public Object deserialize(InputStream input) {
+            return input;
+        }
+
+        @Override
+        public String toString() {
+            return "InputStreamDeserializer{}";
```
carterkozak, line 94: I think the sls mapper will use the name (INSTANCE) since this is an enum, this is helpful for the debugger though

ferozco:
👍

carterkozak (approved):

PR #572, opened 2020-03-27
Escape plus sign in query strings

pkoenig10:
## Before this PR
The `+` character is not escaped when used in query strings.

## After this PR
The `+` character is escaped when used in query strings.

According to [RFC 3986](https://tools.ietf.org/html/rfc3986#section-3.4), `+` is a valid character in a query string and does not need to be escaped. However, the [W3C recommendations for URI addressing](https://www.w3.org/Addressing/URL/4_URI_Recommentations.html) state that `+` is shorthand notation for a space. In practice, nearly all servers interpret `+` as a space and most URL encoding libraries escape `+` in query strings. In order to be compatible with server implementations, we should escape the `+` character in query strings.

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/BaseUrl.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/BaseUrl.java
@@ -221,6 +221,8 @@ public final class BaseUrl {
         private static final CharMatcher IS_HOST = UNRESERVED.or(SUB_DELIMS);
         private static final CharMatcher IS_P_CHAR = UNRESERVED.or(SUB_DELIMS);
         private static final CharMatcher IS_PATH = UNRESERVED.or(SUB_DELIMS).or(CharMatcher.anyOf("/"));
+        // The RFC permits percent-encoding any character. We also percent encode '+' because otherwise most servers
+        // interpret it as an url encoded space.
         private static final CharMatcher IS_QUERY_CHAR =
                 IS_P_CHAR.or(CharMatcher.anyOf("/?")).and(CharMatcher.noneOf("=&+"));
 
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/BaseUrl.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/BaseUrl.java
@@ -222,7 +222,7 @@ public final class BaseUrl {
         private static final CharMatcher IS_P_CHAR = UNRESERVED.or(SUB_DELIMS);
         private static final CharMatcher IS_PATH = UNRESERVED.or(SUB_DELIMS).or(CharMatcher.anyOf("/"));
         private static final CharMatcher IS_QUERY_CHAR =
-                CharMatcher.anyOf("=&").negate().and(IS_P_CHAR.or(CharMatcher.anyOf("?/")));
+                IS_P_CHAR.or(CharMatcher.anyOf("/?")).and(CharMatcher.noneOf("=&+"));
 
         static boolean isHost(String maybeHost) {
             return IS_HOST.matchesAllOf(maybeHost) || isIpv6Host(maybeHost);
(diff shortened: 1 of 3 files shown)


uschi2000:
I'm not convinced we should make this change as it violates the RFC.

pkoenig10:
The RFC only states that it is not required to percent encode `+`. It does not state that it is required to not percent encode `+`.

This is an issue of practicality. Without this change, dialogue is incorrect for nearly every server implementation.

uschi2000:
can we assume that all servers translate percent-encoded `+` back into `+`? if that is true, then I  your proposal does not harm, I agree.

pkoenig10:
Yes, any servers that don't would be incorrect.

dialogue-core/src/main/java/com/palantir/dialogue/core/BaseUrl.java
```diff
@@ -222,7 +222,7 @@ public String toString() {
         private static final CharMatcher IS_P_CHAR = UNRESERVED.or(SUB_DELIMS);
         private static final CharMatcher IS_PATH = UNRESERVED.or(SUB_DELIMS).or(CharMatcher.anyOf("/"));
         private static final CharMatcher IS_QUERY_CHAR =
-                CharMatcher.anyOf("=&").negate().and(IS_P_CHAR.or(CharMatcher.anyOf("?/")));
+                IS_P_CHAR.or(CharMatcher.anyOf("/?")).and(CharMatcher.noneOf("=&+"));
```
carterkozak, line 227: Mind adding a comment describing the flexibility the spec provides around the `+` character here?

carterkozak:
Great find @pkoenig10, thanks for providing a description and fix. It would be nice to add context in a comment in the code, otherwise I think this is good to go.

dialogue-core/src/test/java/com/palantir/dialogue/core/UrlBuilderTest.java
```diff
@@ -160,11 +160,11 @@ public void urlEncoder_encodePathSegment_onlyEncodesNonReservedChars() {
 
     @Test
     public void urlEncoder_encodeQuery_onlyEncodesNonReservedChars() {
-        String nonReserved = "aAzZ09!$'()*+,;?/";
+        String nonReserved = "aAzZ09!$'()*,;/?";
```
uschi2000, line 163: please add an explicit test that shows how we differ from the RFC

pkoenig10: Not sure what you're suggesting. An explicit test would be redundant. This test clearly demonstrates which characters are percent-encoded and which are not.

pkoenig10:
@carterkozak I'm going to need you to add the changelog entry, I opened this PR from a fork because I don't have permissions on this repo and can't edit the comment.

dialogue-core/src/test/java/com/palantir/dialogue/core/UrlBuilderTest.java
uschi2000:
```
@Test
public void urlEncoder_encodeQuery_encodesPlusSign() {
    // the encode behavior differs from the RFC in that it protects against servers that decode + as space
    ...
}
```

uschi2000: it's just a service to future readers/maintainers of this code. deviation from the spec is very surprising and I want to make sure (in two places) that we documented that and why we do it.

carterkozak:
I've checked the box, but I'm not sure the bot has permission to push back to your fork. Any chance you could add the entry manually?

dialogue-core/src/test/java/com/palantir/dialogue/core/UrlBuilderTest.java
carterkozak: Explicit test for behavior where we've decided to take a more specific approach sounds reasonable, though I don't think it differs from the RFC, it chooses a specific path where the rfc gives us the option not to encode `+`.

It would be helpful to call out specifically in the test that support among other libraries is mixed for handling non-encoded `+` in case someone wants to change the code later, they'll be more likely to reconsider :-)

carterkozak (approved):
thanks!

PR #576, opened 2020-03-27
Http 204 and alias of optional

### LINKED ISSUE #573: ConjureBodySerDe does not support alias of optional responses
## What happened?
When an endpoint returns an alias of an empty optional, the [conjure spec](https://github.com/palantir/conjure/blob/master/docs/spec/wire.md#31-status-codes) requires a 204 response with an empty body. When deserializing this response, Dialogue throws the following exception:
https://github.com/palantir/dialogue/blob/bb92eac336ba9e2e0938ed1d6b4248f61114fee9/dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ConjureBodySerDe.java#L218-L219

This is handled in conjure-java-runtime, by using reflection to detect alias of optional types:
https://github.com/palantir/conjure-java-runtime/blob/7761d1063e96fd53e46bf2f76596e14f12856488/conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/feignimpl/EmptyContainerDecoder.java#L106-L123

iamdanfox:
## Before this PR

@pkoenig10 reported the problem (fixes https://github.com/palantir/dialogue/issues/573). This is high-priority because it's a regression compared to c-j-r's deserialization.

## After this PR
==COMMIT_MSG==
When used with conjure-java generated java bindings, dialogue can now deserialize HTTP 204 into `FooAlias.of(Optional.empty())`. Previously it would throw.
==COMMIT_MSG==


dialogue-client-verifier/src/test/java/com/palantir/verification/IntegrationTest.java
```diff
@@ -120,6 +138,13 @@ public void stream_3_gigabytes() throws IOException {
         System.out.printf("%d MB took %d millis%n", megabytes, sw.elapsed(TimeUnit.MILLISECONDS));
     }
 
+    private void set204Response() {
+        undertowHandler = exchange -> {
+            exchange.getResponseHeaders().put(Headers.CONTENT_TYPE, "application/json");
```
carterkozak: This isn't valid, if there's no content it shouldn't have a type!

dialogue-serde/src/test/java/com/palantir/conjure/java/dialogue/serde/JacksonEmptyContainerLoaderTest.java
```diff
@@ -0,0 +1,228 @@
...
+                .isEqualTo(Alias3.of(OptionalDouble.empty()));
+    }
+
+    @Test
+    public void http_204_can_handle_alias_of_alias_of_optional_string() {
+        assertThat(emptyContainerDecoder.getEmptyInstance(new TypeMarker<AliasAlias1>() {}))
+                .isEqualTo(AliasAlias1.of(Alias1.of(Optional.empty())));
+    }
+
+    @Generated("com.palantir.conjure.java.types.AliasGenerator")
```
ferozco: can we avoid checking in generated code? We already generate bindings in other places

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ConjureBodySerDe.java
```diff
@@ -218,11 +218,10 @@ public T deserialize(Response response) {
...
                     throw errorDecoder.decode(response);
                 } else if (response.code() == 204) {
-                    if (!isOptionalType) {
-                        throw new SafeRuntimeException(
-                                "Unable to deserialize non-optional response type from 204", SafeArg.of("type", token));
-                    }
-                    return TypeMarkers.getEmptyOptional(token);
+                    // TODO(dfox): what if we get a 204 for a non-optional type???
+                    // TODO(dfox): support http200 & body=null
+                    // TODO(dfox): what if we were expecting an empty list but got {}?
```
iamdanfox, line 233: adding extra strict tests upstream to the verifier: https://github.com/palantir/conjure-verification/pull/330

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ConjureBodySerDe.java
```diff
@@ -50,6 +51,7 @@
     private final List<Encoding> encodingsSortedByWeight;
     private final ErrorDecoder errorDecoder;
     private final Encoding defaultEncoding;
+    private final EmptyContainerDeserializer empties;
```
ferozco: s/empties/emptyBodyDecoder/

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ConjureBodySerDe.java
```diff
@@ -195,17 +193,19 @@ public void close() {
...
+                ErrorDecoder errorDecoder,
+                EmptyContainerDeserializer empty,
+                TypeMarker<T> token) {
             this.encodings = encodings.stream()
                     .map(encoding -> new EncodingDeserializerContainer<>(encoding, token))
                     .collect(ImmutableList.toImmutableList());
             this.errorDecoder = errorDecoder;
-            this.token = token;
-            this.isOptionalType = TypeMarkers.isOptional(token);
+            this.emptyInstance = Suppliers.memoize(() -> empty.getEmptyInstance(token)); // TODO(dfox): save exception?
```
ferozco: I think by memoizing we will lose the exception if mulitple calls receive a 204 for a non empty type. can we construct the instance at response time?

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ConjureBodySerDe.java
```diff
@@ -218,11 +218,10 @@ public T deserialize(Response response) {
                 if (errorDecoder.isError(response)) {
                     throw errorDecoder.decode(response);
                 } else if (response.code() == 204) {
-                    if (!isOptionalType) {
-                        throw new SafeRuntimeException(
-                                "Unable to deserialize non-optional response type from 204", SafeArg.of("type", token));
-                    }
-                    return TypeMarkers.getEmptyOptional(token);
+                    // TODO(dfox): what if we get a 204 for a non-optional type???
```
ferozco, line 231: should be handled by moving the instance construction

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ConjureBodySerDe.java
```diff
@@ -218,11 +218,10 @@ public T deserialize(Response response) {
...
+                    // TODO(dfox): support http200 & body=null
```
ferozco, line 232: These last two cases are handled by the regular deserializer

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/Encodings.java
```diff
@@ -37,6 +39,9 @@
 
     private Encodings() {}
 
+    private static final Supplier<ObjectMapper> JSON_MAPPER =
```
ferozco, line 42: why memoize? can't we just have a static field?

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ConjureBodySerDe.java
iamdanfox: I'd prefer not to construct the instance every time because we always expect to get the same answer

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/Encodings.java
carterkozak: imo it's a bit cleaner to avoid work until it's requested, we don't want expensive class loading. When failures occur, it's much nicer to have a stack trace through a call tree than a class init error.

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/JacksonEmptyContainerLoader.java
```diff
@@ -0,0 +1,153 @@
...
+    private final ObjectMapper mapper;
+
+    JacksonEmptyContainerLoader(ObjectMapper mapper) {
+        this.mapper = mapper;
+    }
+
+    @Override
+    @SuppressWarnings("unchecked")
+    public <T> T getEmptyInstance(TypeMarker<T> token) {
+        Class<?> clazz = getClass(token.getType());
```
ferozco: Could we do something like:
```
ObjectReader objectReader = mapper.readerFor(mapper.constructType(type.getType()));
return objectReader.readValue(mapper.nullNode());
```

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ConjureBodySerDe.java
carterkozak: In most cases that's correct, unfortunately null->empty collection won't work that way because we can't provide field annotations.

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/JacksonEmptyContainerLoader.java
carterkozak: I created a PR into this branch which does basically that

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/Encodings.java
iamdanfox: In addition to the stacktraces, there's also the perf thing: if everyone did big work in their static fields, then witchcraft startup would get slower and slower.  If we all use suppliers.memoize then the world starts up fast

carterkozak: We use a variation of this in an internal project to memoize encoding.serializer/deserializer creation in conjure-undertow, to avoid relatively expensive ObjectReader/ObjectWriter creation until the first time they're needed. Might be worth implementing by default in both dialogue and conjure-undertow.

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/JacksonEmptyContainerLoader.java
carterkozak: Link #592

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/Encodings.java
ferozco: Ok, its pretty ubiquitous to have static mappers. Perhaps we should write an errorprone rule to try to avoid it. IMO its not that big of deal though

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ConjureBodySerDe.java
ferozco: I was wrong, memoizing won't interfere with the exception so this works well

carterkozak:
fortunately the builds aren't passing, so we've only burned a tag.

carterkozak (approved):

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/JacksonEmptyContainerLoader.java
```diff
@@ -0,0 +1,158 @@
...
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+import javax.annotation.Nonnull;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+final class JacksonEmptyContainerLoader implements EmptyContainerDeserializer {
```
iamdanfox, line 40: Linking to the original impl: https://github.com/palantir/conjure-java-runtime/blob/45d6c9fa668f000bcd86e9f8003789db5d70d843/conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/feignimpl/EmptyContainerDecoder.java#L79

iamdanfox:
:+1:

PR #582, opened 2020-03-28
Top level span provides information about the request being made

ferozco:
## Before this PR
The traces we produced where not very helpful since they did not show you what request was being made.

image

## After this PR
==COMMIT_MSG==
Top level span of a request contains the method, service and endpoint that is being called
==COMMIT_MSG==
image

## Possible downsides?
N/A


dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueTracedRequestChannel.java
```diff
@@ -0,0 +1,39 @@
...
+    private final Channel delegate;
+
+    public DialogueTracedRequestChannel(Channel delegate) {
+        this.delegate = delegate;
+    }
+
+    @Override
+    public ListenableFuture<Response> execute(Endpoint endpoint, Request request) {
+        String operationName =
+                "Dialogue: " + endpoint.httpMethod() + " " + endpoint.serviceName() + "#" + endpoint.endpointName();
```
carterkozak: I don’t think we need the http method here, either http method + template, or service + endpoint name should be enough uniquely identidying data

ferozco: I thought it was nice to provide as much information as possible, but I can remove the method if you want

carterkozak (approved):

dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueTracedRequestChannel.java
```diff
@@ -0,0 +1,38 @@
...
+    private final Channel delegate;
+
+    public DialogueTracedRequestChannel(Channel delegate) {
+        this.delegate = delegate;
+    }
+
+    @Override
+    public ListenableFuture<Response> execute(Endpoint endpoint, Request request) {
+        String operationName = "Dialogue: request " + endpoint.serviceName() + "#" + endpoint.endpointName();
+        return Tracers.wrapListenableFuture(operationName, () -> delegate.execute(endpoint, request));
```
iamdanfox, line 36: @ferozco I think previously we included the HTTP verb too, currently it's just the service & endpoint

carterkozak: https://github.com/palantir/dialogue/pull/582#discussion_r399687592

iamdanfox: fair enough

PR #583, opened 2020-03-28
Migrate to junit 5

ferozco:
## Before this PR
Majority of our tests used junit4

## After this PR
Majority of tests use junit5. We still can't migrate everything over since we don't have support for junit5 parameterized tests in OSS

## Possible downsides?
N/A


dialogue-client-test-lib/src/main/java/com/palantir/dialogue/AbstractSampleServiceClientTest.java
```diff
@@ -195,7 +199,8 @@ public void testBlocking_throwsOnConnectError() throws Exception {
                 .hasMessageMatching(".*((Connection refused)|(Failed to connect)).*");
     }
 
-    @Test(timeout = 5_000) // see client construction: we set a 1s timeout
+    @Test // see client construction: we set a 1s timeout
+    @Timeout(2)
```
carterkozak: intentional reduction from 5 to 2 seconds?

carterkozak:
lgtm aside from the timeout change, unsure if it makes a difference -- feel free to add merge-when-ready if the change is fine.
:+1:

iamdanfox:
Nice refactor, but I think the autorelease was probably unnecessary ;) https://github.com/palantir/dialogue/pull/588

PR #584, opened 2020-03-28
client produces outgoing span

ferozco:
## Before this PR
We did not produce a client outgoing span on requests

## After this PR
==COMMIT_MSG==
Apache channel produces client outgoing span
==COMMIT_MSG==

Also added tests so we can see the end to end trace produced by a channel.

UPDATED:
image

## Possible downsides?
N/A


carterkozak:
The unnamed operation should be named, and come from the tracing wrapped executor used by the blocking channel adapter, I think we have a setup bug.

carterkozak:
Thoughts on #585?

ferozco:
We should definitely name the adapter, but I don't think that addresses the issue that we're not actually producing a client outgoing span as the last span on an outgoing request. So I think we should merge both.

I'm also considering spending some time this week playing around with tracing, since I'm not particularly happy about all the extra spans from wrapping futures.

carterkozak:
> but I don't think that addresses the issue that we're not actually producing a client outgoing span as the last span on an outgoing request

This is an interesting point, we're already producing a client outgoing span which we associate with the request (using parent-span headers to connect spans across service boundaries), is it valid to produce a client outgoing span which cannot be tracked across services? It makes sense for the inner-most span to be client-outgoing and attach data to headers, but following that through I'm not sure it makes sense to disallow other tracing spans in the client infrastructure.

What do you think?

ferozco:
My mistake I had overlooked the fact that we attach current span Id within `TraceEnrichingChannel` and so the span I'm adding would not be the the parent span of the server side trace. I'm not entirely sure whether it is valid or not. I'm relatively confident that nothing we use internal would even enforce that the right annotations are produced.

Agreed, It would be most correct for us to produce a client outgoing span within `TraceEnrichingChannel` and attach that span to the out going request. We can then continue to produce subsequent spans and sort of treat them as internal spans. I'll update the PR accordingly.

ferozco:
@carterkozak should be ready for another look

dialogue-core/src/main/java/com/palantir/dialogue/core/TraceEnrichingChannel.java
```diff
@@ -21,19 +21,32 @@
...
 import com.palantir.dialogue.Response;
+import com.palantir.tracing.CloseableSpan;
+import com.palantir.tracing.DetachedSpan;
 import com.palantir.tracing.Tracer;
+import com.palantir.tracing.api.SpanType;
 import com.palantir.tracing.api.TraceHttpHeaders;
 
 /** A channel that adds Zipkin compatible tracing headers. */
 final class TraceEnrichingChannel implements Channel {
+    public static final String OPERATION = "Dialogue-http-request";
```
carterkozak: Can this be private? The public modifier can be removed either way.

ferozco: thanks intellij

carterkozak:
thanks! 👍

PR #585, opened 2020-03-29
Provide a tracing operation name to the blocking adapter executor

carterkozak:
==COMMIT_MSG==
Provide a tracing operation name to the blocking adapter executor
==COMMIT_MSG==


ferozco:
👍

ferozco:
🤖

PR #591, opened 2020-03-30
Upgrade netflix concurrency limits

ferozco:
Ran into a conflict in atlasdb (since they don't have GCV) where the version of concurrency limits pulled in by CJR is different from the one used by Dialogue.

Should we also autoupdate concurrency limits?


iamdanfox:
I think the only reason we didn't before was because behaviour changes in a patch release of the upstream library silently changed behaviour in a bad way for us, so the intention was that people would read release notes carefully.

ferozco:
Ok, sure! happy to keep it manual (i.e. not upgrade). This just brings us in line with what CJR produces and hence what is actually being used in production

versions.props
```diff
@@ -2,7 +2,7 @@ com.fasterxml.jackson.*:* = 2.10.3
 com.github.ben-manes.caffeine:caffeine = 2.8.1
 com.google.code.findbugs:jsr305 = 3.0.2
 com.google.guava:guava = 27.0.1-jre
-com.netflix.concurrency-limits:* = 0.2.0
+com.netflix.concurrency-limits:* = 0.2.2
```
iamdanfox, line 5: Ok only change that looks like it could affect us is the WindowedLimit, but simulations still look ok so :+1:

https://github.com/Netflix/concurrency-limits/compare/v0.2.0...v0.2.2

iamdanfox (approved):

versions.props
ferozco: We've been using 0.2.2 for the last 11 months: https://github.com/palantir/conjure-java-runtime/blob/develop/versions.props#L8

PR #599, opened 2020-03-30
Defer expensive ObjectReader/ObjectWriter creation

carterkozak:
## Before this PR
Expensive client creation regardless of whether or not components were used.

## After this PR
==COMMIT_MSG==
Defer expensive ObjectReader/ObjectWriter creation
==COMMIT_MSG==

## Possible downsides?
Things don't fail up front.


carterkozak:
some services have thousands of endpoints.

carterkozak:
also opened https://github.com/palantir/conjure-java/pull/859

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ConjureBodySerDe.java
```diff
@@ -76,6 +77,13 @@
         emptyBodyDeserializer = new EmptyBodyDeserializer(errorDecoder);
     }
 
+    private static List<WeightedEncoding> decorateEncodings(List<WeightedEncoding> input) {
```
ferozco, line 85: lets do this wrapping outside of the conjureBodySerde?

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/DeferredEncoding.java
```diff
@@ -0,0 +1,94 @@
...
+    public String toString() {
+        return "DeferredEncoding{delegate=" + delegate + '}';
+    }
+
+    private static final class DeferredSerializer<T> implements Serializer<T> {
+
+        private final Supplier<Serializer<T>> delegate;
+
+        DeferredSerializer(Supplier<Serializer<T>> delegate) {
+            this.delegate = Suppliers.memoize(delegate::get);
```
iamdanfox: given that these creation functions have no side-effects, thoughts on just storing a nullable field and not using Suppliers.memoize with it's `synchronized` block?

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/DeferredEncoding.java
```diff
@@ -0,0 +1,94 @@
...
+
+/**
+ * Wrapper around an {@link Encoding} which allows both {@link Serializer} and {@link Deserializer} instances
+ * to be created when they're first used. This avoids the up-front cost of creating jackson <pre>ObjectReader</pre>
+ * and <pre>ObjectWriter</pre> instances for each request and response body type, for each encoding on service
+ * startup when many endpoints are only used with one encoding based on the clients that make requests.
+ * Note that this results in the first request to a given endpoint being more expensive than it would be
+ * otherwise, though this is already the case to an extent before the JIT compiler can optimize the path.
+ */
+final class DeferredEncoding implements Encoding {
```
iamdanfox: OnDemandEncoding? LazyInitializedEncoding?

ferozco:
Seems like we're trading off start up time vs. runtime. Any idea what the impact of this would be for either case?

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ConjureBodySerDe.java
iamdanfox: not sure that makes sense, because users can in theory supply their own encodings using the builder?

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/DeferredEncoding.java
carterkozak: memoize doesn't synchronize unless another thread is already creating the object, in that case I think the ObjectReader/ObjectWriter are expensive enough that we might as well block and reuse the instance another thread has created for us. After the first creation, the memoized supplier is a simple volatile read.

carterkozak:
> Seems like we're trading off start up time vs. runtime. Any idea what the impact of this would be for either case?

The mirrored change for conjure-undertow shaved a few seconds off startup time for a large internal product. I figured I'd pull the memoized encoding bits upstream since we were discussing memoizing deserializer bits in the fix for 204 response decoding.

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ConjureBodySerDe.java
carterkozak: There's an argument for using minimal logic here and providing memoizing encoding wrappers from WC, which is why I hadn't proposed this yet. Since we've started looking at memoizing these things, I think it's best to centralize memoization in one place than to mix it, and this hedges the worst case (many large services where only a small subset of endpoints are used)

ferozco:
Ok, happy to move forward with this 👍

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/DeferredEncoding.java
carterkozak: :+1: updated to LazyInitializedEncoding here and in the conjure-undertow PR

PR #602, opened 2020-03-30
DialogueChannel behaves when initialized or live-reloaded with zero uris

iamdanfox:
## Before this PR

Saw a strange DataDog graph, one hypothesis is that it's somehow caused by having zero uris. Wanted to capture the expected behaviour, and in writing the test I think it's all a bit broken.

## After this PR
==COMMIT_MSG==
A DialogueChannel can now be constructed with zero uris (or live-reloaded to zero uris).  In this state, all requests are just queued up and drained when uris become available.
==COMMIT_MSG==

I'm not sure this ever really happens in practise, but given that it's an edge case which _could_ happen (especially when there's eventual consistency in our service discovery), I figured we might as well make it behave sensibly.

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueChannel.java
@@ -48,7 +48,9 @@ import javax.annotation.Nullable;
 
 public final class DialogueChannel implements Channel {
     private final Map<String, LimitedChannel> limitedChannelByUri = new ConcurrentHashMap<>();
-    private final AtomicReference<LimitedChannel> nodeSelectionStrategy = new AtomicReference<>();
+    private final AtomicReference<LimitedChannel> nodeSelectionStrategy =
+            new AtomicReference<>(ZeroUriChannel.INSTANCE);
+    private final QueuedChannel queuedChannel; // just so we can process the queue when uris reload
 
     private final String channelName;
     private final ClientConfiguration clientConfiguration;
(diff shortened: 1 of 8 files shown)


dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueChannel.java
```diff
@@ -110,6 +113,9 @@ public void updateUris(Collection<String> uris) {
                 ImmutableList.copyOf(limitedChannelByUri.values()),
                 random,
                 channelName));
+
+        // some stuck requests might be able to make progress now
+        queuedChannel.schedule();
```
iamdanfox, line 118: This also fixes an edge-case where if we had n fully saturated nodes (maybe stuck processing very slow requests) and another node comes online, that new node will now start receiving requests from the queue.

Before this PR, the new node would not be utilized until one of our existing n nodes returned something (or a new request came in).

dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueChannel.java
```diff
@@ -132,11 +138,16 @@ private LimitedChannel createLimitedChannel(String uri, int uriIndex) {
...
     private static LimitedChannel getUpdatedNodeSelectionStrategy(
-            @Nullable LimitedChannel previousNodeSelectionStrategy,
+            LimitedChannel previousNodeSelectionStrategy,
             ClientConfiguration config,
             List<LimitedChannel> channels,
             Random random,
             String channelName) {
+        if (channels.isEmpty()) {
+            log.info("Zero URIs available, all requests will be queued", SafeArg.of("channelName", channelName));
+            return ZeroUriChannel.INSTANCE;
```
carterkozak, line 148: In this case I think we want to fail all requests due to a non-existant server -- draining the queue. We should make sure that the failure doesn't produce retries.

iamdanfox: Hmm, I kinda liked the idea of queuing up requests when we're in a zero uri state just in case there are blips in service discovery?

carterkozak: I think that would be confusing for consumers, WC already uses channels that fail immediately when zero URIs are present.
Waiting for a URI to become available could be reasonable if we had a timeout, and possibly updated the queue logic to rotate requests so that a single request isn't queued forever while subsequent requests past the queue limit are always rejected.

dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueChannel.java
ellisjoe: it seems pretty strange to me that this channel would know about the QueuedChannel at all, and this is why the original queued channel had that scheduled poke every 100ms.

iamdanfox: I've actually just pulled this change out to a separate PR so it doesn't get too conflated from the zero URI behaviour (which is gonna change anyway): https://github.com/palantir/dialogue/pull/605 Shall we discuss there?

iamdanfox:
I've split this into two separate PRs, #605 and #606 - will put any behaviour changes separately, as I think the plan is to try and fail fast if there are zero uris (especially as the queue's capacity is only 1000, so if a server is cranking out thousands of requests per second, the queue won't really save them from anything).

PR #605, opened 2020-03-31
Live reloading an extra uri allows QueuedChannel to make progress

iamdanfox:
## Before this PR

Say you have n upstreams and their concurrency limiters are all saturated so there are a bunch of queued requests, then you live-reload in an extra uri. At the moment, no requests get sent to this URI until the QueuedChannel's edge transitions are triggered (i.e. either a new request is made or an existing request comes back).

## After this PR
==COMMIT_MSG==
Live reloading an extra uri allows QueuedChannel to immediately make progress
==COMMIT_MSG==

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueChannel.java
@@ -49,6 +49,7 @@ import javax.annotation.Nullable;
 public final class DialogueChannel implements Channel {
     private final Map<String, LimitedChannel> limitedChannelByUri = new ConcurrentHashMap<>();
     private final AtomicReference<LimitedChannel> nodeSelectionStrategy = new AtomicReference<>();
+    private final QueuedChannel queuedChannel; // just so we can process the queue when uris reload
 
     private final String channelName;
     private final ClientConfiguration clientConfiguration;
@@ -72,14 +73,10 @@ public final class DialogueChannel implements Channel {
         clientMetrics = DialogueClientMetrics.of(clientConfiguration.taggedMetricRegistry());
         this.clock = clock;
         this.random = random;
+        this.queuedChannel =
+                new QueuedChannel(new SupplierChannel(nodeSelectionStrategy::get), channelName, clientMetrics);
         updateUris(clientConfiguration.uris());
-        this.delegate = wrap(
-                channelName,
-                new SupplierChannel(nodeSelectionStrategy::get),
-                clientConfiguration,
-                scheduler,
-                random,
-                clientMetrics);
+        this.delegate = wrap(queuedChannel, channelName, clientConfiguration, scheduler, random, clientMetrics);
     }
 
     @Override
@@ -110,6 +107,9 @@ public final class DialogueChannel implements Channel {
                 ImmutableList.copyOf(limitedChannelByUri.values()),
                 random,
                 channelName));
+
+        // some queued requests might be able to make progress on a new uri now
+        queuedChannel.schedule();
     }
 
     private LimitedChannel createLimitedChannel(String uri, int uriIndex) {
@@ -209,13 +209,13 @@ public final class DialogueChannel implements Channel {
     }
 
     private static Channel wrap(
+            QueuedChannel queuedChannel,
             String channelName,
-            LimitedChannel delegate,
             ClientConfiguration conf,
             Supplier<ScheduledExecutorService> scheduler,
             Random random,
             DialogueClientMetrics clientMetrics) {
-        Channel channel = new LimitedChannelToChannelAdapter(new QueuedChannel(delegate, channelName, clientMetrics));
+        Channel channel = new LimitedChannelToChannelAdapter(queuedChannel);
         channel = new TracedChannel(channel, "Dialogue-request-attempt");
         channel = retryingChannel(channel, channelName, conf, scheduler, random);
         channel = new UserAgentChannel(channel, conf.userAgent().get());
--- a/dialogue-core/src/test/java/com/palantir/dialogue/core/DialogueChannelsTest.java
+++ b/dialogue-core/src/test/java/com/palantir/dialogue/core/DialogueChannelsTest.java
@@ -23,8 +23,14 @@ import static org.mockito.ArgumentMatchers.eq;
 import static org.mockito.Mockito.lenient;
 import static org.mockito.Mockito.when;
 
+import com.codahale.metrics.Counter;
+import com.codahale.metrics.Metric;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Iterables;
+import com.google.common.collect.Maps;
 import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
+import com.google.common.util.concurrent.SettableFuture;
 import com.palantir.conjure.java.api.config.service.ServiceConfiguration;
 import com.palantir.conjure.java.api.config.service.UserAgent;
 import com.palantir.conjure.java.api.config.ssl.SslConfiguration;
@@ -37,8 +43,10 @@ import com.palantir.dialogue.Request;
 import com.palantir.dialogue.Response;
 import com.palantir.dialogue.UrlBuilder;
 import com.palantir.tracing.TestTracing;
+import com.palantir.tritium.metrics.registry.MetricName;
 import java.nio.file.Paths;
 import java.util.Map;
+import java.util.Objects;
 import java.util.concurrent.ExecutionException;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
@@ -92,7 +100,7 @@ public final class DialogueChannelsTest {
     private Response response;
 
     private Request request = Request.builder().build();
-    private Channel channel;
+    private DialogueChannel channel;
 
     @BeforeEach
     public void before() {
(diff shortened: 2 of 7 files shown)


dialogue-core/src/test/java/com/palantir/dialogue/core/DialogueChannelsTest.java
```diff
@@ -155,6 +163,35 @@ public void bad_channel_throwing_an_error_still_returns_a_future() {
...
+                        + "should be 80 queued up requests")
+                .isEqualTo(numRequests - ConcurrencyLimitedChannel.INITIAL_LIMIT);
+
+        // live-reload from 0 -> 1
+        channel.updateUris(ImmutableList.of(stubConfig.uris().get(0), "https://some-other-uri"));
+
+        assertThat(queuedRequestsCounter())
+                .describedAs("Now that we have two uris, another batch of 20 requests can get out the door")
+                .isEqualTo(numRequests - 2 * ConcurrencyLimitedChannel.INITIAL_LIMIT);
+    }
```
carterkozak, line 188: slick test :-)

iamdanfox: Honestly it does feel a bit weird using metrics to make assertions about behaviour rather than direct java method calls, but I guess if our metrics no longer reflect the ground truth then we have bigger problems!

dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueChannel.java
```diff
@@ -49,6 +49,7 @@
 public final class DialogueChannel implements Channel {
     private final Map<String, LimitedChannel> limitedChannelByUri = new ConcurrentHashMap<>();
     private final AtomicReference<LimitedChannel> nodeSelectionStrategy = new AtomicReference<>();
+    private final QueuedChannel queuedChannel; // just so we can process the queue when uris reload
```
iamdanfox, line 52: Keeping this field around might tempt people to call the `QueuedChannel#maybeExecute` method directly instead of sending everything down through the fully decorated `delegate`. I could just keep around an interface that literally just has the `schedule()` method on it to prevent this... thoughts?

dialogue-core/src/test/java/com/palantir/dialogue/core/DialogueChannelsTest.java
carterkozak: Can we assert on the number of invocations of the delegate channel instead?

dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueChannel.java
carterkozak: The runnable idea protects us from misuse, but I think the abstraction makes the code less clear. imo this is fine as is.

carterkozak (approved):
Thanks Dan!

To address the question about the scheduled executor vs calling schedule ourselves, the event-driven approach allows us to avoid managing perpetually scheduled tasks, which are difficult to reason about and caused resource leaks when they weren't handled properly. this model handles edge cases more reliably as there's no question around waiting for a scheduled task to fire.

dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueChannel.java
```diff
@@ -110,6 +107,9 @@ public void updateUris(Collection<String> uris) {
                 ImmutableList.copyOf(limitedChannelByUri.values()),
                 random,
                 channelName));
+
+        // some queued requests might be able to make progress on a new uri now
+        queuedChannel.schedule();
```
ellisjoe, line 112: as mentioned in the other PR, this seems like a strange thing for the DialogueChannel to have to understand (feels very leaky)

iamdanfox: Ok so theoretically yes dialogue is just one giant stack of Channels all wrapping each other like a big onion and at the top level we only interact with the outer layer, but in practise I think this limits the functionality that can be built.

Given that this is entirely encapsulated inside DialogueChannel, we at least have the freedom to tweak this as nothing is leaked to users

dialogue-core/src/test/java/com/palantir/dialogue/core/DialogueChannelsTest.java
```diff
@@ -155,6 +159,34 @@ public void bad_channel_throwing_an_error_still_returns_a_future() {
         assertThatThrownBy(() -> Futures.getUnchecked(future)).hasCauseInstanceOf(NoClassDefFoundError.class);
     }
 
+    @Test
+    @SuppressWarnings("FutureReturnValueIgnored") // intentionally spawning a bunch of throwaway requests
+    void live_reloading_an_extra_uri_allows_queued_requests_to_make_progress() {
```
ellisjoe, line 164: like this is the sort of test that feels weird to me. we should be able to verify in the queue tests that when the delegate channel has capacity it's able to make progress, and it potentially has to force itself to try to make progress every so often, rather than verifying that each way an underlying channel may start to have new capacity allows the queued channel to make progress.

dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueChannel.java
ellisjoe: In my mind, this is the sort of leakiness that leads to the complexity of CJR. In the original queued implementation, I could describe it's behavior as "It will schedule new requests when they are first submitted, a request completes, and every 100ms". Now I have to describe it as "When they're submitted, a request completes, or when uris are reloaded" which now brings in behavior that has nothing to do with queueing. I think small things like this don't seem bad in isolation, but make understanding the overall behavior extremely difficult because each of the layers has an impact on the others.

carterkozak: I would make the same argument against the scheduler :-)

I think small things like the scheduled executor don't seem bad in isolation, but make understanding the overall behavior extremely difficult because each of the layers has an impact on the others.

Happy to chat if you like.

ellisjoe: Would be more than happy to chat. The thing that scares me is that the next time we add some functionality to update the underlying channels, we'll have to keep in mind that we need to keep poking the queued channel. The arch of this currently means that anytime an underlying channel is updated to increase it's capacity, it should also poke the queued channel which seems like a very strange thing for them to have to think about.

iamdanfox:
Maybe this isn't quite right yet, but I don't think the schedule every 100ms thing is really a viable alternative. The 100ms thing is kinda just sweeping imprecision under the carpet. If we have a chain of say 7 requests which each normally take ~5ms and they all fall into whatever unlucky edge case we didn't handle precisely (so get swept up by the 100ms scheduler) then an RPC call that would normally take 35ms could take 700ms.

ellisjoe:
Under normal conditions requests will get scheduled when they are submitted or when other requests complete at which point space is expected to be available. The only reason the background task is there is to ensure tasks get scheduled when the underlying channels have increased capacity but the first two situations are not met (eg: the uris have been reloaded, the channel has not been used in X seconds due to blacklisting, etc)

carterkozak:
I think that's the point Dan is raising, when the state becomes a bit degraded the tenth of a second jitter on top of a system that's already struggling under pressure is problematic for throughput.

ellisjoe:
I don't understand how this change is any better when the system is struggling? I also don't understand how the situation Dan described would ever occur.

The states in my model are:
1. If I make 7 requests and none can be scheduled: I will wait 100ms and try to schedule the first one.
2. If I make 7 requests and 1 can be scheduled: I will schedule it and each subsequent one after the prior finishes taking 35ms
3. If I can schedule more than one: I will schedule as many as possible concurrently, and schedule subsequent ones when those requests complete

Again, my main concern is that this leaks expectations across Channel boundaries (someone needs to poke the QueuedChannel when they think it may be able to schedule requests). I personally think the scheduled task to poke is easy to understand and reason about, but am not tied to that if there are alternatives that don't leak responsibilities across Channel boundaries.

iamdanfox:
The scenario I was imagining was a chain of 7 services, so service1 -> service2 -> service3 etc, not 7 requests coming from one place.  This is where I worry that in the swiss cheese model of all the badness lining up, you could have one user sitting waiting for a very long time (700ms) when they usually just wait 35ms.

ellisjoe:
What happens today when each service is unable to schedule the initial request? When will it be scheduled again?

carterkozak:
Our current implementation will fail requests immediately if hosts aren't available. If nodes are all limited and a new node is added, we immediately start flushing the queue to the new node to the extent our limiters allow. If we allowed requests to be queued until a first node is added, it would triggered similarly.

carterkozak:
To put it more simply, the request will be sent as soon as possible as permitted by LimitedChannels.

markelliot:
Our service provisioning isn't all that dynamic, the edge case here (all the nodes are busy, all the requests are in the queue waiting, a new node suddenly gets added) seems exceptionally rare. I'd value the encapsulation over the admittedly better behavior here because of the rarity.

markelliot:
In the cases where we're using Dialogue between thick clients and large, backend services, one would also expect to never encounter this, as the URL list consists only of a single proxy URL.

carterkozak:
That's one of several downsides we've noted in this PR, others, and various one-off conversations. Perpetual background tasks must be managed with the lifecycle of DialogueChannel, and must never be leaked. The complexity surrounding that change outweighs readability benefit over the current implementation. It has a similar allure to the CJR static state connection pooling for readability while making the code more silently complex.

> In the cases where we're using Dialogue between thick clients and large, backend services, one would also expect to never encounter this, as the URL list consists only of a single proxy URL.

This is only accurate for external clients, not those running within an enclave.

markelliot:
Making channels closeable, for instance would solve the lifecycle concerns. 

I’m less concerned about readability and more concerned about correctness and explainability — where is it documented that you must always call schedule when changing the supplier state of a QueuedChannel? How are we capturing why QueuedChannel’s implementation must be special for later developer?

The problems of CJR are mostly that the system was so complex and intertwined _and_ that none of it was documented _and_ that none of it was obvious or explainable.

markelliot:
> This is only accurate for external clients, not those running within an enclave.

As a practical matter our largest deployments still navigate through a border proxy with one hostname instead of many, even in enclaves.

iamdanfox:
Ok can I propose we take a step back here and try and write down some of the guiding principles of this project? I think if we had a written down document that said "in general we prioritise explainability over performance" or something like that then there would be fewer of these cases where I write a PR which looks reasonable to me as a maintainer but clearly doesn't sit well with you guys?

PR #606, opened 2020-03-31
Logging around zero uri behaviour

iamdanfox:
## Before this PR

We haven't really intentionally chosen any behaviour when a DialogueChannel live-reloads to/from a zero uri state. (Currently you can't construct one with zero uris, and reloading into this state throws an IndexOutOfBounds exception).

## After this PR
==COMMIT_MSG==
DialogueChannel now produces log.info lines when live-reloading transitions to/from a zero-uri state.
==COMMIT_MSG==

This doesn't change any of the underlying behaviour, it's just supposed to be an uncontroversial baby-step towards a better world.

## Possible downsides?
- we don't have tests for log.info lines, so this could probably regress silently.

### DIFF
--- a/changelog/@unreleased/pr-606.v2.yml
+++ b/changelog/@unreleased/pr-606.v2.yml
@@ -1,0 +1,6 @@
+type: improvement
+improvement:
+  description: DialogueChannel now produces log.info lines when live-reloading transitions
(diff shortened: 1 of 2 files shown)


carterkozak (approved):

ferozco:
👍

PR #608, opened 2020-04-01
Apache client idle and validation timeouts scale with socket timeouts

carterkozak:
When read and write timeouts from ClientConfiguration are low,
connections cannot be kept alive idle beyond the timeout duration.

This change introduces scaling from the idle connection timeout to
the inactivity validation timeout, requiring validation after 40%
of the idle timeout has passed. This doesn't make a difference in
most cases, slightly modifying the default inactivity validation time
from 25 seconds to 22 seconds.

## After this PR
==COMMIT_MSG==
Apache client idle and validation timeouts scale with socket timeouts
==COMMIT_MSG==

### DIFF
--- a/dialogue-apache-hc4-client/src/main/java/com/palantir/dialogue/hc4/ApacheHttpClientChannels.java
+++ b/dialogue-apache-hc4-client/src/main/java/com/palantir/dialogue/hc4/ApacheHttpClientChannels.java
@@ -113,7 +113,13 @@ public final class ApacheHttpClientChannels {
         long socketTimeoutMillis =
                 Math.max(conf.readTimeout().toMillis(), conf.writeTimeout().toMillis());
         int connectTimeout = Ints.checkedCast(conf.connectTimeout().toMillis());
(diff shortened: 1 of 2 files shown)


dialogue-apache-hc4-client/src/main/java/com/palantir/dialogue/hc4/ApacheHttpClientChannels.java
```diff
@@ -138,8 +140,7 @@ public static CloseableClient createCloseableHttpClient(ClientConfiguration conf
         connectionManager.setDefaultMaxPerRoute(Integer.MAX_VALUE);
         // Increased from two seconds to twenty-five seconds because we have strong support for retries
         // and can optimistically avoid expensive connection checks.
```
iamdanfox: javadoc now out of date - shall we put a little sentence explaining the choice of 55 (over say 1 minute) and the 2.5 divisor?

carterkozak: :+1:

iamdanfox (approved):

PR #609, opened 2020-04-01
[proof of concept] cancelling client-side future aborts work on the server

iamdanfox:
I literally just wanted to play around with this, not intending to merge anything.

Was a little surprised that this worked with http1.  Also seems like Futures.transform's cancel behaviour isn't what I wanted when it's already returned a Response (but the server is still doing work to put bytes onto the network).

### DIFF
--- a/dialogue-client-verifier/src/test/java/com/palantir/verification/IntegrationTest.java
+++ b/dialogue-client-verifier/src/test/java/com/palantir/verification/IntegrationTest.java
@@ -44,11 +44,13 @@ import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.OutputStream;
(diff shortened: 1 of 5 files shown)


dialogue-client-verifier/src/test/java/com/palantir/verification/IntegrationTest.java
```diff
@@ -112,6 +121,62 @@ public void conjure_generated_blocking_interface_with_optional_binary_return_typ
         assertThat(maybeBinary.get()).hasSameContentAs(asInputStream("Hello, world"));
     }
 
+    @Test
+    public void can_clients_abort_in_progress_server_work() throws Exception {
```
jkozlowski, line 125: Are we thinking about propagating timeouts for calls to the server? With http1 will cancelling the future not cost us connections?

iamdanfox:
Closing out because this does probably sacrifice a connection.

PR #625, opened 2020-04-08
Nicer error when queue is full

iamdanfox:
## Before this PR

An internal test is failing with this exception:

```
Caused by: com.palantir.logsafe.exceptions.SafeRuntimeException: Failed to make a request
	at com.palantir.dialogue.core.LimitedChannelToChannelAdapter.lambda$static$0(LimitedChannelToChannelAdapter.java:34)
	at java.base/java.util.Optional.orElseGet(Optional.java:369)
	at com.palantir.dialogue.core.LimitedChannelToChannelAdapter.execute(LimitedChannelToChannelAdapter.java:44)
	at com.palantir.dialogue.core.TracedChannel.lambda$execute$0(TracedChannel.java:37)
```

There's only one reason that this can happen (queue full), but I think it would be nicer if we explicitly tell users this.

## After this PR
==COMMIT_MSG==
When the queue is completely full, the returned failure message demonstrates it's because the queue is full
==COMMIT_MSG==


dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
```diff
@@ -80,13 +83,21 @@
         this.maxQueueSize = maxQueueSize;
         this.queueSizeCounter = metrics.requestsQueued(channelName);
         this.queuedTime = metrics.requestQueuedTime(channelName);
+        this.limitedResultSupplier = () -> Futures.immediateFailedFuture(new SafeRuntimeException(
+                "Unable to make a request",
+                SafeArg.of("queueSizeEstimate", queueSizeEstimate.get()),
```
carterkozak: I think we should remove this value, reading from the estimate will race execution and cause confusion.

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
```diff
@@ -80,13 +83,21 @@
         this.maxQueueSize = maxQueueSize;
         this.queueSizeCounter = metrics.requestsQueued(channelName);
         this.queuedTime = metrics.requestQueuedTime(channelName);
+        this.limitedResultSupplier = () -> Futures.immediateFailedFuture(new SafeRuntimeException(
+                "Unable to make a request",
```
carterkozak: Should we mention the queue here?

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
iamdanfox: This atomic read is arguably a bit pointless because we _should_ only hit this exception due to queueSizeEstimate > maxQueueSize, I guess it's just defensive in case we ever end up with a _different_ cause??

carterkozak (approved):

Issue #615, opened 2020-04-03
No good option for black holed server nodes


bwaldrep:
## What happened?

We've seen a few incidents where a client was talking to a 3-node server using a round-robin strategy to spread load evenly. For various reasons (typically a host-level issue) one server node will become unresponsive and black hole all incoming requests until they hit their client timeout. After a minute or two of continued round-robin nearly all client's worker threads are blocked waiting on requests to the bad node to time out, causing an outage.

## What did you want to happen?

None of the current node selection strategies handle this case well, as seen in the `black_hole*` simulation results. Ideally we could be robust to one of multiple server nodes entering this state and still make some progress using the remaining healthy nodes.

@iamdanfox had a promising prototype here: #373

iamdanfox:
So https://github.com/palantir/dialogue/pull/373 has now merged, and people anyone who has set `NodeSelectionStrategy.ROUND_ROBIN` will get it from dialogue 1.28.0 onwards. In the `black_hole` simulation where pin_until_error delivers 56% client-perceived success, the new 'Balanced' implementation delivers 91.3% client-perceived success :D

See the [simulation graphs](https://github.com/palantir/dialogue/blob/9061e48ae0c0b710fb829c75bd93c36d7445e3be/simulation/src/test/resources/report.md#black_holeconcurrency_limiter_round_robin)

PR #515, opened 2020-03-09
Pin until error only switches on errors

ferozco:
## Before this PR
We would switch the pinned channel in response to client side limits

## After this PR
==COMMIT_MSG==
Pin until error strategy only switches the pinned channel in response to 429's and 5xx's
==COMMIT_MSG==

## Possible downsides?
This change will result in a higher number of artificially failed requests since limiting "eats" a retry. Its not too big of a deal and can be further minimized through #512

PR #627, opened 2020-04-08
PinUntilError switches host when a LimitedChannel refuses a request

iamdanfox:
## Before this PR

When using the PinUntilError channel, if a server never sends any failure responses, it is possible to full saturate the concurrency limiter for one host without even trying any other hosts. This means that clients can experience artificial slowness, even when there is is a bunch of capacity that we're not using!

## After this PR
==COMMIT_MSG==
The `PinUntilError` node selection strategy will now consider a concurrency limited response as justification for switching channels.
==COMMIT_MSG==

# [GRAPHS](https://github.com/palantir/dialogue/blob/dfox/less-dumb-pin-until-error/simulation/src/test/resources/report.md)

## Possible downsides?
- the first request to get concurrency limited will burn a retry (and wait for a few artificial milliseconds) without actually making a HTTP request. I _think_ this will be amortized away load is not balanced across hosts (once we've switched host, we hope requests will get out the door without problems)
- when requests are coming in much faster than concurrency limiters allow them to go out (i.e. the queue is filling up), the PinUntilErrorChannel will end up behaving a bit like the RoundRobin/RandomSelection channel.


carterkozak:
I think the original implementation did what you've proposed, but we updated it not to do so.

Consider when a stack starts up several services (in a clean order, so requests are never rejected) and requests start to fire. Assuming we have work to do, we fully expect to high the low initial concurrency limits as we push them for more room. If we did fail over too quickly, services which depend on cache locality (or state shared across requests) will fail unnecessarily.

If the concurrency limiters are pushing back because operations are failing (5xx or QoS responses) the pin-until-error channel will be triggered to try the next host in addition to causing the failing channel to reduce concurrency limits.

Thoughts?

simulation/src/test/resources/report.md
```diff
@@ -4,7 +4,7 @@
                 all_nodes_500[CONCURRENCY_LIMITER_PIN_UNTIL_ERROR].txt:	success=50.0%	client_mean=PT0.6S         	server_cpu=PT20M          	client_received=2000/2000	server_resps=2000	codes={200=1000, 500=1000}
                     all_nodes_500[CONCURRENCY_LIMITER_ROUND_ROBIN].txt:	success=50.0%	client_mean=PT0.6S         	server_cpu=PT20M          	client_received=2000/2000	server_resps=2000	codes={200=1000, 500=1000}
                               all_nodes_500[UNLIMITED_ROUND_ROBIN].txt:	success=50.0%	client_mean=PT0.6S         	server_cpu=PT20M          	client_received=2000/2000	server_resps=2000	codes={200=1000, 500=1000}
-                   black_hole[CONCURRENCY_LIMITER_PIN_UNTIL_ERROR].txt:	success=30.0%	client_mean=PT0.6S         	server_cpu=PT6M           	client_received=600/2000	server_resps=600	codes={200=600}
+                   black_hole[CONCURRENCY_LIMITER_PIN_UNTIL_ERROR].txt:	success=89.0%	client_mean=PT0.600224719S 	server_cpu=PT17M48S       	client_received=1780/2000	server_resps=1780	codes={200=1780}
```
iamdanfox: @bwaldrep unsurprisingly this bugfix makes pin until error behave a lot better in the black hole case (https://github.com/palantir/dialogue/issues/615). We start off pinned to the bad host and chuck 20 requests into the black hole, at which point the concurrency limiter says no more, so we switch over to the non-broken host and everything after that succeeds.

https://github.com/palantir/dialogue/blob/dfox/less-dumb-pin-until-error/simulation/src/test/resources/report.md#black_holeconcurrency_limiter_pin_until_error

carterkozak: If we implement the feature to broadcast preferred node selection from a server, this algorithm could be helpful for services which benefit from cache locality but don't suffer correctness issues when requests don't hit the same host.

iamdanfox:
Huh yeah seems like https://github.com/palantir/dialogue/pull/515 landed while I was OOTO.

iamdanfox:
Hmm I don't think I'm super convinced to be honest - it seems like the tradeoffs here are dependent on how much services _need_ cache affinity, which I don't think we have great data for.

it just seems very odd that outgoing requests could get held up when other channels are literally not being utilized. Admittedly this wasn't the case with my a-client-java tests today (because that only had one node).

Also the majority of foundry services don't have rate limiters in place to send defensive 429s - they just get slower until the server falls over. In this reality, I think clients _need_ to try to distribute load a bit.

simulation/src/test/resources/report.md
```diff
@@ -13,14 +13,14 @@
         fast_500s_then_revert[CONCURRENCY_LIMITER_PIN_UNTIL_ERROR].txt:	success=100.0%	client_mean=PT0.080925333S 	server_cpu=PT5M3.47S      	client_received=3750/3750	server_resps=3750	codes={200=3750}
             fast_500s_then_revert[CONCURRENCY_LIMITER_ROUND_ROBIN].txt:	success=76.0%	client_mean=PT0.055333733S 	server_cpu=PT3M27.501499711S	client_received=3750/3750	server_resps=3750	codes={200=2849, 500=901}
                       fast_500s_then_revert[UNLIMITED_ROUND_ROBIN].txt:	success=76.0%	client_mean=PT0.055333733S 	server_cpu=PT3M27.501499711S	client_received=3750/3750	server_resps=3750	codes={200=2849, 500=901}
-               live_reloading[CONCURRENCY_LIMITER_PIN_UNTIL_ERROR].txt:	success=100.0%	client_mean=PT12.9285176S  	server_cpu=PT2H2M54.3S    	client_received=2500/2500	server_resps=2500	codes={200=2500}
+               live_reloading[CONCURRENCY_LIMITER_PIN_UNTIL_ERROR].txt:	success=65.3%	client_mean=PT4.2242424S   	server_cpu=PT1H58M42.9S   	client_received=2500/2500	server_resps=2500	codes={200=1633, 500=867}
```
iamdanfox: this is weird because one of the servers in this scenario is entirely broken, and we don't have the blacklisting channel to lock it down :(

simulation/src/test/resources/report.md
```diff
@@ -13,14 +13,14 @@
...
                    live_reloading[CONCURRENCY_LIMITER_ROUND_ROBIN].txt:	success=63.7%	client_mean=PT4.210672S    	server_cpu=PT1H58M42.9S   	client_received=2500/2500	server_resps=2500	codes={200=1593, 500=907}
                              live_reloading[UNLIMITED_ROUND_ROBIN].txt:	success=55.8%	client_mean=PT2.84916S     	server_cpu=PT1H58M42.9S   	client_received=2500/2500	server_resps=2500	codes={200=1394, 500=1106}
           one_big_spike[CONCURRENCY_LIMITER_BLACKLIST_ROUND_ROBIN].txt:	success=79.0%	client_mean=PT1.478050977S 	server_cpu=PT1M59.71393673S	client_received=1000/1000	server_resps=790	codes={200=790, Failed to make a request=210}
-                one_big_spike[CONCURRENCY_LIMITER_PIN_UNTIL_ERROR].txt:	success=100.0%	client_mean=PT3.53697893S  	server_cpu=PT2M30S        	client_received=1000/1000	server_resps=1000	codes={200=1000}
+                one_big_spike[CONCURRENCY_LIMITER_PIN_UNTIL_ERROR].txt:	success=100.0%	client_mean=PT1.880893726S 	server_cpu=PT2M30S        	client_received=1000/1000	server_resps=1000	codes={200=1000}
```
iamdanfox, line 20: pretty big improvement in our client_mean in the case of a big spike in requests here: https://github.com/palantir/dialogue/blob/dfox/less-dumb-pin-until-error/simulation/src/test/resources/report.md#one_big_spikeconcurrency_limiter_pin_until_error

simulation/src/test/resources/report.md
```diff
@@ -29,7 +29,7 @@ one_endpoint_dies_on_each_server[CONCURRENCY_LIMITER_PIN_UNTIL_ERROR].txt:	succe
         slow_503s_then_revert[CONCURRENCY_LIMITER_PIN_UNTIL_ERROR].txt:	success=100.0%	client_mean=PT0.075948S    	server_cpu=PT3M47.844S    	client_received=3000/3000	server_resps=3000	codes={200=3000}
             slow_503s_then_revert[CONCURRENCY_LIMITER_ROUND_ROBIN].txt:	success=100.0%	client_mean=PT0.707925608S 	server_cpu=PT34M27.664333313S	client_received=3000/3000	server_resps=3390	codes={200=3000}
                       slow_503s_then_revert[UNLIMITED_ROUND_ROBIN].txt:	success=100.0%	client_mean=PT1.470730224S 	server_cpu=PT1H11M35.535333313S	client_received=3000/3000	server_resps=3825	codes={200=3000}
-slowdown_and_error_thresholds[CONCURRENCY_LIMITER_PIN_UNTIL_ERROR].txt:	success=40.4%	client_mean=PT1M36.106700399S	server_cpu=PT10H20M47.626666524S	client_received=10000/10000	server_resps=10000	codes={200=4043, 500=5957}
+slowdown_and_error_thresholds[CONCURRENCY_LIMITER_PIN_UNTIL_ERROR].txt:	success=1.4%	client_mean=PT25.998591133S	server_cpu=PT10H38M15.47999998S	client_received=10000/10000	server_resps=10000	codes={200=136, 500=9864}
```
iamdanfox: Ok I straight up do not understand why the last one had 40% success and now we have 1% :/

ferozco:
Lets talk about this tomorrow during our stand up

iamdanfox:
Closing out - @ferozco to capture some of the existing rationale in javadoc about why we're refusing this approach because apparently we've tried it multiple times.

PR #628, opened 2020-04-08
maxQueueSize 1k -> 100k

iamdanfox:
## Before this PR

When trying to replace c-j-r with dialogue in @jkozlowski's data storage product, `testPartitionedRangeScan` passes with c-j-r but fails with dialogue because it kicks off a big burst of requests but the queue fills up and 4 retries get exhausted before any capacity becomes available.

https://internal-circle/gh/foundry/a-client-java/14856

## After this PR
==COMMIT_MSG==
The QueuedChannel can now absorb a spike of 100k requests when concurrency limiters are preventing requests from getting out the door (previously it could only absorb 1k).
==COMMIT_MSG==

_I tried this locally and it enabled the above tests to pass._

## Possible downsides?
- this isn't a long term solution, because as soon as a bigger burst of traffic comes along we'll hit the same degenerate case where the queue just refuses requests.

### DIFF
--- a/changelog/@unreleased/pr-628.v2.yml
+++ b/changelog/@unreleased/pr-628.v2.yml
@@ -1,0 +1,7 @@
+type: fix
+fix:
+  description: The QueuedChannel can now absorb a spike of 100k requests when concurrency
+    limiters are preventing requests from getting out the door (previously it could
+    only absorb 1k).
+  links:
+  - https://github.com/palantir/dialogue/pull/628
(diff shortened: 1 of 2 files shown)


carterkozak (approved):
lgtm, we haven't seen many requests get queued in our dashboards, so there shouldn't be much risk of memory pressure. Long term I'd like to bound the queue based on time.

iamdanfox:
🛎

PR #629, opened 2020-04-08
Retry GET & HEAD requests on server 500s

iamdanfox:
## Before this PR

Our simulations show a _lot_ of user-facing failures because whenever a server returns a 500 we just give up and die. Obviously retrying _all_ 500s is too dangerous because we servers may have already performed side-effects before they fail, leading to multiple unintended side-effects. We also don't have a first-class concept of 'idempotence' in conjure.

HOWEVER, it turns out [RFC2731](https://tools.ietf.org/html/rfc7231#section-4.2.1) actually defines which http methods are ok to retry, by specifying a concept of `safe` and `idempotent`.

## After this PR
==COMMIT_MSG==
dialogue now retries GET and HEAD requests after server 500s.
==COMMIT_MSG==

(Amusingly, it seems nginx actually did retry _everything_ for a while and only stopped resending POST requests [in 2016](https://ma.ttias.be/nginx-1-9-13-no-longer-retry-non-idempotent-upstream-requests-default/)).

# [GRAPHS](https://github.com/palantir/dialogue/blob/dfox/retry-5xx/simulation/src/test/resources/report.md)

## Possible downsides?
- people may have defined non-idempotent endpoints using GET/HEAD/PUT/DELETE methods, and this PR would cause them to receive retries, possibly triggering duplicate side-effects. I think this is unlikely, because nginx and squid already retry these.
- for a truly-broken server, this will increase the time it takes a user to receive a 5xx response
- if a server suddenly starts returning 5xxs for everything, it might suddenly get 4 x the request rate because all clients will start retrying a bunch!


simulation/src/test/resources/report.md
```diff
@@ -1,40 +1,40 @@
...
                               one_big_spike[UNLIMITED_ROUND_ROBIN].txt:	success=100.0%	client_mean=PT1.115837367S 	server_cpu=PT8M3.3S       	client_received=1000/1000	server_resps=3222	codes={200=1000}
-one_endpoint_dies_on_each_server[CONCURRENCY_LIMITER_PIN_UNTIL_ERROR].txt:	success=64.3%	client_mean=PT0.6036528S   	server_cpu=PT25M          	client_received=2500/2500	server_resps=2500	codes={200=1608, 500=892}
- one_endpoint_dies_on_each_server[CONCURRENCY_LIMITER_ROUND_ROBIN].txt:	success=65.5%	client_mean=PT0.6S         	server_cpu=PT25M          	client_received=2500/2500	server_resps=2500	codes={200=1638, 500=862}
-           one_endpoint_dies_on_each_server[UNLIMITED_ROUND_ROBIN].txt:	success=65.6%	client_mean=PT0.6S         	server_cpu=PT25M          	client_received=2500/2500	server_resps=2500	codes={200=1639, 500=861}
+one_endpoint_dies_on_each_server[CONCURRENCY_LIMITER_PIN_UNTIL_ERROR].txt:	success=97.6%	client_mean=PT1.970194456S 	server_cpu=PT41M42.6S     	client_received=2500/2500	server_resps=4171	codes={200=2441, 500=59}
```
iamdanfox: Big improvements in user-perceived success rate (at the cost of slightly slower mean): https://github.com/palantir/dialogue/blob/dfox/retry-5xx/simulation/src/test/resources/report.md#one_endpoint_dies_on_each_serverconcurrency_limiter_pin_until_error

simulation/src/test/resources/report.md
```diff
@@ -1,40 +1,40 @@
...
                            drastic_slowdown[UNLIMITED_ROUND_ROBIN].txt:	success=100.0%	client_mean=PT9.158069333S 	server_cpu=PT10H10M32.277333313S	client_received=4000/4000	server_resps=4000	codes={200=4000}
         fast_500s_then_revert[CONCURRENCY_LIMITER_PIN_UNTIL_ERROR].txt:	success=100.0%	client_mean=PT0.080925333S 	server_cpu=PT5M3.47S      	client_received=3750/3750	server_resps=3750	codes={200=3750}
-            fast_500s_then_revert[CONCURRENCY_LIMITER_ROUND_ROBIN].txt:	success=76.0%	client_mean=PT0.055333733S 	server_cpu=PT3M27.501499711S	client_received=3750/3750	server_resps=3750	codes={200=2849, 500=901}
-                      fast_500s_then_revert[UNLIMITED_ROUND_ROBIN].txt:	success=76.0%	client_mean=PT0.055333733S 	server_cpu=PT3M27.501499711S	client_received=3750/3750	server_resps=3750	codes={200=2849, 500=901}
+            fast_500s_then_revert[CONCURRENCY_LIMITER_ROUND_ROBIN].txt:	success=99.2%	client_mean=PT0.173513629S 	server_cpu=PT4M53.194832836S	client_received=3750/3750	server_resps=5217	codes={200=3721, 500=29}
```
iamdanfox: dramatic improvement here too. I know multipass had a campaign of getting all their heavy users to switch to ROUNB_ROBIN, so I think this is not pointless.

https://github.com/palantir/dialogue/blob/dfox/retry-5xx/simulation/src/test/resources/report.md#fast_500s_then_revertconcurrency_limiter_round_robin

carterkozak:
I think our nginx instances only retry 503 and connection failures, not based on other 5xx responses, but I may be misremembering.

I'm generally in favor of this idea, but worried about wrecking non-idempotent PUT endpoints where this could have catastrophic consequences. Worth a broad email and bit of runway before we roll it out broadly. Let's discuss tomorrow.

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -182,29 +195,35 @@ private long getBackoffNanoseconds() {
...
-                            "Exhausted {} retries, returning a retryable response with status {}",
-                            SafeArg.of("retries", maxRetries),
-                            SafeArg.of("status", response.code()));
-                }
-                // not closing the final response body because ConjureBodySerde needs to read it to deserialize
-                return Futures.immediateFuture(response);
+                return incrementFailuresAndMaybeRetry(response, qosThrowable);
+            }
+
+            if (Responses.isServerError(response) && safeAndIdempotent(endpoint.httpMethod())) {
```
iamdanfox: plan is to stick to 500s initially

iamdanfox: Also possibly want a meter so we can watch a graph and see how often this fires.

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -259,6 +278,31 @@ private void logRetry(long backoffNanoseconds, @Nullable Throwable throwable) {
...
+    private static boolean safeAndIdempotent(HttpMethod httpMethod) {
+        switch (httpMethod) {
+            case GET:
+            case HEAD:
+                // As per https://tools.ietf.org/html/rfc7231#section-4.2.1. "Safe" means roughly read-only.
+                // "Of the request methods defined by this specification, the GET, HEAD, OPTIONS, and TRACE methods
+                // are defined to be safe."
+                return true;
+            case PUT:
+            case DELETE:
```
iamdanfox, line 306: Initial plan is to not do PUT/DELETE because we're confident that devs have made them idempotent. Can revisit when we've done some more thorough investigation (e.g. analyzing request logs)

iamdanfox:
Also mention in the README, possibly include the keyword `may` in the conjure spec

dialogue-core/src/main/metrics/dialogue-core-metrics.yml
```diff
@@ -20,6 +20,10 @@ namespaces:
           Number of requests that are actively running. The `stage` may refer to `running` requests actively
           executing over the wire or `processing` which may be awaiting a client or backing off for a retry.
           Note that running requests are also counted as processing.
+      request.retry:
+        type: counter
```
carterkozak: I think a meter makes more sense here, we're generally measuring rate of retires over time rather than the concrete value.

iamdanfox:
Just a thought, do you think it would be interesting to also get metrics for when we fully _exhaust_ retries? This seems kinda interesting in the case of a big spike of requests where concurrency limiters quickly become saturated

carterkozak:
I put this together a while ago but never merged it: #527

iamdanfox:
Mmm interesting. Maybe another way would be to just dial up the `Exhausted {} retries, returning a retryable response with status {}` log message from debug -> info? Seems like it shouldn't be noisy most of the time

carterkozak:
It depends, we have a separate debug log line for the retryable http response path as well as the exception path. I don't think we want to log the exception path at info because the failure will be thrown at the callsite and eventually logged elsewhere, but I suppose we could log the rest of that data at info and leave out the trace?

simulation/src/test/resources/txt/all_nodes_500[CONCURRENCY_LIMITER_ROUND_ROBIN].txt
```diff
@@ -1 +1 @@
-success=73.4%	client_mean=PT2.210265S    	server_cpu=PT20M          	client_received=2000/2000	server_resps=2000	codes={200=1468, 500=532}
\ No newline at end of file
+success=100.0%	client_mean=PT4.921208174S 	server_cpu=PT25M43.2S     	client_received=2000/2000	server_resps=2572	codes={200=2000}
```
Comment (carterkozak) [inline simulation/src/test/resources/txt/all_nodes_500[CONCURRENCY_LIMITER_ROUND_ROBIN].txt]: I'm not sure we want to simulate using GET requests as they're underrepresented across the fleet. Thoughts?

carterkozak:
I'm happy with this :+1:, it's as safe as it can be :-)

PR #631, opened 2020-04-09
Fix PinUntilError to start on a random channel

iamdanfox:
## Before this PR

Our simulations are supposed to show the combined effect of 10 clients all calling one server, which made me wonder why all the PIN_UNTIL_ERROR cases had so consistently locked onto one host and never touched others...

Turns out PinUntilError has a bug where:
1. we pick out the first url, 
2. carefully shuffle the initial node list, and then
3. find the index of that first url in the newly shuffled list and start there.

This is a bit silly because at startup every server will start off unfairly locked onto this same host.

## After this PR
==COMMIT_MSG==
PinUntilError channel now starts off pinned to a randomly chosen url.
==COMMIT_MSG==

## Possible downsides?
- clearly this regressed... how do I write a test for this?


dialogue-core/src/main/java/com/palantir/dialogue/core/PinUntilErrorChannel.java
```diff
@@ -82,16 +83,7 @@
     }
 
     static PinUntilErrorChannel of(
-            NodeSelectionStrategy strategy,
-            List<LimitedChannel> channels,
-            DialoguePinuntilerrorMetrics metrics,
-            Random random,
-            String channelName) {
-        return from(channels.get(0), strategy, channels, metrics, random, channelName);
```
iamdanfox, line 90: this `channels.get(0)` is what wrecked us.

simulation/src/test/resources/report.md
```diff
@@ -4,32 +4,32 @@
...
+                   black_hole[CONCURRENCY_LIMITER_PIN_UNTIL_ERROR].txt:	success=100.0%	client_mean=PT0.6S         	server_cpu=PT20M          	client_received=2000/2000	server_resps=2000	codes={200=2000}
                        black_hole[CONCURRENCY_LIMITER_ROUND_ROBIN].txt:	success=90.0%	client_mean=PT0.600011117S 	server_cpu=PT17M59.4S     	client_received=1799/2000	server_resps=1799	codes={200=1799}
                                  black_hole[UNLIMITED_ROUND_ROBIN].txt:	success=68.3%	client_mean=PT0.6S         	server_cpu=PT13M39.6S     	client_received=1366/2000	server_resps=1366	codes={200=1366}
-             drastic_slowdown[CONCURRENCY_LIMITER_PIN_UNTIL_ERROR].txt:	success=100.0%	client_mean=PT0.075961S    	server_cpu=PT5M3.844S     	client_received=4000/4000	server_resps=4000	codes={200=4000}
+             drastic_slowdown[CONCURRENCY_LIMITER_PIN_UNTIL_ERROR].txt:	success=100.0%	client_mean=PT11.430491833S	server_cpu=PT2H28M50.443999975S	client_received=4000/4000	server_resps=4000	codes={200=4000}
```
iamdanfox, line 10: this changed because all our clients just happened to lock onto the other channel. (Fix for the identical client behaviour coming separately.)

https://github.com/palantir/dialogue/blob/dfox/fix-initial-shuffle/simulation/src/test/resources/report.md#drastic_slowdownconcurrency_limiter_pin_until_error

dialogue-core/src/main/java/com/palantir/dialogue/core/PinUntilErrorChannel.java
```diff
@@ -104,9 +96,12 @@ static PinUntilErrorChannel from(
          * situation where there might be many nodes but all clients have decided to hammer one of them.
          */
         ImmutableList<LimitedChannel> initialShuffle = shuffleImmutableList(channels, random);
-        // We only rely on reference equality since we expect LimitedChannels to be reused across updates
-        int initialHost = initialShuffle.indexOf(initialChannel);
-        initialHost = initialHost == -1 ? 0 : initialHost;
+        int initialHost = 0;
+        if (initialChannel.isPresent()) {
+            // indexOf relies on reference equality since we expect LimitedChannels to be reused across updates
+            initialHost = initialShuffle.indexOf(initialChannel.get());
```
carterkozak: nit, might be easier to read as one line using max instead of two asignments
```suggestion
            initialHost = Math.max(0, initialShuffle.indexOf(initialChannel.get()));
```

carterkozak:
👍

carterkozak (approved):

PR #632, opened 2020-04-09
[test-only] Simulations no longer have 10 identical clients

iamdanfox:
## Before this PR

Our simulations run with 10 clients, but because we seeded them with exactly the same pseudorandom instance, they'd behave _exactly_ the same.  This is leads to somewhat unrepresentative graphs, when one node is fast and one is slow, every single client might coincidentally start locked onto the fast one and just stay there.

## After this PR
==COMMIT_MSG==
Simulations are now more representative because every client's pseudorandom instance is seeded differently.
==COMMIT_MSG==


simulation/src/test/resources/report.md
```diff
@@ -5,36 +5,36 @@
                     all_nodes_500[CONCURRENCY_LIMITER_ROUND_ROBIN].txt:	success=50.0%	client_mean=PT0.6S         	server_cpu=PT20M          	client_received=2000/2000	server_resps=2000	codes={200=1000, 500=1000}
                               all_nodes_500[UNLIMITED_ROUND_ROBIN].txt:	success=50.0%	client_mean=PT0.6S         	server_cpu=PT20M          	client_received=2000/2000	server_resps=2000	codes={200=1000, 500=1000}
                    black_hole[CONCURRENCY_LIMITER_PIN_UNTIL_ERROR].txt:	success=30.0%	client_mean=PT0.6S         	server_cpu=PT6M           	client_received=600/2000	server_resps=600	codes={200=600}
```
iamdanfox: amusingly this PR hasn't changed _any_ of the pin until error ones due to this bug: https://github.com/palantir/dialogue/pull/631

simulation/src/test/java/com/palantir/dialogue/core/Strategy.java
```diff
@@ -29,52 +29,49 @@
...
+    };
 
     private static Channel withDefaults(
             Simulation sim,
             Supplier<Map<String, SimulationServer>> channelSupplier,
+            int clientNumber,
             UnaryOperator<ClientConfiguration.Builder> applyConfig) {
-        Random pseudo = new Random(3218974678L);
+        // ensures each client doesn't behave identically
+        Random pseudo = new Random(3218974678L + clientNumber);
```
carterkozak: What if we add a `Random random()` method to the `Simulation` and update this method to pull the simulation-wide random instead of creating a new one? This way the simulation remains deterministic but e don't need to plumb around metadata that shouldn't be otherwise meaningful.

simulation/src/test/resources/report.md
```diff
@@ -4,37 +4,37 @@
...
-                      fast_500s_then_revert[UNLIMITED_ROUND_ROBIN].txt:	success=76.0%	client_mean=PT0.055333733S 	server_cpu=PT3M27.501499711S	client_received=3750/3750	server_resps=3750	codes={200=2849, 500=901}
-               live_reloading[CONCURRENCY_LIMITER_PIN_UNTIL_ERROR].txt:	success=93.0%	client_mean=PT11.5675456S  	server_cpu=PT2H48.6S      	client_received=2500/2500	server_resps=2500	codes={200=2326, 500=174}
-                   live_reloading[CONCURRENCY_LIMITER_ROUND_ROBIN].txt:	success=63.7%	client_mean=PT4.210672S    	server_cpu=PT1H58M42.9S   	client_received=2500/2500	server_resps=2500	codes={200=1593, 500=907}
-                             live_reloading[UNLIMITED_ROUND_ROBIN].txt:	success=55.8%	client_mean=PT2.84916S     	server_cpu=PT1H58M42.9S   	client_received=2500/2500	server_resps=2500	codes={200=1394, 500=1106}
+                   black_hole[CONCURRENCY_LIMITER_PIN_UNTIL_ERROR].txt:	success=56.9%	client_mean=PT0.6S         	server_cpu=PT11M22.8S     	client_received=1138/2000	server_resps=1138	codes={200=1138}
```
iamdanfox, line 7: this looks a lot more realistic - old version was fake news.

https://github.com/palantir/dialogue/blob/dfox/more-representative/simulation/src/test/resources/report.md#black_holeconcurrency_limiter_pin_until_error

dialogue-core/src/main/java/com/palantir/dialogue/core/NeverThrowChannel.java
```diff
@@ -47,4 +47,9 @@
             return Futures.immediateFailedFuture(e);
         }
     }
+
+    @Override
+    public String toString() {
+        return "NeverThrowChannel{" + delegate + '}';
+    }
```
carterkozak, line 54: nice 👍

carterkozak (approved):
Thanks!

PR #633, opened 2020-04-09
Implement QosException mapping

carterkozak:
Previously QosExceptions that were not retried were thrown
as UnknownRemoteExceptions where we expect QosException types.

Fixed the error decoder to return mapped exceptions rather than
mixing throwing and returning.

## After this PR
==COMMIT_MSG==
Implement QosException mapping
==COMMIT_MSG==

### DIFF
--- a/dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ErrorDecoder.java
+++ b/dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ErrorDecoder.java
@@ -18,18 +18,26 @@ package com.palantir.conjure.java.dialogue.serde;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.io.CharStreams;
+import com.google.common.primitives.Longs;
+import com.palantir.conjure.java.api.errors.QosException;
 import com.palantir.conjure.java.api.errors.RemoteException;
 import com.palantir.conjure.java.api.errors.SerializableError;
 import com.palantir.conjure.java.api.errors.UnknownRemoteException;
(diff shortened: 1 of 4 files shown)


dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ErrorDecoder.java
```diff
@@ -40,36 +48,66 @@
...
+        int code = response.code();
+        switch (code) {
+            case 308:
+                Optional<String> location = response.getFirstHeader(HttpHeaders.LOCATION);
+                if (location.isPresent()) {
+                    String locationHeader = location.get();
+                    try {
+                        return QosException.retryOther(new URL(locationHeader));
+                    } catch (MalformedURLException e) {
+                        log.error(
```
ferozco, line 71: Its a little subtle that we fall through if we fail to parse the location header since the 503 could have a serializable exception body and encoding properly set

carterkozak: I don't think our utilities allow us to send 503 responses with serializable exception bodies, unless I'm missing something.

Agree that the fall-through is subtle, but I matched CJR behavior for now.

ferozco: ya that's fair

ferozco (approved):
Thanks @carterkozak

carterkozak:
:robot:

iamdanfox:
Ooof. Straight up was not implemented 😂

PR #635, opened 2020-04-10
[WIP] Make ConcurrencyLimiters handle a big spike of requests

iamdanfox:
## Before this PR

There is a test in a-client-java which passes with c-j-r but _fails_ with dialogue. It kicks off a big spike of requests (1600 specifically), but the server has a Guava RateLimiter that only tolerates 300 qps and sends 429s if you exceed this rate.

I added some logging to the dialogue ConcurrencyLimiter and it showed that we'd instantly fill up the default 20 permits, but when one of these responses came back 429, a permit would open and the QueuedChannel would immediately try to run the request at head of the queue.

Because the regular netflix WindowedLimit requires a _full_ window (which is time-based, derived from observed RTTs) before it makes any decisions to increase/decrease limits, this head of the queue request gets let out the door.

Here's where the badness happens. This request gets 429'd, response comes back and the queue kicks off the new queue head. This game continues until we've burned all the retries for all queued requests, leading to the final response coming out as "429 throttling suggested". Test fails.

_Amusingly, in James' original PR from 2018 when he introduced concurrencylimiters, he included the solution for this: https://github.com/palantir/conjure-java-runtime/pull/786/files#diff-6d01b85469a332c185d7de7b5dc070ccR33. Somehow it didn't get copied across to dialogue, hence my adventures above._

## After this PR
==COMMIT_MSG==
dialogue's concurrency limiters can now handle a big spike of requests to a defensively rate limited server. Previously, it would fire off requests _too_ quickly, leading to user-facing failures where all retries hit a 429.
==COMMIT_MSG==

It achieves this by _instantly_ decreasing the concurrency limit as soon as a failure or qos response is received (rather than waiting for a full window). This means that when the queue tries to send the first queued request, the max has already decreased to 19, so it doesn't get out the door straight away. This allows the limiter to get more information from the server (possibly downgrading in quick succession if necessary 20 -> 19 -> 18).

## Possible downsides?
- we need unit-tests for this copypasta
- we need a representative integration test in this repo


simulation/src/test/resources/report.md
```diff
@@ -1,36 +1,36 @@
...
                       slow_503s_then_revert[UNLIMITED_ROUND_ROBIN].txt:	success=100.0%	client_mean=PT1.430774859S 	server_cpu=PT1H9M33.661333314S	client_received=3000/3000	server_resps=3798	codes={200=3000}
-slowdown_and_error_thresholds[CONCURRENCY_LIMITER_PIN_UNTIL_ERROR].txt:	success=32.6%	client_mean=PT1M33.640484199S	server_cpu=PT10H34M22.666666616S	client_received=10000/10000	server_resps=10000	codes={200=3261, 500=6739}
-    slowdown_and_error_thresholds[CONCURRENCY_LIMITER_ROUND_ROBIN].txt:	success=1.2%	client_mean=PT24.092490733S	server_cpu=PT10H34M28.19999998S	client_received=10000/10000	server_resps=10000	codes={200=120, 500=9880}
+slowdown_and_error_thresholds[CONCURRENCY_LIMITER_PIN_UNTIL_ERROR].txt:	success=94.2%	client_mean=PT2M19.506447524S	server_cpu=PT6H33M46.759998892S	client_received=10000/10000	server_resps=10000	codes={200=9423, 500=577}
+    slowdown_and_error_thresholds[CONCURRENCY_LIMITER_ROUND_ROBIN].txt:	success=91.8%	client_mean=PT2M10.117350732S	server_cpu=PT9H7M52.566666203S	client_received=10000/10000	server_resps=10000	codes={200=9184, 500=816}
```
iamdanfox, line 33: Guys this is about as dramatic as possible. 1.2% user-perceived success rate turned into 91.8%.

https://github.com/palantir/dialogue/blob/dfox/bring-back-windowed-limit/simulation/src/test/resources/report.md#slowdown_and_error_thresholdsconcurrency_limiter_round_robin

simulation/src/test/resources/report.md
```diff
@@ -1,36 +1,36 @@
 # Report
 <!-- Run SimulationTest to regenerate this report. -->
 ```
-                all_nodes_500[CONCURRENCY_LIMITER_PIN_UNTIL_ERROR].txt:	success=50.0%	client_mean=PT0.6S         	server_cpu=PT20M          	client_received=2000/2000	server_resps=2000	codes={200=1000, 500=1000}
-                    all_nodes_500[CONCURRENCY_LIMITER_ROUND_ROBIN].txt:	success=50.0%	client_mean=PT0.6S         	server_cpu=PT20M          	client_received=2000/2000	server_resps=2000	codes={200=1000, 500=1000}
+                all_nodes_500[CONCURRENCY_LIMITER_PIN_UNTIL_ERROR].txt:	success=79.3%	client_mean=PT12.132645S   	server_cpu=PT20M          	client_received=2000/2000	server_resps=2000	codes={200=1586, 500=414}
```
iamdanfox, line 4: 50% success -> 79% success

https://github.com/palantir/dialogue/blob/dfox/bring-back-windowed-limit/simulation/src/test/resources/report.md#all_nodes_500concurrency_limiter_pin_until_error

simulation/src/test/resources/report.md
```diff
@@ -1,36 +1,36 @@
...
+                    one_big_spike[CONCURRENCY_LIMITER_ROUND_ROBIN].txt:	success=100.0%	client_mean=PT1.80676796S  	server_cpu=PT2M30S        	client_received=1000/1000	server_resps=1000	codes={200=1000}
                               one_big_spike[UNLIMITED_ROUND_ROBIN].txt:	success=100.0%	client_mean=PT1.115837367S 	server_cpu=PT8M3.3S       	client_received=1000/1000	server_resps=3222	codes={200=1000}
-one_endpoint_dies_on_each_server[CONCURRENCY_LIMITER_PIN_UNTIL_ERROR].txt:	success=64.0%	client_mean=PT0.603544S    	server_cpu=PT25M          	client_received=2500/2500	server_resps=2500	codes={200=1601, 500=899}
- one_endpoint_dies_on_each_server[CONCURRENCY_LIMITER_ROUND_ROBIN].txt:	success=64.5%	client_mean=PT0.6S         	server_cpu=PT25M          	client_received=2500/2500	server_resps=2500	codes={200=1612, 500=888}
+one_endpoint_dies_on_each_server[CONCURRENCY_LIMITER_PIN_UNTIL_ERROR].txt:	success=59.0%	client_mean=PT10.986749661S	server_cpu=PT25M          	client_received=2217/2500	server_resps=2217	codes={200=1474, 500=743}
```
iamdanfox, line 23: this is the _only_ scenario that gets worse, and it's because we don't have per-endpoint concurrencylimiters.

carterkozak:
When requests are retried, we add them to the back of the queue, so rather than burning all retries for one request, I think we're burning _all_ retries for _all_ requests, no?

iamdanfox:
Ah yep that makes more sense, also explains why smaller spikes went through fine (presumably because they managed to scrape through on their last request)

dialogue-core/src/main/java/com/palantir/dialogue/core/ConcurrencyLimitedChannel.java
```diff
@@ -116,6 +124,15 @@
             }
             return result;
         } else {
+            if (log.isDebugEnabled()) {
+                log.debug(
+                        "All permits already in use, request would exceed current max concurrency limit",
+                        SafeArg.of("max", limiter.getLimit()),
```
carterkozak, line 130: potentially racy, but in practice it shouldn't be super volatile

dialogue-core/src/main/java/com/palantir/dialogue/core/ConjureWindowedLimit.java
```diff
@@ -0,0 +1,100 @@
...
+        }
+
+        if (endTime > nextUpdateTime || currentSample.didDrop()) {
+            synchronized (lock) {
+                // Double check under the lock
+                if (endTime > nextUpdateTime || sample.get().didDrop()) {
+                    ImmutableSampleWindow current = sample.get();
+                    if (isWindowReady(current)) {
+                        sample.set(new ImmutableSampleWindow());
+                        nextUpdateTime = endTime
```
carterkozak, line 80:
```suggestion
                        nextUpdateTime = endTime + WINDOW_TIME;
```

dialogue-core/src/main/java/com/palantir/dialogue/core/ConjureWindowedLimit.java
```diff
@@ -0,0 +1,100 @@
...
+/**
+ * Changes made from {@link com.netflix.concurrency.limits.limit.WindowedLimit}:
+ *
+ * <p>1. Modify to reduce window size whenever a dropped sample is seen, rather than awaiting the whole window. 2.
+ * Change package and make package private. 3. Code style. 4. Inlined constants.
+ */
+@SuppressWarnings("NullAway") // TODO(dfox): deal with this
+class ConjureWindowedLimit implements Limit {
+    private static final long MIN_WINDOW_TIME = TimeUnit.SECONDS.toNanos(1);
+    private static final long MAX_WINDOW_TIME = TimeUnit.SECONDS.toNanos(1);
```
carterkozak, line 33: these two parameters can be replaced with a single `WINDOW_TIME = TimeUnit.SECONDS.toNanos(1);`

dialogue-core/src/main/java/com/palantir/dialogue/core/ConjureWindowedLimit.java
```diff
@@ -0,0 +1,100 @@
...
+    private static final long MIN_RTT_THRESHOLD = TimeUnit.MICROSECONDS.toNanos(100);
```
carterkozak, line 34: Curious why we want this check. If we fail quickly (new hardware that can fire requests twice as fast?) we may end up ignoring all success but not failures (or the other way around).

carterkozak:
If we don't use a window at all, and instead use the raw AIMD limit algorithm our results look similar to these, and the algorithm is easier for me to reason about.

Comparison: https://github.com/palantir/dialogue/compare/12f2452d96a9a3f7a58f4dfdac40498d73d6a5ad..4d96c2550413c12e20fe0e3b1b41c0c680d14534#diff-7888877ab61501c1e50055394ce9599d

slowdown_and_error_thresholds doesn't show quite as much improvement, but 1%-> 60% is a strong showing. Most values are identical with several simulations showing three to four times reduced client perceived time to response.
The ability to reduce our dependence on the external library and remove unnecessary synchronization and allocations in tight CaS loops is an added bonus.

iamdanfox:
Closing out because we landed a fix for this in https://github.com/palantir/dialogue/releases/tag/1.17.0

PR #638, opened 2020-04-13
Remove the windowed limit from the Dialogue concurrency limiter configuration

carterkozak:
This is an alternative to #635

[graphs](https://github.com/palantir/dialogue/blob/ckozak/remove_concurrency_limiter_window/simulation/src/test/resources/report.md#all_nodes_500concurrency_limiter_pin_until_error)

CJR uses a customized windowed limit which drops through to the delegate
any time an error is encountered, where the dialogue implementation may
wait 1 second before errors result in reduced limits.
We can simplify the design by disabling the sliding window which allows
the happy scenario to scale up more quickly based on real time data
without arbitrary time-based limitations.

This approach is preferable over the customized conditional window
design because it's much simpler to reason about, and allows the
system to react and self-regulate more quickly.

## After this PR
==COMMIT_MSG==
Remove the windowed limit from the Dialogue concurrency limiter configuration
==COMMIT_MSG==


ferozco (commented):
Nice! I quite prefer the simplicity of removing the windowed limit entirely and we seem to see very similar changes to the simulations as #635

simulation/src/test/resources/report.md
```diff
@@ -1,36 +1,36 @@
...
                              live_reloading[UNLIMITED_ROUND_ROBIN].txt:	success=60.2%	client_mean=PT2.84698S     	server_cpu=PT1H58M37.45S  	client_received=2500/2500	server_resps=2500	codes={200=1504, 500=996}
           one_big_spike[CONCURRENCY_LIMITER_BLACKLIST_ROUND_ROBIN].txt:	success=79.0%	client_mean=PT1.478050977S 	server_cpu=PT1M59.71393673S	client_received=1000/1000	server_resps=790	codes={200=790, Failed to make a request=210}
-                one_big_spike[CONCURRENCY_LIMITER_PIN_UNTIL_ERROR].txt:	success=100.0%	client_mean=PT3.53697893S  	server_cpu=PT2M30S        	client_received=1000/1000	server_resps=1000	codes={200=1000}
-                    one_big_spike[CONCURRENCY_LIMITER_ROUND_ROBIN].txt:	success=100.0%	client_mean=PT1.880861693S 	server_cpu=PT2M30S        	client_received=1000/1000	server_resps=1000	codes={200=1000}
+                one_big_spike[CONCURRENCY_LIMITER_PIN_UNTIL_ERROR].txt:	success=100.0%	client_mean=PT1.286733552S 	server_cpu=PT2M48.75S     	client_received=1000/1000	server_resps=1125	codes={200=1000}
```
iamdanfox, line 20: @carterkozak do you understand why the "one_big_spike" graph now looks so weird?? https://github.com/palantir/dialogue/blob/ckozak/remove_concurrency_limiter_window/simulation/src/test/resources/report.md#one_big_spikeconcurrency_limiter_pin_until_error

iamdanfox:
Ok so clearly this would unblock the last failing test I have internally, I think the risk is now that the concurrencylimiters can now ramp up so quickly that one particular high volume client might be able to overewhelm a server. This shows up in the graphs' y-axis as servers end up handling a higher number of simultaneous requests (which in reality is bounded by how many WC worker threads we have.)

We do have good metrics to be able to see what these concurrency limiters actually do in production though, so maybe we just give it a shot for now?

simulation/src/test/resources/report.md
carterkozak: I think it’s able to ramp high enough that the server sends a 503 and the pin changes. This should also happen on develop if we run the simulation longer

iamdanfox: Oooh I forgot I put an error threshold in there!

```
        int capacity = 100;
        servers = servers(
                SimulationServer.builder()
                        .serverName("node1")
                        .simulation(simulation)
                        .handler(h -> h.respond200UntilCapacity(429, capacity).responseTime(Duration.ofMillis(150)))
                        .build(),
                SimulationServer.builder()
                        .serverName("node2")
                        .simulation(simulation)
                        .handler(h -> h.respond200UntilCapacity(429, capacity).responseTime(Duration.ofMillis(150)))
                        .build());
```

Cool beans.

iamdanfox (approved):

PR #639, opened 2020-04-13
Sentence explaining pin until error decision

ferozco:
Add a sentence to state the decision around pin until error in the face of limited responses

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/PinUntilErrorChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/PinUntilErrorChannel.java
@@ -45,7 +45,9 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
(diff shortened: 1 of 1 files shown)


dialogue-core/src/main/java/com/palantir/dialogue/core/PinUntilErrorChannel.java
```diff
@@ -45,7 +45,9 @@
 import org.slf4j.LoggerFactory;
 
 /**
- * Routes all requests to one host until an error is received, then we'll move on to the next host.
+ * Routes all requests to one host until an error is received, then we'll move on to the next host. We explicitly
+ * do not consider limited responses as errors when determining whether to change hosts, trading off improved locality
+ * over overall throughput.
```
carterkozak, line 50: We could expand on this to note that events which reduce the limit also result in changing the pinned channel.

carterkozak:
:+1: Up to you whether or not we want to elaborate.

PR #641, opened 2020-04-14
Simpler AIMD implementation

carterkozak:
## Before this PR
Previously we depended on the external netflix concurrency limiters library for a very small subset of features it provided. This opened us to risk when the upstream library changes, and reduced our ability to experiment with new versions due to a shared dependency with CJR.

## After this PR
This implementation is lock-free and avoids unnecessary code that would support features we don't use.

==COMMIT_MSG==
Dialogue provides its own concurrency limiter implementation
==COMMIT_MSG==


carterkozak:
Note that the simulations are not impacted.

dialogue-core/src/main/java/com/palantir/dialogue/core/AimdConcurrencyLimiter.java
```diff
@@ -0,0 +1,110 @@
...
+        return (currentLimit, inFlightSnapshot) -> {
+            if (didDrop) {
+                currentLimit = (int) (currentLimit * BACKOFF_RATIO);
+            } else if (inFlightSnapshot * 2 >= currentLimit) {
+                currentLimit = currentLimit + 1;
+            }
+            if (currentLimit >= MAX_LIMIT) {
+                currentLimit = currentLimit / 2;
+            }
+            return Math.min(MAX_LIMIT, Math.max(MIN_LIMIT, currentLimit));
```
carterkozak: these 9 lines are the important part, everything else is scaffolding.

dialogue-core/src/main/java/com/palantir/dialogue/core/AimdConcurrencyLimiter.java
```diff
@@ -0,0 +1,110 @@
...
+            if (Responses.isQosStatus(result) || Responses.isServerError(result)) {
+                dropped();
+            } else {
+                success();
+            }
+        }
+
+        @Override
+        public void onFailure(Throwable _throwable) {
+            ignore();
```
carterkozak, line 68: This is preserved from our initial implementation, but we might want to hit `dropped()` if `_throwable` is an `IOException`.

dialogue-core/src/main/java/com/palantir/dialogue/core/AimdConcurrencyLimiter.java
```diff
@@ -0,0 +1,110 @@
...
+        }
+
+        void success() {
+            inFlight.decrementAndGet();
+            limit.accumulateAndGet(inFlightSnapshot, limitUpdaterSuccess);
+        }
+    }
+
+    private static IntBinaryOperator newLimitUpdater(boolean didDrop) {
+        return (currentLimit, inFlightSnapshot) -> {
```
ferozco: would we split this into two separate methods? seems like the only shared code is the final return statement

dialogue-core/src/main/java/com/palantir/dialogue/core/AimdConcurrencyLimiter.java
```diff
@@ -0,0 +1,110 @@
...
+    // Effectively unlimited, reduced from MAX_VALUE to prevent overflow
+    private static final int MAX_LIMIT = Integer.MAX_VALUE / 2;
+
+    private static final IntBinaryOperator limitUpdaterDidDrop = newLimitUpdater(true);
+    private static final IntBinaryOperator limitUpdaterSuccess = newLimitUpdater(false);
+
+    private final AtomicInteger limit = new AtomicInteger(INITIAL_LIMIT);
+    private final AtomicInteger inFlight = new AtomicInteger();
+
+    public Optional<Listener> acquire() {
```
carterkozak:
```suggestion
    Optional<Listener> acquire() {
```

dialogue-core/src/main/java/com/palantir/dialogue/core/AimdConcurrencyLimiter.java
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    int getLimit() {
+        return limit.get();
+    }
+
+    int getInflight() {
+        return inFlight.get();
+    }
+}
```
carterkozak, line 116: should add a toString for debugging

dialogue-core/src/main/java/com/palantir/dialogue/core/AimdConcurrencyLimiter.java
```diff
@@ -0,0 +1,110 @@
...
+                success();
+            }
+        }
+
+        @Override
+        public void onFailure(Throwable _throwable) {
+            ignore();
+        }
+
+        void ignore() {
```
ferozco, line 71: inline?

dialogue-core/src/main/java/com/palantir/dialogue/core/AimdConcurrencyLimiter.java
carterkozak: Good point, the min/max checks aren't both necessary in either code path, these can be much simpler.

dialogue-core/src/main/java/com/palantir/dialogue/core/AimdConcurrencyLimiter.java
carterkozak: I think this is a tad more self-documenting, in a future change I'd like to trigger `dropped()` on IOException where this would get pulled out again.

dialogue-core/src/main/java/com/palantir/dialogue/core/AimdConcurrencyLimiter.java
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    enum LimitUpdater implements IntBinaryOperator {
+        SUCCESS() {
+            @Override
+            public int applyAsInt(int originalLimit, int inFlightSnapshot) {
+                if (inFlightSnapshot * 2 >= originalLimit) {
+                    int updatedLimit = originalLimit + 1;
+                    if (updatedLimit >= MAX_LIMIT) {
+                        updatedLimit = Math.max(MIN_LIMIT, updatedLimit / 2);
```
carterkozak: this is odd, I'm not sure why we cut the limit in half instead of sticking with MAX_LIMIT.

carterkozak: While I wanted to avoid potential changes in this PR, this is wonky and worth cleaning up in this change for both cleaner code and better behavior.

ferozco:
Thanks carter! 👍

iamdanfox:
This is very pretty, and I'm so happy the simulations are exactly the same :D

PR #643, opened 2020-04-14
Instrument the BlockingChannelAdapter ThreadFactory

carterkozak:
## Before this PR
No idea how many threads were available or rate of creation and termination.

## After this PR
==COMMIT_MSG==
Instrument the BlockingChannelAdapter ThreadFactory
==COMMIT_MSG==

## Possible downsides?
metric $. compute is more expensive.


ferozco (approved):
Nice

PR #645, opened 2020-04-14
IOExceptions push back concurrency limits

carterkozak:
## Before this PR
Previously all exceptions were ignored, and the limiters were
only impacted by response codes.

## After this PR
This will resolve issues with handshake storms.
==COMMIT_MSG==
IOExceptions decrease concurrency limits
==COMMIT_MSG==

## Possible downsides?
Might impact us in unexpected ways.

### DIFF
--- a/changelog/@unreleased/pr-645.v2.yml
+++ b/changelog/@unreleased/pr-645.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: IOExceptions decrease concurrency limits
+  links:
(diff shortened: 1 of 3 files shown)


iamdanfox:
Ok so specifically we're expecting things like `ConnectException` and `org.apache.http.conn.HttpHostConnectException` to reduce concurrency here?

carterkozak:
Those are the most exciting and potentially impactful examples, but in general anything that suggests a degraded ability to communicate with other servers should reduce concurrency.

iamdanfox (approved):

PR #648, opened 2020-04-14
Implement naive 308 retry-other handling

carterkozak:
Dialogue considers 308 responses the same as other QoS codes in
order to approximate the expected behavior. After enough 308s
we expect calls to land on the right server.
This allows us to avoid adding cruft to node selection strategies
in support of a single use case.

==COMMIT_MSG==
Implement naive 308 retry-other handling
==COMMIT_MSG==

## Possible downsides?
It doesn't use the `Location` header.

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
@@ -208,15 +208,15 @@ final class RetryingChannel implements Channel {
         }
 
         ListenableFuture<Response> handleHttpResponse(Response response) {
-            if (Responses.isQosStatus(response)) {
+            if (Responses.isQosStatus(response) && !shouldPropagateQos(serverQoS)) {
+                // n.b. 308 retry-other responses are handled the same as other qos status codes,
+                // the provided location is not taken into account.
                 return incrementFailuresAndMaybeRetry(response, qosThrowable, retryDueToQosResponse);
             }
 
             if (response.code() == 500 && safeToRetry(endpoint.httpMethod())) {
                 return incrementFailuresAndMaybeRetry(response, serverErrorThrowable, retryDueToServerError);
             }
-
-            // TODO(dfox): if people are using 308, we probably need to support it too
 
             return Futures.immediateFuture(response);
         }
@@ -286,9 +286,7 @@ final class RetryingChannel implements Channel {
 
         private ListenableFuture<Response> wrap(ListenableFuture<Response> input) {
             ListenableFuture<Response> result = input;
-            if (!shouldPropagateQos(serverQoS)) {
-                result = Futures.transformAsync(result, this::handleHttpResponse, MoreExecutors.directExecutor());
-            }
+            result = Futures.transformAsync(result, this::handleHttpResponse, MoreExecutors.directExecutor());
             result = Futures.catchingAsync(
                     result, Throwable.class, this::handleThrowable, MoreExecutors.directExecutor());
             return result;
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
@@ -208,9 +208,7 @@ final class RetryingChannel implements Channel {
         }
 
         ListenableFuture<Response> handleHttpResponse(Response response) {
-            if (Responses.isQosStatus(response) && !shouldPropagateQos(serverQoS)) {
-                // n.b. 308 retry-other responses are handled the same as other qos status codes,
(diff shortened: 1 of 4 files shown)


dialogue-core/src/test/java/com/palantir/dialogue/core/RetryingChannelTest.java
```diff
@@ -169,7 +169,50 @@ public void retries_503s() throws Exception {
...
+        Response mockResponse = mock(Response.class);
+        when(mockResponse.code()).thenReturn(308);
+        when(channel.execute(any(), any())).thenReturn(Futures.immediateFuture(mockResponse));
+
+        Channel retryer = new RetryingChannel(
+                channel,
+                "my-channel",
+                3,
+                Duration.ZERO,
+                // Okay. This is odd. It doesn't say 308, but we consider it the same as other server QoS exceptions.
```
carterkozak: Thoughts on this?

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -286,9 +286,7 @@ private void logRetry(long backoffNanoseconds, @Nullable Throwable throwable) {
 
         private ListenableFuture<Response> wrap(ListenableFuture<Response> input) {
             ListenableFuture<Response> result = input;
-            if (!shouldPropagateQos(serverQoS)) {
```
carterkozak, line 289: subtle bug where this prevented 500s from being retried when qos propagation was enabled.

dialogue-core/src/test/java/com/palantir/dialogue/core/RetryingChannelTest.java
iamdanfox: Hmmm I think this was only added for the time-series folks, I don't think they use 308s at all.  To that end, I think I'd make it do what it says on the tin, and only propagate 429 and 503.

iamdanfox:
Nice. This should unblock timelock!

dialogue-core/src/test/java/com/palantir/dialogue/core/RetryingChannelTest.java
ferozco: Ya I think I agree with dan here

carterkozak: That's fair, the code will be a bit crufty, but it will be less astonishing for consumers.

carterkozak: Updated, I tried to make it as readable as possible.

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -208,19 +208,28 @@ private long getBackoffNanoseconds() {
...
+        private boolean isRetryableQosStatus(Response response) {
+            if (!Responses.isQosStatus(response)) {
+                // Not QoS
+                return false;
+            }
+            // ServerQos.PROPAGATE_429_and_503_TO_CALLER should only propagate too-many-requests and unavailable.
+            boolean shouldPropagate = shouldPropagate429And503(serverQoS)
+                    && (Responses.isTooManyRequests(response) || Responses.isUnavailable(response));
+            return !shouldPropagate;
+        }
```
iamdanfox, line 233: I think I'd pitch:
```java
        private boolean isRetryableQosStatus(Response response) {
            switch (serverQoS) {
                case AUTOMATIC_RETRY:
                    return Responses.isQosStatus(response);
                case PROPAGATE_429_and_503_TO_CALLER:
                    return Responses.isRetryOther(response);
            }
        }
```

carterkozak: I’m worried that if we add new QoS statuses they would be erroneously retried

iamdanfox: I see what you mean, subtracting the 429 and 503 is a little more defensive... although we have got a long way without adding any more statuses! How about

```java
        private boolean isRetryableQosStatus(Response response) {
            switch (serverQoS) {
                case AUTOMATIC_RETRY:
                    return Responses.isQosStatus(response);
                case PROPAGATE_429_and_503_TO_CALLER:
                    return Responses.isQosStatus(response)
                            && !Responses.isTooManyRequests(response)
                            && !Responses.isUnavailable(response);
            }
        }
```

carterkozak: sgtm 👍

iamdanfox (approved):

PR #650, opened 2020-04-15
Slightly more info from logging & metrics

iamdanfox:
## Before this PR

Was browsing our logs to see if anything funny stood out, seems to be mainly just a couple of log origins.

## After this PR
==COMMIT_MSG==
Slightly more info from logging & metrics
==COMMIT_MSG==

## Possible downsides?
- tag cardinality increases if we have a ton of different throwable causes. in practise, I think it'll be just one or two.

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
@@ -94,7 +94,7 @@ final class RetryingChannel implements Channel {
     private final DoubleSupplier jitter;
     private final Meter retryDueToServerError;
     private final Meter retryDueToQosResponse;
-    private final Meter retryDueToThrowable;
+    private final DialogueClientMetrics.RequestRetryBuilderReasonStage retryDueToThrowable;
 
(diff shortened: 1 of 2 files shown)


dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -144,10 +145,10 @@
                 .channelName(channelName)
                 .reason("qosResponse")
                 .build();
-        this.retryDueToThrowable = DialogueClientMetrics.of(metrics)
+        this.retryDueToThrowable = throwable -> DialogueClientMetrics.of(metrics)
```
carterkozak: DialogueClientMetrics should be created up front and reused.
```suggestion
        DialogueClientMetrics dialogueClientMetrics = DialogueClientMetrics.of(metrics);
        this.retryDueToThrowable = throwable -> dialogueClientMetrics
```

iamdanfox:
autolabeler is relentless

carterkozak (approved):

carterkozak:
:robot::popcorn:

PR #651, opened 2020-04-15
Add tests for AimdConcurrencyLimiter; rename Listener to Token

uschi2000:
## Before this PR
No tests, no docs for AimdConcurrencyLimiter

## After this PR
==COMMIT_MSG==
Tests and docs for AimdConcurrencyLimiter
==COMMIT_MSG==

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/AimdConcurrencyLimiter.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/AimdConcurrencyLimiter.java
@@ -89,14 +89,16 @@ final class AimdConcurrencyLimiter {
             }
         }
 
-        /** Indicates that the effect of this request on concurrency limits should be ignored. */
+        /**
+         * Indicates that the effect of the request corresponding to this token on concurrency limits should be ignored
+         */
         void ignore() {
             inFlight.decrementAndGet();
         }
 
         /**
-         * Indicates that this request was dropped and that the concurrency limit should be multiplicatively
-         * decreased.
+         * Indicates that the request corresponding to this token was dropped and that the concurrency limit should be
+         * multiplicatively decreased.
          */
         void dropped() {
             inFlight.decrementAndGet();
@@ -104,8 +106,8 @@ final class AimdConcurrencyLimiter {
         }
 
         /**
-         * Indicates that this request was successful and that the concurrency limit should be additively
-         * increased.
+         * Indicates that the request corresponding to this token was successful and that the concurrency limit should
+         * be additively increased.
          */
         void success() {
             inFlight.decrementAndGet();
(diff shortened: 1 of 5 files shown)


carterkozak (approved):
Thanks!

dialogue-core/src/main/java/com/palantir/dialogue/core/AimdConcurrencyLimiter.java
```diff
@@ -35,23 +39,35 @@
...
+     * eventually releasing the token by calling exactly one of the {@link Token#ignore}, {@link Token#dropped},
+     * or {@link Token#success} methods.
+     *
+     * If the token
+     * is used in the context of a {@link Response Future&lt;Response&gt;} object, then passing the {@link Token} as a
+     * {@link FutureCallback callback} to the future will invoke either {@link Token#onSuccess} or
+     * {@link Token#onFailure} which delegate to
+     * ignore/dropped/success depending on the success or failure state of the response.
+     * */
+    Optional<Token> acquire() {
```
iamdanfox: could we use Permit instead of Token? Just because BearerToken is such a different concept here

dialogue-core/src/main/java/com/palantir/dialogue/core/AimdConcurrencyLimiter.java
```diff
@@ -23,7 +23,11 @@
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.IntBinaryOperator;
 
-/** Simple lock-free additive increase multiplicative decrease concurrency limiter. */
+/**
+ * Simple lock-free additive increase multiplicative decrease concurrency limiter. Typically, a dispatching
+ * {@link com.palantir.dialogue.Request} tries to {@link #acquire} a new token and releases it when the
+ * corresponding {@link Response} is retrieved.
```
iamdanfox, line 29: Some javadoc is a great idea here! Maybe worth referring to netflix's AIMD limit as the inspiration, and that this one just avoids needing `synchronized` blocks?

carterkozak: This avoids a lot beyond synchronization, reading from the precise system clock for example. Their library is great, but has a lot of code to support extension and configurability in ways we don’t need.

uschi2000: added link

dialogue-core/src/main/java/com/palantir/dialogue/core/AimdConcurrencyLimiter.java
uschi2000: renamed to Permit

uschi2000:
is there a flake with `UserAgentChannelTest > injectsDialogueVersionAndEndpointVersion() FAILED` ?

iamdanfox:
No I think that test is a little silly - it checks for the CI environment variable, which means it behaves weirdly when run from gradle locally.

PR #661, opened 2020-04-17
PinUntilErrorChannel doesn't switch on 429

iamdanfox:
## Before this PR

In PDS-117063, a user of our internal atlas-replacement switched from c-j-r -> dialogue and saw server errors.

Looking at the `pinuntilerror.nextNode` metric, it seemed we switched channels 5 times during a supposedly transactional workflow. This meant that some requests landed on one node and others landed on a different node, which caused the second node to return a hard error.

cc @LucasIME and @jkozlowski 

## After this PR
==COMMIT_MSG==
PinUntilErrorChannel doesn't switch on 429, to unblock transactional workflows
==COMMIT_MSG==

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/PinUntilErrorChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/PinUntilErrorChannel.java
@@ -129,11 +129,12 @@ final class PinUntilErrorChannel implements LimitedChannel {
                 .map(future -> DialogueFutures.addDirectCallback(future, new FutureCallback<Response>() {
                     @Override
                     public void onSuccess(Response response) {
-                        if (Responses.isQosStatus(response) || Responses.isServerError(response)) {
+                        // We *only* switch nodes on server *errors*, NOT QoS responses to support transactional
+                        // workflows where it is important for a large number of requests to all land on the same node,
+                        // even if a couple of them get rate limited in the middle.
+                        if (Responses.isServerError(response)) {
                             OptionalInt next = incrementHostIfNecessary(currentIndex);
                             instrumentation.receivedErrorStatus(currentIndex, channel, response, next);
-                            // TODO(dfox): handle 308 See Other somehow, as we currently don't have a host -> channel
-                            // mapping
                         } else {
                             instrumentation.successfulResponse(currentIndex);
                         }
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/PinUntilErrorChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/PinUntilErrorChannel.java
@@ -129,10 +129,11 @@ final class PinUntilErrorChannel implements LimitedChannel {
                 .map(future -> DialogueFutures.addDirectCallback(future, new FutureCallback<Response>() {
                     @Override
                     public void onSuccess(Response response) {
-                        // We *only* switch nodes on server *errors*, NOT QoS responses to support transactional
+                        // We specifically don't switch  429 responses to support transactional
                         // workflows where it is important for a large number of requests to all land on the same node,
                         // even if a couple of them get rate limited in the middle.
-                        if (Responses.isServerError(response)) {
+                        if (Responses.isServerError(response)
+                                || (Responses.isQosStatus(response) && !Responses.isTooManyRequests(response))) {
                             OptionalInt next = incrementHostIfNecessary(currentIndex);
                             instrumentation.receivedErrorStatus(currentIndex, channel, response, next);
                         } else {
--- a/dialogue-core/src/test/java/com/palantir/dialogue/core/PinUntilErrorChannelTest.java
+++ b/dialogue-core/src/test/java/com/palantir/dialogue/core/PinUntilErrorChannelTest.java
@@ -35,9 +35,12 @@ import java.util.Optional;
 import java.util.Random;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.IntStream;
+import java.util.stream.Stream;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.MethodSource;
 import org.mockito.Mock;
 import org.mockito.Mockito;
 import org.mockito.junit.jupiter.MockitoExtension;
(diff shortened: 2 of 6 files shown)


simulation/src/test/resources/report.md
```diff
@@ -17,7 +17,7 @@
                    live_reloading[CONCURRENCY_LIMITER_ROUND_ROBIN].txt:	success=89.9%	client_mean=PT3.8310808S   	server_cpu=PT1H55M21.66S  	client_received=2500/2500	server_resps=2500	codes={200=2248, 500=252}
                              live_reloading[UNLIMITED_ROUND_ROBIN].txt:	success=60.2%	client_mean=PT2.84698S     	server_cpu=PT1H58M37.45S  	client_received=2500/2500	server_resps=2500	codes={200=1504, 500=996}
           one_big_spike[CONCURRENCY_LIMITER_BLACKLIST_ROUND_ROBIN].txt:	success=79.0%	client_mean=PT1.478050977S 	server_cpu=PT1M59.71393673S	client_received=1000/1000	server_resps=790	codes={200=790, Failed to make a request=210}
-                one_big_spike[CONCURRENCY_LIMITER_PIN_UNTIL_ERROR].txt:	success=100.0%	client_mean=PT1.286733552S 	server_cpu=PT2M48.75S     	client_received=1000/1000	server_resps=1125	codes={200=1000}
+                one_big_spike[CONCURRENCY_LIMITER_PIN_UNTIL_ERROR].txt:	success=100.0%	client_mean=PT1.135007332S 	server_cpu=PT2M49.65S     	client_received=1000/1000	server_resps=1131	codes={200=1000}
```
iamdanfox, line 20: As expected, the graph shows one big spike of requests (i.e. this exact use case) remains pinned to one uri. https://github.com/palantir/dialogue/blob/dfox/pin-until-error-fix/simulation/src/test/resources/report.md#one_big_spikeconcurrency_limiter_pin_until_error

dialogue-core/src/main/java/com/palantir/dialogue/core/PinUntilErrorChannel.java
```diff
@@ -129,11 +129,12 @@ LimitedChannel getCurrentChannel() {
...
                     @Override
                     public void onSuccess(Response response) {
-                        if (Responses.isQosStatus(response) || Responses.isServerError(response)) {
+                        // We *only* switch nodes on server *errors*, NOT QoS responses to support transactional
+                        // workflows where it is important for a large number of requests to all land on the same node,
+                        // even if a couple of them get rate limited in the middle.
+                        if (Responses.isServerError(response)) {
                             OptionalInt next = incrementHostIfNecessary(currentIndex);
                             instrumentation.receivedErrorStatus(currentIndex, channel, response, next);
-                            // TODO(dfox): handle 308 See Other somehow, as we currently don't have a host -> channel
```
carterkozak, line 135: This comment should have been removed, 308 is covered by isQosStatus.I think we want to specifically opt out of the too many requests status here instead of checking isServerError.

dialogue-core/src/test/java/com/palantir/dialogue/core/PinUntilErrorChannelTest.java
```diff
@@ -87,12 +90,34 @@ public void channels_are_shuffled_initially_and_successful_requests_stay_on_chan
...
     public void various_error_status_codes_cause_node_switch() {
-        testStatusCausesNodeSwitch(429);
         for (int errorStatus = 500; errorStatus < 600; errorStatus++) {
             testStatusCausesNodeSwitch(errorStatus);
         }
     }
 
+    @ParameterizedTest
+    @MethodSource("qosCodes")
+    void qos_responses_do_not_cause_node_switch(int status) {
```
carterkozak: 308 is specifically meant to force us to switch, 503 as well, I think this should verify only 429

dialogue-core/src/main/java/com/palantir/dialogue/core/PinUntilErrorChannel.java
iamdanfox: like an `if (Responses.isQoS(reposnse)) { return; }`?

dialogue-core/src/main/java/com/palantir/dialogue/core/PinUntilErrorChannel.java
```diff
@@ -129,11 +129,12 @@ LimitedChannel getCurrentChannel() {
                 .map(future -> DialogueFutures.addDirectCallback(future, new FutureCallback<Response>() {
                     @Override
                     public void onSuccess(Response response) {
-                        if (Responses.isQosStatus(response) || Responses.isServerError(response)) {
+                        // We *only* switch nodes on server *errors*, NOT QoS responses to support transactional
+                        // workflows where it is important for a large number of requests to all land on the same node,
+                        // even if a couple of them get rate limited in the middle.
+                        if (Responses.isServerError(response)) {
```
ferozco: I think it makes sense to remain on the same node in the face of a 429, but it seems wrong to stay pinned after a 503

dialogue-core/src/main/java/com/palantir/dialogue/core/PinUntilErrorChannel.java
carterkozak: I think we want to add “&& !isTooManyRequests” to the original conditional

dialogue-core/src/test/java/com/palantir/dialogue/core/PinUntilErrorChannelTest.java
iamdanfox: good point - was a bit hasty!

simulation/src/test/resources/report.md
carterkozak: We may want to update this simulation to respond 429 instead of 503

iamdanfox: It's intended to be representative of this exact workflow, so it responds 429 above some threshold:
```java
 public void one_big_spike() {
        int capacity = 100;
        servers = servers(
                SimulationServer.builder()
                        .serverName("node1")
                        .simulation(simulation)
                        .handler(h -> h.respond200UntilCapacity(429, capacity).responseTime(Duration.ofMillis(150)))
                        .build(),
                SimulationServer.builder()
                        .serverName("node2")
                        .simulation(simulation)
                        .handler(h -> h.respond200UntilCapacity(429, capacity).responseTime(Duration.ofMillis(150)))
                        .build());
```

carterkozak (approved):

PR #662, opened 2020-04-20
Rename `dialogue.client.response` metric to `client.response`

carterkozak:
This matches the metric reported by CJR to avoid unnecessary churn
on dashboards and alerts.

## After this PR
==COMMIT_MSG==
Rename `dialogue.client.response` metric to `client.response` to match CJR and avoid unnecessary churn on dashboards and alerts.
==COMMIT_MSG==

## Possible downsides?
breaks dialogue boards.

### DIFF
--- a/dialogue-core/src/main/metrics/dialogue-core-metrics.yml
+++ b/dialogue-core/src/main/metrics/dialogue-core-metrics.yml
@@ -10,6 +10,10 @@ namespaces:
         type: timer
         tags: [channel-name, service-name]
         docs: Request time, note that this does not include time spent reading the response body.
+      response.error:
+        type: meter
+        tags: [channel-name, service-name, reason]
+        docs: Rate of errors received by reason and service-name. Currently only errors with reason `IOException` are reported.
       deprecations:
         type: meter
         tags: [service-name]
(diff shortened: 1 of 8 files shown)


carterkozak:
Not going to rev the major version for this one since we're relatively early in rollout.

dialogue-core/src/main/metrics/dialogue-core-metrics.yml
```diff
@@ -2,13 +2,17 @@ options:
   javaPackage: com.palantir.dialogue.core
   javaVisibility: packagePrivate
 namespaces:
-  dialogue.client:
-    docs: Dialogue client response metrics.
+  client:
```
iamdanfox, line 5: linking to c-j-r's metric schema: https://github.com/palantir/conjure-java-runtime/blob/8de52360deb5a05cd9ff68c11e6655767cdd9cb2/okhttp-clients/src/main/metrics/okhttp-metrics.yml#L34-L48

wanna put `deprecations` under here too?

carterkozak: Good call!

iamdanfox:
For the last one c-j-r produces, we don't really produce an equivalent metric do we?

```yml
response.error:
        type: meter
        tags: [reason, service-name]
        docs: Rate of errors received by reason and service-name. Currently only errors with reason `IOException` are reported.
```

carterkozak:
Not yet, but we can add it

iamdanfox (approved):
This is gonna wreck my DD dashboard 😭 but probably a worthwhile sacrifice to avoid wrecking everyone else's.

PR #664, opened 2020-04-21
The ZeroUriChannel provides additional context for failures

carterkozak:
==COMMIT_MSG==
The ZeroUriChannel provides additional context for failures
==COMMIT_MSG==

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/ZeroUriChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/ZeroUriChannel.java
@@ -21,16 +21,26 @@ import com.google.common.util.concurrent.ListenableFuture;
 import com.palantir.dialogue.Endpoint;
 import com.palantir.dialogue.Request;
 import com.palantir.dialogue.Response;
+import com.palantir.logsafe.Preconditions;
+import com.palantir.logsafe.SafeArg;
 import com.palantir.logsafe.exceptions.SafeIllegalStateException;
 import java.util.Optional;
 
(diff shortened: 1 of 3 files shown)


ferozco (approved):

dialogue-core/src/main/java/com/palantir/dialogue/core/ZeroUriChannel.java
```diff
@@ -21,16 +21,26 @@
...
 import java.util.Optional;
 
 /** When we literally have zero URIs, no request can get out the door. */
-enum ZeroUriChannel implements LimitedChannel {
-    INSTANCE;
+final class ZeroUriChannel implements LimitedChannel {
+
+    private final String channelName;
+
+    ZeroUriChannel(String channelName) {
```
iamdanfox, line 34: I'm a little worried that people might not realize when they construct a DialogueChannel that the channel name must be safe to log (we don't even mention it in javadoc)...

would it be crazy to make the builder:

```java
        public Builder channelName(SafeArg<String> value) {
            this.channelName = value;
            return this;
        }
```

carterkozak: I'd prefer to use the `@Safe` annotation, though it doesn't provide validation. SafeArg for values feels off because it's not clear that the key is ignored.

carterkozak: Added #665

PR #665, opened 2020-04-21
Add documentation/annotations describing log-safety

carterkozak:
==COMMIT_MSG==
Add documentation/annotations describing log-safety of channel names
==COMMIT_MSG==


iamdanfox (approved):
Wish we had some kind of static analysis for this!

carterkozak:
I've thought about it but I'm really not sure what we would validate. The idea behind the annotation is that humans are required to distinguish safe and unsafe data.

PR #668, opened 2020-04-23
Apache HttpClient is configured not to normalize URIs

carterkozak:
## Before this PR
Previously this resulted in failures to match endpoints when
empty path parameters were applied to non-final path segments.

## After this PR
==COMMIT_MSG==
Apache HttpClient is configured not to normalize URIs
==COMMIT_MSG==

## Possible downsides?
We may be relying on normalization in some way that we aren't aware of.

### DIFF
--- a/dialogue-test-common/src/main/java/com/palantir/dialogue/AbstractChannelTest.java
+++ b/dialogue-test-common/src/main/java/com/palantir/dialogue/AbstractChannelTest.java
@@ -370,6 +370,29 @@ public abstract class AbstractChannelTest {
     }
 
     @Test
+    public void supports_empty_path_parameter() throws InterruptedException, ExecutionException {
+        endpoint.method = HttpMethod.GET;
+        endpoint.renderPath =
+                (params, url) -> url.pathSegment("a").pathSegment("").pathSegment("b");
+        ListenableFuture<Response> result = channel.execute(endpoint, request);
+        RecordedRequest recorded = server.takeRequest();
+        assertThat(recorded.getMethod()).isEqualTo("GET");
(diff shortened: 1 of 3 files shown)


iamdanfox:
Think it needs a ./gradlew test -Drecreate=true to fix the traces_on_retries.log

dialogue-test-common/src/main/java/com/palantir/dialogue/AbstractChannelTest.java
```diff
@@ -369,6 +369,29 @@ public void supportsGzipEncodedResponse() throws Exception {
...
+    @Test
+    public void emptyTrailingPathParameterResultsInTrailingSlash() throws InterruptedException, ExecutionException {
+        endpoint.method = HttpMethod.GET;
+        endpoint.renderPath = (params, url) -> url.pathSegment("foo").pathSegment("");
+        ListenableFuture<Response> result = channel.execute(endpoint, request);
+        RecordedRequest recorded = server.takeRequest();
+        assertThat(recorded.getMethod()).isEqualTo("GET");
+        assertThat(recorded.getPath()).isEqualTo("/foo/");
+        assertThat(result.get().code()).isEqualTo(200);
+    }
```
iamdanfox, line 393: nice that these get run for all our other implementations too :)

iamdanfox (approved):

carterkozak:
So the tracing test results in the `actual` category are just missing the top-level span! It should we expect `Dialogue: request service#endpoint initial` at the top level where really we want `Dialogue: request service#endpoint` (non-initial parent span). Not sure how that could happen.

PR #674, opened 2020-04-24
[test-only] make 'fast_503s_then_revert' simulation fail faster

iamdanfox:
As requested, this changes our "fail fast" simulation scenario so that failures come back _really_ fast.

It is quite expensive to simulate a cumulative 1h30ish of server time for all three strategies, so this test does end up taking ~20s or so.

# [GRAPHS](https://github.com/palantir/dialogue/blob/dfox-patch-19373/simulation/src/test/resources/report.md#fast_503s_then_revertconcurrency_limiter_round_robin)


simulation/src/test/java/com/palantir/dialogue/core/SimulationTest.java
```diff
@@ -198,26 +198,30 @@ public void slow_503s_then_revert() {
...
-                        .until(Duration.ofSeconds(3), "fast 500s")
-                        .handler(h -> h.response(500).linearResponseTime(Duration.ofMillis(10), capacity))
+                        .handler(h -> h.response(200).responseTime(Duration.ofMillis(120)))
+                        .build(),
+                SimulationServer.builder()
+                        .serverName("fast_503s_then_revert")
+                        .simulation(simulation)
+                        .handler(h -> h.response(200).responseTime(Duration.ofMillis(120)))
+                        .until(Duration.ofSeconds(3), "fast 503s")
+                        .handler(h -> h.response(503).responseTime(Duration.ofNanos(10)))
```
carterkozak, line 213: I think we want one sever that always 503s.

iamdanfox: I actually think seeing the transitions are also very useful, and representative of the real world. You get a solid 7 seconds of pure brokenness in the middle

iamdanfox:
@carterkozak so here's the current behaviour on develop: https://github.com/palantir/dialogue/blob/dfox-patch-19373/simulation/src/test/resources/report.md#fast_503s_then_revertconcurrency_limiter_round_robin.

We get 100% success for all strategies, with client-perceived responsetime of 120 and 165ms.

simulation/src/test/java/com/palantir/dialogue/core/SimulationTest.java
carterkozak: That's fair, good point :+1:

carterkozak:
> We get 100% success for all strategies, with client-perceived responsetime of 120 and 165ms.

I wonder if reducing nodes from 3 to 2 better captures the failure scenario we're looking at...

iamdanfox:
I can dial up the simulation length, then some of our retries will definitely get exhausted.

carterkozak (approved):

simulation/src/test/resources/report.md
```diff
@@ -10,9 +10,9 @@
...
-        fast_500s_then_revert[CONCURRENCY_LIMITER_PIN_UNTIL_ERROR].txt:	success=99.9%	client_mean=PT0.078624177S 	server_cpu=PT4M54.840666665S	client_received=3750/3750	server_resps=3750	codes={200=3746, 500=4}
-            fast_500s_then_revert[CONCURRENCY_LIMITER_ROUND_ROBIN].txt:	success=78.5%	client_mean=PT0.056930933S 	server_cpu=PT3M33.490999698S	client_received=3750/3750	server_resps=3750	codes={200=2944, 500=806}
-                      fast_500s_then_revert[UNLIMITED_ROUND_ROBIN].txt:	success=76.4%	client_mean=PT0.055505822S 	server_cpu=PT3M28.146833043S	client_received=3750/3750	server_resps=3750	codes={200=2866, 500=884}
+        fast_503s_then_revert[CONCURRENCY_LIMITER_PIN_UNTIL_ERROR].txt:	success=100.0%	client_mean=PT0.120012009S 	server_cpu=PT1H30M0.00000004S	client_received=45000/45000	server_resps=45004	codes={200=45000}
+            fast_503s_then_revert[CONCURRENCY_LIMITER_ROUND_ROBIN].txt:	success=98.1%	client_mean=PT0.277457972S 	server_cpu=PT1H28M17.7602765S	client_received=45000/45000	server_resps=71798	codes={200=44148, 503=852}
```
iamdanfox, line 14: this 98.1% success rate shows how some requests hit the unlucky 4 coin flips in a row.

PR #677, opened 2020-04-27
Changelog for new Balanced impl

iamdanfox:
I forgot to put a changelog in https://github.com/palantir/dialogue/pull/373, and @dtobin urgently wants a release


iamdanfox:
Power merging because no java changes

PR #678, opened 2020-04-27
Only count IOExceptions toward failures in the balanced strategy

carterkozak:
Missed this in review:
==COMMIT_MSG==
Only count IOExceptions toward failures in the balanced strategy
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-678.v2.yml
+++ b/changelog/@unreleased/pr-678.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Only count IOExceptions toward failures in the balanced strategy
+  links:
+  - https://github.com/palantir/dialogue/pull/678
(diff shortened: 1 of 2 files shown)


iamdanfox:
I'm not actually convinced this is better. What if there's actually an IllegalStateException thrown by some inner channel (yes programmer error) but if only happens on one node then it's kinda nice we could route around it??

carterkozak:
I'm not sure how that could happen in a way that indicates status of the connection to the target host.

iamdanfox:
But if the outcome is that we route around that brokenness (irrespective of whether it was a connection problem or just a bug in one of channels), isn't that good?

carterkozak:
I'm not confident that the failure would have anything to do with the channel, but rather the request and endpoint implementations passed in. Perhaps that's not true though.

iamdanfox:
I can imagine possible cases, e.g. 1 url has 5 trailing slashes on it for some reason, but others don't?

carterkozak:
That's a good point :-)

PR #681, opened 2020-04-27
Logs & metrics for the Balanced node selection strategy

iamdanfox:
## Before this PR

We landed new code in https://github.com/palantir/dialogue/pull/373 and I'd like to see graphs in DD of how it's behaving.

## After this PR
==COMMIT_MSG==
Metrics for the Balanced node selection strategy
==COMMIT_MSG==

## Possible downsides?
- every PR we add increases the DD cost...


dialogue-core/src/main/java/com/palantir/dialogue/core/WeakReducingGauge.java
```diff
@@ -48,19 +55,20 @@ public synchronized void add(T sourceElement) {
...
 
-    public static <T> WeakSummingGauge<T> getOrCreate(
-            ToIntFunction<T> toIntFunction,
-            T initialValue,
+    public static <T> WeakReducingGauge<T> getOrCreate(
             TaggedMetricRegistry taggedMetricRegistry,
-            MetricName metricName) {
+            MetricName metricName,
+            ToLongFunction<T> toLongFunc,
+            Function<LongStream, Number> reducingFunction,
```
carterkozak: I think this will be a bit cleaner if we leave streams out of the API, much like the standard collectors. Standard accumulator/combiner model is nice to work with.

iamdanfox: I dunno it's actually really concise, e.g. `LongStream::sum` or `LongStream::average`? It also lets me stick em all in an array so that I can log something if there are multiple.

dialogue-core/src/main/java/com/palantir/dialogue/core/WeakReducingGauge.java
```diff
@@ -32,14 +35,18 @@
  * the final summary integer.
  */
 @ThreadSafe
-final class WeakSummingGauge<T> implements Gauge<Integer> {
-    private final ToIntFunction<T> gaugeFunction;
+public final class WeakReducingGauge<T> implements Gauge<Number> {
```
iamdanfox: Had to make this public because it's needed in `com.palantir.dialogue.{hc4, core}`. Don't really think it's worth moving to tritium, but I also don't really want people to rely on it :/

carterkozak: We can add `internal` to either the package or class name ¯\_(ツ)_/¯

dialogue-core/src/main/java/com/palantir/dialogue/core/WeakReducingGauge.java
carterkozak: No strong opinion since this isn't a reusable library :-)

dialogue-core/src/main/metrics/dialogue-core-metrics.yml
```diff
@@ -72,3 +72,22 @@ namespaces:
...
+  dialogue.roundrobin:
+    docs: Instrumentation for the ROUND_ROBIN node selection strategy (currently implemented by BalancedChannel).
+    metrics:
+      success:
+        type: meter
+        tags: [channel-name, hostIndex]
+        docs: Meter of the requests that were successfully made, tagged by the index of the host. (Note if there are >10
+          nodes this metric will not be recorded).
+
+  dialogue.balanced:
```
carterkozak, line 85: Thoughts on including score in the name for clarity?

iamdanfox: yep the metric comes out as `dialogue.balanced.score` (defined a bit below), this is just the namespace

carterkozak: Oh geez... I can't read.

dialogue-core/src/main/java/com/palantir/dialogue/core/BalancedNodeSelectionStrategyChannel.java
```diff
@@ -161,25 +178,130 @@ public String toString() {
...
+                    longStream -> {
+                        long[] longs = longStream.toArray();
+                        if (longs.length > 1) {
+                            log.warn(
+                                    "Multiple objects contribute to the same gauge, taking the average "
+                                            + "(beware this may be misleading)",
+                                    SafeArg.of("metricName", metricName),
+                                    SafeArg.of("values", Arrays.toString(longs)));
+                        }
+                        return Arrays.stream(longs).average().orElse(0);
```
iamdanfox, line 228: thoughts on `.orElse(-1)` ? Just occurred to me that we could in theory get into a state where these have all been GC'd? and then a score of 0 might seem really desirable?

iamdanfox:
@carterkozak I'm wondering whether it would actually be more useful to write these things to a histogram, so that when the metric-logger samples every 30 seconds, the number would actually represent the score over the entire time period, rather than just at that instant?

It would also let me throw away the weak summing gauge thing, because if multiple channels all updated the same histogram then I think the final number would just represent both?

iamdanfox:
I think we only get `.max` whitelisted though, so might actually need to histogram `1 / (1 + score)`?

carterkozak:
@iamdanfox I'm not sure how that would work, since the histogram must be updated when some sort of event occurs. Do we update the histogram after every `get` and `update` call? Both the histogram and the score have a decaying component, do we need to differentiate the two?

iamdanfox:
mmm the decaying view of a decaying value might be a little confusing for folks... maybe we just run with this initially and if the graphs are garbage we try something else?

carterkozak:
:+1:

PR #682, opened 2020-04-28
Balanced: track RunningTimers instead of just inflight AtomicInteger

iamdanfox:
## Before this PR

The heuristic in BalancedNodeSelectionStrategyChannel is already _pretty good_, but I think it still makes silly decisions when servers suddenly get slower. This can be seen intuitively in the following two simulation cases:

(1) In the `black_hole` scenario (https://github.com/palantir/dialogue/issues/615) below, it keeps chucking requests into the black hole (the yellow line) long after the server has gone bad.

<image width=400 src="https://media.githubusercontent.com/media/palantir/dialogue/9061e48ae0c0b710fb829c75bd93c36d7445e3be/simulation/src/test/resources/black_hole[CONCURRENCY_LIMITER_ROUND_ROBIN].png" />

(2) In the `drastic_slowdown` scenario, when the yellow server gets slow, it doesn't actually achieve the _balance_ that we wanted, as it ends up tying up 55 worker threads on the slow yellow server, while the fast server is humming along with a steady ~15 worker threads.

<image width=400 src="https://media.githubusercontent.com/media/palantir/dialogue/9061e48ae0c0b710fb829c75bd93c36d7445e3be/simulation/src/test/resources/drastic_slowdown%5BCONCURRENCY_LIMITER_ROUND_ROBIN%5D.png" />

I think we can do better, and it's not even gonna cost more 😄

## After this PR
==COMMIT_MSG==
The 'Balanced' implementation will now more quickly react a single host starts responding slowly, and will more aggressively send traffic to other nodes.
==COMMIT_MSG==

The gist of this change is we can make smart decisions when some unfinished requests seem slow, _before_ we actually receive a single bad response.

# [GRAPHS](https://github.com/palantir/dialogue/blob/dfox/running-timers/simulation/src/test/resources/report.md)

Check out the dramatic improvement in:

- [black_hole](https://github.com/palantir/dialogue/blob/dfox/running-timers/simulation/src/test/resources/report.md#black_holeconcurrency_limiter_round_robin), which now delivers a 95% client-perceived success rate (notice the _blue_ server actually hits the same peak utilization of serving ~120 concurrent requests, but we back off a lot faster from yellow)
- [drastic_slowdown](https://github.com/palantir/dialogue/blob/dfox/running-timers/simulation/src/test/resources/report.md#drastic_slowdownconcurrency_limiter_round_robin) where we've _halved_ the client-perceived response time.

The fun part is that the book-keeping is actually quite cheap. I'm using one `long` and one `int` to represent all of the running timers (TODO: make these threadsafe), wherease previously we just had a single AtomicInteger representing the number of inflight requests. No concurrent skip list sets thankfully.

## Possible downsides?
- I had to move the FutureCallback, so we now allocate it per-request instead of upfront. This seems pretty minor given that pretty much all our other channels also allocate a FutureCallback.


dialogue-core/src/main/java/com/palantir/dialogue/core/RunningTimersLongImpl.java
```diff
@@ -0,0 +1,71 @@
...
+
+package com.palantir.dialogue.core;
+
+import com.github.benmanes.caffeine.cache.Ticker;
+
+final class RunningTimersLongImpl implements RunningTimers {
+
+    private final Ticker ticker;
+
+    // TODO(dfox): binpack both of these into an AtomicLong
```
iamdanfox, line 25: if we like the overall idea of this PR, then I'll put together a threadsafe implementation of this and actually write some tests to check it has sufficient capacity

simulation/src/test/resources/report.md
```diff
@@ -2,35 +2,35 @@
...
+                     simplest_possible_case[UNLIMITED_ROUND_ROBIN].txt:	success=100.0%	client_mean=PT0.784590909S 	server_cpu=PT2H52M36.6S   	client_received=13200/13200	server_resps=13200	codes={200=13200}
         slow_503s_then_revert[CONCURRENCY_LIMITER_PIN_UNTIL_ERROR].txt:	success=100.0%	client_mean=PT0.137438962S 	server_cpu=PT6M39.829999979S	client_received=3000/3000	server_resps=3077	codes={200=3000}
-            slow_503s_then_revert[CONCURRENCY_LIMITER_ROUND_ROBIN].txt:	success=100.0%	client_mean=PT0.093312988S 	server_cpu=PT4M35.008666652S	client_received=3000/3000	server_resps=3039	codes={200=3000}
-                      slow_503s_then_revert[UNLIMITED_ROUND_ROBIN].txt:	success=100.0%	client_mean=PT0.093312988S 	server_cpu=PT4M35.008666652S	client_received=3000/3000	server_resps=3039	codes={200=3000}
+            slow_503s_then_revert[CONCURRENCY_LIMITER_ROUND_ROBIN].txt:	success=100.0%	client_mean=PT0.081574996S 	server_cpu=PT4M2.768666661S	client_received=3000/3000	server_resps=3017	codes={200=3000}
```
iamdanfox, line 29: these graphs are also kinda fun to look at, as the kinda illustrate how when a server's performance suddenly changes, this 'running timers' heuristic stops us charging off a cliff at our old request rate (you can see this in the big yellow spike on the left).

https://github.com/palantir/dialogue/blob/dfox/running-timers/simulation/src/test/resources/report.md#slow_503s_then_revertconcurrency_limiter_round_robin

dialogue-core/src/main/java/com/palantir/dialogue/core/RunningTimersLongImpl.java
```diff
@@ -0,0 +1,71 @@
...
+        long now = ticker.read();
+        fromBeginningOfTime += now;
+        count += 1;
+        return new Timer(now);
+    }
+
+    @Override
+    public long getRunningNanos() {
+        long now = ticker.read();
+        return (now * count) - fromBeginningOfTime;
```
carterkozak, line 45: I'm not sure this can work outside of tests which assume zero nanoTime start. In practice nanoTime is an arbitrary long value that may wrap.

iamdanfox: Yeah this is very much just an MVP, I think to make it real I'd just decrease the precision this stores (there's no need to keep track of nanoseconds), and probably try System.currentTimeMillis() and also add in some rescaling when we get too close to filling up a long.

PR #683, opened 2020-04-28
Balanced node selection strategy uses Power of 2 choices (P2C) instead of full sorting

iamdanfox:
## Before this PR

Currently, every request that goes through the BalancedNodeSelectionStrategy involves sorting the entire list of nodes (and thereby computing the score for each of them too, which includes a bunch of atomic reads).

I saw this P2C trick mentioned in twitter's daperture blog post: https://twitter.github.io/finagle/guide/Clients.html#p2c-least-loaded, where they found it was an easy way to avoid spending CPU sorting big lists of hosts.

In general it seems nice to make things O(constant) instead of O(nodes) where possible :)

## After this PR
==COMMIT_MSG==
The Round Robin node selection strategy is now slightly more CPU efficient because it employs a 'power of 2 choices' comparison rather than sorting the entire list of nodes.
==COMMIT_MSG==

## Possible downsides?
- if two nodes are borked but a third is just fine, this optimization would not consider the third one and just pick one of the borked nodes.  
- streams were very readable, but the new imperative code is more confusing :(
- the simulations don't really give us any meaningful signal because most of them just use two nodes anyway 
- in general we have quite a small number of nodes, so maybe optimizing away sorting CPU cycles is pointless?


iamdanfox:
Closing out so we don't need to worry about cases where multiple upstreams are broken.  Keeps things simpler for users

PR #684, opened 2020-04-28
Balanced channels use a multiplier on input for less jagged decay

carterkozak:
Failure-based penalties are smoothed out and no longer multiples
of ten.

==COMMIT_MSG==
Balanced channels use a multiplier on input for less jagged decay
==COMMIT_MSG==


iamdanfox:
Seems reasonable, although I can't actually spot the jaggedness in any of the graphs... I did quite like having the entire heuristic contained in one function

carterkozak:
I don't think our simulations capture the case where fewer than ten but more than one concurrent requests are executing, and one node dies, which is fairly reasonable as simulations already take a while to run, but it means we don't capture the case where this is most meaningful.

ferozco:
👍

PR #688, opened 2020-04-28
Client respect `Node-Selection-Strategy` response header

ferozco:
## Before this PR
Clients would declare (or default to) a static node selection strategy. Clients would have to guess which node selection strategy was the right one to use for a given service, resulting in a leaky abstraction since they would have to know implementation details of the service they were calling to choose the right one.

## After this PR
==COMMIT_MSG==
Clients respect `Node-Selection-Strategy` response header
==COMMIT_MSG==

Servers are now able to instruct clients on which strategy to use by adding `Node-Selection-Strategy` response header. The header is structured similarly to `Content-Type` as a `;` delimited list of strategies, ordered by preference. The current supported strategies are:
- PIN_UNTIL_ERROR
- PIN_UNTIL_ERROR_WITHOUT_RESHUFFLE
- BALANCED

The expectation is that for a given server a single node selection strategy will be defined. 

The hope is that by consolidating node selection strategy selection into the server we will see improved perf across the fleet and clients will have to think less about the services they communicate with.

Currently the `NodeSelectionStrategySelector` decides which node selection strategy to use, basically the first known strategy from the most recent response.

## Possible downsides?
This raises some questions about how we handle client creation internally, since we currently create separate DialogueChannel's (and underlying Apache client) if two logic services have different static client configuration (including node selection strategy) but they will converge to the same strategy if the server responds with `Node-Selection-Strategy` header.

There is also currently no way for the client to override the server suggested strategy, some may view that as a feature rather than a bug however 🤷‍♂️


simulation/src/test/resources/report.md
```diff
@@ -13,7 +13,7 @@
         fast_503s_then_revert[CONCURRENCY_LIMITER_PIN_UNTIL_ERROR].txt:	success=100.0%	client_mean=PT0.120012009S 	server_cpu=PT1H30M0.00000004S	client_received=45000/45000	server_resps=45004	codes={200=45000}
             fast_503s_then_revert[CONCURRENCY_LIMITER_ROUND_ROBIN].txt:	success=100.0%	client_mean=PT0.120126613S 	server_cpu=PT1H30M0.0000004S	client_received=45000/45000	server_resps=45040	codes={200=45000}
                       fast_503s_then_revert[UNLIMITED_ROUND_ROBIN].txt:	success=100.0%	client_mean=PT0.120126613S 	server_cpu=PT1H30M0.0000004S	client_received=45000/45000	server_resps=45040	codes={200=45000}
-               live_reloading[CONCURRENCY_LIMITER_PIN_UNTIL_ERROR].txt:	success=97.2%	client_mean=PT4.8253936S   	server_cpu=PT2H46.79S     	client_received=2500/2500	server_resps=2500	codes={200=2430, 500=70}
+               live_reloading[CONCURRENCY_LIMITER_PIN_UNTIL_ERROR].txt:	success=94.0%	client_mean=PT5.2613496S   	server_cpu=PT1H56M35.39S  	client_received=2500/2500	server_resps=2500	codes={200=2350, 500=150}
```
ferozco: Not sure why this simulation was impacted

carterkozak: Looks like we're not preserving pinned state on reload: https://github.com/palantir/dialogue/blob/fo/server-side-node-selection/simulation/src/test/resources/report.md#live_reloadingconcurrency_limiter_pin_until_error

ferozco: 🤦 looks like the node selection wrapper is wrecking our equality check

dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueNodeSelectionStrategy.java
```diff
@@ -0,0 +1,57 @@
...
+import java.util.List;
+
+public enum DialogueNodeSelectionStrategy {
+    PIN_UNTIL_ERROR,
+    PIN_UNTIL_ERROR_WITHOUT_RESHUFFLE,
+    BALANCED,
+    UNKNOWN;
+
+    static List<DialogueNodeSelectionStrategy> fromHeader(String header) {
+        return Splitter.on(",").splitToList(header).stream()
```
carterkozak: We can extract out the splitter to a static field, and include trimming whitespace: `Splitter.on(',').trimResults().omitEmptyStrings()`

dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueNodeSelectionStrategy.java
```diff
@@ -0,0 +1,57 @@
...
+                .map(DialogueNodeSelectionStrategy::safeValueOf)
+                .collect(ImmutableList.toImmutableList());
```
carterkozak, line 45: string -> list -> stream -> list.

Should we filter out unknowns here?

There's a [splitToStream](https://guava.dev/releases/snapshot-jre/api/docs/com/google/common/base/Splitter.html#splitToStream-java.lang.CharSequence-) function, or we could write an old fashion for loop ;-)

dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueNodeSelectionStrategy.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    private static DialogueNodeSelectionStrategy safeValueOf(String value) {
+        try {
+            return valueOf(value.trim().toUpperCase());
+        } catch (Exception e) {
+            return UNKNOWN;
```
carterkozak: nit: prefer to avoid exceptions for control flow, can we switch on value?

dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueNodeSelectionStrategy.java
```diff
@@ -0,0 +1,57 @@
...
+
+    static List<DialogueNodeSelectionStrategy> fromHeader(String header) {
+        return Splitter.on(",").splitToList(header).stream()
+                .map(DialogueNodeSelectionStrategy::safeValueOf)
+                .collect(ImmutableList.toImmutableList());
+    }
+
+    private static DialogueNodeSelectionStrategy safeValueOf(String value) {
+        try {
+            return valueOf(value.trim().toUpperCase());
```
carterkozak: Do we want to support mixed case strategies?

dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueNodeSelectionStrategy.java
carterkozak: Might as well info/debug log unknown string values?

dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueNodeSelectionStrategy.java
ferozco: The Http spec only required header names be case insensitive so its up to each implementer to decide how it interprets the values. For simplicity I figured it was fine to be completely case insensitive.

dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueNodeSelectionStrategy.java
ferozco: Ahh ya, I think our guava isn't modern enough to get splitToStream. I'll upgrade and update

dialogue-core/src/main/java/com/palantir/dialogue/core/NodeSelectionStrategyChannel.java
```diff
@@ -0,0 +1,191 @@
...
+        private final FutureCallback<Response> callback;
+
+        WrapperChannel(LimitedChannel delegate) {
+            this.delegate = delegate;
+            this.callback = new FutureCallback<Response>() {
+                @Override
+                public void onSuccess(Response result) {
+                    result.getFirstHeader(NODE_SELECTION_HEADER)
+                            .ifPresent(strategy -> updateRequestedStrategies(
+                                    delegate, DialogueNodeSelectionStrategy.fromHeader(strategy)));
```
carterkozak: Can we extract the reused lambdas? The magbeExecute result below can avoid the response-> addDirectCallback allocation as well.

The callback and transformer registrations showed up pretty heavily in the profiler when I took a quick look at our simulation performance, long term we might want to do something about that.

dialogue-core/src/main/java/com/palantir/dialogue/core/DefaultNodeSelectionStrategySelector.java
```diff
@@ -0,0 +1,97 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+@SuppressWarnings("NullAway")
+public final class DefaultNodeSelectionStrategySelector implements NodeSelectionStrategySelector {
```
carterkozak: It would be helpful to mark a meter each time a strategy change occurs, possibly including the strategy we're moving to.

dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueChannel.java
```diff
@@ -129,14 +130,7 @@ public void updateUris(Collection<String> uris) {
...
 
-        nodeSelectionStrategy.getAndUpdate(previous -> getUpdatedNodeSelectionStrategy(
-                previous,
-                clientConfiguration,
-                ImmutableList.copyOf(limitedChannelByUri.values()),
-                random,
-                ticker,
-                channelName));
-
+        nodeSelectionStrategy.updateChannels(ImmutableList.copyOf(limitedChannelByUri.values()));
```
carterkozak: We could keep things simpler by putting our complete trust in the most recent response we've received, that way the nodeSelectionStrategy wouldn't need to wrap individual uri channels or be notified of uri updates.

No real proposal here, just weighing implementing this in a more naive way to begin with.

dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueChannel.java
```diff
@@ -150,57 +144,12 @@ private LimitedChannel createLimitedChannel(String uri, int uriIndex) {
...
-                        random,
-                        channelName);
-            case ROUND_ROBIN:
-                // When people ask for 'ROUND_ROBIN', they usually just want something to load balance better.
-                // We used to have a naive RoundRobinChannel, then tried RandomSelection and now use this heuristic:
-                return new BalancedNodeSelectionStrategyChannel(
-                        channels, random, tick, config.taggedMetricRegistry(), channelName);
-        }
-        throw new SafeRuntimeException(
-                "Unknown NodeSelectionStrategy", SafeArg.of("unknown", config.nodeSelectionStrategy()));
```
carterkozak: DialogueChannel is much easier to follow without the node selection logic :+1:

dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueChannel.java
ferozco: Ya I had thought of that originally, but it would lead to pretty bad behaviour during rolling upgrades with RR/Balanced. I think the extra bit of complexity is worth handling the case

dialogue-core/src/main/java/com/palantir/dialogue/core/NodeSelectionStrategyChannel.java
```diff
@@ -0,0 +1,191 @@
...
+    public void updateChannels(ImmutableList<LimitedChannel> newChannels) {
+        DialogueNodeSelectionStrategy updatedStrategy = strategySelector.setActiveChannels(newChannels);
+        nodeChannels.set(newChannels);
+        nodeSelectionStrategy.getAndUpdate(previousChannel -> getUpdatedNodeSelectionStrategy(
+                previousChannel.channel(), newChannels, updatedStrategy, metrics, random, tick, channelName));
+    }
+
+    private void updateRequestedStrategies(LimitedChannel channel, List<DialogueNodeSelectionStrategy> strategies) {
+        DialogueNodeSelectionStrategy updatedStrategy = strategySelector.updateChannelStrategy(channel, strategies);
+        nodeSelectionStrategy.getAndUpdate(currentStrategy -> {
```
carterkozak: cas operations are relatively expensive, I think we want to do a quick check to avoid the cas in the expected case where the strategy doesn't change.

dialogue-core/src/main/java/com/palantir/dialogue/core/NodeSelectionStrategyChannel.java
```diff
@@ -0,0 +1,191 @@
...
+            @Nullable LimitedChannel previousNodeSelectionStrategy,
+            @Nullable ImmutableList<LimitedChannel> nullableChannels,
+            DialogueNodeSelectionStrategy updatedStrategy,
+            TaggedMetricRegistry metrics,
+            Random random,
+            Ticker tick,
+            String channelName) {
+
+        ImmutableList<LimitedChannel> channels =
+                Preconditions.checkNotNull(nullableChannels, "channels must not be null");
```
carterkozak: This doesn't look very nullable

dialogue-core/src/main/java/com/palantir/dialogue/core/NodeSelectionStrategyChannel.java
```diff
@@ -0,0 +1,191 @@
...
+        }
+
+        switch (updatedStrategy) {
+            case PIN_UNTIL_ERROR:
+            case PIN_UNTIL_ERROR_WITHOUT_RESHUFFLE:
+                DialoguePinuntilerrorMetrics pinuntilerrorMetrics = DialoguePinuntilerrorMetrics.of(metrics);
+                // Previously pin until error, so we should preserve our previous location
+                if (previousNodeSelectionStrategy instanceof PinUntilErrorNodeSelectionStrategyChannel) {
+                    PinUntilErrorNodeSelectionStrategyChannel previousPinUntilError =
+                            (PinUntilErrorNodeSelectionStrategyChannel) previousNodeSelectionStrategy;
```
carterkozak, line 145: I wish we could make this cleaner, but likely out of scope for this change.

dialogue-core/src/main/java/com/palantir/dialogue/core/NodeSelectionStrategySelector.java
```diff
@@ -0,0 +1,28 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.dialogue.core;
+
+import java.util.List;
+
+public interface NodeSelectionStrategySelector {
```
carterkozak: Can this be package private?
```suggestion
interface NodeSelectionStrategySelector {
```

dialogue-core/src/main/java/com/palantir/dialogue/core/NodeSelectionStrategyChannel.java
```diff
@@ -0,0 +1,191 @@
...
+import com.palantir.logsafe.exceptions.SafeRuntimeException;
+import com.palantir.tritium.metrics.registry.TaggedMetricRegistry;
+import java.util.List;
+import java.util.Optional;
+import java.util.Random;
+import java.util.concurrent.atomic.AtomicReference;
+import org.checkerframework.checker.nullness.qual.Nullable;
+import org.immutables.value.Value;
+
+public final class NodeSelectionStrategyChannel implements LimitedChannel {
```
carterkozak: package private?

dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueNodeSelectionStrategy.java
```diff
@@ -0,0 +1,57 @@
...
+package com.palantir.dialogue.core;
+
+import com.google.common.base.Splitter;
+import com.google.common.collect.ImmutableList;
+import com.palantir.conjure.java.client.config.NodeSelectionStrategy;
+import com.palantir.logsafe.SafeArg;
+import com.palantir.logsafe.exceptions.SafeIllegalStateException;
+import java.util.List;
+
+public enum DialogueNodeSelectionStrategy {
```
carterkozak: package private?

dialogue-core/src/test/java/com/palantir/dialogue/core/DefaultNodeSelectionStrategySelectorTest.java
```diff
@@ -0,0 +1,100 @@
...
+    @Test
+    void falls_back_to_previous_on_conflict() {
+        DialogueNodeSelectionStrategy strategy;
+
+        strategy = strategySelector.updateChannelStrategy(
+                channelA, ImmutableList.of(DialogueNodeSelectionStrategy.BALANCED));
+        assertThat(strategy).isEqualTo(DialogueNodeSelectionStrategy.BALANCED);
+
+        strategy = strategySelector.updateChannelStrategy(
+                channelB, ImmutableList.of(DialogueNodeSelectionStrategy.PIN_UNTIL_ERROR_WITHOUT_RESHUFFLE));
```
carterkozak: It doesn't look like we consider any of these channels active, if a stuck request ends up completing for an old channel, do we want to resurrect the strategy?

dialogue-core/src/main/java/com/palantir/dialogue/core/NodeSelectionStrategyChannel.java
ferozco: You and I both think that, but NullAway insisted that it could be. The alternative was to disable null away on the class, but I went with this so that we would keep the safety if any changes were made

carterkozak: Doh. The name and check confused me, I wonder if we could move the null check to call sites instead, if that doesn’t look too messy?

dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueNodeSelectionStrategy.java
```diff
@@ -0,0 +1,68 @@
...
+
+    static List<DialogueNodeSelectionStrategy> fromHeader(String header) {
+        return SPLITTER.splitToStream(header)
+                .map(DialogueNodeSelectionStrategy::safeValueOf)
+                .collect(ImmutableList.toImmutableList());
+    }
+
+    private static DialogueNodeSelectionStrategy safeValueOf(String value) {
+        String normalizedValue = value.toUpperCase();
+        if (PIN_UNTIL_ERROR.name().equals(normalizedValue)) {
```
carterkozak, line 50: meganit: We can avoid toUpperCase in favor of equalsIgnoreCase

carterkozak (approved):
I like it, thanks Felipe!
@iamdanfox Have you had a chance to read over this? This is a substantial enough change that I'd like to hold off merging until you've had a chance to review as well.

dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueNodeSelectionStrategy.java
```diff
@@ -0,0 +1,68 @@
...
+import com.palantir.logsafe.exceptions.SafeIllegalStateException;
+import java.util.List;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+enum DialogueNodeSelectionStrategy {
+    PIN_UNTIL_ERROR,
+    PIN_UNTIL_ERROR_WITHOUT_RESHUFFLE,
+    BALANCED,
+    UNKNOWN;
```
iamdanfox, line 37: I think we can probably ditch the UNKNOWN variant here, as this library is the source of truth for what strategies exist?

ferozco: But you could have an old client communicating with a new server and hence receive an unknown strategy

iamdanfox: Also can we put a little comment saying why this is copy&pasted version of the regular remoting-api NodeSelectionStrategy? (iirc this was so that we could try out implementation details without having to promote them to a full user-selectable option?)

dialogue-core/src/main/java/com/palantir/dialogue/core/NodeSelectionStrategyChannel.java
```diff
@@ -101,21 +134,68 @@ private static LimitedChannel getUpdatedNodeSelectionStrategy(
...
+        return Optional.empty();
+    }
+
+    @Value.Immutable
+    interface NodeSelectionChannel {
+        DialogueNodeSelectionStrategy strategy();
+
+        LimitedChannel channel();
+
+        ImmutableList<LimitedChannel> hostChannels();
```
iamdanfox, line 188: ~why are hostChannels saved here rather than just using an instance field?~ ah nvm it's live reloading again isn't it?

ferozco: Ya we need to keep track of the host channels so that we can recreate the underlying channel in response to a server's node-selection-strategy response

dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueNodeSelectionStrategy.java
ferozco: sure

iamdanfox: > old client communicating with a new server and hence receive an unknown strategy

in this case though we just ignore it right? i'm not sure what the `unknown` variant really buys us.

ferozco: It becomes important if you decide to have more complicated logic (what was previously implemented) for determining the node selection strategy to use. If you really don't like it, i'm happy to remove it

dialogue-core/src/main/java/com/palantir/dialogue/core/NodeSelectionStrategyChannel.java
```diff
@@ -101,21 +134,68 @@ private static LimitedChannel getUpdatedNodeSelectionStrategy(
...
+            case BALANCED:
                 // When people ask for 'ROUND_ROBIN', they usually just want something to load balance better.
                 // We used to have a naive RoundRobinChannel, then tried RandomSelection and now use this heuristic:
-                return new BalancedNodeSelectionStrategyChannel(channels, random, tick, metrics, channelName);
+                return channelBuilder
+                        .channel(new BalancedNodeSelectionStrategyChannel(channels, random, tick, metrics, channelName))
+                        .build();
+            case UNKNOWN:
         }
         throw new SafeRuntimeException("Unknown NodeSelectionStrategy", SafeArg.of("unknown", updatedStrategy));
```
iamdanfox, line 168: woah if we got an unknown value surely we'd want to just preserve the old thing??? as written doesn't this mean that if a server started recommending `Node-Selection-Strategy: BALANCED2` but clients didn't know what that was then all their clients could suddenly die??? this would be a mega P0

dialogue-core/src/main/metrics/dialogue-core-metrics.yml
```diff
@@ -91,3 +91,12 @@ namespaces:
         docs: The score that the BalancedChannel currently assigns to each host (computed based on inflight requests and
           recent failures). Requests are routed to the channel with the lowest score. (Note if there are >10 nodes this
           metric will not be recorded).
+
+  dialogue.nodeselection:
+    docs: Instrumentation for which node selection strategy is used
+    metrics:
+      strategy:
+        type: meter
+        tags: [strategy]
```
iamdanfox: needs a `channel-name` tag?

dialogue-core/src/main/java/com/palantir/dialogue/core/NodeSelectionStrategyChannel.java
```diff
@@ -17,48 +17,62 @@
...
 import com.palantir.logsafe.exceptions.SafeRuntimeException;
 import com.palantir.tritium.metrics.registry.TaggedMetricRegistry;
+import java.util.List;
 import java.util.Optional;
 import java.util.Random;
 import java.util.concurrent.atomic.AtomicReference;
 import org.checkerframework.checker.nullness.qual.Nullable;
+import org.immutables.value.Value;
 
+@SuppressWarnings("NullAway")
```
iamdanfox, line 37: i've found it's usually possible to jiggle things around until NullAway passes - what was it complaining about here?

dialogue-core/src/main/java/com/palantir/dialogue/core/NodeSelectionStrategyChannel.java
ferozco: The logic for picking which node selection strategy to use currently ignores `UNKNOWN` strategy requests. This line is only here to keep errorprone happy

dialogue-core/src/main/java/com/palantir/dialogue/core/NodeSelectionStrategyChannel.java
ferozco: it gets upset about atomic references

dialogue-core/src/main/java/com/palantir/dialogue/core/NodeSelectionStrategySelector.java
```diff
@@ -0,0 +1,24 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.dialogue.core;
+
+import java.util.List;
+import java.util.Optional;
+
+interface NodeSelectionStrategySelector {
```
iamdanfox: thoughts on calling this `NodeSelectionStrategyChooser` ? the repetition of 'selection' and 'selector' kinda messes with my head

ferozco: sure

iamdanfox: alternatively maybe it would actually help readability to just collapse this into the class given that we're not doing the fancy voting just yet?

ferozco: renamed

dialogue-core/src/test/java/com/palantir/dialogue/core/NodeSelectionStrategyChannelTest.java
```diff
@@ -0,0 +1,107 @@
...
+        channel.maybeExecute(null, null).get();
+        verify(strategySelector, times(1))
+                .updateAndGet(eq(ImmutableList.of(
+                        DialogueNodeSelectionStrategy.BALANCED, DialogueNodeSelectionStrategy.UNKNOWN)));
+    }
+
+    private static void setResponse(LimitedChannel mockChannel, Optional<String> header) {
+        Mockito.clearInvocations(mockChannel);
+        Mockito.reset(mockChannel);
+        Response resp = mock(Response.class);
```
iamdanfox, line 103: would prefer to use TestResponse.withheader rather than mocking, but not a big deal

dialogue-core/src/test/java/com/palantir/dialogue/core/NodeSelectionStrategyChannelTest.java
```diff
@@ -0,0 +1,107 @@
...
+    void handles_zero_to_one_uri_update() {
+        channel.updateChannels(ImmutableList.of());
+        channel.updateChannels(ImmutableList.of(channel1));
+    }
+
+    @Test
+    void handles_one_to_many_uri_update() {
+        channel.updateChannels(ImmutableList.of(channel1));
+        channel.updateChannels(ImmutableList.of(channel1, channel2));
+    }
```
iamdanfox, line 88: Somewhat concerned that these tests could pretty easily vacuously pass, but maybe we have enough unit tests elsewhere that prove we can make requests correctly after a reload?

iamdanfox (approved):
Ok let's give it a shot! V. excited to be able to get our Balanced implementation out to more of the fleet. I expect it's gonna be a bit weird when blue-greening between nodes that suggest different Node-Selection-Strategies, but this should be pretty transient and I guess we can watch the metrics to see when it happens :)

PR #689, opened 2020-04-29
PinUntilError stable hostindex metrics

iamdanfox:
## Before this PR

Graphs look wrong for atlas -> timelock api calls because only one of these hosts is the leader so there should be no other calls to non leader hosts.

image

## After this PR
==COMMIT_MSG==
The `dialogue.pinuntilerror.success` metric had a `hostIndex` tag which was pretty useless because it was the index into the _shuffled_ list of nodes.  Now, this index is preserved from the _original_ list of nodes, so we should expect to see a square wave style pattern as nodes are reshuffled every 10 minutes.
==COMMIT_MSG==

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/PinUntilErrorNodeSelectionStrategyChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/PinUntilErrorNodeSelectionStrategyChannel.java
@@ -41,6 +41,7 @@ import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.stream.IntStream;
 import javax.annotation.Nullable;
+import org.immutables.value.Value;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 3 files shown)


dialogue-core/src/main/java/com/palantir/dialogue/core/PinUntilErrorNodeSelectionStrategyChannel.java
```diff
@@ -155,28 +163,41 @@ public void onFailure(Throwable throwable) {
...
 
         int size();
     }
 
+    @Value.Immutable
+    interface PinChannel extends LimitedChannel {
+        LimitedChannel delegate();
+
+        @Value.Auxiliary
+        int stableIndex();
```
carterkozak, line 183: I suppose when channels are removed  we end up with holes, but that's okay

dialogue-core/src/main/java/com/palantir/dialogue/core/PinUntilErrorNodeSelectionStrategyChannel.java
```diff
@@ -91,14 +91,22 @@ static PinUntilErrorNodeSelectionStrategyChannel of(
             DialoguePinuntilerrorMetrics metrics,
             Random random,
             String channelName) {
+        // We preserve the 'stableIndex' so that calls can be attributed to one host even across reshuffles
```
ferozco, line 94: the "stableIndex" is only stable across shuffles not reloads right?

iamdanfox: yep that's true.

carterkozak: That doesn't help us with the atlas case, does it? Those use several hundred unique clients

iamdanfox: assuming all DialogueChannels get created with uris in the same input order, then I think this would solve their problem. (This is _pre_ the initial shuffle)

I didn't want to just sort everything on the way in so that people could open up the raw config and see what index 0 actually is if they wanted to.

carterkozak (approved):
looks like a clear improvement to me.

ferozco:
👍

PR #690, opened 2020-04-29
[pure refactor] simplify DialogueChannel a little

iamdanfox:
My ulterior motive here is I'm a little scared of our wc ClientFactory & live reloading DialogueChannel and I want to explore how much we could simplify these.

This is just a pre-req.


dialogue-core/src/main/java/com/palantir/dialogue/core/Channels.java
```diff
@@ -0,0 +1,70 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.dialogue.core;
+
+import com.google.common.collect.ImmutableList;
+import com.palantir.dialogue.Channel;
+
+final class Channels {
```
iamdanfox: This is a big part of what I was trying to get at here - I wanted there to be a single place where we compose all the channels.

dialogue-core/src/main/java/com/palantir/dialogue/core/ActiveRequestInstrumentationChannel.java
```diff
@@ -35,13 +36,13 @@
             Channel delegate,
             String channelName,
             final @CompileTimeConstant String stage,
-            DialogueClientMetrics metrics) {
+            TaggedMetricRegistry metrics) {
```
ferozco, line 39: I think we should prefer passing the Metrics wrapper instead of full metric registry to be more explicit about what metrics are produce by what components

dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueChannel.java
```diff
@@ -27,70 +27,40 @@
...
-        clientMetrics = ClientMetrics.of(clientConfiguration.taggedMetricRegistry());
-        dialogueClientMetrics = DialogueClientMetrics.of(clientConfiguration.taggedMetricRegistry());
-        this.random = random;
-        this.ticker = ticker;
+    // we keep around internals purely for live-reloading
+    private final Config cf;
+
+    private final QueuedChannel queuedChannel; // just so we can process the queue when uris reload
+    private final Map<String, LimitedChannel> limitedChannelByUri = new ConcurrentHashMap<>();
+    private final AtomicReference<LimitedChannel> nodeSelectionChannel = new AtomicReference<>();
```
iamdanfox, line 55: I think having fewer fields helps emphasise what's immutable and what we end up mutating.

dialogue-core/src/main/java/com/palantir/dialogue/core/Channels.java
ferozco: #688 pulls out the channel updating logic from DialogueChannel which I think results in DialogueChannel being the sole place that does the composition

ferozco (commented):
I think we should block on #688 before merging this in since they both touch DialogueChannel pretty closely and result in pretty similar outcome

dialogue-core/src/main/java/com/palantir/dialogue/core/ActiveRequestInstrumentationChannel.java
iamdanfox: if we had a single DialogueMetrics class then I'd agree, but i really don't like keeping around so many different FooMetrics BarMetrics BazMetrics classes, and I think it makes the plumbing worse.

realistically, we're not in danger of adding any non-metric-schema metrics because they'll get killed in CR by one of us.

dialogue-core/src/main/java/com/palantir/dialogue/core/Channels.java
```diff
@@ -0,0 +1,66 @@
...
+                channel, cf.channelName(), "running", cf.clientConf().taggedMetricRegistry());
+        // TracedChannel must wrap TracedRequestChannel to ensure requests have tracing headers.
+        channel = new TraceEnrichingChannel(channel);
+
+        ChannelToLimitedChannelAdapter limited = new ChannelToLimitedChannelAdapter(channel);
+        return ConcurrencyLimitedChannel.create(
+                cf, limited, cf.clientConf().uris().indexOf(uri));
+    }
+
+    static Channel wrapQueuedChannel(Config cf, Channel queuedChannel) {
```
carterkozak: Why doesn't this take a QueuedChannel?

ferozco (approved):

ferozco:
🤖

PR #691, opened 2020-04-29
Expose convenience method to create a named apache channel

ferozco:
## Before this PR
@JiahuiJiang had noted that it was a bit annoying to create a simple channel because of all the boilerplate around it.

## After this PR
==COMMIT_MSG==
Expose utility for creating a closeable channel
==COMMIT_MSG==

## Possible downsides?
More API, probably only want people to use this in cases where WC is not around.

### DIFF
--- a/dialogue-apache-hc4-client/src/main/java/com/palantir/dialogue/hc4/CloseableChannel.java
+++ b/dialogue-apache-hc4-client/src/main/java/com/palantir/dialogue/hc4/CloseableChannel.java
@@ -1,0 +1,22 @@
+/*
+ * (c) Copyright 2020 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.dialogue.hc4;
+
+import com.palantir.dialogue.Channel;
+import java.io.Closeable;
+
+public interface CloseableChannel extends Channel, Closeable {}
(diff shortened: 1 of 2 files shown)


dialogue-apache-hc4-client/src/main/java/com/palantir/dialogue/hc4/CloseableChannel.java
```diff
@@ -0,0 +1,22 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.dialogue.hc4;
+
+import com.palantir.dialogue.Channel;
+import java.io.Closeable;
+
+public interface CloseableChannel extends Channel, Closeable {}
```
iamdanfox, line 22: I'm worried this is too much of a general API and we're gonna regret this.

iamdanfox:
I'm actively exploring how we could make a clientfactory that is exposable in OSS land to abstract away all the DialogueChannel & apache wiring & wc ClientFactory stuff, can we hold tight for a few days and if I can't do it we'll go down this route?

iamdanfox:
@JiahuiJiang can you take a look at the new DialogueClients.create interface provided by `com.palantir.dialogue:dialogue-clients` and see if it works for your use case?

jiahuijiang:
yeah! this is great!

Just wondering, what's the core difference that made the original JsxRsClient and Retrofit2Client didn't have this issue?

carterkozak:
The old CJR clients had loads of static state, which is one of the reasons they were so difficult to understand, monitor, and debug.

iamdanfox:
Great q. The old jaxrs & retrofit clients used a ton of _static_ state, so there was a static connection pool, static dispatcher etc. This was suboptimal because it meant that entirely unrelated outbound RPC calls would all contend with each other as they go through this one central bottleneck per JVM.

In dialogue, we've avoided _all_ static state, so we gave people the power to create connectionpools, but this comes with the responsibility of cleaning them up. Clearly this turned out to be a bit too much of a foot gun as failing to clean these up leads to dangling threads and a couple of P0s. This new DialogueClients thing handles that for you, re-using apache clients where possible and also closing them when they're evicted by live-reloading new config.

jiahuijiang:
ohhh that part is in 'ChannelCache'? 

Yeah totally agree that relying on caller to properly close all the resources is probably not very feasible and is likely to cause lot of resource leaks. This looks great :D thank you so much!

iamdanfox:
Yep that's it. Let me know if you have any other q's, I think we should be able to extend this abstraction pretty easily by adding more fluent 'with' methods if necessary.

PR #692, opened 2020-04-29
[Pure-refactor] Factor out node selection strategy logic

ferozco:
## Before this PR
Breaking apart #688 so that the real changes are easier to review

## After this PR
==COMMIT_MSG==

==COMMIT_MSG==


carterkozak (approved):
love it, thanks!

PR #693, opened 2020-04-29
consolidate on single callback

ferozco:
Instead of keeping around the queued and node selection strategy channel, we keep a single callback for when the underlying channels change.

You could conceivably plumb something to register a Uri event handler with through the builder instead of calling `updateUri` directly on a `DialogueChannel`


carterkozak (commented):
While this makes it easier to glance at the code and understand the big picture, I think it becomes harder to understand what precisely is happening.

carterkozak:
I don't have a terribly strong preference though, can be convinced otherwise.

ferozco:
I want to make this change as a stepping stone towards removing `updateUris` from `DialogueChannel` in favour of having a `CallbackManager` (bad name, but basically refreshable.subscribe) passed in through the builder with which we can register our callback

carterkozak:
Would it be easier to follow using a private method that we can pass as a reference? `private onChannelUpdate(ImmutableList<LimitedChannel> channels) {`

PR #705, opened 2020-05-01
Helpful toString implementations for debugging

iamdanfox:
## Before this PR

When debugging stuff in IntelliJ, there's a nice helpful toString by the side of any variables in scope, e.g.

image

Unfortunately, most of our channels just give you a standard system identity hashcode thing.

## After this PR
==COMMIT_MSG==
Many internal Dialogue channels now have helpful toString implementations to aid debugging
==COMMIT_MSG==

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueTracedRequestChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueTracedRequestChannel.java
@@ -23,10 +23,10 @@ import com.palantir.dialogue.Request;
 import com.palantir.dialogue.Response;
 import com.palantir.tracing.Tracers;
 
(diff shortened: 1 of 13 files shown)


dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueTracedRequestChannel.java
```diff
@@ -23,10 +23,10 @@
 import com.palantir.dialogue.Response;
 import com.palantir.tracing.Tracers;
 
-public final class DialogueTracedRequestChannel implements Channel {
+final class DialogueTracedRequestChannel implements Channel {
```
iamdanfox, line 26: Also I think this was unintentionally public

ferozco (approved):

PR #709, opened 2020-05-04
Apache channels allow requests after close has been called

carterkozak:
Rather than closing the full client, the close method releases
idle pooled connections. When requests that have already been
queued are sent, they will still execute normally.

==COMMIT_MSG==
Apache channels allow requests after close has been called
==COMMIT_MSG==


iamdanfox (approved):
Thanks, I think this should prevent the P0 that @raiju hit

PR #710, opened 2020-05-04
[test-only] simulation for blue/green upgrades when the new node returns 400s

iamdanfox:
## Before this PR

@tpetracca reported in #dev-rpc that there have been some P0s recently (PDS-118387, PDS-118388, PDS-118389) where a new node started erroneously returning 4xx responses (e.g. 400 or even 401/403), when the old node continued to serve 200s.

## After this PR
==COMMIT_MSG==
simulation for blue/green upgrades when the new node returns 400s
==COMMIT_MSG==

Interestingly it seems pin_until_error currently performs the best in this scenario.

# [GRAPHS](https://github.com/palantir/dialogue/blob/dfox/more-simulations/simulation/src/test/resources/report.md#fast_400s_then_revertconcurrency_limiter_pin_until_error)

### DIFF
--- a/simulation/src/test/java/com/palantir/dialogue/core/SimulationTest.java
+++ b/simulation/src/test/java/com/palantir/dialogue/core/SimulationTest.java
@@ -225,6 +225,33 @@ public class SimulationTest {
     }
 
     @Test
+    public void fast_400s_then_revert() {
+        servers = servers(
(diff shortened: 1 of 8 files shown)


carterkozak (approved):

iamdanfox:
🧙‍♂️

PR #711, opened 2020-05-04
Balanced now considers 4xx responses when ranking channels

iamdanfox:
## Before this PR

All dialogue node selection strategies performed pretty badly when one node started returning 400s, with pin_until_error leading the field with 82% client-perceived success.

## After this PR
==COMMIT_MSG==
The new ROUND_ROBIN algorithm now considers the number of recent 4xx responses when computing a per-host 'score', which allows us to route traffic away from buggy nodes that might send excessive 401/403 responses.
==COMMIT_MSG==

cc @tpetracca 

# [GRAPHS](https://github.com/palantir/dialogue/blob/dfox/balanced-4xx/simulation/src/test/resources/report.md#fast_400s_then_revertconcurrency_limiter_round_robin)

Note: this PR does _NOT_ retry 4xxs, it just means that as soon as we spot one we become suspicious of that node, and try to route requests around it.


simulation/src/test/resources/report.md
```diff
@@ -11,8 +11,8 @@
...
                            drastic_slowdown[UNLIMITED_ROUND_ROBIN].txt:	success=100.0%	client_mean=PT0.275952999S 	server_cpu=PT18M23.81199998S	client_received=4000/4000	server_resps=4000	codes={200=4000}
         fast_400s_then_revert[CONCURRENCY_LIMITER_PIN_UNTIL_ERROR].txt:	success=82.2%	client_mean=PT0.102233333S 	server_cpu=PT10M13.4S     	client_received=6000/6000	server_resps=6000	codes={200=4934, 400=1066}
-            fast_400s_then_revert[CONCURRENCY_LIMITER_ROUND_ROBIN].txt:	success=71.0%	client_mean=PT0.091S       	server_cpu=PT9M6S         	client_received=6000/6000	server_resps=6000	codes={200=4260, 400=1740}
-                      fast_400s_then_revert[UNLIMITED_ROUND_ROBIN].txt:	success=71.0%	client_mean=PT0.091S       	server_cpu=PT9M6S         	client_received=6000/6000	server_resps=6000	codes={200=4260, 400=1740}
+            fast_400s_then_revert[CONCURRENCY_LIMITER_ROUND_ROBIN].txt:	success=99.8%	client_mean=PT0.119833333S 	server_cpu=PT11M59S       	client_received=6000/6000	server_resps=6000	codes={200=5990, 400=10}
```
iamdanfox: @tpetracca this simulation runs with 10 clients, and the `codes={200=5990, 400=10}` bit shows that clients only perceive ten 400s in total, because all other requests happen to get routed to working nodes.

carterkozak:
What happens when we add validation for requests that previously created bad data?

iamdanfox:
Not sure I understand the suggestion - could you flesh out an example a bit?

carterkozak:
If I have an endpoint that stores a dog name (for dog-account creation), and find out that empty or blank strings can be provided, I might release a version that makes the endpoint more strict and disallows empty and blank values in addition to null. In this case, I _want_ to create more 400s instead of 200s in the presence of bad data.
I agree that the opposite case is also possible, but I'm not sure we should be opinionated here because there are valid scenarios in both directions.

iamdanfox:
Mmm interesting. In that scenario, if a blue-green rollout kept going, this PR would cause all traffic to concentrate to the last remaining old node, until that old node eventually gets switched off and then it would equalize across all our new (stricter) nodes.

My hunch is that there are more cases where product teams write bugs into their auth code and erroneously return (i.e. 200 -> 4xxs) compared to the number of cases where teams intentionally want to add strictness and ban some previously accepted input (4xx -> 200).

carterkozak:
Let's discuss tomorrow, I'd like to take some time to think through this :-)

dialogue-core/src/main/java/com/palantir/dialogue/core/BalancedNodeSelectionStrategyChannel.java
```diff
@@ -128,9 +128,17 @@
...
+
                     if (Responses.isQosStatus(response) || Responses.isServerError(response)) {
                         recentFailuresReservoir.update(FAILURE_WEIGHT);
                         observability.debugLogStatusFailure(response);
+                    } else if (Responses.isClientError(response)) {
+                        // We track 4xx responses because bugs in the server might cause one node to erroneously
+                        // throw 401/403s when another node could actually return 200s. Empirically, healthy servers
+                        // do actually return a continuous background rate of 4xx responses, so we weight these
+                        // drastically less than 5xx responses.
+                        recentFailuresReservoir.update(FAILURE_WEIGHT / 100);
```
iamdanfox, line 152: @carterkozak this weights 4xx responses as 0.1 as discussed

iamdanfox:
~Ooh so I think where we were previously casting to integer scores, the fact that we have doubles now might mean it takes a lot longer for things to decay to _fully_ zero, meaning the sorting order would stay as it is for a lot longer? It seems silly to be sorting one node ahead of another because it has 0.0001 and the other has 0.0002~ Fixed.

dialogue-core/src/main/java/com/palantir/dialogue/core/CoarseExponentialDecayReservoir.java
```diff
@@ -47,20 +46,18 @@
...
         // Potential optimization if the value is zero
         decayIfNecessary();
-        // Results must be rounded otherwise a single increment will be effectively erased after the first
-        // decay cycle.
-        return Ints.saturatedCast(Math.round(value.get()));
+        return value.get();
     }
 
-    void decayIfNecessary() {
+    private void decayIfNecessary() {
```
carterkozak, line 60: :+1:

dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueInternalWeakReducingGauge.java
```diff
@@ -39,38 +38,44 @@
...
-    public synchronized void add(T sourceElement) {
+    public void add(T sourceElement) {
         weakSet.add(sourceElement);
     }
 
     @Override
-    public synchronized Number getValue() {
-        return operator.apply(weakSet.stream().mapToLong(gaugeFunction));
+    public Number getValue() {
+        return summarize.apply(weakSet.stream());
```
carterkozak: I don’t think this is valid without method level synchronization, the synchronized wrappers don’t provide any protection for iterators or streams.

carterkozak:
I think we can keep the scores as integers retaining rounding, we would just need to do it outside of the reservoir

dialogue-core/src/test/java/com/palantir/dialogue/core/CoarseExponentialDecayReservoirTest.java
```diff
@@ -42,9 +42,9 @@ void testDecay_byHalf() {
         decay.update(2);
         assertThat(decay.get()).isEqualTo(2);
         clock.set(10);
-        assertThat(decay.get()).isOne();
+        assertThat(decay.get()).isBetween(0.9999, 1.0);
```
carterkozak: Could use something like `assertThat(decay.get()).isEqualTo(1, Offset.offset(.0001));` I think `isEqualTo` with an offset is an alias to `isCloseTo`.

dialogue-core/src/main/java/com/palantir/dialogue/core/BalancedNodeSelectionStrategyChannel.java
```diff
@@ -158,11 +167,11 @@ public void onFailure(Throwable throwable) {
 
         SortableChannel computeScore() {
             int requestsInflight = inflight.get();
-            int recentFailures = recentFailuresReservoir.get();
+            double failureReservoir = recentFailuresReservoir.get();
 
-            int score = requestsInflight + recentFailures;
+            int score = requestsInflight + Ints.saturatedCast(Math.round(failureReservoir));
```
carterkozak, line 185: nit: might want to add a comment describing the case where a small fraction impacts selection long beyond its intended lifespan in the absence of other data.

carterkozak:
lgtm, a comment would be nice on the rounding code.

carterkozak:
:+1:

dialogue-core/src/main/java/com/palantir/dialogue/core/BalancedNodeSelectionStrategyChannel.java
iamdanfox: Even better, have captured it in a test.

carterkozak: Tests are great, but it's also nice to provide inline context at a glance ;-)

PR #715, opened 2020-05-06
ApacheHttpClientChannels create/close metrics

iamdanfox:
## Before this PR

I'm replacing WC's clientfactory with my new fluent one, but it would mean we lose some instrumentation around client creation & closes.

## After this PR
==COMMIT_MSG==
Two new metrics instrument when apache clients are created/closed.
==COMMIT_MSG==

I actually prefer this approach because I was always slightly suspicious that people were manually creating these things and we couldn't see them in the metrics.

### DIFF
--- a/dialogue-apache-hc4-client/build.gradle
+++ b/dialogue-apache-hc4-client/build.gradle
@@ -1,5 +1,6 @@
 apply from: "$rootDir/gradle/publish-jar.gradle"
 apply plugin: 'com.palantir.metric-schema'
+apply plugin: 'com.palantir.metric-schema-markdown'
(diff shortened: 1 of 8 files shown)


dialogue-apache-hc4-client/build.gradle
```diff
@@ -1,5 +1,6 @@
 apply from: "$rootDir/gradle/publish-jar.gradle"
 apply plugin: 'com.palantir.metric-schema'
+apply plugin: 'com.palantir.metric-schema-markdown'
```
carterkozak, line 3: What if we only generate markdown in the dialogue-clients module?

carterkozak (approved):

PR #716, opened 2020-05-06
New `dialogue.client.create` metric when DialogueChannel is created

iamdanfox:
## Before this PR

We have seen some OOMs from people endlessly creating clients (as these kick off threads under the hood). Unfortunately I don't fully trust the graphs because the instrumentation is done at the Witchcraft layer, so people could (in theory) bypass it by creating their own clients.

## After this PR
==COMMIT_MSG==
Creating DialogueChannel instances now marks a new `dialogue.client.create` meter so we can detect any services that are creating unexpectedly large numbers of these clients.
==COMMIT_MSG==

## Possible downsides?
- mo metrics mo money
- more statements to execute on the finalizer thread


carterkozak (approved):

iamdanfox:
👨‍🚀

PR #717, opened 2020-05-06
Dialogue can record info to a HostEventsSink

iamdanfox:
## Before this PR

Up until now, we didn't plumb HostEventsSink into dialogue because it had `okhttp` in the package name. From c-j-r 5.9.0 onwards this is no longer the case, and it's plumbed in via the ClientConfiguration.

https://github.com/palantir/conjure-java-runtime/pull/1591

## After this PR
==COMMIT_MSG==
Dialogue now records info to a `HostEventsSink` provided in the `ClientConfiguration`. If none is specified, there is no runtime cost.
==COMMIT_MSG==


dialogue-core/src/main/java/com/palantir/dialogue/core/HostMetricsChannel.java
```diff
@@ -0,0 +1,101 @@
...
+        }
+
+        @Override
+        public void onFailure(Throwable throwable) {
+            if (throwable instanceof IOException) {
+                hostEventCallback.recordIoException();
+            }
+        }
+    }
+}
```
carterkozak, line 101: assuming this is more or less identical to the WC implementation

carterkozak:
👍

dialogue-core/src/main/java/com/palantir/dialogue/core/HostMetricsChannel.java
iamdanfox: Yep, I injected a clock, but otherwise this is the same. I think we _might_ get slightly non-human readable info because channelName can be quite long like `foo-ROUND_ROBIN`, but I think something is better than nothing

dialogue-clients/src/main/java/com/palantir/dialogue/clients/DialogueClients.java
```diff
@@ -71,7 +72,10 @@ public static ReloadingFactory create(Refreshable<ServicesConfigBlock> scb) {
                     WithDialogueOptions<ReloadingFactory>,
                     ConjureClients.NonReloadingClientFactory,
                     ConjureClients.ToReloadingFactory<ReloadingFactory>,
-                    ReloadingChannelFactory {}
+                    ReloadingChannelFactory {
+        // TODO(dfox): c-j-r should really define this method in WithClientOptions
```
iamdanfox: I think I just wanna get this right from the beginning. https://github.com/palantir/conjure-java-runtime/pull/1592

PR #718, opened 2020-05-07
DialogueClients produces versioned metrics

iamdanfox:
## Before this PR

I tried to read out the `dialogue.client.create` metric in my Witchcraft PR and found that the `dialogueVersion` tag was missing on one of the metrics. 

## After this PR
==COMMIT_MSG==
Clients constructed using the `DialogueClients` fluent factories now have metrics tagged with the proper `dialogueVersion`.
==COMMIT_MSG==

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueInternalVersionedTaggedMetricRegistry.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueInternalVersionedTaggedMetricRegistry.java
@@ -22,30 +22,34 @@ import com.codahale.metrics.Histogram;
 import com.codahale.metrics.Meter;
 import com.codahale.metrics.Metric;
 import com.codahale.metrics.Timer;
+import com.palantir.logsafe.Preconditions;
 import com.palantir.tritium.metrics.registry.MetricName;
 import com.palantir.tritium.metrics.registry.TaggedMetricRegistry;
 import com.palantir.tritium.metrics.registry.TaggedMetricSet;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.function.BiConsumer;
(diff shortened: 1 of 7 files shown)


dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueInternalVersionedTaggedMetricRegistry.java
```diff
@@ -22,30 +22,34 @@
...
 import com.palantir.tritium.metrics.registry.TaggedMetricRegistry;
 import com.palantir.tritium.metrics.registry.TaggedMetricSet;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.function.BiConsumer;
 import java.util.function.Supplier;
 
-final class VersionedTaggedMetricRegistry implements TaggedMetricRegistry {
+public final class DialogueInternalVersionedTaggedMetricRegistry implements TaggedMetricRegistry {
```
iamdanfox, line 35: It's kinda heartbreaking making stuff like this public just so we can use it from one of our other jars 😭

carterkozak: What if we don’t, but only provide version metrics via the standard factory?
We’ll lose them if we get dialogue upgrades prior to wc taking advantage of the new goodness, but I think it’s worth it.

iamdanfox: it's actually such a useful part of my dashboard though I _really_ don't want to lose it.

I think my preference would actually be to add a metric-schema feature so you can opt-in to embedding a library version (and specify that tag you want, in this case "dialogueVersion")

PR #720, opened 2020-05-08
Failing test for ConcurrencyLimiters talking to a Guava RateLimiter server

iamdanfox:
## Before this PR

In @jkozlowski's internal-ski-product, we haven't been able to fully replace c-j-r with dialogue because there's one test (`testPartitionedRangeScan`) which has been flaking with the dialogue implementation but rock solid with c-j-r.

In this particular test they kick off a big spike of requests to the ski-product server and then wait for them all to succeed, relying on concurrencylimiters to throttle sensibly on the client side. On the server-side, ski-product uses guava ratelimiters, with a default of 300 reads per datastore.

## After this PR
==COMMIT_MSG==
IntegrationTest tests against a real server with a Guava ratelimiter set to allow 300 queries per second.
==COMMIT_MSG==

Roughly, I think that we're too quick to increase the limit - say we have 20/20 permits used, and then three return successfully all at pretty much the same time. Suddenly we have 17 inflight and the permits additively increased to 23. Six queued requests get fired off and surprise surprise the server tells us to get wrecked.

# [GRAPHS](https://github.com/palantir/dialogue/blob/dfox/realistic-concurrency-limiter-test/simulation/src/test/resources/report.md)

## Possible downsides?
- The test uses a real thread.sleep and it currently takes 10 seconds to complete.
- We're making a tradeoff here by more cautiously increasing limits (to avoid blowing out a whole batch of requests), which results in a potentially slower client-perceived mean.

### DIFF
--- a/dialogue-client-verifier/src/test/java/com/palantir/verification/IntegrationTest.java
+++ b/dialogue-client-verifier/src/test/java/com/palantir/verification/IntegrationTest.java
@@ -24,6 +24,7 @@ import com.google.common.io.ByteStreams;
 import com.google.common.primitives.Ints;
 import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
+import com.google.common.util.concurrent.RateLimiter;
 import com.palantir.conjure.java.api.config.service.ServiceConfiguration;
 import com.palantir.conjure.java.api.config.service.UserAgent;
 import com.palantir.conjure.java.api.config.ssl.SslConfiguration;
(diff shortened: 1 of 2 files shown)


dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -339,8 +339,7 @@ private void logRetry(long backoffNanoseconds, @Nullable Throwable throwable) {
                         SafeArg.of("backoffMillis", TimeUnit.NANOSECONDS.toMillis(backoffNanoseconds)),
                         SafeArg.of("channelName", channelName),
                         SafeArg.of("serviceName", endpoint.serviceName()),
-                        SafeArg.of("endpoint", endpoint.endpointName()),
-                        throwable);
```
carterkozak: what's happening here?

iamdanfox: So these stacktraces are actually always identical (they don't include the cause), and they made the SimulationTest logs impossible to read. I'll put it back to keep this PR focussed, but I think it's worth considering eliding them.

iamdanfox:
Bad news, the integration test immediately flaked. https://circleci.com/gh/palantir/dialogue/9981

carterkozak:
This isn't as strict as the original implementation, which we expected, no?

iamdanfox:
Ok have force pushed so this branch is once again just a failing test.

iamdanfox:
Fixed in https://github.com/palantir/dialogue/pull/738

PR #721, opened 2020-05-08
[stacked] Bring back netflix lib + ConjureWindowedLimit

iamdanfox:
## Before this PR

The failing test in https://github.com/palantir/dialogue/pull/720 means we can't really declare victory with dialogue yet.

## After this PR
==COMMIT_MSG==
Bring back netflix lib + ConjureWindowedLimit
==COMMIT_MSG==

`do not merge` because I don't think we want to bring back this heavyweight synchronized implementation, I just wanted to demonstrate that it fixes the problem.

Interestingly, you can really see the window resulting in stepped limit increases in the graphs:

# [GRAPHS](https://github.com/palantir/dialogue/blob/dfox/option1/simulation/src/test/resources/report.md#drastic_slowdownconcurrency_limiter_pin_until_error)


dialogue-core/src/main/java/com/palantir/dialogue/core/ConjureWindowedLimit.java
```diff
@@ -0,0 +1,100 @@
...
+import java.util.function.Consumer;
+
+/**
+ * Changes made from {@link com.netflix.concurrency.limits.limit.WindowedLimit}:
+ *
+ * <p>1. Modify to reduce window size whenever a dropped sample is seen, rather than awaiting the whole window. 2.
+ * Change package and make package private. 3. Code style. 4. Inlined constants.
+ */
+@SuppressWarnings("NullAway") // TODO(dfox): deal with this
+class ConjureWindowedLimit implements Limit {
```
iamdanfox, line 31: For any bystanders I didn't just write this it's just lifted from https://github.com/palantir/conjure-java-runtime/blob/develop/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/ConjureWindowedLimit.java

simulation/src/test/resources/report.md
```diff
@@ -16,23 +16,23 @@
...
                       slow_503s_then_revert[UNLIMITED_ROUND_ROBIN].txt:	success=100.0%	client_mean=PT0.093312988S 	server_cpu=PT4M35.008666652S	client_received=3000/3000	server_resps=3039	codes={200=3000}
-slowdown_and_error_thresholds[CONCURRENCY_LIMITER_PIN_UNTIL_ERROR].txt:	success=99.8%	client_mean=PT3M9.775496352S	server_cpu=PT13H15M43.393332383S	client_received=10000/10000	server_resps=15825	codes={200=9978, 500=22}
-    slowdown_and_error_thresholds[CONCURRENCY_LIMITER_ROUND_ROBIN].txt:	success=99.8%	client_mean=PT3M20.092857357S	server_cpu=PT14H41M17.066666225S	client_received=10000/10000	server_resps=15426	codes={200=9982, 500=18}
+slowdown_and_error_thresholds[CONCURRENCY_LIMITER_PIN_UNTIL_ERROR].txt:	success=100.0%	client_mean=PT2M34.920963057S	server_cpu=PT7H4M18.946665555S	client_received=10000/10000	server_resps=10578	codes={200=10000}
+    slowdown_and_error_thresholds[CONCURRENCY_LIMITER_ROUND_ROBIN].txt:	success=100.0%	client_mean=PT2M31.613638514S	server_cpu=PT9H44M51.153332803S	client_received=10000/10000	server_resps=10901	codes={200=10000}
```
iamdanfox, line 35: I think the different between 99.9% success and 100% success is pretty big, because one failure can ruin an entire batch.

PR #722, opened 2020-05-08
[stacked] AimdConcurrencyLimiter increases at most once per second

iamdanfox:
## Before this PR

Failing test in https://github.com/palantir/dialogue/pull/720

## After this PR
==COMMIT_MSG==
AimdConcurrencyLimiter increases at most once per second
==COMMIT_MSG==

Hard-coding a window feels pretty janky, but it does pass the internal-ski-product test! Auto-scaling the window size based on recent rtt would be much nicer.

Graphs show that we impose less spiky traffic on servers

# [GRAPHS](https://github.com/palantir/dialogue/blob/dfox/option2/simulation/src/test/resources/report.md)


sfackler:
Another potential option that could involve fewer moving parts is to do something like the PinUntilError channel and only adjust the limit upwards if it's not already higher than the observed limit when the request started. Then if you have a limit of e.g. 20 and 20 requests are made concurrently you'd end up at a limit of 21 after they complete rather than 30.

It should allow the limit to ramp up significantly more quickly than 1 per second but also more accurately respect the actual concurrency that the server was observed to be able to handle.

iamdanfox:
Closing out in favour of https://github.com/palantir/dialogue/pull/738

PR #723, opened 2020-05-08
[stacked] Just fiddle with the numbers

iamdanfox:
## Before this PR

Failing test in #720

## After this PR
==COMMIT_MSG==
ConcurrencyLimiters increase more tentatively and backoff more aggressively.
==COMMIT_MSG==

# [GRAPHS](https://github.com/palantir/dialogue/blob/dfox/option3/simulation/src/test/resources/report.md#drastic_slowdownconcurrency_limiter_pin_until_error)


dialogue-core/src/main/java/com/palantir/dialogue/core/AimdConcurrencyLimiter.java
```diff
@@ -38,7 +38,7 @@
 
     private static final Logger log = LoggerFactory.getLogger(AimdConcurrencyLimiter.class);
     private static final int INITIAL_LIMIT = 20;
-    private static final double BACKOFF_RATIO = .9D;
+    private static final double BACKOFF_RATIO = .6D;
```
carterkozak: I wonder if we can reduce the backoff change a bit without sacrificing rates of success? In an ideal world I think the value would be pretty close to the increment saturation threshold.

carterkozak:
If I'm reading the graphs correctly this seems like the best performer. I like that it doesn't use a hard-coded window-size.

dialogue-core/src/main/java/com/palantir/dialogue/core/AimdConcurrencyLimiter.java
carterkozak: .6 is pretty low, it basically cuts the permits in half any time there's a 500 or connection failure. What if we leave this value at .9, update the `originalLimit * 0.95` below to reference this value, and also reduce the increment value from 1 to ~.5?

carterkozak: Thinking about the problem a bit more, increasing the limit from 1 to 2 should only require a single successful request, but increasing the limit from 100 to 101 should probably require several successful requests without failures while the limiter is mostly saturated.

carterkozak (approved):

PR #725, opened 2020-05-11
simple dynamic concurrency limiter smoothing

carterkozak:
## [graphs](https://github.com/palantir/dialogue/blob/ckozak/limits/simulation/src/test/resources/report.md#drastic_slowdownconcurrency_limiter_pin_until_error)

==COMMIT_MSG==
simple dynamic concurrency limiter smoothing
==COMMIT_MSG==


iamdanfox:
@carterkozak can you re-run simulations so we can compare the graphs? I think it currently they're currently just the ones from my 'fiddle with the numbers' PR.

dialogue-core/src/main/java/com/palantir/dialogue/core/AimdConcurrencyLimiter.java
```diff
@@ -120,25 +122,40 @@ void dropped() {
...
+
+    enum LimitUpdater implements DoubleBinaryOperator {
         SUCCESS() {
             @Override
-            public int applyAsInt(int originalLimit, int inFlightSnapshot) {
-                if (inFlightSnapshot >= originalLimit * 0.95) {
-                    return Math.min(MAX_LIMIT, originalLimit + 1);
+            public double applyAsDouble(double originalLimit, double inFlightSnapshot) {
+                if (inFlightSnapshot >= originalLimit * BACKOFF_RATIO) {
+                    return Math.min(MAX_LIMIT, originalLimit + (1D / originalLimit));
```
iamdanfox, line 150: Wanna put a little comment to say we intentionally want to do (proportionally) bigger increments when our concurrency limit is low (i.e. jumping from 1 -> 2 in one hop is desirable), but we want to be a bit more cautious about jumping from 100 -> 101?

dialogue-core/src/main/java/com/palantir/dialogue/core/AimdConcurrencyLimiter.java
```diff
@@ -37,13 +39,13 @@
 final class AimdConcurrencyLimiter {
```
iamdanfox, line 39: This is not really additive increase multiplicative decrease is it... CautiousIncreaseAggressiveDecrease?

dialogue-core/src/main/java/com/palantir/dialogue/core/AimdConcurrencyLimiter.java
```diff
@@ -148,7 +165,7 @@ public int applyAsInt(int originalLimit, int _inFlightSnapshot) {
      * permits such that another permit can be {@link #acquire acquired}.
      */
     int getLimit() {
-        return limit.get();
+        return Ints.saturatedCast(Math.round(limit.get()));
```
iamdanfox, line 168: slight preference to expose this number as a double and do the cast later so that datadog dashboards could show the _precise_ number? (might help with debugging if there are a bunch of requests and we see the limit go from 120.9 -> 120.7)

carterkozak:
I think the results are a tad different, they're updated in my commit: https://github.com/palantir/dialogue/pull/725/commits/465beddf995ba787e69faafb6220455a6ff9d7f3#diff-7888877ab61501c1e50055394ce9599d

iamdanfox:
Oooh this one is pretty https://github.com/palantir/dialogue/blob/ckozak/limits/simulation/src/test/resources/report.md#one_big_spikeconcurrency_limiter_pin_until_error

carterkozak:
Cleaned up and moved to #738

PR #726, opened 2020-05-11
Dialogue should not propagate retry-other

carterkozak:
Currently dialogue rethrows qos exceptions, which doesn't quite match CJR: CJR only rethrows throttle and unavailable when `ServerQos.PROPAGATE_429_and_503_TO_CALLER` is set.
The dialogue runtime isn't at all aware of the client configuration at the moment, so we can't emulate that precisely.
This change brings dialogue closer to CJR, but isn't meant to bring the two exactly together.
There's an argument to be made that when a 503 or 429 is returned for all retries, it _should_ be propagated in order to apply backpressure across the entire system.

==COMMIT_MSG==
Dialogue does not propagate retry-other
==COMMIT_MSG==

### DIFF
--- a/dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ErrorDecoder.java
+++ b/dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ErrorDecoder.java
@@ -66,7 +66,9 @@ enum ErrorDecoder {
                 if (location.isPresent()) {
                     String locationHeader = location.get();
                     try {
-                        return toRemoteException(code, QosException.retryOther(new URL(locationHeader)));
+                        UnknownRemoteException remoteException = new UnknownRemoteException(code, "");
+                        remoteException.initCause(QosException.retryOther(new URL(locationHeader)));
+                        return remoteException;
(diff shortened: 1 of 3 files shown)


dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ErrorDecoder.java
```diff
@@ -115,4 +115,10 @@ private static String toString(InputStream body) throws IOException {
             return CharStreams.toString(reader);
         }
     }
+
+    private static UnknownRemoteException toRemoteException(int code, QosException qos) {
```
iamdanfox: slight preference to inline this to reassure people there's nothing too crazy going on here

iamdanfox:
So anyone that had a `catch` block expecting to catch 308s as `QosException.RetryOther` will no longer work here right? Are we confident that the atlasdb folks don't rely on this for their timelock calls?

carterkozak:
CJR throws a FeignException in that case, only dialogue (and dialogue-feign) throw the RetryOther from a client.

iamdanfox (approved):

PR #727, opened 2020-05-11
Idle connections are checked and timed out more frequently

carterkozak:
Previously the idle timeout and check interval were the same,
resulting in the potential for connections to live twice as long
as we expected. While this isn't a big problem, it introduces a
potential race when the server closes an idle connection as the
client attempts to check it out from the pool. The liveness check
can succeed while the subsequent request occurs after the
server-side has closed.

## After this PR
==COMMIT_MSG==
Idle connections are checked and timed out more frequently
==COMMIT_MSG==

## Possible downsides?
More state to manage ourselves

### DIFF
--- a/dialogue-apache-hc4-client/src/main/java/com/palantir/dialogue/hc4/ApacheHttpClientChannels.java
+++ b/dialogue-apache-hc4-client/src/main/java/com/palantir/dialogue/hc4/ApacheHttpClientChannels.java
@@ -17,7 +17,9 @@ package com.palantir.dialogue.hc4;
 
 import com.codahale.metrics.Meter;
 import com.google.common.collect.ImmutableSet;
+import com.google.common.io.Closer;
 import com.google.common.primitives.Ints;
+import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import com.palantir.conjure.java.api.config.service.BasicCredentials;
 import com.palantir.conjure.java.client.config.CipherSuites;
 import com.palantir.conjure.java.client.config.ClientConfiguration;
@@ -48,6 +50,7 @@ import java.util.Map;
 import java.util.Queue;
 import java.util.Set;
 import java.util.concurrent.ExecutorService;
+import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.LongStream;
 import javax.annotation.Nullable;
@@ -75,6 +78,7 @@ import org.apache.http.impl.auth.BasicSchemeFactory;
 import org.apache.http.impl.client.CloseableHttpClient;
 import org.apache.http.impl.client.HttpClientBuilder;
 import org.apache.http.impl.client.HttpClients;
+import org.apache.http.impl.client.IdleConnectionEvictor;
 import org.apache.http.impl.client.ProxyAuthenticationStrategy;
 import org.apache.http.impl.conn.PoolingHttpClientConnectionManager;
 import org.apache.http.impl.conn.SystemDefaultRoutePlanner;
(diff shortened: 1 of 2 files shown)


dialogue-apache-hc4-client/src/main/java/com/palantir/dialogue/hc4/ApacheHttpClientChannels.java
```diff
@@ -336,7 +357,7 @@ public CloseableClient build() {
             // Most of our servers use a keep-alive timeout of one minute, by using a slightly lower value on the
             // client side we can avoid unnecessary retries due to race conditions when servers close idle connections
             // as clients attempt to use them.
-            long idleConnectionTimeoutMillis = Math.min(Duration.ofSeconds(55).toMillis(), socketTimeoutMillis);
+            long idleConnectionTimeoutMillis = Math.min(Duration.ofSeconds(50).toMillis(), socketTimeoutMillis);
```
carterkozak, line 360: Slight reduction so that `idleConnectionTimeoutMillis` + sweep time remains below 1m to avoid racing the server.

dialogue-apache-hc4-client/src/main/java/com/palantir/dialogue/hc4/ApacheHttpClientChannels.java
```diff
@@ -410,10 +433,31 @@ public CloseableClient build() {
             });
 
             CloseableHttpClient apacheClient = builder.build();
-            return CloseableClient.wrap(apacheClient, name, connectionManager, conf, executor);
+            IdleConnectionEvictor connectionEvictor = new IdleConnectionEvictor(
+                    connectionManager,
+                    idleConnectionEvictorThreadFactory(name, conf.taggedMetricRegistry()),
+                    // Use a shorter check duration than idle connection timeout duration in order to avoid allowing
+                    // stale connections to race the server-side timeout.
+                    Math.min(idleConnectionTimeoutMillis, 5_000),
```
carterkozak, line 438: Basically 5 seconds with the caveat that a very, very low idle timeout will result in more frequent sweeps.

dialogue-apache-hc4-client/src/main/java/com/palantir/dialogue/hc4/ApacheHttpClientChannels.java
```diff
@@ -385,7 +406,9 @@ public CloseableClient build() {
                             .build())
                     .setDefaultSocketConfig(socketConfig)
                     .evictIdleConnections(idleConnectionTimeoutMillis, TimeUnit.MILLISECONDS)
```
ferozco: how does this interact with the idleConnectionEvictor we provide?

carterkozak: it's actually ignored at the moment because we create the idleConnectionEvictor ourselves. I left it for specificity, but we can go either way.

ferozco: can we remove it? Its a bit confusing to set config that we intend to ignore

carterkozak: Removed a few other duplicates as well.

ferozco (approved):

iamdanfox:
This class has got a quite a lot of subtle timeouts flying around... how does the setValidateAfterInactivity thing differ from the IdleConnectionEvictor we have here?

carterkozak:
setValidateAfterInactivity is a pre-flight check that's less expensive than creating a new connection, but relatively expensive compared to a small request. It's nearly identical to the okhttp check, but the apache client allows us to opt into it after an idle period rather than all requests.

PR #729, opened 2020-05-11
fix #728 308 retry-other responses don't use exponential backoff

### LINKED ISSUE #728: Following 308 redirects go through the exponential backoff
## What happened?

Dialogue 1.38.0

As the title suggests. I have a test on AtlasDB that verifies that a server that gives you looping 308s for ten seconds and then starts returning 200s has us recover quickly.

```java
    @Test
    public void recoversQuicklyOnceServerCeasesToSpam308s() {
        server.stubFor(TIMESTAMP_MAPPING.willReturn(aResponse()
                .withStatus(308)
                .withHeader("Location", getUriForPort(serverPort))));

        Instant start = Instant.now();
        ExecutorService ex = PTExecutors.newSingleThreadExecutor(true);
        ex.submit(this::scheduleServerRecoveryAfterTenSeconds);

        makeTimestampsRequest();
        assertThat(Instant.now())
                .as("should recover in a second after things are good again")
                .isBefore(start.plus(Duration.ofSeconds(11)));
        ex.shutdown();
    }

    private ConjureGetFreshTimestampsResponse makeTimestampsRequest() {
        return conjureTimelockService.getFreshTimestamps(
                AuthHeader.valueOf("Bearer unused"), CLIENT, ConjureGetFreshTimestampsRequest.of(10));
    }
```

This succeeds with a conjureTimelockService provided via CJR but fails with one from Dialogue (in particular, the assertion fails, though no exception is thrown).

Poking around in the debugger it looks like 308s go through the same code path as 429s and 503s, including the exponential backoff.

I'd have liked the 308s to be followed immediately - these aren't indicative of QoS problems that 429s or 503s would seem to be, so unclear if waiting is generally helpful here.

carterkozak:
==COMMIT_MSG==
308 retry-other responses don't use exponential backoff
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-729.v2.yml
+++ b/changelog/@unreleased/pr-729.v2.yml
@@ -1,0 +1,5 @@
+type: fix
+fix:
+  description: 308 retry-other responses don't use exponential backoff
+  links:
+  - https://github.com/palantir/dialogue/pull/729
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/Responses.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/Responses.java
@@ -20,7 +20,7 @@ import com.palantir.dialogue.Response;
 /** Utility functionality for {@link Response} handling. */
 final class Responses {
 
-    private static boolean isRetryOther(Response response) {
+    static boolean isRetryOther(Response response) {
         return response.code() == 308;
     }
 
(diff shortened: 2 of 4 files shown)


ferozco:
Since we ignore `Location`, 308 and 503 are effecitvely the same. Should we immediately retry both? If we do immediately with RR NSS we could end up hammering the same host.

carterkozak:
I think 503 is different, 308 means we're asking the wrong node, where 503 means the node isn't in working order right now (but may be later). When services are starting we can respond 503, in the single node case we absolutely want to back off.

iamdanfox (approved):
LGTM

(just took off merge when ready to avoid cutting any discussion short)

ferozco:
sure lets ship it 👍

PR #730, opened 2020-05-11
[pure refactor] re-order RetryingChannel methods for readability

iamdanfox:
## Before this PR

I found RetryingChannel pretty hard to read. It has 11 fields, quite a lot of behaviour, this mutable `failures` variable and some fancy recursive scheduling.

## After this PR
==COMMIT_MSG==
re-order RetryingChannel methods for top-down readability, so that the most important methods come first and their implementation details follow.
==COMMIT_MSG==

## Possible downsides?
<!-- Please describe any way users could be negatively affected by this PR. --


iamdanfox:
Blargh think i'm too tired for this

PR #732, opened 2020-05-12
[test-only] Speed up simulations

iamdanfox:
## Before this PR ([3m11](https://circleci.com/gh/palantir/dialogue/9934) builds)

Locally and on CI, SimulationTest was pretty slow to run. By adding some trusty log lines, I noticed that adding stuff to the DeterministicScheduler got slower and slower. Intuitively this makes sense, because their DeltaQueue thing is a linked list, and our `fast_503s_then_revert` case sends 500rps for 90sec, i.e. it puts 45k objects onto that linked list.

Also we were logging a lot, which was kinda slow.

## After this PR ([1m26](https://circleci.com/gh/palantir/dialogue/9964) builds)
==COMMIT_MSG==
SimulationTest is now faster, because as soon as it schedules a request, it advances the simulation clock to this time. This minimizes the number of items inside the DeterministicScheduler's linked list.
==COMMIT_MSG==

~`do not merge` because there's more optimizations I think we can do.~


carterkozak:
Writing to the console is expensive, and simulations run on a single thread. What happens if we log using asynchronous components instead?
How large is the simulation stdout/stderr? Should we set the threshold to error for simulations?

iamdanfox:
With warn we only get 3 log lines per run.

carterkozak (approved):

PR #733, opened 2020-05-12
Reduce allocation in InstrumentedChannel

iamdanfox:
## Before this PR

SimulationTest is slow for many reasons. One of my hypothesis is that we actually allocate quite a lot of objects in order to process 45k events.

## After this PR
==COMMIT_MSG==
InstrumentedChannel no longer uses metric-schema's staged builders to access a Timer on every request, instead it does a lookup in a caffeine cache.
==COMMIT_MSG==

This achieved a 20% throughput increase, from 9k -> 11k req/sec

```diff
-com.palantir.dialogue.core.Benchmark - Fired off all requests (4864 ms, 9251req/sec)
+com.palantir.dialogue.core.Benchmark - Fired off all requests (4055 ms, 11097req/sec)
```

image

## Possible downsides?
- This isn't JMH precise, so maybe the difference is actually bigger/smaller than my crappy Stopwatch log lines show

### DIFF
--- a/changelog/@unreleased/pr-733.v2.yml
+++ b/changelog/@unreleased/pr-733.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: '`InstrumentedChannel` now allocates fewer objects per RPC call.'
+  links:
+  - https://github.com/palantir/dialogue/pull/733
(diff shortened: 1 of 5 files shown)


ferozco:
I think there's a trade off between performance and simplicity and I'm worried that the additional complexity of a cache and managing the timing calculations ourselves may not be worth avoiding an allocation

carterkozak:
I liked the model based on conjure-undertow where we build EndpointChannels which are constructed for a particular endpoint, allowing up-front calculation isolated from caching individual channel components. Thoughts?

iamdanfox:
Yep I think the per-endpoint build-upfront thing is really what would make life nice here. Just wanted to put this up to show that there is a demonstrable cost (even in crappy benchmarks) to these allocations.

iamdanfox:
Closing out because we've moved this instrumentation to get bound up-front as part of the EndpointChannel refactor.

PR #738, opened 2020-05-12
Concurrency limiters have more backpressure against increases

carterkozak:
This change updates the limiter algorithm to increment by smaller
amounts based on the current limit, so increasing from 1 to 2 takes
one successful response, however higher limits require more
successful responses to increase.

### [Graphs](https://github.com/palantir/dialogue/blob/ckozak/cautious_increase_aggressive_decrease/simulation/src/test/resources/report.md#drastic_slowdownconcurrency_limiter_pin_until_error)
==COMMIT_MSG==
Concurrency limiters have more backpressure against increasing limits
==COMMIT_MSG==

## Possible downsides?
This may end up queueing requests much longer than we would like. This is a significant change to the concurrency limiter implementation.
For example, around ~100 concurrent requests with a limit of 100 requires around 1,000 successful requests to make up for one failure where it would previously only take ten. In most cases our concurrency limiters aren't this high, and when they are we don't expect 1k requests to take very long.


dialogue-core/src/main/java/com/palantir/dialogue/core/CautiousIncreaseAggressiveDecreaseConcurrencyLimiter.java
```diff
@@ -122,36 +126,52 @@ void dropped() {
...
-            public int applyAsInt(int originalLimit, int _inFlightSnapshot) {
-                return Math.max(MIN_LIMIT, (int) (originalLimit * BACKOFF_RATIO));
+            public double applyAsDouble(double originalLimit, double _inFlightSnapshot) {
+                return Math.max(MIN_LIMIT, originalLimit * BACKOFF_RATIO);
             }
         };
     }
 
+    private double accumulateAndGetLimit(int value, DoubleBinaryOperator accumulatorFunction) {
+        AtomicDouble localLimit = limit;
```
iamdanfox, line 165: why save the AtomicDouble into another variable rather than accessing the field?

carterkozak: under contention this can become a very hot loop, the jit doesn't actually assume that final fields are truly final in most cases due to potential reflection hackery (part of the motivation for jigsaw!) and we don't opt into the experimental `-XX:+TrustFinalNonStaticFields`.

Relevant post describing performance improvements to Atomic*FieldUpdaters which special-cased trusting final non-static fields:
https://shipilev.net/blog/2015/faster-atomic-fu/

Some context:
https://bugs.openjdk.java.net/browse/JDK-8132243
https://bugs.openjdk.java.net/browse/JDK-8058164

dialogue-core/src/main/java/com/palantir/dialogue/core/CautiousIncreaseAggressiveDecreaseConcurrencyLimiter.java
```diff
@@ -122,36 +126,52 @@ void dropped() {
...
                 }
                 return originalLimit;
             }
         },
         DROPPED() {
             @Override
-            public int applyAsInt(int originalLimit, int _inFlightSnapshot) {
-                return Math.max(MIN_LIMIT, (int) (originalLimit * BACKOFF_RATIO));
+            public double applyAsDouble(double originalLimit, double _inFlightSnapshot) {
+                return Math.max(MIN_LIMIT, originalLimit * BACKOFF_RATIO);
```
iamdanfox: interestingly I think taking away the cast has made this actually backoff _slower_ than before, as going 3d * 0.9 comes out as a double 2.7, whereas previously the integer cast would have taken this down to 2.

I think you can see this happening in the `all_nodes_500` case: https://github.com/palantir/dialogue/blob/ckozak/cautious_increase_aggressive_decrease/simulation/src/test/resources/report.md#all_nodes_500concurrency_limiter_pin_until_error

carterkozak:
I imagine the tighter concurrency limiters will help us reuse more connections instead of creating new ones, but the drawback is the potential to p0 folks if it turns out we got too aggressive. We should keep a close eye on metrics once we start to roll these out.

iamdanfox:
Can you cherry-pick the test from https://github.com/palantir/dialogue/pull/720 or target the PR onto that? I really want to make sure whatever solution we go with actually works for jakub's internal-ski-product

carterkozak:
Ack, cherry-picked.

dialogue-core/src/main/java/com/palantir/dialogue/core/CautiousIncreaseAggressiveDecreaseConcurrencyLimiter.java
iamdanfox: I _knew_ there was gonna be some beautiful java perf reason under all this 😂. Could we capture that first sentence in a comment maybe?

dialogue-core/src/main/java/com/palantir/dialogue/core/CautiousIncreaseAggressiveDecreaseConcurrencyLimiter.java
iamdanfox: @carterkozak thoughts on putting a `Math.floor` on this? it would have a big effect when the limit is small but not change much when someone is going from 101 -> 90.9?

carterkozak: Hah, was already running simulations on it when this came through ;-)

iamdanfox (approved):
Nice, thanks @carterkozak! I'm really enjoying being able to CR these subtle concurrency limiter PRs in a sane way with graphs to look at - think this would have been a lot more painful in c-j-r land.

iamdanfox:
I think this will also make the concurrency limiters we observe on prod a _lot_ more sane... I don't really believe that a lot of these servers are actually using 800 concurrency connections!

raiju:
Can we broadcast this (type of) change at least to GLs? Mainly as a heads-up.

carterkozak:
:+1:

PR #739, opened 2020-05-12
[test-only] Run all simulations in parallel

iamdanfox:
## Before this PR

Each simulation case is single-threaded, but we run all of them serially. Given that we're CPU bound, this is a bit lame and it means that one core heats up while others do nothing.

## After this PR
==COMMIT_MSG==
By switching to JUnit5, we can run each simulation test independently from each other. Running a single simulation case still happens on one thread, we can just run n of them at the same time.
==COMMIT_MSG==

Locally, I can now run the entire SimulationTest in ~30 sec.

Screenshot 2020-05-12 at 18 46 19


carterkozak:
`BUILD SUCCESSFUL in 13s` I like it! :+1:

PR #741, opened 2020-05-12
Declare precise deps with api/implementation

iamdanfox:
## Before this PR

I ran a `./gradlew build --scan` locally and saw a bunch of deprecation warnings including this: 

> The compile configuration has been deprecated for dependency declaration. 
> This will fail with an error in Gradle 7.0.

https://docs.gradle.org/6.4/userguide/upgrading_version_5.html#dependencies_should_no_longer_be_declared_using_the_compile_and_runtime_configurations

`compile` is bad because it allows library consumers to start relying on our internal implementation details.

## After this PR
==COMMIT_MSG==
Published jars now more precisely declare their dependencies with Gradle's `api` and `implementation`, so that implementation details will appear in the `runtime` POM scope and be unavailable at compile time for consumers by default.
==COMMIT_MSG==


dialogue-clients/build.gradle
```diff
@@ -22,7 +22,7 @@ dependencies {
     implementation 'org.slf4j:slf4j-api'
 
     annotationProcessor 'org.immutables:value'
-    compile 'org.immutables:value::annotations'
+    api 'org.immutables:value::annotations'
```
ferozco: shouldn't this be compileOnly?

dialogue-core/build.gradle
```diff
@@ -26,8 +26,9 @@ dependencies {
     testImplementation project(':dialogue-test-common')
 
     annotationProcessor 'org.immutables:value'
+    api 'org.immutables:value::annotations'
```
ferozco: same here

ferozco (commented):
just question about annotions otherwise looks great

dialogue-clients/build.gradle
iamdanfox: You'd think that, but it doesn't play nicely with `-Werror`:
```
> Task :dialogue-apache-hc4-client:compileJava FAILED
warning: unknown enum constant ImplementationVisibility.PACKAGE
  reason: class file for org.immutables.value.Value$Style$ImplementationVisibility not found
error: warnings found and -Werror specified
Note: /Volumes/git/dialogue/dialogue-apache-hc4-client/src/main/java/com/palantir/dialogue/hc4/ApacheHttpClientBlockingChannel.java uses or overrides a deprecated API.
Note: Recompile with -Xlint:deprecation for details.
1 error
1 warning
```

Also

```
> Task :dialogue-core:compileTestJava FAILED
warning: unknown enum constant ImplementationVisibility.PACKAGE
  reason: class file for org.immutables.value.Value$Style$ImplementationVisibility not found
error: warnings found and -Werror specified
1 error
1 warning
```

iamdanfox: EDIT nvm I just added a bunch compileOnly's and it works now :)

ferozco:
👍

PR #748, opened 2020-05-14
Dialogue clients do not fail when a Content-Length header is provided

carterkozak:
==COMMIT_MSG==
Dialogue clients do not fail when a Content-Lenght header is provided
==COMMIT_MSG==

### DIFF
--- a/dialogue-apache-hc4-client/src/main/java/com/palantir/dialogue/hc4/ApacheHttpClientBlockingChannel.java
+++ b/dialogue-apache-hc4-client/src/main/java/com/palantir/dialogue/hc4/ApacheHttpClientBlockingChannel.java
@@ -26,6 +26,7 @@ import com.palantir.dialogue.Response;
 import com.palantir.dialogue.blocking.BlockingChannel;
 import com.palantir.dialogue.core.BaseUrl;
 import com.palantir.logsafe.Preconditions;
+import com.palantir.logsafe.SafeArg;
 import com.palantir.logsafe.exceptions.SafeRuntimeException;
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
@@ -33,6 +34,7 @@ import java.io.InputStream;
 import java.io.OutputStream;
 import java.net.URL;
 import java.util.Optional;
+import java.util.OptionalLong;
 import javax.annotation.Nullable;
 import org.apache.http.Header;
 import org.apache.http.HeaderIterator;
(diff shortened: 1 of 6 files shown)


carterkozak:
Not sure this is ideal, PR for discussion.

carterkozak:
Note that this would allow us to set the known content-length from the feign jaxrs shim as well, but we may want to add an OptionalLong to the RequestBody interface instead of pulling data from the request headers (same as we do for content-type) though that wouldn't be helpful for @jackwickham.

dialogue-target/src/main/java/com/palantir/dialogue/RequestBody.java
```diff
@@ -31,6 +32,13 @@
     /** Returns <pre>true</pre> if {@link #writeTo(OutputStream)} may be invoked multiple times. */
     boolean repeatable();
 
+    /** Returns the {@code Content-Length} of the request body, if known. Note that there is no guarantee that
+     * this will result in a {@code Content-Length} request rather than {@code Transfer-Encoding: chunked}.
+     */
+    default OptionalLong length() {
+        return OptionalLong.empty();
+    }
```
iamdanfox: I thought we weren't intending to expose this in `:dialogue-target`, but just say if users happen to set a `Content-Length` header we might choose to respect it?

carterkozak: I could go either way on this — the changes to this class are in an isolated commit.
We definitely aren’t including any changes to BinaryRequestBody, conjure consumers don’t interact with this interface though.

ferozco: I think for now I'd prefer to go with respecting the `Content-Length` header if the user provides it rather that extending the Request interface

carterkozak: reverted that commit.

dialogue-apache-hc4-client/src/main/java/com/palantir/dialogue/hc4/ApacheHttpClientBlockingChannel.java
```diff
@@ -215,7 +239,9 @@ public boolean isStreaming() {
         }
 
         @Override
-        public void consumeContent() {}
+        public void consumeContent() {
+            requestBody.close();
```
ferozco, line 243: cheeky little fix

ferozco (approved):

PR #751, opened 2020-05-15
Eliminate an Optional#map from a PinUntilErrorChannel hot codepath

iamdanfox:
## Before this PR

While trying to make the simulations go faster (so I can simulate hours of requests instead of minutes), I noticed this method was burning a little bit of our CPU time.

## After this PR
==COMMIT_MSG==
PinUntilError no longer allocates a lambda per request
==COMMIT_MSG==

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/PinUntilErrorNodeSelectionStrategyChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/PinUntilErrorNodeSelectionStrategyChannel.java
@@ -131,28 +131,33 @@ final class PinUntilErrorNodeSelectionStrategyChannel implements LimitedChannel 
         int pin = currentPin.get();
         PinChannel channel = nodeList.get(pin);
 
-        return channel.maybeExecute(endpoint, request)
-                .map(future -> DialogueFutures.addDirectCallback(future, new FutureCallback<Response>() {
-                    @Override
-                    public void onSuccess(Response response) {
-                        // We specifically don't switch  429 responses to support transactional
-                        // workflows where it is important for a large number of requests to all land on the same node,
-                        // even if a couple of them get rate limited in the middle.
-                        if (Responses.isServerError(response)
-                                || (Responses.isQosStatus(response) && !Responses.isTooManyRequests(response))) {
-                            OptionalInt next = incrementHostIfNecessary(pin);
-                            instrumentation.receivedErrorStatus(pin, channel, response, next);
-                        } else {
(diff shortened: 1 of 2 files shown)


dialogue-core/src/main/java/com/palantir/dialogue/core/PinUntilErrorNodeSelectionStrategyChannel.java
```diff
@@ -131,28 +131,33 @@ LimitedChannel getCurrentChannel() {
...
-
-                    @Override
-                    public void onFailure(Throwable throwable) {
-                        OptionalInt next = incrementHostIfNecessary(pin);
-                        instrumentation.receivedThrowable(pin, channel, throwable, next);
-                    }
-                }));
+        Optional<ListenableFuture<Response>> maybeResponse = channel.maybeExecute(endpoint, request);
+        if (!maybeResponse.isPresent()) {
+            return Optional.empty();
```
carterkozak, line 136: nice

dialogue-core/src/main/java/com/palantir/dialogue/core/PinUntilErrorNodeSelectionStrategyChannel.java
```diff
@@ -131,28 +131,33 @@ LimitedChannel getCurrentChannel() {
         int pin = currentPin.get();
         PinChannel channel = nodeList.get(pin);
```
carterkozak, line 132: If we want to take this another step, the only state the FutureCallback requires is the pin, so we could allocate one callback for each PinChannel up front and reuse them.

carterkozak (approved):
lgtm, thanks!

dialogue-core/src/main/java/com/palantir/dialogue/core/PinUntilErrorNodeSelectionStrategyChannel.java
iamdanfox: I _almost_ did this, but I think the upside is gonna be pretty marginal compared for quite a big readability cost - all of these optimizations are dwarfed by how slow the instrumented channel is

carterkozak: I haven't actually found the instrumented channel helpful, it's too broad to provide per-endpoint information, and not helpful from a high level because it's almost entirely dependent upon the server handling.
What if we removed the timer? Alternatively what if we removed the service-name tag from the timer? Either way we can cut dd spend and I don't think we lose any actionable data.

PR #752, opened 2020-05-15
Eliminate stream, findFirst from BalancedNodeSelectionStrategyChannel

iamdanfox:
## Before this PR

Again, I'm trying to make SimulationTests faster so I can reasonably simulate larger volumes of fast requests over longer amounts of time. Turns out, this method showed up in my flamegraphs.

## After this PR
==COMMIT_MSG==
BalancedNodeSelectionStrategyChannel is slightly more performant as it uses plain loops instead of java Streams (which necessitate the use of lambdas)
==COMMIT_MSG==

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/BalancedNodeSelectionStrategyChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/BalancedNodeSelectionStrategyChannel.java
@@ -91,13 +91,25 @@ final class BalancedNodeSelectionStrategyChannel implements LimitedChannel {
 
         // TODO(dfox): P2C optimization when we have high number of nodes to save CPU?
         // http://www.eecs.harvard.edu/~michaelm/NEWWORK/postscripts/twosurvey.pdf
-        return preShuffled.stream()
-                .map(MutableChannelWithStats::computeScore)
-                .sorted(Comparator.comparingInt(SortableChannel::getScore))
-                .map(channel -> channel.delegate.maybeExecute(endpoint, request))
-                .filter(Optional::isPresent)
-                .map(Optional::get)
-                .findFirst();
+        List<SortableChannel> sortedByScore = sortByScore(preShuffled);
+
+        for (SortableChannel channel : sortedByScore) {
+            Optional<ListenableFuture<Response>> maybe = channel.delegate.maybeExecute(endpoint, request);
+            if (maybe.isPresent()) {
+                return maybe;
+            }
+        }
+
+        return Optional.empty();
+    }
+
+    private static List<SortableChannel> sortByScore(List<MutableChannelWithStats> preShuffled) {
+        List<SortableChannel> sortedByScore = new ArrayList<>();
+        for (MutableChannelWithStats channel : preShuffled) {
+            sortedByScore.add(channel.computeScore());
+        }
+        sortedByScore.sort(Comparator.comparingInt(SortableChannel::getScore));
+        return sortedByScore;
     }
 
     /** Returns a new shuffled list, without mutating the input list (which may be immutable). */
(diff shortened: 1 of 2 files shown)


dialogue-core/src/main/java/com/palantir/dialogue/core/BalancedNodeSelectionStrategyChannel.java
```diff
@@ -91,13 +91,25 @@
...
-        return preShuffled.stream()
-                .map(MutableChannelWithStats::computeScore)
-                .sorted(Comparator.comparingInt(SortableChannel::getScore))
-                .map(channel -> channel.delegate.maybeExecute(endpoint, request))
-                .filter(Optional::isPresent)
-                .map(Optional::get)
-                .findFirst();
+        List<SortableChannel> sortedByScore = sortByScore(preShuffled);
+
+        for (SortableChannel channel : sortedByScore) {
```
carterkozak: If we're going this far, we might as well avoid the iterator as well :-)
```suggestion
        for (int i = 0; i < sortedByScore.size(); i++) {
            SortableChannel channel = sortedByScore.get(i);
```

dialogue-core/src/main/java/com/palantir/dialogue/core/BalancedNodeSelectionStrategyChannel.java
```diff
@@ -91,13 +91,25 @@
...
+            if (maybe.isPresent()) {
+                return maybe;
+            }
+        }
+
+        return Optional.empty();
+    }
+
+    private static List<SortableChannel> sortByScore(List<MutableChannelWithStats> preShuffled) {
+        List<SortableChannel> sortedByScore = new ArrayList<>();
```
carterkozak: nit: presize
```suggestion
        List<SortableChannel> sortedByScore = new ArrayList<>(preShuffled.size());
```

dialogue-core/src/main/java/com/palantir/dialogue/core/BalancedNodeSelectionStrategyChannel.java
```diff
@@ -91,13 +91,25 @@
...
+        for (MutableChannelWithStats channel : preShuffled) {
+            sortedByScore.add(channel.computeScore());
+        }
+        sortedByScore.sort(Comparator.comparingInt(SortableChannel::getScore));
```
carterkozak: We can extract the comparator to a static field

dialogue-core/src/main/java/com/palantir/dialogue/core/BalancedNodeSelectionStrategyChannel.java
```diff
@@ -91,13 +91,25 @@
...
+                return maybe;
+            }
+        }
+
+        return Optional.empty();
+    }
+
+    private static List<SortableChannel> sortByScore(List<MutableChannelWithStats> preShuffled) {
+        List<SortableChannel> sortedByScore = new ArrayList<>();
+        for (MutableChannelWithStats channel : preShuffled) {
```
carterkozak: could use a standard for loop + index based accessor

dialogue-core/src/main/java/com/palantir/dialogue/core/BalancedNodeSelectionStrategyChannel.java
iamdanfox: Or maybe even just implement comparable?

carterkozak (approved):
lgtm, this approach is better than the original, defer to your judgement for the additional potential optimizations.

dialogue-core/src/main/java/com/palantir/dialogue/core/BalancedNodeSelectionStrategyChannel.java
carterkozak: Yep, the two should be equivalent from a performance point of view. I slightly prefer comparators over implementing compareTo, but no strong opinion.

carterkozak (approved):
👍 thanks!

PR #753, opened 2020-05-15
[test-only] speed up simulations

iamdanfox:
## Before this PR

Surprise surprise, a big chunk of slowness of our simulations was due to metric reporting and the actual benchmark harness. Reducing the amount of thrown exceptions and actually the `asString` method gives us a little speedup.

Turning off trace sampling is a decent speedup too.

## After this PR
==COMMIT_MSG==
SimulationTest is a bit faster (now completes in ~5 seconds)
==COMMIT_MSG==

### DIFF
--- a/dialogue-test-common/src/main/java/com/palantir/dialogue/TestResponse.java
+++ b/dialogue-test-common/src/main/java/com/palantir/dialogue/TestResponse.java
@@ -23,7 +23,7 @@ import com.palantir.logsafe.exceptions.SafeRuntimeException;
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
-import java.util.Optional;
+import java.util.concurrent.atomic.AtomicBoolean;
 import javax.annotation.Nullable;
 import javax.ws.rs.core.HttpHeaders;
 
@@ -31,7 +31,7 @@ public final class TestResponse implements Response {
 
     private final CloseRecordingInputStream inputStream;
 
-    private Optional<Throwable> closeCalled = Optional.empty();
+    private AtomicBoolean closeCalled = new AtomicBoolean(false);
     private int code = 0;
     private ListMultimap<String, String> headers = ImmutableListMultimap.of();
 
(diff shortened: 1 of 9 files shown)


dialogue-test-common/src/main/java/com/palantir/dialogue/TestResponse.java
```diff
@@ -23,15 +23,15 @@
...
+import java.util.concurrent.atomic.AtomicBoolean;
 import javax.annotation.Nullable;
 import javax.ws.rs.core.HttpHeaders;
 
 public final class TestResponse implements Response {
 
     private final CloseRecordingInputStream inputStream;
 
-    private Optional<Throwable> closeCalled = Optional.empty();
+    private AtomicBoolean closeCalled = new AtomicBoolean(false);
```
carterkozak: Might as well just use a bog-standard non-volatile boolean field given simulations are individually single-threaded.

dialogue-test-common/src/main/java/com/palantir/dialogue/TestResponse.java
```diff
@@ -72,20 +72,20 @@ public TestResponse code(int value) {
     public void close() {
         checkNotClosed();
         try {
-            closeCalled = Optional.of(new SafeRuntimeException("Close called here"));
+            closeCalled.compareAndSet(false, true);
```
carterkozak: boolean cas isn't necessary, once this is called we're always in the `true` state ;-)

simulation/src/main/java/com/palantir/dialogue/core/MetricNames.java
```diff
@@ -23,20 +23,27 @@
...
+            MetricName.builder().safeName("responseClose").build();
+    private static final MetricName GLOBAL_RESPONSES =
+            MetricName.builder().safeName("globalResponses").build();
+    private static final MetricName GLOBAL_SERVER_TIME =
+            MetricName.builder().safeName("globalServerTime").build();
+
     /** Counter incremented every time a {@code Response} is closed. */
     static Counter responseClose(TaggedMetricRegistry reg) {
-        return reg.counter(MetricName.builder().safeName("responseClose").build());
+        return reg.counter(RESPONSE_CLOSE);
```
carterkozak, line 36: This is cleaner than before imo, but we should update the call sites to look up the metric once per simulation rather than per-request.

carterkozak:
👍

PR #757, opened 2020-05-15
Remove `service-name` tag from `client.response` timer

carterkozak:
This reduces pressure on our indexers, timers and histograms
are relatively heavy weight, and without per-endpoint tags
these timers aren't terribly helpful to service developers.

==COMMIT_MSG==
Remove `service-name` tag from `client.response` timer
==COMMIT_MSG==

## Possible downsides?
service-name is no longer included in the metric. It wasn't very helpful before because we didn't include the endpoint-name either, which would strain the observability pipeline.


ferozco (commented):
Generally supportive of this change 👍

iamdanfox:
Chatted over zoom after discussion in #dev-rpc - resolved to stick with the metrics we have now because it provides exact parity with c-j-r and allows dashboards to seamlessly continue working. Additionally, the CPU cost is now minimized because we're doing the endpointchannel upfront thing.

PR #763, opened 2020-05-18
[test only] Simulations don't sample traces

carterkozak:
==COMMIT_MSG==
Simulations don't sample traces
==COMMIT_MSG==

### DIFF
--- a/simulation/src/test/java/com/palantir/dialogue/core/SimulationTest.java
+++ b/simulation/src/test/java/com/palantir/dialogue/core/SimulationTest.java
@@ -52,6 +52,7 @@ import java.util.stream.IntStream;
 import java.util.stream.Stream;
 import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeAll;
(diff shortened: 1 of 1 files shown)


simulation/src/test/java/com/palantir/dialogue/core/SimulationTest.java
```diff
@@ -103,6 +104,11 @@
     @ParameterizedTest
     @interface SimulationCase {}
 
+    @BeforeAll
+    static void beforeAll() {
+        Tracer.setSampler(() -> false);
+    }
```
iamdanfox, line 110: I had this in @Before? https://github.com/palantir/dialogue/blob/707b92aaeac49ff316b770837770a2edca07504d/simulation/src/test/java/com/palantir/dialogue/core/SimulationTest.java#L624

carterkozak: doh, missed that.

carterkozak: Ressurrected this from an old branch, I think that may not have existed when I wrote this, also very possible I missed it.

PR #764, opened 2020-05-18
EndpointChannels allow pre-computing stuff upfront

iamdanfox:
## Before this PR

We're leaving performance on the table by recomputing things based on endpoint info (e.g. in the UserAgent channel, we do some string concatenation on every single request based on the Endpoint serviceName). This pattern happens in a bunch of places (e.g. building up tagged metrics using metric schema). Given that we know the complete set of endpoints at codegen time, we should be able to do something like Witchcraft's approach of binding everything upfront.

## After this PR: 30% speedup
==COMMIT_MSG==
Codegen can take advantage of a new methods to do endpoint-specific work once upfront, instead of spending CPU cycles doing it again and again on every request.
==COMMIT_MSG==

This does introduce some new public API:

- `EndpointChannel`
- `ChannelEndpointStage`
- some methods on the `runtime.clients()` object

After merging this, I'd need to go and change the conjure-java codegen to use this.

```
Benchmark                             Mode  Cnt      Score      Error  Units
EndToEndBenchmark.dialogue2Blocking  thrpt   16  25647.619 ± 3542.047  ops/s
EndToEndBenchmark.dialogueBlocking   thrpt   16  19812.490 ± 3081.670  ops/s
```

## Possible downsides?
- From an API-design perspective, we now have an alternative way of doing the same thing as `Channel`, which could be confusing from a maintenance perspective
- We have one `instanceof` check which sends people down a different codepath if they pass in a bog standard `Channel` vs if they pass in a `ChannelEndpointStage`
- Until I can land the corresponding codegen in conjure-java, this will cause every RPC call to allocate a O(tens) more objects, so I think there will be a slight perf hit.


dialogue-example/dialogue-example-dialogue/src/main/java/com/palantir/dialogue/example/SampleServiceAsync2.java
```diff
@@ -0,0 +1,132 @@
...
+            private final Deserializer<AliasOfOptional> getMyAliasDeserializer =
+                    runtime.bodySerDe().deserializer(new TypeMarker<AliasOfOptional>() {});
+
+            private final Deserializer<AliasOfAliasOfOptional> getMyAlias2Deserializer =
+                    runtime.bodySerDe().deserializer(new TypeMarker<AliasOfAliasOfOptional>() {});
+
+            @Override
+            public ListenableFuture<Void> voidToVoid() {
+                Request.Builder request = Request.builder();
+                return runtime.clients().call(voidToVoidChannel, request.build(), voidToVoidDeserializer);
```
iamdanfox, line 88: open question - should we actually just bind the `voidToVoidDeserializer` upfront too? This would save us 1 (one) allocation of `deserializer::deserialize`

dialogue-target/src/main/java/com/palantir/dialogue/BindEndpoint.java
```diff
@@ -0,0 +1,21 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.dialogue;
+
+public interface BindEndpoint {
+    EndpointChannel bindEndpoint(Endpoint endpoint);
+}
```
iamdanfox: this is public because `DefaultClients` lives in `:dialogue-serde` and doesn't depend on `:dialogue-core`, but needs to be able to bind DialogueClients.

carterkozak: Could we write things through the runtime object? I'd like the runtime to be capable of taking a channel and endpoint, and returning an EndpointChannel.

dialogue-example/dialogue-example-dialogue/src/main/java/com/palantir/dialogue/example/SampleServiceAsync2.java
```diff
@@ -0,0 +1,132 @@
...
+    /**
+     * Creates an asynchronous/non-blocking client for a SampleService service.
+     */
+    static SampleServiceAsync of(Channel channel, ConjureRuntime runtime) {
+        return new SampleServiceAsync() {
+            private final PlainSerDe plainSerDe = runtime.plainSerDe();
+
+            // THIS IS THE BIND THINGY
+            private final EndpointChannel voidToVoidChannel =
+                    runtime.clients().bindEndpoint(channel, DialogueSampleEndpoints.voidToVoid);
```
iamdanfox: @carterkozak I think the PR does the thing you describe? here we're saying to the runtime here's a `Channel` and and `Endpoint`, please give me an EndpointChannel.

dialogue-target/src/main/java/com/palantir/dialogue/BindEndpoint.java
iamdanfox: Is this kinda what you were imagining? https://github.com/palantir/dialogue/pull/764/files#r426760705

iamdanfox: We could move this particular interface if we're happy to change some inter-project dependencies (e.g. if we made `:core` depend on `:serde` or vice versa)

dialogue-example/dialogue-example-dialogue/src/main/java/com/palantir/dialogue/example/SampleServiceAsync2.java
carterkozak: Got it. I'm a bit worried about the instanceof checks, I think it may be safer to take a semantic break in the runtime and require the bind method instead of directly executing against the channel.

.palantir/revapi.yml
```diff
@@ -186,3 +186,15 @@ acceptedBreaks:
...
+    com.palantir.dialogue:dialogue-target:
+    - code: "java.method.addedToInterface"
+      new: "method <T> com.google.common.util.concurrent.ListenableFuture<T> com.palantir.dialogue.Clients::call(com.palantir.dialogue.EndpointChannel,\
+        \ com.palantir.dialogue.Request, com.palantir.dialogue.Deserializer<T>)"
+      justification: "Nobody is implementing ConjureRuntime yet, and this is a compile-time\
+        \ break not ABI break"
+    - code: "java.method.addedToInterface"
+      new: "method com.palantir.dialogue.EndpointChannel com.palantir.dialogue.Clients::bindEndpoint(com.palantir.dialogue.Channel,\
+        \ com.palantir.dialogue.Endpoint)"
+      justification: "Nobody is implementing ConjureRuntime yet, and this is a compile-time\
```
carterkozak: we also document `no guarantees are made for custom implementations` :-)

dialogue-core/src/main/java/com/palantir/dialogue/core/DeprecationWarningChannel.java
```diff
@@ -37,42 +38,47 @@
...
-                                        SafeArg.of(
-                                                "service",
-                                                response.getFirstHeader("server")
-                                                        .orElse("no server header provided")));
-                            }
-                        });
+    public EndpointChannel endpoint(Endpoint endpoint) {
+        EndpointChannel proceed = delegate.endpoint(endpoint);
+        FutureCallback<Response> callback = createCallback(endpoint);
+        return req -> DialogueFutures.addDirectCallback(proceed.execute(req), callback);
```
carterkozak: super minor nit: may want to use something concrete with a toString for debugging.

dialogue-example/dialogue-example-dialogue/src/main/java/com/palantir/dialogue/example/SampleServiceAsync2.java
iamdanfox: Ok I think I've got rid of all the `instanceof` checks apart from the one in the runtime (which seems pretty unavoidable to me), as we still want people to be able to pass in a bog-standard `Channel` and have their FooBlocking still work

```java
        if (channel instanceof ChannelEndpointStage) {
            return ((ChannelEndpointStage) channel).endpoint(endpoint);
        }
```

dialogue-core/src/main/java/com/palantir/dialogue/core/DeprecationWarningChannel.java
iamdanfox: Yep I'm very much aware that I added all these nice toStrings and now lambdas & anonymous classes will have ruined them all.

dialogue-target/src/main/java/com/palantir/dialogue/EndpointChannel.java
```diff
@@ -0,0 +1,29 @@
...
+
+package com.palantir.dialogue;
+
+import com.google.common.util.concurrent.ListenableFuture;
+
+/**
+ * {@code EndpointChannel} is an abstraction of a transport layer (e.g., HTTP) that is used client stubs.
+ *
+ * This is a version of {@link Channel} that always sends the given request to a predefined {@link Endpoint}.
+ * Constructed using {@link ChannelEndpointStage#endpoint}.
```
iamdanfox: maybe the class used to construct one of these should actually be called an `EndpointChannelBuilder`?

dialogue-target/src/main/java/com/palantir/dialogue/Clients.java
```diff
@@ -30,6 +30,10 @@
      */
     <T> ListenableFuture<T> call(Channel channel, Endpoint endpoint, Request request, Deserializer<T> deserializer);
```
iamdanfox, line 38: Thoughts on just deprecating this guy?

dialogue-target/src/main/java/com/palantir/dialogue/EndpointChannel.java
ferozco: Perhaps `EndpointChannelFactory`?

dialogue-core/src/main/java/com/palantir/dialogue/core/NeverThrowChannel.java
```diff
@@ -52,4 +61,27 @@
     public String toString() {
         return "NeverThrowChannel{" + delegate + '}';
     }
+
+    private static final class NeverThrowEndpointChannel implements EndpointChannel {
```
ferozco, line 61: I'm not a huge fan of all of this duplication :(

dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueChannel.java
```diff
@@ -250,7 +255,21 @@ public Channel buildNonLiveReloading() {
...
+
+                @Override
+                public ListenableFuture<Response> execute(Endpoint endpoint, Request request) {
+                    return channel.endpoint(endpoint).execute(request);
+                }
+            };
         }
     }
+
+    interface Channel2 extends Channel, ChannelEndpointStage {}
```
ferozco: nit: rename to something more descriptive, perhaps: `CompositeChannel`?

ferozco (commented):
just some quick thoughts, going to play around with it for a bit

dialogue-core/src/main/java/com/palantir/dialogue/core/NeverThrowChannel.java
iamdanfox: Yeah this is because this class is used in both the per-uri 'channel' style (below the NSS) and also in the 'precompiled endpoint' style above the NSS.

I _think_ it might actually make sense to move the NeverThrow logic into the runtime, because actually if users provide a non-sensical Channel implementation it would actually be nice to still catch mistakes and turn them into failedFutures...

ferozco:
quick sketch of a different approach that consolidates the channel factory logic into a single location instead of being within each channel: https://github.com/palantir/dialogue/compare/dfox/endpoint-channel...fo/proposal?expand=1

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/DefaultClients.java
```diff
@@ -52,13 +54,38 @@
...
 
+    @Override
+    public EndpointChannel bind(Channel channel, Endpoint endpoint) {
+        if (channel instanceof EndpointChannelFactory) {
+            return ((EndpointChannelFactory) channel).endpoint(endpoint);
+        }
+
+        if (log.isInfoEnabled()) {
+            log.info(
+                    "Channel of type {} does not implement ChannelEndpointStage, "
```
iamdanfox: type name changed. let's switch to debug

dialogue-target/src/main/java/com/palantir/dialogue/EndpointChannel.java
```diff
@@ -0,0 +1,29 @@
...
+
+package com.palantir.dialogue;
+
+import com.google.common.util.concurrent.ListenableFuture;
+
+/**
+ * {@code EndpointChannel} is an abstraction of a transport layer (e.g., HTTP) that is used client stubs.
+ *
+ * This is a version of {@link Channel} that always sends the given request to a predefined {@link Endpoint}.
+ * Constructed using {@link EndpointChannelFactory#endpoint}.
```
carterkozak: nit: wouldn't necessarily note how this is constructed

dialogue-core/src/main/java/com/palantir/dialogue/core/ActiveRequestInstrumentationChannel.java
```diff
@@ -56,8 +57,41 @@
         return DialogueFutures.addDirectListener(delegate.execute(endpoint, request), counter::dec);
     }
 
+    static EndpointChannel create(
```
ferozco, line 60: why do we need an EndpointChannel and Channel version of this? Should they just be separate classes rather than nested?

dialogue-example/dialogue-example-dialogue/src/main/java/com/palantir/dialogue/example/SampleServiceAsync2.java
```diff
@@ -0,0 +1,132 @@
...
+    @Override
+    ListenableFuture<AliasOfOptional> getMyAlias();
+
+    @Override
+    ListenableFuture<AliasOfAliasOfOptional> getMyAlias2();
+
+    /**
+     * Creates an asynchronous/non-blocking client for a SampleService service.
+     */
+    static SampleServiceAsync of(Channel channel, ConjureRuntime runtime) {
```
ferozco, line 55: did we decide to go with a separate factory method in codegen?

dialogue-target/src/main/java/com/palantir/dialogue/EndpointChannelFactory.java
```diff
@@ -0,0 +1,31 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.dialogue;
+
+/**
+ * A 'staged' version of {@link Channel}, that allows implementors to precompute necessary objects (e.g. counter
```
ferozco, line 20: nit: update docs

dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueChannel.java
```diff
@@ -250,7 +259,21 @@ public Channel buildNonLiveReloading() {
...
+
+                @Override
+                public ListenableFuture<Response> execute(Endpoint endpoint, Request request) {
+                    return channel.endpoint(endpoint).execute(request);
+                }
+            };
         }
     }
+
+    interface Channel2 extends Channel, EndpointChannelFactory {}
```
ferozco: nit: rename to `CompositeChannel`

dialogue-core/src/test/java/com/palantir/dialogue/core/ContentDecodingChannelTest.java
```diff
@@ -45,21 +44,21 @@ public void testDecoding() throws Exception {
             throw new IllegalStateException(e);
         }
         Response response = new ContentDecodingChannel(
-                        (endpoint, request) -> Futures.immediateFuture(new TestResponse(out.toByteArray())
+                        request -> Futures.immediateFuture(new TestResponse(out.toByteArray())
```
ferozco, line 47: it is nice to reduce a bit of boilerplate in tests :)

dialogue-core/src/test/java/com/palantir/dialogue/core/RetryingChannelTest.java
```diff
@@ -55,140 +55,144 @@
     private static final Request REQUEST = Request.builder().build();
 
     @Mock
-    private Channel channel;
+    private EndpointChannel channel;
 
     @Test
     public void testNoFailures() throws ExecutionException, InterruptedException {
-        when(channel.execute(any(), any())).thenReturn(SUCCESS);
+        when(channel.execute(any())).thenReturn(SUCCESS);
```
ferozco, line 62: assume the test changes are just refactors, so i'm just skimming

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/DefaultClients.java
```diff
@@ -52,13 +54,38 @@
...
+                            + "which is recommended for maximum performance. Falling back to lambda impl.",
+                    SafeArg.of("type", channel.getClass().getSimpleName()),
+                    SafeArg.of("serviceName", endpoint.serviceName()),
+                    SafeArg.of("endpointName", endpoint.endpointName()),
+                    UnsafeArg.of("channel", channel),
+                    new SafeRuntimeException("Exception for stacktrace"));
+        }
+        return new EndpointChannelAdapter(endpoint, channel);
```
ferozco, line 86: could avoid the extra allocation by inlining the dispatch in `call`

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/DefaultClients.java
```diff
@@ -164,4 +191,24 @@ private static Request accepting(Request original, String acceptValue) {
                 .putHeaderParams(HttpHeaders.ACCEPT, acceptValue)
                 .build();
     }
+
+    private static final class EndpointChannelAdapter implements EndpointChannel {
```
ferozco, line 195: duplicate class?

dialogue-example/dialogue-example-dialogue/src/main/java/com/palantir/dialogue/example/SampleServiceBlocking2.java
```diff
@@ -0,0 +1,79 @@
...
+import com.palantir.dialogue.ConjureRuntime;
+import com.palantir.ri.ResourceIdentifier;
+import java.io.InputStream;
+import java.time.OffsetDateTime;
+import java.util.List;
+import java.util.Optional;
+import javax.annotation.Generated;
+
+@Generated("com.palantir.conjure.java.services.dialogue.DialogueInterfaceGenerator")
+public interface SampleServiceBlocking2 {
```
iamdanfox, line 29: Corresponding codegen change: https://github.com/palantir/conjure-java/pull/910

dialogue-core/src/main/java/com/palantir/dialogue/core/ActiveRequestInstrumentationChannel.java
carterkozak: We use this for two separate stages, once around wire requests, and once around top-level 'processing' requests that includes those waiting in a queue/retry.

iamdanfox: I think separately I'd like to revaluate all our metrics and figure out the cost/benefit for each of them (both in $ and CPU)

dialogue-core/src/test/java/com/palantir/dialogue/core/RetryingChannelTest.java
iamdanfox: yup, didn't add/remove any of em.

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/DefaultClients.java
iamdanfox: Not quite, one is in :serde and the other is in :core, we don't have a dep in either direction. I named em the same just because they're identical

ferozco:
Looks reasonable to me, I'll let @carterkozak take a look too before we ship this

dialogue-target/src/main/java/com/palantir/dialogue/Clients.java
```diff
@@ -24,12 +24,21 @@
...
+     * Makes a request to the specified {@link EndpointChannel} and deserializes the response using a provided
+     * deserializer.
+     */
+    <T> ListenableFuture<T> call(EndpointChannel channel, Request request, Deserializer<T> deserializer);
 
     /**
      * Makes a request to the specified {@link Endpoint} and deserializes the response using a provided deserializer.
+     * @deprecated prefer {@link #bind} as this allows pre-computing values to save CPU
      */
+    @Deprecated
```
carterkozak, line 37: This may cause issues compiling generated code before the generator is upgraded. May want to hold off a day or two before adding a deprecation.

dialogue-target/src/main/java/com/palantir/dialogue/EndpointChannelFactory.java
```diff
@@ -0,0 +1,31 @@
...
+ * lookups) based on the provided {@code endpoint}, saving CPU cycles on each call.
+ */
+public interface EndpointChannelFactory {
+
+    /**
+     * Construct a new {@link EndpointChannel} which will send all requests to the given {@link Endpoint}. This
+     * method is expected to be called once at startup time. Behaviour is undefined if called with the same endpoint
+     * many times.
```
carterkozak, line 28: Not sure how undefined this can really be -- I think we'll need to do something along those lines for the feign shim, but we're not terribly concerned with feign performance as everything moves to pure dialogue.

carterkozak (approved):

PR #768, opened 2020-05-18
[test-only] JMH microbenchmarks

iamdanfox:
## Before this PR

We're starting to land PRs that fiddle with small optimizations, without an easy way to demonstrate the before/after improvements. This runs the risk of us doing big refactors that aren't actually helpful.

## After this PR
==COMMIT_MSG==
A new `:dialogue-jmh` project allows running microbenchmarks from IntellIJ or from the command line using `./gradlew jmh`
==COMMIT_MSG==

Initial results show that PIN_UNTIL_ERROR is ~2x faster than BALANCED
```
Benchmark                           (headerDriven)  (numChannels)  (selectionStrategy)   Mode  Cnt     Score     Error   Units
NodeSelectionBenchmark.postRequest            true              2      PIN_UNTIL_ERROR  thrpt   12   467.786 ±  28.286  ops/ms
NodeSelectionBenchmark.postRequest            true              2          ROUND_ROBIN  thrpt   12  3616.095 ± 201.541  ops/ms
NodeSelectionBenchmark.postRequest            true              8      PIN_UNTIL_ERROR  thrpt   12   141.798 ±  25.086  ops/ms
NodeSelectionBenchmark.postRequest            true              8          ROUND_ROBIN  thrpt   12  1824.192 ±  74.748  ops/ms
NodeSelectionBenchmark.postRequest           false              2      PIN_UNTIL_ERROR  thrpt   12  8020.820 ± 221.295  ops/ms
NodeSelectionBenchmark.postRequest           false              2          ROUND_ROBIN  thrpt   12  4672.392 ± 101.247  ops/ms
NodeSelectionBenchmark.postRequest           false              8      PIN_UNTIL_ERROR  thrpt   12  6694.895 ± 195.309  ops/ms
NodeSelectionBenchmark.postRequest           false              8          ROUND_ROBIN  thrpt   12  2031.100 ±  40.723  ops/ms
```

Also i've got one that just runs our generated code against undertow.

```
Benchmark                             Mode  Cnt      Score      Error  Units
EndToEndBenchmark.dialogueBlocking   thrpt   16  24013.938 ± 2141.187  ops/s
EndToEndBenchmark.rawApacheBlocking  thrpt   16  29994.941 ± 3443.332  ops/s
```

## Possible downsides?
- it doesn't run on CI, so could decay. In practise I don't really care because we can easily fix it when we want to run it again.
- we don't have an easy way to graph changes over O(months)


dialogue-jmh/build.gradle
```diff
@@ -0,0 +1,52 @@
...
+    classpath = sourceSets.main.runtimeClasspath
+    maxHeapSize = '1024m'
+
+    // e.g. ./gradlew jmv -PjmhOptions='-rf json'
+    if (project.hasProperty('jmhOptions')) {
+        args(project.jmhOptions.split(' '))
+    }
+    doFirst {
+        standardOutput new org.apache.tools.ant.util.TeeOutputStream(
+                new FileOutputStream(benchmarkFile), System.out)
```
carterkozak: Unless we're using the output I'd prefer to leave this out, iirc it interacts poorly with ctrl+c and the benchmarks continue running.

dialogue-jmh/src/main/java/com/palantir/dialogue/core/EndToEndBenchmark.java
```diff
@@ -0,0 +1,145 @@
...
+import org.openjdk.jmh.annotations.TearDown;
+import org.openjdk.jmh.annotations.Threads;
+import org.openjdk.jmh.annotations.Warmup;
+import org.openjdk.jmh.runner.Runner;
+import org.openjdk.jmh.runner.options.Options;
+import org.openjdk.jmh.runner.options.OptionsBuilder;
+
+@State(Scope.Benchmark)
+@Warmup(iterations = 10, time = 500, timeUnit = TimeUnit.MILLISECONDS)
+@Measurement(iterations = 3, time = 500, timeUnit = TimeUnit.MILLISECONDS)
```
carterkozak: increasing from 3 to 4 measurement iterations reduces the result variance. I think we want a longer interval as well, the WC numbers got a great deal tighter when we bumped measurement to 4*4 seconds. I think we use 4*4s warmup there, but we can probably get away with 2*4 warmup.

I like to add default fork args specifying  heap and gc to reduce impact of defaults that differ between jvms.

dialogue-jmh/src/main/java/com/palantir/dialogue/core/EndToEndBenchmark.java
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    @TearDown
+    public void after() {
+        undertow.stop();
+        MoreExecutors.shutdownAndAwaitTermination(RetryingChannel.sharedScheduler.get(), 1, TimeUnit.SECONDS);
+        MoreExecutors.shutdownAndAwaitTermination(blockingExecutor, 1, TimeUnit.SECONDS);
+    }
+
+    @Threads(1)
```
carterkozak: We can specify a sane number of threads everywhere and run with `-t 1` for other counts to avoid duplication.

dialogue-jmh/src/main/java/com/palantir/dialogue/core/EndToEndBenchmark.java
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    @Benchmark
+    public void parallelBlocking() {
+        blocking.voidToVoid();
+    }
+
+    @Benchmark
+    public ListenableFuture<Void> parallelAsync() {
+        return async.voidToVoid();
```
carterkozak: I think we need to wait for the future to complete, otherwise we're benchmarking our executor

carterkozak (approved):
lgtm, thanks!

PR #776, opened 2020-05-19
Reduce mutability by deleting DialogueChannel#updateUris

iamdanfox:
## Before this PR

Previously, we had quite a lot of code hanging around internally to support a model of live-reloading where users would be expected to call `#updateUris()`. HOWEVER, this functionality is not actually exposed by DialogueClients, so we're paying a maintenance & readability cost without getting the upsides.

(Also, DataDog graphs showed that things pretty much never live reload including during a blue/green, so this 'smart' live-reloading wasn't being exercised when it _was_ exposed.)

## After this PR
==COMMIT_MSG==
DialogueChannel no longer allows users to mutate its internals by calling `updateUris`. If the URIs change, throw it away and make a new one.
==COMMIT_MSG==

## Possible downsides?
- I think there's a decent chance that at some point we might want to implement smart live-reloading again, so we might end up re-implementing some of this stuff.  I _think_ that might be OK, because for now it makes the codebase a little simpler.


.palantir/revapi.yml
```diff
@@ -198,3 +198,9 @@ acceptedBreaks:
         \ com.palantir.dialogue.Endpoint)"
       justification: "We make no guarantees for custom implementations of ConjureRuntime, and this is a compile-time\
         \ break not ABI break."
+  "1.46.0":
+    com.palantir.dialogue:dialogue-core:
+    - code: "java.method.removed"
+      old: "method void com.palantir.dialogue.core.DialogueChannel::updateUris(java.util.Collection<java.lang.String>)"
+      justification: "No remaining users of this method, everyone has switched to\
+        \ 'DialogueClients' instead"
```
iamdanfox, line 206: https://sourcegraph.p.b/search?q=updateUris%28&patternType=literal

dialogue-core/src/test/java/com/palantir/dialogue/core/DialogueChannelTest.java
```diff
@@ -140,35 +137,6 @@ public void bad_channel_throwing_an_error_still_returns_a_future() {
         assertThatThrownBy(() -> Futures.getUnchecked(future)).hasCauseInstanceOf(NoClassDefFoundError.class);
     }
 
-    @Test
-    @SuppressWarnings("FutureReturnValueIgnored") // intentionally spawning a bunch of throwaway requests
-    void live_reloading_an_extra_uri_allows_queued_requests_to_make_progress() {
```
iamdanfox, line 145: this test was fake news anyway, because it didn't work for anyone using DialogueClients (which is most people now)

dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueChannel.java
```diff
@@ -31,137 +30,28 @@
...
+        this.cf = cf;
+        this.delegate = delegate;
     }
 
     @Override
     public ListenableFuture<Response> execute(Endpoint endpoint, Request request) {
         return delegate.endpoint(endpoint).execute(request);
     }
 
-    public void updateUris(Collection<String> uris) {
```
iamdanfox, line 112: @carterkozak @ferozco I could leave this method around by just delete the method body to avoid any danger of ABI break? what do you think?

carterkozak: I think it would be worse if it was used and didn't do anything than if it disappears.

carterkozak (approved):

PR #777, opened 2020-05-20
Optimize tracing channels

iamdanfox:
## Before this PR

Looking at yourkit, it seems like we were spending a decent chunk of time faffing around with DetachedSpans, even when sampling is turned off. This seemed silly. There were also some lambdas which seemed a bit unnecessary. Ideally I'd like to get to the stage where a dialogue request involves zero lambdas.

## After this PR
==COMMIT_MSG==
When the current trace is not sampled (i.e. 99% of the time), Dialogue short-circuits a bunch of work.
==COMMIT_MSG==

```
Benchmark                                  Mode  Cnt       Score      Error  Units
EndToEndBenchmark.dialogueBlocking        thrpt   16   27695.216 ± 2796.739  ops/s
EndToEndBenchmark.rawApacheBlocking       thrpt   16   25994.526 ± 5569.710  ops/s
EndToEndBenchmark.setZeroNetworkDialogue  thrpt   16  206281.409 ± 4193.327  ops/s
```

Also we didn't have tests for this particular channel, and now we do.


dialogue-core/src/main/java/com/palantir/dialogue/core/TraceEnrichingChannel.java
```diff
@@ -40,11 +41,18 @@
 
     @Override
     public ListenableFuture<Response> execute(Endpoint endpoint, Request request) {
+        if (!Tracer.isTraceObservable()) {
```
carterkozak: I think we want to ensure observability has been decided.

Tracer.hasTraceId() && !Tracer.isTraceObservable()

dialogue-core/src/main/java/com/palantir/dialogue/core/TracedChannel.java
```diff
@@ -17,23 +17,54 @@
...
+
+    static EndpointChannel requestAttempt(EndpointChannel delegate) {
+        return new TracedChannel(delegate, "Dialogue-request-attempt");
     }
 
     @Override
     public ListenableFuture<Response> execute(Request request) {
-        return Tracers.wrapListenableFuture(operationName, () -> delegate.execute(request));
+        if (!Tracer.isTraceObservable()) {
+            // no point spending CPU cycles setting up a DetachedSpan if we're not sampling span info
```
carterkozak: Here as well

dialogue-core/src/main/java/com/palantir/dialogue/core/TracedChannel.java
```diff
@@ -17,23 +17,53 @@
...
+
+        DetachedSpan span = DetachedSpan.start(operationName);
+        ListenableFuture<Response> future = null;
+        try (CloseableSpan ignored = span.childSpan(operationNameInitial)) {
+            future = delegate.execute(request);
+        } finally {
+            if (future != null) {
+                future.addListener(span::complete, MoreExecutors.directExecutor());
+            } else {
+                span.complete();
```
carterkozak, line 63: null path should be impossible here since we are confident channel.execute won’t throw.

dialogue-jmh/src/main/java/com/palantir/dialogue/core/EndToEndBenchmark.java
```diff
@@ -155,4 +176,13 @@ public static void main(String[] _args) throws Exception {
                 .build();
         new Runner(opt).run();
     }
+
+    private enum InstantChannel implements Channel {
+        INSTANCE;
+
+        @Override
+        public ListenableFuture<Response> execute(Endpoint _endpoint, Request _request) {
+            return Futures.immediateFuture(new TestResponse().code(200));
```
carterkozak, line 185: I think we can pre-allocate this

carterkozak (approved):

PR #779, opened 2020-05-20
Make Simulations more representative of prod usage (i.e. EndpointChannel)

iamdanfox:
## Before this PR

New conjure-java uses the `bind` method (for perf gainz) but we weren't using it in our simulations.

## After this PR
==COMMIT_MSG==
Simulations now bind endpoints to channels upfront, which is more representative of prod usage
==COMMIT_MSG==

My main excitement is we were previously simulating at ~10k requests/sec and now we can do 30k requests/sec.


.palantir/revapi.yml
```diff
@@ -204,3 +204,9 @@ acceptedBreaks:
       old: "method void com.palantir.dialogue.core.DialogueChannel::updateUris(java.util.Collection<java.lang.String>)"
       justification: "No remaining users of this method, everyone has switched to\
         \ 'DialogueClients' instead"
+  "1.47.0":
+    com.palantir.dialogue:dialogue-core:
+    - code: "java.class.removed"
+      old: "class com.palantir.dialogue.core.RefreshingChannelFactory"
+      justification: "No remaining users of this method, everyone has switched to\
+        \ 'DialogueClients' instead"
```
iamdanfox, line 212: https://sourcegraph.p.b/search?q=RefreshingChannelFactory&patternType=literal

carterkozak (approved):
looks reasonable to me, most changes only impact simulation code

PR #780, opened 2020-05-20
`client.response` timer now measures user-perceived time

iamdanfox:
## Before this PR

Our datadog graphs have a nice `client.response` timer, but this can be a bit unhelpful because it doesn't measure the user-perceived time (i.e. graphs could currently all show nice quick responses, but users might still perceived slow responses due to lots of backoff & retrying, which wouldn't show up in our graphs).

cc @tpetracca who mentioned this in #dev-rpc a while ago

## After this PR
==COMMIT_MSG==
`client.response` timer now measures user-perceived time
==COMMIT_MSG==

## Possible downsides?
- graphs might _look_ worse, but I think they're actually more accurate/actionable/useful.


dialogue-core/src/main/java/com/palantir/dialogue/core/TimingEndpointChannel.java
```diff
@@ -0,0 +1,82 @@
...
+import java.io.IOException;
+
+final class TimingEndpointChannel implements EndpointChannel {
+    private final EndpointChannel delegate;
+    private final Timer responseTimer;
+    private final Meter ioExceptionMeter;
+
+    private TimingEndpointChannel(EndpointChannel delegate, Timer responseTimer, Meter ioExceptionMeter) {
+        this.delegate = delegate;
+        this.responseTimer = responseTimer;
```
carterkozak: I thought we weren’t using a timer anymore because it doesn’t give us specific enough (per-endpoint) data to be useful?

iamdanfox: Hmm I'm always hesitant to delete metrics because of how salty people got before... I guess the thing I find most interesting is actually the rate of `client.response.count`.

Could turn this into a meter too?

carterkozak: I thought it was already a meter

iamdanfox: https://github.com/palantir/dialogue/blob/4619ce4d2ea8fb8715e21d9ddb796ca075cd4a4e/dialogue-core/src/main/metrics/dialogue-core-metrics.yml#L10

carterkozak: oof, the thumbs up on this never resulted in a merge :-(
https://github.com/palantir/dialogue/pull/757

iamdanfox: Ah dang, I can cherry-pick your change across here if you like?

carterkozak: works for me, otherwise I can fix up the conflicts ¯\_(ツ)_/¯

carterkozak (approved):

dialogue-core/src/main/java/com/palantir/dialogue/core/TimingEndpointChannel.java
iamdanfox: Chatter over VTC, resolved to keep this as a timer as this gives us parity with c-j-r and people have established dashboards around this.

PR #781, opened 2020-05-20
Tracing around dialogue encodings

carterkozak:
==COMMIT_MSG==
Tracing around dialogue encodings describing serde and I/O time per request, matching data produced by conjure-undertow on the server
==COMMIT_MSG==

### DIFF
--- a/dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/TracedEncoding.java
+++ b/dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/TracedEncoding.java
@@ -1,0 +1,112 @@
+/*
+ * (c) Copyright 2020 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.conjure.java.dialogue.serde;
+
+import com.palantir.dialogue.TypeMarker;
+import com.palantir.logsafe.Preconditions;
+import com.palantir.tracing.CloseableTracer;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.lang.reflect.Type;
+
+/**
+ * Wrapper around an {@link Encoding} which adds tracing spans around serialization and deserialization operations
+ * including I/O.
+ */
+final class TracedEncoding implements Encoding {
+
+    private final Encoding delegate;
+
+    TracedEncoding(Encoding delegate) {
+        this.delegate = Preconditions.checkNotNull(delegate, "Encoding is required");
+    }
+
+    @Override
+    public <T> Serializer<T> serializer(TypeMarker<T> type) {
+        String operation = "Dialogue: serialize " + toString(type) + " to " + getContentType();
+        return new TracedSerializer<>(delegate.serializer(type), operation);
+    }
+
+    @Override
+    public <T> Deserializer<T> deserializer(TypeMarker<T> type) {
+        String operation = "Dialogue: deserialize " + toString(type) + " from " + getContentType();
+        return new TracedDeserializer<>(delegate.deserializer(type), operation);
+    }
+
+    @Override
+    public String getContentType() {
+        return delegate.getContentType();
+    }
+
+    @Override
+    public boolean supportsContentType(String contentType) {
+        return delegate.supportsContentType(contentType);
+    }
+
+    @Override
+    public String toString() {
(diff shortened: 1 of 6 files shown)


dialogue-serde/src/test/java/com/palantir/conjure/java/dialogue/serde/TracedEncodingTest.java
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    @Test
+    void testSerializerOperationName() {
+        Tracer.setSampler(AlwaysSampler.INSTANCE);
+        Tracer.getAndClearTrace();
+        Encoding.Serializer<String> serializer =
+                new TracedEncoding(new StubEncoding()).serializer(new TypeMarker<String>() {});
+        SpanObserver mockObserver = mock(SpanObserver.class);
+        Tracer.subscribe("test", mockObserver);
```
iamdanfox, line 61: could we hoist all this setup code to an `@Before` so that it's easier to skim the tests and see what the difference is in each of them?

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/TracedEncoding.java
```diff
@@ -0,0 +1,112 @@
...
+
+    private final Encoding delegate;
+
+    TracedEncoding(Encoding delegate) {
+        this.delegate = Preconditions.checkNotNull(delegate, "Encoding is required");
+    }
+
+    @Override
+    public <T> Serializer<T> serializer(TypeMarker<T> type) {
+        String operation = "Dialogue: serialize " + toString(type) + " to " + getContentType();
```
iamdanfox, line 40: this is v elegant

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/TracedEncoding.java
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    @Override
+    public boolean supportsContentType(String contentType) {
+        return delegate.supportsContentType(contentType);
+    }
+
+    @Override
+    public String toString() {
+        return "TracedEncoding{delegate=" + delegate + '}';
```
iamdanfox: i've actually started omitting the `delegate=` when there's only one field, especially because we compose so much stuff you end up with a 
`Foo{delegate=Bar{delegate=Baz{delegate=`
vs 
`Foo{Bar{Baz{`

dialogue-serde/src/test/java/com/palantir/conjure/java/dialogue/serde/TracedEncodingTest.java
```diff
@@ -0,0 +1,109 @@
...
+
+    @Test
+    void testTypeString_stringList() {
+        // It's more complicated to generate 'List<String>' from Type, that can be done later if necessary
+        assertThat(TracedEncoding.toString(new TypeMarker<List<String>>() {}))
+                .isEqualTo("java.util.List<java.lang.String>");
+    }
+
+    @Test
+    void testSerializerOperationName() {
```
ferozco, line 57: these tests will not parallelize well :(

iamdanfox: luckily `./gradlew build` currently takes 1m22 so maybe 🤷 
https://circleci.com/gh/palantir/dialogue/10416?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link

ferozco: didn't you configure everything to run in parallel by default?

iamdanfox: Nah i gave up because of our `@TestTracing` methods, which can't run in parallel with _anything_ else.

The only `junit-platform.properties` is in :simulation
```
junit.jupiter.execution.parallel.enabled=true
junit.jupiter.execution.parallel.config.strategy=dynamic
```

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/TracedEncoding.java
carterkozak: that's much  more concise, updated this along with the lazily initialized encoding.

iamdanfox (approved):
LGTM, would like to just make the test a bit easier on the eyes

dialogue-serde/src/test/java/com/palantir/conjure/java/dialogue/serde/TracedEncodingTest.java
carterkozak: Unfortunately it only applies to two of the four tests in this class. We could split them into separate classes, but it becomes more difficult to jump from TracedEncoding to its tests :-/
Fwiw this test is identical to the traced encoding test for conjure-undertow with the "Undertow" substring replaced with "Dialogue" ;-)

iamdanfox: Personally I'd still hoist them. Things like `Tracer.setSample(AlwaysSample.INSTANCE)` don't hurt your first two tests, they just seem like a sane default for this whole file?

PR #784, opened 2020-05-21
Fallback default impl for Clients.bind

iamdanfox:
We've had some unfortunate problems with the rollout of the new Clients.bind optimization, causing some users' servers to fail to start.

This solution is a little gross, but I think it should at least tide people over until they pick up the newer `:dialogue-serde` jar and actually get the meat of the optimization!

cc @natgabb

### DIFF
--- a/dialogue-target/src/main/java/com/palantir/dialogue/Clients.java
+++ b/dialogue-target/src/main/java/com/palantir/dialogue/Clients.java
@@ -37,7 +37,16 @@ public interface Clients {
     @Deprecated
     <T> ListenableFuture<T> call(Channel channel, Endpoint endpoint, Request request, Deserializer<T> deserializer);
 
-    EndpointChannel bind(Channel channel, Endpoint endpoint);
+    default EndpointChannel bind(Channel channel, Endpoint endpoint) {
+        return new EndpointChannel() {
+            @Override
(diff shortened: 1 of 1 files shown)


carterkozak (approved):

ferozco:
👍

dialogue-target/src/main/java/com/palantir/dialogue/Clients.java
```diff
@@ -37,7 +37,16 @@
...
-    EndpointChannel bind(Channel channel, Endpoint endpoint);
+    default EndpointChannel bind(Channel channel, Endpoint endpoint) {
+        return new EndpointChannel() {
+            @Override
+            public ListenableFuture<Response> execute(Request request) {
+                // this default implementation exists just in case people take new :dialogue-target but have an old
+                // version of :dialogue-serde
+                return channel.execute(endpoint, request);
+            }
+        };
```
carterkozak, line 48: we probably want something like a platform long term to keep all library versions coupled.

iamdanfox: Yeah, so Felipe mentioned that gradle module metadata might actually be exactly what we need here, as I think we could define that all dialogue jars should be the same version _in this repo_ and have all consumers magically pick it up!

PR #791, opened 2020-05-22
Optimize Request.builder().from()

iamdanfox:
## Before this PR

As per a discussion in #dev-foundry-infra, a decent proportion of dialogue overhead is actually spent copying Requests just to add/modify a header. 

Rather than trying to optimize multimaps, we can actually get a bit of a speedup by just re-using stuff where possible. (There's no need to allocate and populate an identical map to an existing one!)

```
Benchmark                                                                     Mode  Cnt     Score     Error   Units
RequestBuilderBenchmark.addUserAgentToEmpty                                  thrpt   12  9310.828 ± 736.043  ops/ms
RequestBuilderBenchmark.addUserAgentToNonEmpty                               thrpt   12   936.356 ±  71.790  ops/ms

RequestBuilderBenchmark.addUserAgentToEmpty:·gc.alloc.rate.norm              thrpt   12   592.000 ±   0.001    B/op
RequestBuilderBenchmark.addUserAgentToNonEmpty:·gc.alloc.rate.norm           thrpt   12  2688.003 ±   0.008    B/op
```

## After this PR
==COMMIT_MSG==
Reduce allocation and eliminate some unnecessary copying from `Request.builder().from()`
==COMMIT_MSG==

Less GC due to dumb copying means servers will have more CPU available to do real work.

```
Benchmark                                                                     Mode  Cnt      Score      Error   Units
RequestBuilderBenchmark.addUserAgentToEmpty                                  thrpt   12  23070.647 ± 3011.135  ops/ms
RequestBuilderBenchmark.addUserAgentToNonEmpty                               thrpt   12   2175.030 ±  212.811  ops/ms

RequestBuilderBenchmark.addUserAgentToEmpty:·gc.alloc.rate.norm              thrpt   12    264.000 ±    0.001    B/op
RequestBuilderBenchmark.addUserAgentToNonEmpty:·gc.alloc.rate.norm           thrpt   12    744.001 ±    0.004    B/op
```

## Possible downsides?

- the _vast_ majority of dialogue users will not care about this, it's really only helpful in the case where something is making tons of network requests and is pretty maxed out on CPU...
- whenever we're micro-optimizing, there's a danger that making the code harder to understand is actually worse than a small performance hit.  I _hope_ this implementation is still easy to understand, but would like to hear if not.

### DIFF
--- a/dialogue-target/src/main/java/com/palantir/dialogue/Request.java
+++ b/dialogue-target/src/main/java/com/palantir/dialogue/Request.java
@@ -226,7 +226,7 @@ public final class Request {
         }
 
         public Request.Builder pathParams(Map<String, ? extends String> entries) {
-            pathParams = null;
+            pathParams = ImmutableMap.builder();
             return putAllPathParams(entries);
         }
 
--- a/dialogue-target/src/main/java/com/palantir/dialogue/Request.java
+++ b/dialogue-target/src/main/java/com/palantir/dialogue/Request.java
@@ -201,7 +201,8 @@ public final class Request {
         }
 
         public Request.Builder queryParams(Multimap<String, ? extends String> entries) {
-            queryParams = ImmutableListMultimap.builder();
+            queryParams = null;
+            existingQueryParams = null;
             return putAllQueryParams(entries);
         }
 
@@ -226,7 +227,8 @@ public final class Request {
         }
 
         public Request.Builder pathParams(Map<String, ? extends String> entries) {
-            pathParams = ImmutableMap.builder();
+            pathParams = null;
+            existingPathParams = null;
             return putAllPathParams(entries);
         }
 
--- a/dialogue-target/src/main/java/com/palantir/dialogue/Request.java
+++ b/dialogue-target/src/main/java/com/palantir/dialogue/Request.java
@@ -116,7 +116,7 @@ public final class Request {
     }
 
     public static final class Builder {
-        private static final ListMultimap<String, String> EMPTY_HEADERS =
+        private static final ImmutableListMultimap<String, String> EMPTY_HEADERS =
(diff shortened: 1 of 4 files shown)


dialogue-target/src/main/java/com/palantir/dialogue/Request.java
```diff
@@ -113,22 +117,44 @@ public static Builder builder() {
...
     }
 
+    @NotThreadSafe
     public static final class Builder {
-        private ListMultimap<String, String> headerParams = MultimapBuilder.treeKeys(String.CASE_INSENSITIVE_ORDER)
-                .arrayListValues()
-                .build();
-        private ImmutableListMultimap.Builder<String, String> queryParams = ImmutableListMultimap.builder();
-        private ImmutableMap.Builder<String, String> pathParams = ImmutableMap.builder();
+        private static final ImmutableListMultimap<String, String> EMPTY =
```
ferozco, line 122: as @carterkozak suggested in standup today, it would be pretty cool to use a bit set to track whether we have already done the copy-on-write for a given collection

iamdanfox: Yeah bitset thing would be even nicer, but at the moment I quite like that we're still storing immutable collections internally, so we're really protected from any bugs in the implementation.

Could we land this one as an incremental improvement and consider the bitset thing separately? I'm just worried the bitset approach make it harder to read/understand as we'd need to move away from ImmutableMap.Builder and store mutable things internally...

carterkozak: Sure, no need to crank it up to 11 at once.

ferozco (approved):

carterkozak (approved):

PR #792, opened 2020-05-22
Thread Fairness Queue

iamdanfox:
## Before this PR

We're almost done with dialogue rollout, but internal-ski-product is the last unsolved usecase. We have a viable design for the load balancing problem, which leaves the 'fairness' problem, which JB solved in c-j-r with the [ThreadWorkQueue](https://github.com/palantir/conjure-java-runtime/blob/develop/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/ThreadWorkQueue.java).

## After this PR
==COMMIT_MSG==
When all nodes are concurrency limited and requests queue up, they'll be processed in a fairer way so that a single heavy thread can't starve out all other threads. Note the vast majority of the time, this code will be short-circuited (so have no impact) because the queue is empty. 
==COMMIT_MSG==

I really don't want to leave internal-ski-product hanging on c-j-r, and given that we don't have any other viable proposals for how to avoid this, I've put up this PR. As mentioned above, I think this won't actually affect the vast majority of users _at all_. It's only really when a product is exceeding their server ratelimits that this would have any effect at all.

## Possible downsides?
- "this logic doesn't really belong in a pure RPC library" - i think this logic has to be colocated with the concurrencylimiter & queue logic, and we've found these two components subtle enough that I don't really want internal-ski-product to copy and paste them.
- concurrent code often contains bugs

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
@@ -32,11 +32,8 @@ import com.palantir.logsafe.exceptions.SafeRuntimeException;
 import com.palantir.tracing.CloseableSpan;
 import com.palantir.tracing.DetachedSpan;
 import com.palantir.tritium.metrics.registry.TaggedMetricRegistry;
-import java.util.Deque;
 import java.util.Optional;
-import java.util.concurrent.ConcurrentLinkedDeque;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Supplier;
 import org.immutables.value.Value;
 import org.slf4j.Logger;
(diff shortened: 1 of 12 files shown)


dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
```diff
@@ -164,93 +156,64 @@ void schedule() {
...
-                                            + "logging",
-                                    SafeArg.of("channel", channelName),
-                                    SafeArg.of("service", endpoint.serviceName()),
-                                    SafeArg.of("endpoint", endpoint.endpointName()));
-                        }
-                    }
-                });
-                return true;
-            } else {
-                if (!queuedCalls.offerFirst(queueHead)) {
```
iamdanfox, line 224: moved this to the `restore` method on line 183

iamdanfox:
From discussions in today's standup, it sounds like @ferozco and @carterkozak would prefer to hold off until it's demonstrated that the internal-ski-product client really needs this

Issue #686, opened 2020-04-28
Prefer routing traffic within an AWS availability zone to save $$$


iamdanfox:
Most of our services have nodes in different availability zones. Given that there's nothing constraining traffic in any aws-specific way, every time a node of email-service wants to talk to a node of MP, it might pick a node in any region. This means we're probably paying $$$ in cross-AZ traffic when we don't need to.

### Pricing diagram from this [blog post](https://www.apptio.com/emerge/aws-data-transfer-costs/)

![image](https://user-images.githubusercontent.com/3473798/80515195-75e1c580-8979-11ea-8c35-2b28991c3880.png)

It seems like if we can slightly bias connections towards staying in their region (e.g.  `eu-west-1a` <-> `eu-west-1a`) then we'd be able to cut down on our spend a bit.

## Proposal

When a server is running on AWS, there's a magic IP address we can call to find out which region it's currently in, e.g.

```
$ curl -s http://169.254.169.254/latest/meta-data/placement/availability-zone
eu-west-1a
```

(Atlasdb currently [uses this](https://github.com/palantir/atlasdb/blob/9f54f73aee6689640a6b7e99a4808e649e1f408f/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/Ec2HostLocationSupplier.java)).

Then, servers could either advertise this information somehow. Either using a header, or a dedicated metadata endpoint, or perhaps even plumbed through yaml somehow. We might even be able to DNS resolve the hosts we're given and match them against amazon's published IP ranges: https://ip-ranges.amazonaws.com/ip-ranges.json.

With this information, I'd suggest that we add a tiny constant bias to the Balanced Channel's scores, so rather than starting everything off at 0, we'd say hosts that are in other availability zones get a minimum score of 1.  This would mean that under zero utilization, the first request would always go intra AZ.

## Possible downsides?

Obviously this would need to fail gracefully when running locally, in docker or on Azure.

carterkozak:
How do client preceived latencies differ between nodes in different AZs? I'd rather use that data to rank targets than to target specific cloud vendors in an rpc library.
Another option is for deployment infrastructure to provide a quality-factor based on availability zones along with URIs, centralizing that discovery.

iamdanfox:
So the idea here is more about $ savings than latencies tbh

carterkozak:
Right, we can solve the problem without vendor-specific implementation.

j-baker:
[...]

PR #794, opened 2020-05-27
Balanced channel measures RTT, hoping to stay within AZ (and save $)

iamdanfox:
## Before this PR

There's an outstanding FR to make dialogue smart enough to reduce our AWS spend (https://github.com/palantir/dialogue/issues/686).

## After this PR

There are a variety of ways to achieve this, calling the AWS API directly, baking in a hostname convention, but this PR implements a version which just measures the RTT for a [HTTP OPTIONS](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/OPTIONS) call.

==COMMIT_MSG==
Balanced channel should now bias towards whichever node has the lowest latency, which should reduce AWS spend by routing requests within AZ.
==COMMIT_MSG==

TODO:
- [x] validate whether we can actually identify different AZs by RTT (validated using snapshots)
- ~maybe just go back to HEAD, OPTIONS doesn't buy us much~
- [x] don't send a RTT-sampling request for _every_ RPC
- [x] do we need to expire the RTTs eventually?
- [x] tests

## Possible downsides?
- if dialogue is used to talk to a server that doesn't support OPTIONS requests, their logs might look weird
- ~after merging this, the graphs will probably show 'balanced' sending lots of traffic to one node (within AZ) and very little to another. this might confuse people.~ EDIT it's now opt-in
- simulations don't exercise this as they all return OPTIONS instantly.


dialogue-core/src/main/java/com/palantir/dialogue/core/BalancedNodeSelectionStrategyChannel.java
```diff
@@ -217,20 +246,26 @@ public String toString() {
...
         }
 
+        public long getRtt() {
+            return rtt;
+        }
+
         @Override
         public String toString() {
-            return "SortableChannel{score=" + score + ", delegate=" + delegate + '}';
+            return "SortableChannel{" + "score=" + score + ", rtt=" + rtt + ", delegate=" + delegate + '}';
```
carterkozak:
```suggestion
            return "SortableChannel{score=" + score + ", rtt=" + rtt + ", delegate=" + delegate + '}';
```

dialogue-core/src/main/java/com/palantir/dialogue/core/BalancedNodeSelectionStrategyChannel.java
```diff
@@ -180,15 +189,35 @@ public void onFailure(Throwable throwable) {
...
+            //     return;
+            // }
+
+            Optional<ListenableFuture<Response>> maybe = delegate.maybeExecute(RttEndpoint.INSTANCE, BLANK_REQUEST);
+            if (maybe.isPresent()) {
+                // nextRttMeasurement = initialNanos + TimeUnit.SECONDS.toNanos(1);
+                DialogueFutures.addDirectListener(maybe.get(), () -> {
+                    rttNanos = clock.read() - initialNanos;
+                    if (log.isInfoEnabled()) {
+                        log.info("rttNanos {}", SafeArg.of("rtt", rttNanos), UnsafeArg.of("channel", delegate));
```
carterkozak: needs a second `{}`

iamdanfox:
~Ok so I've stumbled across an edge case that is probably worth solving:~

~what if we're talking to 10 upstreams, 5 of which are local to my AZ and 5 of which are in another AZ. We probably don't want to firehose all our traffic at whichever of the local 5 happens to be the fastest.~

EDIT solved.

dialogue-core/src/test/java/com/palantir/dialogue/core/BalancedNodeSelectionStrategyChannelTest.java
```diff
@@ -127,25 +128,118 @@ void constant_4xxs_do_eventually_move_the_needle_but_we_go_back_to_fair_distribu
...
         channel.maybeExecute(endpoint, request);
-        assertThat(channel.getScores())
+        assertThat(channel.getScoresForTesting().map(c -> c.getScore()))
                 .describedAs("%s: Constant 4xxs did move the needle %s", Duration.ofNanos(clock.read()), channel)
                 .containsExactly(1, 0);
 
         incrementClockBy(Duration.ofSeconds(5));
 
-        assertThat(channel.getScores())
+        assertThat(channel.getScoresForTesting().map(c -> c.getScore()))
```
carterkozak: peanut gallery: I want an error-prone rule that can refactor this automatically
```suggestion
        assertThat(channel.getScoresForTesting().map(BalancedNodeSelectionStrategyChannel::getScore))
```

dialogue-target/src/main/java/com/palantir/dialogue/HttpMethod.java
```diff
@@ -24,4 +24,5 @@
     PATCH,
     POST,
     PUT,
+    OPTIONS
```
carterkozak: Mind breaking this out into a separate PR?

dialogue-core/src/main/java/com/palantir/dialogue/core/BalancedNodeSelectionStrategyChannel.java
```diff
@@ -342,4 +436,105 @@ public void markRequestMade() {
...
+            this.clock = clock;
+        }
+
+        Optional<RttMeasurementPermit> acquireNonBlocking() {
+            long last = lastMeasured.get();
+            if (last + BETWEEN_SAMPLES > clock.read()) {
+                return Optional.empty();
+            }
+
+            if (currentlySampling.compareAndSet(false, true)) {
```
carterkozak: Pre-check because there will be (relatively) long periods of time when every invocation fails to cas

```suggestion
            if (!currentlySampling.get() && currentlySampling.compareAndSet(false, true)) {
```

dialogue-core/src/main/java/com/palantir/dialogue/core/BalancedNodeSelectionStrategyChannel.java
```diff
@@ -342,4 +436,105 @@ public void markRequestMade() {
...
+                lastMeasured.set(clock.read());
+                return Optional.of(() -> currentlySampling.compareAndSet(true, false));
```
carterkozak: No need to check the current value, any time this executes the resulting value is `false`.
The value returned here could be a singleton.
```suggestion
                return Optional.of(() -> currentlySampling.set(false));
```

dialogue-core/src/main/java/com/palantir/dialogue/core/BalancedNodeSelectionStrategyChannel.java
```diff
@@ -342,4 +436,105 @@ public void markRequestMade() {
...
+        public String toString() {
+            return "RttMeasurement{" + "bestRttNanos=" + bestRttNanos + ", samples=" + Arrays.toString(samples) + '}';
+        }
+    }
+
+    private static final class RttMeasurementRateLimiter {
+        private static final long BETWEEN_SAMPLES = Duration.ofSeconds(1).toNanos();
+
+        private final Ticker clock;
+        private final AtomicLong lastMeasured = new AtomicLong(0);
```
carterkozak: If we're only using get/set we might as well use a volatile long.

dialogue-core/src/main/java/com/palantir/dialogue/core/BalancedNodeSelectionStrategyChannel.java
```diff
@@ -342,4 +436,105 @@ public void markRequestMade() {
...
+        private final AtomicBoolean currentlySampling = new AtomicBoolean(false);
+
+        private RttMeasurementRateLimiter(Ticker clock) {
+            this.clock = clock;
+        }
+
+        Optional<RttMeasurementPermit> acquireNonBlocking() {
```
carterkozak: Perhaps rename to `tryAcquire`? I think that's more concise and makes it clear this won't block.

dialogue-core/src/main/java/com/palantir/dialogue/core/BalancedNodeSelectionStrategyChannel.java
```diff
@@ -189,48 +291,52 @@ public void onFailure(Throwable throwable) {
...
 
     /**
-     * A dedicated immutable class ensures safe sorting, as otherwise there's a risk that the inflight AtomicInteger
+     * A dedicated value class ensures safe sorting, as otherwise there's a risk that the inflight AtomicInteger
      * might change mid-sort, leading to undefined behaviour.
      */
-    private static final class SortableChannel {
-        private final int score;
+    @VisibleForTesting
+    static final class ChannelStats {
```
ferozco: I think having 7 inner classes is bit too many, I think it might improve legibility if we pulled some of these out

dialogue-core/src/main/java/com/palantir/dialogue/core/BalancedNodeSelectionStrategyChannel.java
```diff
@@ -56,13 +68,24 @@
 final class BalancedNodeSelectionStrategyChannel implements LimitedChannel {
```
ferozco: I think we should update the description so that it includes the additional parameters to our cost function

dialogue-core/src/main/java/com/palantir/dialogue/core/BalancedNodeSelectionStrategyChannel.java
```diff
@@ -342,4 +436,105 @@ public void markRequestMade() {
...
+        }
+
+        @Override
+        public String endpointName() {
+            return "rtt";
+        }
+
+        @Override
+        public String version() {
+            return "0.0.0";
```
carterkozak: maybe try to pull the dialogue version?

dialogue-core/src/main/java/com/palantir/dialogue/core/BalancedNodeSelectionStrategyChannel.java
iamdanfox: Yup agreed. commit incoming

dialogue-core/src/main/java/com/palantir/dialogue/core/BalancedNodeSelectionStrategyChannel.java
```diff
@@ -92,25 +116,101 @@
...
+        }
+
+        // Latency (rtt) is measured in nanos, which is a tricky unit to include in our 'score' because adding it
+        // would dominate all the other data (which has the unit of 'num requests'). To avoid the need for a conversion
+        // fudge-factor, we instead figure out where each rtt lies on the spectrum from bestRttNanos to worstRttNanos,
+        // with 0 being best and 1 being worst. This ensures that if several nodes are all within the same AZ and
+        // can return in ~1 ms but others return in ~5ms, the 1ms nodes will all have a similar rttScore (near zero).
+        // Note, this can only be computed when we have all the snapshots in front of us.
+        long rttRange = worstRttNanos - bestRttNanos;
+        if (bestRttNanos != Long.MAX_VALUE && worstRttNanos != 0 && rttRange > 0) {
```
carterkozak: perhaps a negative sentinel value, extracted to a static field to make the code easier to read? `bestRttNanos != UNKNOWN_RTT`

dialogue-core/src/main/java/com/palantir/dialogue/core/BalancedNodeSelectionStrategyChannel.java
```diff
@@ -189,48 +291,52 @@ public void onFailure(Throwable throwable) {
...
-    private static final class SortableChannel {
-        private final int score;
+    @VisibleForTesting
+    static final class ChannelStats {
         private final MutableChannelWithStats delegate;
+        private final int requestsInflight;
+        private final double recentBadness;
+
+        private final long rttNanos;
+        private float rttSpectrum = 0; // assigned later once we've figured out the range of rttNanos across channels
```
ferozco: its a bit weird that we construct an instance of this and then later mutate the private field. Perhaps we can invert the flow and require a spectrum be provided when constructed?

ferozco: also its probably worth documenting (and checking) that this is a value between 0 and 1

carterkozak (commented):
What would it take to support a flag to opt in/out of this feature for initial rollout now that we have excavated out BALANCED?

Added a few comments from an initial read through, still need to do a higher level pass.

dialogue-core/src/main/java/com/palantir/dialogue/core/BalancedNodeSelectionStrategyChannel.java
iamdanfox: We can only compute this number when we have full information of all the rttNanos (cos we need to know the min and max across all channels), so we can't initialize this at the same times as the other fields.

If we introduced an extra type, maybe this would be clearer, i.e. ChannelSnapshot and ChannelSnapshot2 (derived from all the ChannelSnapshots)?

dialogue-core/src/main/java/com/palantir/dialogue/core/BalancedNodeSelectionStrategyChannel.java
```diff
@@ -92,25 +116,101 @@
...
 
         return Optional.empty();
     }
 
-    private static SortableChannel[] sortByScore(List<MutableChannelWithStats> preShuffled) {
-        SortableChannel[] sorted = new SortableChannel[preShuffled.size()];
+    private static ChannelStats[] computeScores(List<MutableChannelWithStats> preShuffled) {
+        ChannelStats[] snapshotArray = new ChannelStats[preShuffled.size()];
+
+        long bestRttNanos = Long.MAX_VALUE;
```
ferozco: I'd decouple the logic of constructing the `snapshotArray` from computing the `rttSpecturom` and factor this out so we have a single function that returns to you an OptionalLong which is the `rttSpectrum`

dialogue-core/src/main/java/com/palantir/dialogue/core/BalancedNodeSelectionStrategyChannel.java
```diff
@@ -92,25 +107,49 @@
...
+        // Latency (rtt) is measured in nanos, which is a tricky unit to include in our 'score' because adding it
+        // would dominate all the other data (which has the unit of 'num requests'). To avoid the need for a conversion
+        // fudge-factor, we instead figure out where each rtt lies on the spectrum from bestRttNanos to worstRttNanos,
+        // with 0 being best and 1 being worst. This ensures that if several nodes are all within the same AZ and
+        // can return in ~1 ms but others return in ~5ms, the 1ms nodes will all have a similar rttScore (near zero).
+        // Note, this can only be computed when we have all the snapshots in front of us.
+        long rttRange = worstRttNanos - bestRttNanos;
+        if (bestRttNanos != Long.MAX_VALUE && worstRttNanos != 0 && rttRange > 0) {
+            for (ChannelStats snapshot : snapshotArray) {
+                snapshot.rttSpectrum = ((float) snapshot.rttNanos - bestRttNanos) / rttRange;
```
ferozco: Could this not result in an rttspectrum larger than 1 if we failed to make a request due to concurrency limits?

dialogue-core/src/main/java/com/palantir/dialogue/core/BalancedNodeSelectionStrategyChannel.java
ferozco: Right you could expose a way of getting `rtt.getNanos()` from each `MutableChannelWithStats` and then have `getSnapshot` take a `rttSpectrum` as a parameter.

dialogue-core/src/main/java/com/palantir/dialogue/core/BalancedNodeSelectionStrategyChannel.java
iamdanfox: So this is actually the result of doing the `min` and `max` things above, so it's not actually an arbitrary sentinel.  I tried using streams to compute these:
```java
Arrays.stream(snapshotArray)
                .filter(snapshot -> snapshot.rttNanos.isPresent())
                .mapToLong(snapshot -> snapshot.rttNanos.getAsLong())
                .summaryStatistics();
```

but the summaryStatistics _also_ gives you a Long.MIN_VALUE/MAX_VALUE if there is no inputs.

dialogue-core/src/test/java/com/palantir/dialogue/core/BalancedNodeSelectionStrategyChannelTest.java
ferozco: That would be pretty cool! Think we just need to improve https://github.com/palantir/gradle-baseline/blob/15661f333a76997d0a5b79a52af8767e01a86ca1/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/LambdaMethodReference.java#L45 to handle more than suppliers

ferozco: there ya go: https://github.com/palantir/gradle-baseline/pull/1359

dialogue-core/src/main/java/com/palantir/dialogue/core/RttSampler.java
```diff
@@ -0,0 +1,280 @@
...
+
+        synchronized void addMeasurement(long newMeasurement) {
+            Preconditions.checkArgument(newMeasurement > 0, "Must be greater than zero");
+            Preconditions.checkArgument(newMeasurement < Long.MAX_VALUE, "Must be less than MAX_VALUE");
+
+            if (samples[0] == Long.MAX_VALUE) {
+                Arrays.fill(samples, newMeasurement);
+                bestRttNanos = newMeasurement;
+            } else {
+                System.arraycopy(samples, 1, samples, 0, NUM_MEASUREMENTS - 1);
```
ferozco, line 232: instead of copying the array could we treat the array as a circular buffer keeping  track of our current location

ferozco (commented):
overall looks great! excited to try this out

dialogue-core/src/main/java/com/palantir/dialogue/core/RttSampler.java
iamdanfox: Yeah I considered this, but figured it was actually gonna be more readable to try and keep it simple here - especially when this codepath only gets invoked at most once per second, and in a non-blocking way :)

ferozco:
👍

PR #795, opened 2020-05-27
Optimize Request.Builder size and allocations

carterkozak:
Track which fields are borrowed from a request (immutable) and which have
been copied (mutable).
```
Benchmark                                         Mode  Cnt      Score     Error   Units
RequestBuilderBenchmark.addUserAgentToEmpty      thrpt   12  24146.195 ± 159.934  ops/ms
RequestBuilderBenchmark.addUserAgentToEmpty2     thrpt   12  31716.135 ± 184.990  ops/ms
RequestBuilderBenchmark.addUserAgentToNonEmpty   thrpt   12   2278.744 ±   5.884  ops/ms
RequestBuilderBenchmark.addUserAgentToNonEmpty2  thrpt   12   2408.309 ±  11.029  ops/ms
```

## After this PR
==COMMIT_MSG==
Optimize Request.Builder
==COMMIT_MSG==

### DIFF
--- a/dialogue-target/src/main/java/com/palantir/dialogue/Request.java
+++ b/dialogue-target/src/main/java/com/palantir/dialogue/Request.java
@@ -16,18 +16,22 @@
 
 package com.palantir.dialogue;
 
+import com.google.common.collect.ArrayListMultimap;
 import com.google.common.collect.ImmutableListMultimap;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ListMultimap;
 import com.google.common.collect.Multimap;
-import com.google.common.collect.MultimapBuilder;
 import com.google.common.collect.Multimaps;
 import com.palantir.logsafe.Preconditions;
+import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
-import javax.annotation.Nullable;
+import java.util.TreeMap;
 import javax.annotation.concurrent.NotThreadSafe;
 import javax.annotation.concurrent.ThreadSafe;
 
@@ -36,8 +40,8 @@ import javax.annotation.concurrent.ThreadSafe;
 public final class Request {
 
     private final ListMultimap<String, String> headerParams;
-    private final ImmutableListMultimap<String, String> queryParams;
-    private final ImmutableMap<String, String> pathParams;
+    private final ListMultimap<String, String> queryParams;
+    private final Map<String, String> pathParams;
     private final Optional<RequestBody> body;
 
     private Request(Builder builder) {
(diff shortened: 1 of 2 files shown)


dialogue-target/src/main/java/com/palantir/dialogue/Request.java
```diff
@@ -186,7 +182,7 @@ private Builder() {}
         }
 
         public Request.Builder putQueryParams(String key, String... values) {
-            mutableQueryParams().putAll(key, values);
+            mutableQueryParams().putAll(key, Arrays.asList(values));
```
ferozco, line 185: should we avoid the extra allocation and iterate over the array calling put? elsewhere you seem to suggest that is more performant than putAll

carterkozak: I don’t think this method is ever used

dialogue-target/src/main/java/com/palantir/dialogue/Request.java
```diff
@@ -119,40 +123,32 @@ public static Builder builder() {
...
-        private ListMultimap<String, String> existingUnmodifiableHeaderParams;
+        private Map<String, String> pathParams = ImmutableMap.of();
 
-        @Nullable
-        private ImmutableListMultimap<String, String> existingQueryParams;
+        private Optional<RequestBody> body = Optional.empty();
 
-        @Nullable
-        private ImmutableMap<String, String> existingPathParams;
+        private int mutableCollections = 0;
```
iamdanfox: would kinda like to include 'bitset' or 'bitflags' in the name here so that people don't think this is a count of mutable collections...

dialogue-target/src/main/java/com/palantir/dialogue/Request.java
ferozco: ok, should we mark it as deprecated then?

dialogue-target/src/main/java/com/palantir/dialogue/Request.java
```diff
@@ -249,81 +243,81 @@ private Builder() {}
...
+            return isMutable(MUTABLE_PATH);
         }
 
-        public Request build() {
-            return new Request(this);
+        private void setPathMutable() {
+            setMutable(MUTABLE_PATH);
+        }
+
+        private boolean isMutable(int mask) {
```
iamdanfox: thoughts on calling these methods `getBitFlag` and `setBitFlag` ?

dialogue-target/src/main/java/com/palantir/dialogue/Request.java
```diff
@@ -119,40 +123,32 @@ public static Builder builder() {
...
-                new ImmutableListMultimap.Builder<String, String>().build();
 
-        @Nullable
-        private ListMultimap<String, String> headerParams;
+        @SuppressWarnings("UnnecessaryLambda") // Avoid unnecessary allocation
+        private static final com.google.common.base.Supplier<List<String>> MAP_VALUE_FACTORY = () -> new ArrayList<>(1);
 
-        @Nullable
-        private ImmutableListMultimap.Builder<String, String> queryParams;
+        private static final int MUTABLE_HEADERS = 1;
```
iamdanfox: MUTABLE_HEADERS_BITMASK ?

dialogue-target/src/main/java/com/palantir/dialogue/Request.java
```diff
@@ -249,81 +243,81 @@ private Builder() {}
...
-                pathParams = ImmutableMap.builder();
-
-                if (existingPathParams != null) {
-                    existingPathParams.forEach(pathParams::put);
-                    existingPathParams = null;
-                }
+        private Map<String, String> mutablePathParams() {
+            if (!isPathMutable()) {
+                setPathMutable();
+                pathParams = new HashMap<>(pathParams);
```
iamdanfox, line 270: I believe ImmutableMap preserves insertion order... should this be a LinkedHashMap just in case?

dialogue-target/src/main/java/com/palantir/dialogue/Request.java
```diff
@@ -249,81 +243,81 @@ private Builder() {}
...
             }
             return pathParams;
         }
 
         private ListMultimap<String, String> unmodifiableHeaderParams() {
-            if (existingUnmodifiableHeaderParams != null) {
-                return existingUnmodifiableHeaderParams;
-            }
+            return isHeaderMutable() ? Multimaps.unmodifiableListMultimap(headerParams) : headerParams;
```
iamdanfox, line 276: fwiw, I found this unmodifiable wrapper degrades the `forEach` performance pretty hard, because `AbstractMapBasedMultimap#forEach` has got a very natural implementation that just zooms through its inner map:
```java
  @Override
  public void forEach(BiConsumer<? super K, ? super V> action) {
    checkNotNull(action);
    map.forEach(
        (key, valueCollection) -> valueCollection.forEach(value -> action.accept(key, value)));
  }
```

Unfortunately `UnmodifiableListMultimap` drops through to the `default` implementation of `Multimap#forEach` which goes through the whole indirection of building up an entryset and then iterating through that 
```java
  default void forEach(BiConsumer<? super K, ? super V> action) {
    checkNotNull(action);
    entries().forEach(entry -> action.accept(entry.getKey(), entry.getValue()));
  }
```

dialogue-target/src/main/java/com/palantir/dialogue/Request.java
carterkozak: Eh, I wouldn't go though the trouble at this point, but I don't have a strong preference

dialogue-target/src/main/java/com/palantir/dialogue/Request.java
iamdanfox: Given that no methods in Request actually mutate these maps, maybe we can just be brave and omit this?

dialogue-target/src/main/java/com/palantir/dialogue/Request.java
carterkozak: If order is important for any of these collections we have some bigger problems!

dialogue-target/src/main/java/com/palantir/dialogue/Request.java
carterkozak: Ya, these names are all awful, sorry. Will update them in a moment!

dialogue-target/src/main/java/com/palantir/dialogue/Request.java
carterkozak: Let's fix this upstream: https://github.com/google/guava/issues/3911

carterkozak: Hmm, I'd prefer not to remove the wrappers in this change, not sure it's the best idea long term. I wouldn't mind fully-mutable request objects to be honest, our servers use modifiable request objects and it hasn't really come back to bite us. Realistically we have  a pretty strong thread model for dialogue requests, after a request is created it's handed off to the next channel which takes over ownership... probably out of scope for this change but worth considering.

iamdanfox (approved):

PR #802, opened 2020-05-29
Support http OPTIONS

iamdanfox:
## Before this PR

https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/OPTIONS is an HTTP verb that all of our servers implement because browser CORS preflight requests require it. Just like 'PATCH', I'm not intending this to be usable from conjure defined APIs, but it will be a useful implementation detail of the $ saving PR: https://github.com/palantir/dialogue/pull/794

## After this PR
==COMMIT_MSG==
Endpoints can now specify http OPTIONS verb
==COMMIT_MSG==

### DIFF
--- a/dialogue-okhttp-client/src/main/java/com/palantir/dialogue/OkHttpChannel.java
+++ b/dialogue-okhttp-client/src/main/java/com/palantir/dialogue/OkHttpChannel.java
@@ -80,6 +80,11 @@ public final class OkHttpChannel implements Channel {
                 Preconditions.checkArgument(!request.body().isPresent(), "HEAD endpoints must not have a request body");
                 okRequest = okRequest.head();
                 break;
+            case OPTIONS:
+                Preconditions.checkArgument(
+                        !request.body().isPresent(), "OPTIONS endpoints must not have a request body");
+                okRequest = okRequest.method("OPTIONS", null);
+                break;
             case POST:
                 okRequest = okRequest.post(toOkHttpBody(request.body()));
                 break;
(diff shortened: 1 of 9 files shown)


dialogue-okhttp-client/src/main/java/com/palantir/dialogue/OkHttpChannel.java
```diff
@@ -93,6 +93,11 @@ public void writeTo(BufferedSink sink) throws IOException {
             case PATCH:
                 okRequest = okRequest.patch(toOkHttpBody(request.body()));
                 break;
+            case OPTIONS:
+                Preconditions.checkArgument(
+                        !request.body().isPresent(), "OPTIONS endpoints must not have a request body");
```
iamdanfox: https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/OPTIONS
>Request has body: No

ferozco:
👍

simulation/src/main/java/com/palantir/dialogue/core/SimulationServer.java
```diff
@@ -74,6 +75,10 @@ public static Builder builder() {
 
     @Override
     public ListenableFuture<Response> execute(Endpoint endpoint, Request request) {
+        if (endpoint.httpMethod() == HttpMethod.OPTIONS) {
+            return Futures.immediateFuture(new TestResponse().code(204));
+        }
```
carterkozak: Not sure I'd add this block quite yet

dialogue-target/src/main/java/com/palantir/dialogue/HttpMethod.java
```diff
@@ -24,4 +24,5 @@
     PATCH,
     POST,
     PUT,
+    OPTIONS
```
carterkozak: Alphabetize!

nit: last element should have a trailing comma so we get a nice one-line diff next time as well :-)
```suggestion
    OPTIONS,
```

dialogue-okhttp-client/src/main/java/com/palantir/dialogue/OkHttpChannel.java
carterkozak: We have tests for failing GET/HEAD requests with bodies, mind adding a variant to test OPTIONS with a request body?

carterkozak (approved):

PR #806, opened 2020-06-01
Fix BALANCED_RTT (and add simulations)

iamdanfox:
## Before this PR

My original PR contained a bug where I zipped up two arrays by index, but one of them was shuffled and the other wasn't. This made all the fancy RTT measuring _100% pointless_ 🤦.

Luckily we haven't cut a release yet, so I think the first release with BALANCED_RTT should still work!

## After this PR
==COMMIT_MSG==
Fix BALANCED_RTT (and add simulations)
==COMMIT_MSG==

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/BalancedNodeSelectionStrategyChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/BalancedNodeSelectionStrategyChannel.java
@@ -92,6 +92,7 @@ final class BalancedNodeSelectionStrategyChannel implements LimitedChannel {
         this.rttSampler = samplingEnabled == RttSampling.DEFAULT_OFF ? null : new RttSampler(channels, clock);
         this.channels = IntStream.range(0, channels.size())
                 .mapToObj(index -> new MutableChannelWithStats(
+                        index,
                         channels.get(index),
                         clock,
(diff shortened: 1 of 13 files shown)


dialogue-core/src/main/java/com/palantir/dialogue/core/BalancedNodeSelectionStrategyChannel.java
```diff
@@ -120,28 +121,34 @@
...
-            MutableChannelWithStats channel = chans.get(i);
-            float rttSpectrum = rttSpectrums[i];
+        float[] rttSpectrums =
+                rttSampler != null ? rttSampler.computeRttSpectrums() : new float[shuffledChannels.size()];
+
+        SortableChannel[] snapshotArray = new SortableChannel[shuffledChannels.size()];
+        for (int i = 0; i < shuffledChannels.size(); i++) {
+            MutableChannelWithStats channel = shuffledChannels.get(i);
+            // beware rttSpectrums have the original order, but shuffledChannels are shuffled, so 'i' won't line up
+            float rttSpectrum = rttSpectrums[channel.hostIndex];
```
iamdanfox, line 145: this was the bug - changing i -> hostIndex made it work. Annoyingly since all the unit tests use a `getScoresForTesting` codepath, they never hit this.

I think it's literally just the simulations that reveal this.

carterkozak (approved):

PR #809, opened 2020-06-02
Try to avoid embedding 0.0.0

iamdanfox:
## Before this PR

Looking at graphs of user-agents in our internal log browsing product, there are a _ton_ of things that say `... LogReceiverService/0.0.0 dialogue/1.54.0`. This is because we mistakenly embedded 0.0.0 in all generated code (until https://github.com/palantir/conjure-java/pull/879) and those products have not had a reason to pick up a new log-receiver-api-dialogue jar.

This seems kinda lame, and easy to backfill.

## After this PR
==COMMIT_MSG==
Embedded user-agents will no longer include 0.0.0 if a version can be found in the relevant dialogue jar
==COMMIT_MSG==

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/UserAgentEndpointChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/UserAgentEndpointChannel.java
@@ -56,8 +56,17 @@ final class UserAgentEndpointChannel implements EndpointChannel {
     }
 
     private static UserAgent augmentUserAgent(UserAgent baseAgent, Endpoint endpoint) {
+        String endpointDeclaredVersion = endpoint.version();
+        String endpointJarVersion = endpoint.getClass().getPackage().getImplementationVersion();
+
+        // Until conjure-java 5.14.2, we mistakenly embedded 0.0.0 in everything. This logic attempts
+        // to work-around this and produce a more helpful user agent
+        String endpointVersion = "0.0.0".equals(endpointDeclaredVersion) && endpointJarVersion != null
+                ? endpointJarVersion
+                : endpointDeclaredVersion;
+
         return baseAgent
-                .addAgent(UserAgent.Agent.of(endpoint.serviceName(), endpoint.version()))
+                .addAgent(UserAgent.Agent.of(endpoint.serviceName(), endpointVersion))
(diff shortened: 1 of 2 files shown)


dialogue-core/src/main/java/com/palantir/dialogue/core/UserAgentEndpointChannel.java
```diff
@@ -56,8 +56,17 @@ static EndpointChannel create(EndpointChannel delegate, Endpoint endpoint, UserA
     }
 
     private static UserAgent augmentUserAgent(UserAgent baseAgent, Endpoint endpoint) {
+        String endpointDeclaredVersion = endpoint.version();
+        String endpointJarVersion = endpoint.getClass().getPackage().getImplementationVersion();
```
carterkozak: Can we avoid executing this unless the declared version is all zeros?

carterkozak (approved):

dialogue-core/src/main/java/com/palantir/dialogue/core/UserAgentEndpointChannel.java
```diff
@@ -56,8 +56,19 @@ static EndpointChannel create(EndpointChannel delegate, Endpoint endpoint, UserA
     }
 
     private static UserAgent augmentUserAgent(UserAgent baseAgent, Endpoint endpoint) {
+        String endpointVersion = endpoint.version();
+
+        // Until conjure-java 5.14.2, we mistakenly embedded 0.0.0 in everything. This fallback logic attempts
+        // to work-around this and produce a more helpful user agent
+        if (endpointVersion.equals("0.0.0")) {
```
ferozco, line 63: I wonder if we could make the endpoint version be optional and then exclude it from code gen by default. It would allow us to cache conjure generation much more aggressively rather than force regen on each commit/version

carterkozak: I think we codegen the impl version lookup

iamdanfox: Yep, see https://github.com/palantir/conjure-java/pull/879/files#diff-9fd8c6f3733ade117b02507eef44be37R45-R47

PR #812, opened 2020-06-03
Improve deprecation slf4j logging

carterkozak:
Previously our implementation had several issues:
* Logging was rate limited by service-name only, not including
  endpoint name.
* The logging did not include channel name.
* Multiple instances failed to preserve logging rate limiting,
  for example using the jaxrs shim which creates a new EndpointChannel
  for every request.

==COMMIT_MSG==
Improve deprecation slf4j logging
==COMMIT_MSG==

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/DeprecationWarningChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/DeprecationWarningChannel.java
@@ -29,6 +29,7 @@ import com.palantir.logsafe.SafeArg;
 import com.palantir.tritium.metrics.registry.TaggedMetricRegistry;
 import java.time.Duration;
 import java.util.Optional;
+import org.immutables.value.Value;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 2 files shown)


dialogue-core/src/main/java/com/palantir/dialogue/core/DeprecationWarningChannel.java
```diff
@@ -82,9 +85,10 @@ static EndpointChannel create(Config cf, EndpointChannel delegate, Endpoint endp
             }
 
             meter.mark();
-            if (tryAcquire(endpoint.serviceName())) {
+            if (tryAcquire(channelName, endpoint)) {
```
iamdanfox: could also wrap this in a `log.isWarnEnabled()`

dialogue-core/src/main/java/com/palantir/dialogue/core/DeprecationWarningChannel.java
```diff
@@ -82,9 +85,10 @@ static EndpointChannel create(Config cf, EndpointChannel delegate, Endpoint endp
...
                 log.warn(
```
iamdanfox, line 89: also would you mind renaming the SafeArg.of("service" to "server"? the log screenshot I saw in slack looked v odd

dialogue-core/src/main/java/com/palantir/dialogue/core/DeprecationWarningChannel.java
```diff
@@ -101,9 +105,11 @@ static EndpointChannel create(Config cf, EndpointChannel delegate, Endpoint endp
      * penalty for failing to rate limit correctly is a few extra log lines, and we choose that penalty over the cost
      * of synchronization.
      */
-    private boolean tryAcquire(String serviceName) {
-        if (loggingRateLimiter.getIfPresent(serviceName) == null) {
-            loggingRateLimiter.put(serviceName, SENTINEL);
+    private boolean tryAcquire(String channelName, Endpoint endpoint) {
+        LoggingRateLimiterKey key =
+                LoggingRateLimiterKey.of(channelName, endpoint.serviceName(), endpoint.endpointName());
```
iamdanfox, line 110: we could create this key upfront to save recreating it on every RPC? (like ~L76?)

dialogue-core/src/main/java/com/palantir/dialogue/core/DeprecationWarningChannel.java
```diff
@@ -42,23 +43,25 @@
  */
 final class DeprecationWarningChannel implements EndpointChannel {
     private static final Logger log = LoggerFactory.getLogger(DeprecationWarningChannel.class);
+    // Static cache avoids poor interactions with the jaxrs shim and consumers which recreate clients too aggressively.
+    private static final Cache<LoggingRateLimiterKey, Object> loggingRateLimiter =
```
ferozco, line 47: I'm a bit worried about adding static state, would lazily constructing a cache per instance help alleviate your concerns while avoiding static state?

dialogue-core/src/main/java/com/palantir/dialogue/core/DeprecationWarningChannel.java
carterkozak: yes and no -- I considered it but I think it would be harmful for the jaxrs shim. I don't want to optimize for the shim, but I don't expect most calls to hit deprecated endpoints.

dialogue-core/src/main/java/com/palantir/dialogue/core/DeprecationWarningChannel.java
carterkozak: we already have a cache per instance, no?

carterkozak: fwiw logging is all static state, and this is only used for logging

ferozco: currently the cache key is likely to be such that there is a single entry per instance, but it could easily be changed down the line. I guess since the cache is just for logging its fine

carterkozak: Sorry, I neglected to tag you into a slack thread -- we;ve got users who are seeing a warning logged for every rpc instead of rate limited because they're using a legacy codepath (used by both the jaxrs shim and older generated dialogue) which creates a new EndpointChannel instance for every call.

ferozco:
👍

carterkozak:
🤖 🍿

PR #823, opened 2020-06-08
Fix the Encodings.AbstractJacksonEncoding serializer error message

carterkozak:
It's not necessarily a bug when serialization fails, we write
directly to the socket which may fail if there's a problem
in the network layer or the remote server rejects the
request.

==COMMIT_MSG==
Fix the Encodings.AbstractJacksonEncoding serializer error message
==COMMIT_MSG==


ferozco:
👍

carterkozak:
:robot: :popcorn:

PR #827, opened 2020-06-09
Trace DNS lookups

carterkozak:
==COMMIT_MSG==
Trace DNS lookups
==COMMIT_MSG==

Too much? Has been a pain point in several environments, but we need to balance against too much overhead as well.

2020-06-09_1426x607_screen


carterkozak:
Upon reflection I don't think this is too much tracing -- if every request creates a new connection we have bigger problems, if not there's no dns lookup.

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/InstrumentedDnsResolver.java
```diff
@@ -0,0 +1,51 @@
...
+
+    InstrumentedDnsResolver(DnsResolver delegate) {
+        this.delegate = delegate;
+    }
+
+    @Override
+    public InetAddress[] resolve(String host) throws UnknownHostException {
+        try (CloseableTracer ignored = CloseableTracer.startSpan("DnsResolver.resolve")) {
+            return delegate.resolve(host);
+        }
```
iamdanfox, line 37: I like having this tracing available to debug stuff - my concern is that the trace log viewer can already get really noisy when there are hundreds of RPC calls.

Would it be crazy to short-circuit this based on the log.isDebugEnabled? I'm thinking under 'normal' conditions, even when a trace is sampled, most people don't care about the DNS resolving.  If we're really trying to dig into a perf thing though we could set `com.palantir.dialogue.hc5: DEBUG` and then get these sweet granular spans??

iamdanfox: Actually scratch that the trace-log-viewer can easily be configured to hide spans that are very small, so I think we might as well just record this stuff. It looks incredibly fast in the unsampled case!

iamdanfox (approved):
Yep I'm game

PR #828, opened 2020-06-10
Don't impose c-j-r 6.0

iamdanfox:
I just saw a very strange looking PR which was trying to upgrade dev-env on an internal repo:

https://github.p.b/foundry/people/pull/1631/files

but the dev-env upgrade pulled in a dialogue upgrade which upgraded past a c-j-r major rev.

This seems a bit undesirable, because I really wanted witchcraft's changelog to announce the change to people, so that they had one clear thing to revert if something went wrong.


versions.props
```diff
@@ -40,4 +39,7 @@ org.openjdk.jmh:* = 1.23
 # dependency-upgrader:OFF
 # Match conjure-java-runtime okhttp version
 com.squareup.okhttp3:* = 3.13.1
+
+# Don't want to impose a major upgrade on people
+com.palantir.conjure.java.runtime:* = 5.10.0
```
iamdanfox, line 44: This is actually consistent with what we've done in tracing-java: 

https://github.com/palantir/tracing-java/blob/develop/versions.props#L2-L3

carterkozak (approved):

PR #830, opened 2020-06-10
BlockingChannelAdapter never throws

iamdanfox:
## Before this PR

Part of the contract of a Channel is that `execute` should never throw, but instead send back a failed future. We recently saw some stacktraces from `delivery-metadata` internally that violated this, and were only caught by the NeverThrowChannel:

```
java.util.concurrent.RejectedExecutionException:
	at org.jboss.threads.RejectingExecutor.execute(RejectingExecutor.java:38)
	at org.jboss.threads.EnhancedQueueExecutor.rejectShutdown(EnhancedQueueExecutor.java:2089)
	at org.jboss.threads.EnhancedQueueExecutor.execute(EnhancedQueueExecutor.java:767)
	at java.util.concurrent.AbstractExecutorService.submit(AbstractExecutorService.java:118)
	at com.palantir.tritium.metrics.TaggedMetricsExecutorService.submit(TaggedMetricsExecutorService.java:61)
	at com.palantir.tracing.WrappingExecutorService.submit(WrappingExecutorService.java:102)
	at com.google.common.util.concurrent.ForwardingExecutorService.submit(ForwardingExecutorService.java:109)
	at com.palantir.dialogue.blocking.BlockingChannelAdapter$BlockingChannelAdapterChannel.execute(BlockingChannelAdapter.java:82)
	at com.palantir.dialogue.core.HostMetricsChannel.execute(HostMetricsChannel.java:76)
	at com.palantir.dialogue.core.NeverThrowChannel.execute(NeverThrowChannel.java:49)
```

## After this PR
==COMMIT_MSG==
BlockingChannelAdapter never throws, even if the user-provided `executor` does something unexpected
==COMMIT_MSG==

### DIFF
--- a/dialogue-blocking-channels/src/main/java/com/palantir/dialogue/blocking/BlockingChannelAdapter.java
+++ b/dialogue-blocking-channels/src/main/java/com/palantir/dialogue/blocking/BlockingChannelAdapter.java
@@ -76,11 +76,6 @@ public final class BlockingChannelAdapter {
 
         @Override
         public ListenableFuture<Response> execute(Endpoint endpoint, Request request) {
-            if (Thread.currentThread().isInterrupted()) {
-                // is someone asked to interrupt this thread, we abort early to avoid kicking off network requests
-                return Futures.immediateCancelledFuture();
-            }
-
             SettableFuture<Response> settableFuture = SettableFuture.create();
             BlockingChannelAdapterTask runnable =
                     new BlockingChannelAdapterTask(delegate, endpoint, request, settableFuture);
--- a/dialogue-blocking-channels/src/main/java/com/palantir/dialogue/blocking/BlockingChannelAdapter.java
+++ b/dialogue-blocking-channels/src/main/java/com/palantir/dialogue/blocking/BlockingChannelAdapter.java
@@ -76,6 +76,11 @@ public final class BlockingChannelAdapter {
 
         @Override
         public ListenableFuture<Response> execute(Endpoint endpoint, Request request) {
+            if (Thread.currentThread().isInterrupted()) {
+                // is someone asked to interrupt this thread, we abort early to avoid kicking off network requests
+                return Futures.immediateCancelledFuture();
+            }
+
             SettableFuture<Response> settableFuture = SettableFuture.create();
             BlockingChannelAdapterTask runnable =
                     new BlockingChannelAdapterTask(delegate, endpoint, request, settableFuture);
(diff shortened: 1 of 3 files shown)


carterkozak (approved):

dialogue-blocking-channels/src/main/java/com/palantir/dialogue/blocking/BlockingChannelAdapter.java
```diff
@@ -76,26 +76,36 @@ private BlockingChannelAdapter() {}
 
         @Override
         public ListenableFuture<Response> execute(Endpoint endpoint, Request request) {
+            if (Thread.currentThread().isInterrupted()) {
+                // is someone asked to interrupt this thread, we abort early to avoid kicking off network requests
+                return Futures.immediateCancelledFuture();
+            }
```
iamdanfox: @carterkozak I've also added this little chunk here because I _think_ the exception we observed was when things were shutting down.

iamdanfox: Actually nvm I think we can land this separately.

dialogue-blocking-channels/src/main/java/com/palantir/dialogue/blocking/BlockingChannelAdapter.java
```diff
@@ -79,23 +79,28 @@ private BlockingChannelAdapter() {}
...
-                        }
-                    },
-                    MoreExecutors.directExecutor());
-            return settableFuture;
+                        },
+                        MoreExecutors.directExecutor());
+                return settableFuture;
+            } catch (RuntimeException | Error e) {
+                // user-provided executor could throw exceptions when we try to submit runnables
+                return Futures.immediateFailedFuture(e);
```
carterkozak, line 102: We could specifically catch RejectedExecutionException and return an immediate cancelled future.

ferozco:
I thought the point of the never throw channel was to handle cases like this so that we didn't have to add try/catch logic into every channel.

carterkozak:
Our failure message notes that when never-throw-channel catches a failure it's a bug, in this case the server is shutting down and has stopped the executor.

iamdanfox:
👍  yep I'd prefer to keep any message from the NeverThrowChannel a high signal indication that something unforeseen has gone wrong inside Dialogue. Think it's simpler to just say the contract of channels is that they don't throw and stick to that everywhere

PR #831, opened 2020-06-10
Thread interruption behaviour

iamdanfox:
## Before this PR

Saw some logs recently from `delivery-metadata` which I _think_ are coming from around the time that things are shutting down.  This made me realise that because we don't check `Thread.isInterrupted()` we might be running a bunch of code even after a user has asked a thread to stop, which might result in confusing exceptions.

## After this PR
==COMMIT_MSG==
When the `interrupted` flag is set on a thread, any attempt to make Dialogue network calls from that thread will immediately short-circuit and return a `SafeRuntimeException`.
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-831.v2.yml
+++ b/changelog/@unreleased/pr-831.v2.yml
@@ -1,0 +1,7 @@
+type: improvement
+improvement:
+  description: When the `interrupted` flag is set on a thread, any attempt to make
+    Dialogue network calls from that thread will immediately short-circuit and return
+    a `SafeRuntimeException`.
+  links:
+  - https://github.com/palantir/dialogue/pull/831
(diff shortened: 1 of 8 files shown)


iamdanfox:
I think there are unlimited edge cases here, e.g. if we just got a failure back from the network and then a user sets `interrupted` - should the RetryingChannel stop retrying immediately?

I figured this fixes the low hanging fruit initially, and maybe we can add that extra complexity if we really need it?

changelog/@unreleased/pr-831.v2.yml
```diff
@@ -0,0 +1,7 @@
+type: improvement
+improvement:
+  description: When the `interrupted` flag is set on a thread, any attempt to make
+    Dialogue network calls from that thread will immediately short-circuit and return
+    a `SafeRuntimeException`.
```
carterkozak: Could we use DialogueException with an InterruptedException cause?

carterkozak:
> if we just got a failure back from the network and then a user sets interrupted - should the RetryingChannel stop retrying immediately?

I believe we already do in the blocking case, in the async case it's a non issue until they begin to block because the call no longer has an owner thread.

carterkozak (approved):

PR #832, opened 2020-06-10
Response implementations provide a delegating getFirstHeader override

carterkozak:
This method is substantially more performant in some cases as it
can avoid building an entire ListMultimap

## After this PR
==COMMIT_MSG==
Response implementations provide a delegating getFirstHeader override
==COMMIT_MSG==

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/ContentDecodingChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/ContentDecodingChannel.java
@@ -118,6 +118,14 @@ final class ContentDecodingChannel implements EndpointChannel {
             return headers;
         }
 
+        @Override
+        public Optional<String> getFirstHeader(String header) {
+            if (!allowHeader(header)) {
+                return Optional.empty();
+            }
+            return delegate.getFirstHeader(header);
+        }
+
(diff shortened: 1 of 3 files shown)


dialogue-core/src/main/java/com/palantir/dialogue/core/ContentDecodingChannel.java
```diff
@@ -118,6 +118,14 @@ public int code() {
             return headers;
         }
 
+        @Override
+        public Optional<String> getFirstHeader(String header) {
+            if (!allowHeader(header)) {
+                return Optional.empty();
+            }
+            return delegate.getFirstHeader(header);
+        }
```
carterkozak, line 127: This one is the :hot_pepper:.

iamdanfox: Nice. Do we see a speedup in any of our benchmarks?

carterkozak: I haven't checked, just noticed this while looking at some other code. Our local benchmarks don't hit anything that responds with compressed data, so they wouldn't hit this path.

iamdanfox (approved):

PR #839, opened 2020-06-12
Drop the log level of RttSampler bounds checking

carterkozak:
This check used to throw, but small variances allowed values to be
a hair beyond 1.0. There's no reason to alarm product owners
when this happens.

==COMMIT_MSG==
Drop the log level of RttSampler bounds checking because there's nothing to worry about.
==COMMIT_MSG==

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/RttSampler.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/RttSampler.java
@@ -104,7 +104,7 @@ final class RttSampler {
             OptionalLong rtt = snapshots[i];
             float rttSpectrum = rtt.isPresent() ? ((float) (rtt.getAsLong() - bestRttNanos)) / rttRange : 0;
             if (rttSpectrum < 0f || rttSpectrum > 1f) {
-                log.warn(
+                log.info(
                         "rttSpectrum should be between 0 and 1",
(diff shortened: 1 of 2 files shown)


ferozco:
👍

dialogue-core/src/main/java/com/palantir/dialogue/core/RttSampler.java
```diff
@@ -104,7 +104,7 @@
             OptionalLong rtt = snapshots[i];
             float rttSpectrum = rtt.isPresent() ? ((float) (rtt.getAsLong() - bestRttNanos)) / rttRange : 0;
             if (rttSpectrum < 0f || rttSpectrum > 1f) {
-                log.warn(
+                log.info(
```
iamdanfox, line 107: possibly even down to `debug` since we're confident it wasn't wildly over here?

iamdanfox: and I guess an `isDebugEnabled` maybe?

carterkozak: I wanted to leave it at a level that's default enabled for rollout so we can validate against outliers just in case. Perhaps in a week we could reduce it.

PR #840, opened 2020-06-12
Disable hc5 revapi until we've published a release

carterkozak:
==COMMIT_MSG==
Fix hc5 module publications
==COMMIT_MSG==


ferozco:
👍

PR #843, opened 2020-06-12
Promote BALANCED_RTT2

iamdanfox:
## Before this PR

I introduced some new logic in https://github.com/palantir/dialogue/pull/794, behind the `BALANCED_RTT` feature flag. We later discovered that it contained a bug, so we fixed it an made a new enum variant `BALANCED_RTT2`. (We rolled back all the servers that were pumping out the bad `BALANCED_RTT` value).

We're now trying out the `BALANCED_RTT2` strategy on one big product internally, with promising results so far.

## After this PR
==COMMIT_MSG==
When users select ROUND_ROBIN, they now get Dialogue's BALANCED_RTT2 strategy, which measures the round trip time (RTT) to each host using OPTIONS requests and prefers the lowest latency host.
==COMMIT_MSG==

`do not merge` until we're happy that this is behaving well on catalog & MP.


iamdanfox:
Closing out because we had to roll back balanced_rtt2 internally due to catalog getting flattened by too many TLS handshakes.

PR #853, opened 2020-06-19
StickyChannelFactory

iamdanfox:
## Before this PR

The last big user of c-j-r & retrofit internally is @j-baker & @jkozlowski's ~internal ski product~ atlasdb proxy. We've brainstormed a ton of different options, and the current winner is to publish something from the dialogue repo so that they can benefit from our 'balanced' heuristic, while still keeping requests within one transaction landing on one upstream (the 'sticky' behaviour).

## After this PR
==COMMIT_MSG==
A new `.getStickyChannels()` method in `:dialogue-clients` allows users with transactional requirements to get a channel which will route requests to one host (sticky) even if it 429s.
==COMMIT_MSG==

This PR introduces a new class, `StickyEndpointChannels`, which is intended to be used by anyone who needs this balanced+sticky behaviour (i.e. only internal-ski-product initially). My intention is that they'd create single-uri DialogueChannels, then use `StickyEndpointChannels` to pick the current best:

```groovy
StickyEndpointChannels channels = new StickyEndpointChannels(host1, host2, host3);

// in each transaction
Channel channel = channels.getSticky();
FooServiceBlocking client = FooServiceBlocking.of(channel, runtime);
client.foo();
client.bar();
client.whatever();

// next transaction
Channel channel = channels.getSticky();
...
```

## Possible downsides?
- ~it's more public API, we'd need to figure out a way of exposing it sanely (perhaps a builder??)~
- this means there would be a queue per host, which might make metrics more confusing (instead of the single queue)


dialogue-core/src/main/java/com/palantir/dialogue/core/BalancedScoreTracker.java
```diff
@@ -0,0 +1,214 @@
...
+ * Chooses nodes based on stats about each channel, i.e. how many requests are currently
+ * being served, how many failures have been seen in the last few seconds and (optionally) also what the best latency
+ * to each node is. Use {@link RttSampling#ENABLED} to switch this on.
+ *
+ * This is intended to be a strict improvement over Round Robin and Random Selection which can leave fast servers
+ * underutilized, as it sends the same number to both a slow and fast node. It is *not* appropriate for transactional
+ * workloads (where n requests must all land on the same server) or scenarios where cache warming is very important.
+ * {@link PinUntilErrorNodeSelectionStrategyChannel} remains the best choice for these.
+ */
+final class BalancedScoreTracker {
```
iamdanfox: I just copied our balanced channel and deleted the RTT stuff for readability

dialogue-core/src/main/java/com/palantir/dialogue/core/StickyEndpointChannels.java
```diff
@@ -0,0 +1,105 @@
...
+
+    @ThreadSafe
+    private static final class Sticky implements EndpointChannelFactory, Channel {
+
+        private final StickyEndpointChannels parent;
+        private final Supplier<ScoreTracker> currentHost;
+
+        private Sticky(StickyEndpointChannels parent) {
+            this.parent = parent;
+            this.currentHost = Suppliers.memoize(parent.tracker::getBestHost);
```
iamdanfox: `Supplies.memoize` is what makes it sticky. It uses the `getBestHost` method to pick a channel once, and then we remain on that channel.

j-baker: and if the host is fucktored, I assume that just becomes a transaction manager level retry. That seems fine.

dialogue-core/src/main/java/com/palantir/dialogue/core/BalancedScoreTracker.java
```diff
@@ -0,0 +1,331 @@
...
+
+        registerGauges(taggedMetrics, channelName, channelStats);
+    }
+
+    /**
+     * Returns all channels, ordered with the best score first. Called on every request, so needs to be performant!
+     * Callers *must* use the {@link ChannelScoreInfo#startRequest} and {@link ChannelScoreInfo#onSuccess} etc
+     * methods to feed information back into the tracker.
+     */
+    public ChannelScoreInfo[] getChannelsByScore() {
```
carterkozak: nit 
```suggestion
    ChannelScoreInfo[] getChannelsByScore() {
```

dialogue-core/src/main/java/com/palantir/dialogue/core/BalancedScoreTracker.java
```diff
@@ -0,0 +1,331 @@
...
+
+        ChannelScoreInfo[] returnArray = new ChannelScoreInfo[snapshotArray.length];
+        for (int i = 0; i < snapshotArray.length; i++) {
+            returnArray[i] = snapshotArray[i].delegate;
+        }
+        return returnArray;
+    }
+
+    public ChannelScoreInfo getSingleBestChannelByScore() {
+        // TODO(dfox): in theory we could optimize this by just looping manually and keeping track of the max
```
carterkozak: Would we need to break ties in random order to prevent affinity? I think we considered this when we implemented the balanced channel originally.

changelog/1.62.1/pr-861.v2.yml
```diff
@@ -0,0 +1,5 @@
+type: fix
+fix:
+  description: Work around http proxy timeouts caused by HTTPCLIENT-1478
+  links:
+  - https://github.com/palantir/dialogue/pull/861
```
carterkozak: something wonky is going on

dialogue-core/src/main/java/com/palantir/dialogue/core/BalancedScoreTracker.java
carterkozak: I see, avoiding the ChannelScoreInfo and operate directly on ScoreSnapshots

iamdanfox:
So before merging:

- [x] i think i'm gonna split out a separate PR (https://github.com/palantir/dialogue/pull/867) to pull out the 'balanced score tracker' thing, because it currently involves throwing the balanced_RTT stuff in the bin
- [x] I think it would be best to expose this via the DialogueClients interface somehow and handle all live-reloading concerns for people.

dialogue-core/src/main/java/com/palantir/dialogue/core/StickyEndpointChannels.java
```diff
@@ -0,0 +1,163 @@
...
+            DialogueFutures.addDirectCallback(future, tracker);
+            return future;
+        }
+    }
+
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    public static class Builder {
```
carterkozak: nit: private constructor. I get snagged by this all the time, I'll try to carve out time to write an error-prone rule at some point.

carterkozak:
> I think it would be best to expose this via the DialogueClients interface somehow and handle all live-reloading concerns for people.

It would be nice to have a single exposed interface for this with the implementation hidden.

I wonder if we'd want to add support for sticky session cookies in this mode to support stable targets behind reverse proxies? Not for this PR, just a longer term thought.

iamdanfox:
@carterkozak ok I think this is ready for another look

dialogue-core/src/main/java/com/palantir/dialogue/core/StickyEndpointChannels.java
```diff
@@ -0,0 +1,165 @@
...
+
+/** Allows requests for a 'transaction' should land on one host. */
+public final class StickyEndpointChannels {
+
+    private final ImmutableList<? extends EndpointChannelFactory> channels;
+    private final BalancedScoreTracker tracker;
+
+    private StickyEndpointChannels(Builder params) {
+        this.channels = params.channels;
+        // TODO(dfox): handle 'channels' of size 0/1 (balanced tracker will barf)
```
carterkozak: Can we replace this todo with a preconditions check, or potentially move handling for the zero/one channel cases into Builder.build?

iamdanfox: Ah I should delete this. BalancedScoreTracker already has a `Preconditions.checkState(channelCount >= 2, "At least two channels required");`, and in ReloadingClientFactory we avoid even constructing one of these if we detect 0 or 1 uris.

dialogue-core/src/main/java/com/palantir/dialogue/core/StickyEndpointChannels.java
```diff
@@ -0,0 +1,165 @@
...
+
+        /**
+         * .
+         * @deprecated prefer {@link #endpoint}, as it allows binding work upfront
+         */
+        @Deprecated
+        @Override
+        public ListenableFuture<Response> execute(Endpoint _endpoint, Request _request) {
+            // TODO(dfox): could we delete this entirely?
+            throw new UnsupportedOperationException("Not implemented");
```
carterkozak, line 101: Since this ends up exposing a Channel/EndpointChannel, I think we should provide the naive implementation of this method to prevent surprises.

dialogue-core/src/main/java/com/palantir/dialogue/core/StickyEndpointChannels.java
```diff
@@ -0,0 +1,165 @@
...
+        }
+
+        @Override
+        public ListenableFuture<Response> execute(Request request) {
+            tracker.startRequest();
+            ListenableFuture<Response> future = delegate.execute(request);
+            tracker.observability().markRequestMade();
+            DialogueFutures.addDirectCallback(future, tracker);
+            return future;
+        }
```
carterkozak, line 126: nit: tostring

carterkozak (approved):
lgtm, I think this will eventually be helpful for pg long pooling as well.

Let's update the commit message block away from `[proof-of-concept] balanced + sticky channels` before merging ;-)

dialogue-core/src/main/java/com/palantir/dialogue/core/StickyEndpointChannels.java
iamdanfox: So I had this originally, but I mistakenly ended up going through _two_ backcompat layers and missing the fast path (cos no tests _failed_).  Ideally, I'd like to get conjure-java to codegen a new factory method `.of(EndpointChannelFactory, Runtime)` and then we wouldn't have to pretend everything is a Channel anymore.

iamdanfox:
I'm gonna do a quick pTML and verify this actually passes tests in internal-ski-product before we actually release this and are committed to the API!

EDIT https://github.p.b/foundry/a-c-j/pull/1011/files

dialogue-clients/src/main/java/com/palantir/dialogue/clients/DialogueClients.java
```diff
@@ -65,19 +66,25 @@ public static ReloadingFactory create(Refreshable<ServicesConfigBlock> scb) {
         Channel getChannel(String serviceName);
     }
 
-    public interface ReloadingStickyFactory {
-        StickyChannels getStickyChannels(String serviceName);
+    public interface ToPerHostClientFactory {
```
carterkozak: Name is odd, but should only be used fluently. Maybe `ToPerHostClientFactory` -> `PinnableClientFactory` an `PerHostClientFactory` -> `PinningClientFactory`? Naming is difficult.

iamdanfox: yeah I got so fed up with trying to come up with a good name for this interface I threw it away completely. The two methods are now just tacked onto the ReloadingClientFactory

dialogue-clients/src/main/java/com/palantir/dialogue/clients/ReloadingClientFactory.java
```diff
@@ -126,97 +127,76 @@ public Channel getChannel(String serviceName) {
...
-                        .build();
-
-                return new AbstractStickyChannels() {
-                    @Override
-                    public Channel getSingleHostChannel() {
-                        return stickyEndpointChannels.getStickyChannel();
-                    }
-                };
+            if (singleHostChannels.size() == 1) {
+                return () -> singleHostChannels.get(0);
```
carterkozak, line 192: :+1: this refactor to centralize zero and one uri cases makes the code a lot easier to read, thanks!

carterkozak (approved):

PR #854, opened 2020-06-19
Dialogue url encoding escapes uri sub-delimiter characters

carterkozak:
==COMMIT_MSG==
Dialogue url encoding escapes Uri sub-delimiter characters.
The BaseUrl encoder was implemented to strictly follow rfc3986
URI generic syntax, however it did not take into account the
[section-3.3](https://tools.ietf.org/html/rfc3986#section-3.3)
which describes scheme-specific (http, in our case) reserved
characters.
We ensure broader compatibility with existing http infrastructure
by encoding characters that may be used by some http client and
server implementations.
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-854.v2.yml
+++ b/changelog/@unreleased/pr-854.v2.yml
@@ -1,0 +1,14 @@
+type: fix
+fix:
+  description: |-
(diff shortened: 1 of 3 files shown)


ferozco (approved):

carterkozak:
:robot: :popcorn:

PR #858, opened 2020-06-23
RetryingChannel scheduler uses daemon threads

carterkozak:
==COMMIT_MSG==
RetryingChannel scheduler uses daemon threads and no longer keeps the JVM alive on its own.
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-858.v2.yml
+++ b/changelog/@unreleased/pr-858.v2.yml
@@ -1,0 +1,6 @@
+type: fix
+fix:
+  description: RetryingChannel scheduler uses daemon threads and no longer keeps the
+    JVM alive on its own.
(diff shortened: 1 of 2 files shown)


iamdanfox:
So the only possible downside I could think of here was if someone had a 'shutdown task' that wanted to notify someone or persist something then this change could mean the JVM could shut down with it failing to go through.

In practise, we recommend all teams build their software to be resilient to unexpected interruption _anyway_, so people shouldn't be relying on this.

iamdanfox (approved):

carterkozak:
Right, that would only impact the async client, and only if nothing blocked waiting for the call to complete, which would be dubious at best.

PR #859, opened 2020-06-23
Reduce peak thread count using a scheduled executor for eviction

carterkozak:
## Before this PR

Previously each connection pool had its own idle connection eviction
thread which slept in a loop.

## After this PR
Now we reuse a single threaded
scheduled executor to schedule all evictions.
==COMMIT_MSG==
Reduce peak thread count using a scheduled executor for eviction
==COMMIT_MSG==

## Possible downsides?
This scheduler isn't passed in, and is a global singleton. Singletons are bad. However, this singleton is shared between all consumers (jaxrsclients, DialogueClients, wc factory, etc) to reduce thread churn.
It's possible a misbehaving pool could have a bug which results in the eviction thread blocking for longer than we expect.

### DIFF
--- a/dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ScheduledIdleConnectionEvictor.java
+++ b/dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ScheduledIdleConnectionEvictor.java
@@ -1,0 +1,77 @@
+/*
+ * (c) Copyright 2020 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.dialogue.hc5;
+
+import com.google.common.base.Suppliers;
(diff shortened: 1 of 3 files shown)


dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ScheduledIdleConnectionEvictor.java
```diff
@@ -0,0 +1,79 @@
...
+            ConnPoolControl<?> connectionManager, Duration delayBetweenChecks, Duration maxIdleTime) {
+        return schedule(connectionManager, delayBetweenChecks, maxIdleTime, sharedScheduler.get());
+    }
+
+    private static ScheduledFuture<?> schedule(
+            ConnPoolControl<?> connectionManager,
+            Duration delayBetweenChecks,
+            Duration maxIdleTime,
+            ScheduledExecutorService scheduler) {
+        return scheduler.scheduleWithFixedDelay(
```
iamdanfox, line 64: is it worth putting a little sanity precondition here to make sure we can never try to set delayBetweenChecks to something ridiculous (even faster than every 1 second seems a bit excessive?).

iamdanfox (approved):
LGTM. It is one more divergence between hc4 and hc5 - guess there's just an incentive to keep people on latest!

carterkozak:
I think we can remove hc4 at this point. Thoughts?

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ScheduledIdleConnectionEvictor.java
carterkozak: woop, missed this before adding merge-when-ready.

The check interval is based in part on the socket timeout, a very low socket timeout will be self-limiting to an extent, but if it works we need to ensure we're removing old connections quickly as well, otherwise we may see a spike in nohttpresponse exceptions.

PR #860, opened 2020-06-23
use metric-schema for our lazy gauges

iamdanfox:
## Before this PR

We'd landed the fancy new `libraryName` and `libraryOrigin` functionality in metric-schema, but some of our metrics weren't using these because they'd been manually created.  Now that metric-schema offers a getter (https://github.com/palantir/metric-schema/pull/293), we can make _everything_ use metric schema.

## After this PR
==COMMIT_MSG==
All Dialogue metrics now use metric-schema generated code, ensuring that they all have libraryName & libraryVersion tags.
==COMMIT_MSG==

cc @carterkozak


carterkozak (approved):
It's beautiful!

PR #861, opened 2020-06-23
Work around http proxy timeouts caused by HTTPCLIENT-1478

carterkozak:
==COMMIT_MSG==
Work around http proxy timeouts caused by HTTPCLIENT-1478
==COMMIT_MSG==

### DIFF
--- a/dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientChannels.java
+++ b/dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientChannels.java
@@ -40,6 +40,7 @@ import com.palantir.tritium.metrics.registry.TaggedMetricRegistry;
 import java.io.Closeable;
 import java.io.IOException;
 import java.net.MalformedURLException;
+import java.net.ProxySelector;
 import java.net.URL;
 import java.time.Duration;
 import java.util.ArrayList;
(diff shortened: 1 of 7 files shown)


dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientChannels.java
```diff
@@ -394,7 +395,12 @@ public CloseableClient build() {
 
             HttpClientBuilder builder = HttpClients.custom()
                     .setDefaultRequestConfig(RequestConfig.custom()
-                            .setConnectTimeout(connectTimeout)
+                            // HTTPCLIENT-1478: Work around the connection timeout being used to negotiate proxy
+                            // traffic.
+                            .setConnectTimeout(
+                                    ProxySelector.getDefault().equals(conf.proxy())
+                                            ? connectTimeout
+                                            : Timeout.ofMilliseconds(socketTimeoutMillis))
```
carterkozak, line 403: this block is the non-test change.

iamdanfox: Ok so just to make sure i understand - this behaviour was working fine on hc4 and seems to be borked on hc5, which is why we've got this workaround in one place only?

iamdanfox (approved):

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientChannels.java
carterkozak: That’s correct, I opened an upstream PR to fix the bug a few minutes ago as well.

PR #864, opened 2020-06-24
eliminate false positive warning logs

iamdanfox:
We had a complaint in `#dev-foundry-infra` of too much logspam from this line, and in the case where the scores are all the same number anyway, the warning isn't necessary.

cc @robert3005 who also hit this

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/BalancedNodeSelectionStrategyChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/BalancedNodeSelectionStrategyChannel.java
@@ -40,7 +40,6 @@ import java.util.List;
 import java.util.Optional;
 import java.util.Random;
 import java.util.concurrent.atomic.AtomicInteger;
-import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 import javax.annotation.Nullable;
 import org.slf4j.Logger;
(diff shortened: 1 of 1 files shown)


dialogue-core/src/main/java/com/palantir/dialogue/core/BalancedNodeSelectionStrategyChannel.java
```diff
@@ -310,6 +311,10 @@ private static void registerGauges(
         }
     }
 
+    private static int countDistinct(long[] longs) {
+        return Arrays.stream(longs).boxed().collect(Collectors.toSet()).size();
```
carterkozak:
```suggestion
        return Arrays.stream(longs).distinct().count();
```

Alternatively:

```suggestion
    private static boolean containsMultipleDistinctValues(long[] longs) {
        if (longs.length < 2) return false;
        for (int i = 1; i < longs.length; i++) {
            if (longs[0] != longs[i]) return true;
        }
        return false;
    }
```

carterkozak:
approving pending improvement to the `countDistinct` implementation :+1:

dialogue-core/src/main/java/com/palantir/dialogue/core/BalancedNodeSelectionStrategyChannel.java
iamdanfox: 🤦 yep definitely going for the simple distinct - this method is only called when we sample gauges (i.e. once every 30 seconds), so I'm ok with a stream here

PR #867, opened 2020-06-25
[prereq] Pull out Balanced logic to a BalancedScoreTracker

iamdanfox:
## Before this PR

I have a big old PR here https://github.com/palantir/dialogue/pull/853, so just want to land a refactor independently to make the diff easier to read.

## After this PR
==COMMIT_MSG==
Pull out Balanced logic to a BalancedScoreTracker class
==COMMIT_MSG==

## Possible downsides?

- this does chop out the BALANCED_RTT2 functionality, but I think this is OK because we had to turn it off everywhere and it's gonna need a little more work to get it into a useable state. figured there's not much point polishing somethng that isn't used right now!


dialogue-core/src/main/java/com/palantir/dialogue/core/BalancedNodeSelectionStrategyChannel.java
```diff
@@ -58,47 +47,19 @@
...
+    private final ImmutableList<LimitedChannel> channels;
 
     BalancedNodeSelectionStrategyChannel(
             ImmutableList<LimitedChannel> channels,
             Random random,
             Ticker ticker,
             TaggedMetricRegistry taggedMetrics,
             String channelName,
-            RttSampling samplingEnabled) {
+            RttSampling _samplingEnabled) {
```
iamdanfox: I'm intending to re-write this _after_ we finish the internal-ski-product use case.

carterkozak: Might as well remove this flag entirely for now, can always plumb it through again later.

dialogue-core/src/main/java/com/palantir/dialogue/core/BalancedNodeSelectionStrategyChannel.java
```diff
@@ -109,283 +70,33 @@
...
-
     @VisibleForTesting
     IntStream getScoresForTesting() {
-        return Arrays.stream(computeScores(rttSampler, channels)).mapToInt(SortableChannel::getScore);
+        return tracker.getScoresForTesting();
     }
 
     @Override
     public String toString() {
         return "BalancedNodeSelectionStrategyChannel{channels=" + channels + '}';
```
carterkozak: Let's add BalancedScoreTracker to the toString

dialogue-core/src/main/java/com/palantir/dialogue/core/BalancedNodeSelectionStrategyChannel.java
```diff
@@ -58,47 +47,19 @@
...
             Random random,
             Ticker ticker,
             TaggedMetricRegistry taggedMetrics,
             String channelName,
-            RttSampling samplingEnabled) {
+            RttSampling _samplingEnabled) {
         Preconditions.checkState(channels.size() >= 2, "At least two channels required");
-        this.random = random;
-        this.clock = ticker;
-        this.rttSampler = samplingEnabled == RttSampling.DEFAULT_OFF ? null : new RttSampler(channels, clock);
```
carterkozak, line 92: I'd prefer to remove the RttSampler and related tests entirely, git won't forget about them.

carterkozak:
lgtm aside from a few nits :+1:
I prefer to delete unused code entirely, after spending a few years in a large, old project, but if you plan to reimplement rtt very soon I could be swayed.

PR #870, opened 2020-06-26
Allow dialogue-hc5 to support TLSv1.3 when ciphers are available

carterkozak:
## Before this PR
Only TLSv1.2

## After this PR
==COMMIT_MSG==
Allow dialogue-hc5 to support TLSv1.3 when ciphers are available
==COMMIT_MSG==

Draft: I need to do some verification with this change.

### DIFF
--- a/dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientChannels.java
+++ b/dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientChannels.java
@@ -370,6 +370,10 @@ public final class ApacheHttpClientChannels {
             SSLSocketFactory rawSocketFactory = conf.sslSocketFactory();
             SSLConnectionSocketFactory sslSocketFactory = new SSLConnectionSocketFactory(
                     MetricRegistries.instrument(conf.taggedMetricRegistry(), rawSocketFactory, name),
+                    // Use 'null' to allow the supported protocols that hc5 considers secure: TLSv1.2 and (if
+                    // available) TLSv1.3.
+                    // We don't have the data to discover supported protocols directly here, which would require
+                    // either an SSLContext instance, or a Socket.
                     null,
                     supportedCipherSuites(
                             conf.enableGcmCipherSuites()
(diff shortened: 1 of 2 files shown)


dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientChannels.java
```diff
@@ -370,7 +370,7 @@ public CloseableClient build() {
             SSLSocketFactory rawSocketFactory = conf.sslSocketFactory();
             SSLConnectionSocketFactory sslSocketFactory = new SSLConnectionSocketFactory(
                     MetricRegistries.instrument(conf.taggedMetricRegistry(), rawSocketFactory, name),
-                    new String[] {"TLSv1.2"},
+                    null,
```
ferozco, line 377: seems like setting `null` will result in all TLS protocols being enabled. What happens if we declare 1.3 but do not have the corresponding ciphers? i.e. could we be more explicit and set both 1.2 and 1.3?

carterkozak: If we explicitly set both, we'll throw exceptions on java versions which do not support 1.3. Using null delegates to hc5 to support all available versions, and exclude the weak versions, which results in either `[TLSv1.2]` or `[TLSv1.3, TLSv1.2]`.

We would have to either get an instance of the SSLContext (which isn't provided by the ClientConfiguration API) or create a socket to access this information.

ferozco: ok cool makes sense, mind adding a comment for posterity?

ferozco:
👍

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientChannels.java
carterkozak: :+1:

PR #873, opened 2020-06-29
Resolve low success rate of `short_outage_on_one_node` with limiter+balanced

carterkozak:
## Before this PR
See the simulations

### [Simulations](https://github.com/palantir/dialogue/blob/ckozak/fix_short_failure/simulation/src/test/resources/report.md#short_outage_on_one_nodeconcurrency_limiter_round_robin)

==COMMIT_MSG==
Resolve low success rate of `short_outage_on_one_node` with limiter+balanced
==COMMIT_MSG==

## Possible downsides?
The values may not be tuned exactly right.


dialogue-core/src/main/java/com/palantir/dialogue/core/BalancedNodeSelectionStrategyChannel.java
```diff
@@ -47,6 +49,11 @@
 final class BalancedNodeSelectionStrategyChannel implements LimitedChannel {
     private static final Logger log = LoggerFactory.getLogger(BalancedNodeSelectionStrategyChannel.class);
 
+    private static final int INFLIGHT_COMPARISON_THRESHOLD = 10;
```
iamdanfox: So the only thing I'm struggling to get an intuition for is why this threshold makes sense - I totally get that we can't enable this short-circuiting if we don't have at least one request in flight, but it seems odd to get different behaviour if we have a few very slow requests (exceeding the threshold) vs many very fast requests (under the threshold).

Maybe we're really trying to put a lower-bound on the giveUpThreshold and we should just use `giveUpThreshold = Math.min(newGiveUpThreshhold, 10)` or so?

carterkozak: It's coupled to the concurrency limiter default value. The goal is to avoid the issues that arise when we shift load off an unhealthy node to a healthy one, which may reject requests because its concurrency limiter hasn't expanded yet. We can probably reduce this threshold to avoid issues from nodes that were unhealthy at one point and have healed, but still have a low limit due to (previously) low volume.

Not sure I'd want to use the minimum of giveUpThreshold and this value because it (in part) works to keep the giveUpThreshold high.

iamdanfox (approved):
LGTM

carterkozak:
:+1:

PR #879, opened 2020-06-30
DialogueException is public, may be caught and handled by consumers

carterkozak:
==COMMIT_MSG==
DialogueException is public, may be caught and handled by consumers
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-879.v2.yml
+++ b/changelog/@unreleased/pr-879.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: DialogueException is public, may be caught and handled by consumers
+  links:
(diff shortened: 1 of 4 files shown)


ferozco (approved):

carterkozak:
:robot: :popcorn:

iamdanfox:
Wait did we really want to do this? Unwinding exceptions as public API has been _really_ painful historically, and I kinda liked that there was nothing that people could compile against.

Now we're no better than FeignException :(

carterkozak:
There were a number of cases where this resulted in bloated code, and didn't provide wc a reasonable way to attribute rpc failures to the rpc system.

PR #884, opened 2020-07-01
Remove dialogue-apache-hc4-client in favor of hc5

carterkozak:
We have migrated all known consumers to hc5, leaving this module
around represents technical debt, and an alluring trap for
unsuspecting service authors.

## After this PR
==COMMIT_MSG==
Remove dialogue-apache-hc4-client in favor of hc5. Known consumers have already been migrated.
==COMMIT_MSG==

## Possible downsides?
module is gone.


dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientChannels.java
```diff
@@ -90,9 +89,23 @@
...
+    @Deprecated
+    public static Channel create(ClientConfiguration conf) {
+        return create(conf, "apache-channel");
+    }
+
+    /**
+     * Simple cjr-style factory method. Prefer using {@code dialogue-clients} general purpose factory where possible.
+     * This method exists to ease of migration from hc4 to hc5.
+     */
+    public static Channel create(ClientConfiguration conf, String channelName) {
```
carterkozak, line 108: Added these two methods to facilitate moving existing code from hc4 to hc5 -- there are still a few lingering consumers and I don't want to cause problems for them.

carterkozak:
Opting not to call this a break because an entire module is being removed, any issues will present at build time, not runtime. Older versions of the hc4 client will continue to work if they are still used. The vast majority of consumers are not impacted because they use DialogueClients factories.

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientChannels.java
ferozco: Maybe we can use this as an opportunity to push existing consumers to use fancy client factory interfaces that we expose instead?

ferozco:
👍 LGTM, I would prefer to avoid exposing the legacy methods hc5 ApacheHttpClientChannels but its not a big deal either way

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientChannels.java
carterkozak: The blocker is the ServiceConfiguration API requiring a path to key material where we only have a SSLSocketFactory/trustManager in some cases. That's why we haven't been able to move the JaxrsClient.create code to the fancy factory yet :-(

carterkozak:
I'd considered adding an implementation-agnostic ClientConfiguration-based factory, but I really want to avoid adding more coupling to client-configuration. An ideal solution would use SSLContext instead of TrustManager+SSLSocketFactory, but that wouldn't work as a direct replacement for ClientConfiguration without manual refactoring... I think merging this as is allows us to move forward, and doesn't prevent us from making progress.

PR #891, opened 2020-07-08
Dialogue POST requests with no body set `Content-Length: 0`

carterkozak:
==COMMIT_MSG==
Dialogue POST requests with no body set `Content-Length: 0`

https://tools.ietf.org/html/rfc7230#section-3.3.2 recommends setting
a content-length on POST requests with no body. Some proxies strictly
validate this and respond 411 when the content-length is not specified.
==COMMIT_MSG==


ferozco:
👍

PR #895, opened 2020-07-09
Better concurrency limiter metrics for StickyEndpointChannels

iamdanfox:
## Before this PR

In PDS-125841, @diogoholanda flagged that we're seeing more throttling behaviour internally when using the "internal atlasdb-proxy" with dialogue compared to when it was using c-j-r. I'm a little confused as to _why_ dialogue was actually less good than before, and the graphs are hard to interpret because the sticky endpoint channels (and per host) all ascribe their concurrency limits to 'hostIndex: 0', so we only get one line when we should get n lines.

## After this PR
==COMMIT_MSG==
For users of `getStickyChannels`, ConcurrencyLimiter metrics are now correctly attributed to each hostIndex. Previously, they were all attributed to `hostIndex:0`.
==COMMIT_MSG==

## Possible downsides?
- a gross extra API method on the DialogueChannel builder

### DIFF
--- a/dialogue-clients/src/main/java/com/palantir/dialogue/clients/ChannelCache.java
+++ b/dialogue-clients/src/main/java/com/palantir/dialogue/clients/ChannelCache.java
@@ -31,6 +31,7 @@ import java.io.IOException;
 import java.util.Collections;
 import java.util.Map;
 import java.util.Optional;
+import java.util.OptionalInt;
 import java.util.Set;
 import java.util.WeakHashMap;
 import java.util.concurrent.ConcurrentHashMap;
@@ -89,7 +90,8 @@ final class ChannelCache {
     DialogueChannel getNonReloadingChannel(
             ReloadingClientFactory.ReloadingParams reloadingParams,
             ServiceConfiguration serviceConf,
-            @Safe String channelName) {
+            @Safe String channelName,
+            OptionalInt overrideHostIndex) {
         if (log.isWarnEnabled() && channelCache.estimatedSize() >= MAX_CACHED_CHANNELS * 0.75) {
             log.warn("channelCache nearing capacity - possible bug? {}", SafeArg.of("cache", this));
         }
@@ -99,6 +101,7 @@ final class ChannelCache {
                 .blockingExecutor(reloadingParams.blockingExecutor())
                 .serviceConf(serviceConf)
                 .channelName(channelName)
+                .overrideHostIndex(overrideHostIndex)
                 .build());
     }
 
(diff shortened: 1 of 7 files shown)


dialogue-clients/src/main/java/com/palantir/dialogue/clients/ReloadingClientFactory.java
```diff
@@ -142,18 +144,20 @@ public PerHostClientFactory perHost(String serviceName) {
...
-                                    .build())
-                            .map(singleUriServiceConf -> {
-                                // subtle gotcha here is that every single one of these has the same channelName,
-                                // which means metrics like the QueuedChannel counter will end up being the sum of all
-                                // of them.
-                                return cache.getNonReloadingChannel(params, singleUriServiceConf, channelName);
-                            })
-                            .collect(ImmutableList.toImmutableList());
+                    ImmutableList.Builder<DialogueChannel> list = ImmutableList.builder();
+                    for (int i = 0; i < serviceConfiguration.uris().size(); i++) {
```
carterkozak, line 148: :heart: 4 loops 4 life

dialogue-clients/src/main/java/com/palantir/dialogue/clients/ChannelCache.java
```diff
@@ -189,9 +193,11 @@ public String toString() {
     interface ChannelCacheKey extends AugmentClientConfig {
         ServiceConfiguration serviceConf();
 
+        Optional<ExecutorService> blockingExecutor();
+
         String channelName();
 
-        Optional<ExecutorService> blockingExecutor();
+        OptionalInt overrideHostIndex();
```
carterkozak, line 200: I'm a little worried about this fields (and the blockingExecutor for that matter) resulting in cache misses. Would bad things happen if we annotated `@Auxiliary`?

iamdanfox: Don't think we want to add `@Auxuliary`, because if we ask for something with the hostIndex set to 6 and get a cache hit with the hostIndex set to 0 then we'll get confusing metrics.

However I think the cache misses are not very serious here because this class has both the `channelCache` (full of DialogueChannels) and the `apacheCache` (full of `ApacheHttpClientChannels.CloseableClient`s). Even if we get misses on the channelCache, we'll still get cache hits on the apache clients, so I don't think this PR will result in problems with connection pools.

carterkozak (approved):

PR #897, opened 2020-07-09
ConcurrencyLimiter tries harder to stay under the (floating point) limit

iamdanfox:
## Before this PR

This is mainly motivated by my other PR (#896) where I was surprised to see so many 429s:

```
Completed 30 transactions, each making 400 requests.
successResponses = {node2=3200, node3=3200, node0=2800, node1=2800}
limitedResponses = {node2=684, node3=680, node0=587, node1=587}
```

## After this PR
==COMMIT_MSG==
ConcurrencyLimiter tries harder to stay under the (floating point) limit
==COMMIT_MSG==

With this change, it seems like the concurrency limiter is _slightly_ better at estimating the QPS limit on the server,  resulting in fewer 429s sent.

```
Completed 30 transactions, each making 400 requests.
successResponses = {node2=2800, node3=3200, node0=3200, node1=2800}
limitedResponses = {node2=563, node3=640, node0=622, node1=566}
```


dialogue-core/src/main/java/com/palantir/dialogue/core/CautiousIncreaseAggressiveDecreaseConcurrencyLimiter.java
```diff
@@ -64,10 +64,15 @@
...
+
+        // We don't want to hand out a permit if there are 4 inflight and a limit of 4.1, as this will immediately send
+        // our inflight number to 5, which is clearly above the limit.  Instead, we wait until there is capacity for
+        // one whole request before handing out a permit. In the worst-case scenario with zero inflight and a limit of
+        // 1, we'll still hand out a permit
+        if (currentInFlight <= getLimit() - 1) {
+            return Optional.of(createToken());
         }
-        return Optional.of(createToken());
+        return Optional.empty();
```
carterkozak, line 76: There's a race between our check, and the increment in createToken. Not sure if it's worth fixing, but could be done by deconstructing the increment into a cas loop, or optimistically incrementing and decrementing if the value failed our check (though this approach could impact metrics depending when they're collected, and should be protected by an initial pessimistic check to avoid unnecessary cas when everything is limited).

carterkozak: In that case we'd still race against changes to the limit, which isn't necessarily worth thinking about.

dialogue-core/src/main/java/com/palantir/dialogue/core/CautiousIncreaseAggressiveDecreaseConcurrencyLimiter.java
```diff
@@ -137,7 +142,7 @@ void success() {
         SUCCESS() {
             @Override
             public double applyAsDouble(double originalLimit, double inFlightSnapshot) {
-                if (inFlightSnapshot >= originalLimit * BACKOFF_RATIO) {
+                if (inFlightSnapshot >= Math.floor(originalLimit) * BACKOFF_RATIO) {
```
carterkozak: nice

carterkozak: Do we want to match the `DROPPED` case and wrap the result instead?
```suggestion
                if (inFlightSnapshot >= Math.floor(originalLimit * BACKOFF_RATIO)) {
```

carterkozak: shouldn't make much difference either way

carterkozak (approved):

dialogue-core/src/main/java/com/palantir/dialogue/core/CautiousIncreaseAggressiveDecreaseConcurrencyLimiter.java
iamdanfox: I've updated it just for consistency - I think the difference this makes is we'd increase a little bit more eagerly when the limit is low, but overall yeah doesn't seem that big.

dialogue-core/src/main/java/com/palantir/dialogue/core/CautiousIncreaseAggressiveDecreaseConcurrencyLimiter.java
iamdanfox: Yeah to be honest I'm ok with all of this being quite approximate - the concurrency limiter is always just trying to infer a sensible rate limit and even after we've handed out a permit we go through a few more layers before a network request actually gets out the door!

PR #898, opened 2020-07-09
The system default proxy selector is never used

carterkozak:
==COMMIT_MSG==
The system default proxy selector is never used

Previously jvm configuration could result in dialogue traffic
using an unexpected proxy. Now when the system-default proxy
selector is provided it is ignored to avoid failures due to
unexpected environment configuration.
==COMMIT_MSG==

This will probably break someone, I imagine something depends on this in an odd way...


dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientChannels.java
```diff
@@ -425,14 +426,13 @@ public CloseableClient build() {
 
             setupConnectionPoolMetrics(conf.taggedMetricRegistry(), name, connectionManager);
 
+            ProxySelector configuredSelector = conf.proxy();
+            boolean noProxy = ProxySelector.getDefault().equals(configuredSelector);
```
carterkozak, line 430: Ideally ClientConfigurations would provide a no-proxy selector instance by default, however that would break our workaround for the incorrect timeout being used with forward proxies if cjr was upgraded prior to dialogue. It can be safely done after the next hc5 release.

carterkozak:
going to reset the changelog to 'fix' for now to avoid creating major rev rc tags that may be confusing later. If this is merged I'll reset to break for ~~deniability~~ semver.

iamdanfox:
Mmm interesting, so before this PR if someone had ` -Dhttps.proxyHost=host -Dhttps.proxyPort=port` system properties set (or had used System.setProperty) then the apache client would magically respect that, which is inconsistent with all our other config?

And now we'd just ignore it?

carterkozak:
Well, it's not necessarily inconsistent because our ClientConfigurations factory uses `ProxySelector.getDefault()` when we don't define a proxy.

carterkozak:
Certainly unexpected behavior though.

carterkozak:
no beans.

PR #901, opened 2020-07-10
Fix unexpected socket timeouts when configured as unlimited

carterkozak:
https://issues.apache.org/jira/browse/HTTPCLIENT-2099

## After this PR
==COMMIT_MSG==
Fix unexpected socket timeouts when configured as unlimited by using a 1-day socket timeout
==COMMIT_MSG==

## Possible downsides?
Requests that expect to sit idle waiting for a response longer than a day won't work. Our metrics don't show any requests lasting this long, so we should be safe.

### DIFF
--- a/dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientChannels.java
+++ b/dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientChannels.java
@@ -39,7 +39,6 @@ import com.palantir.tritium.metrics.registry.TaggedMetricRegistry;
 import java.io.Closeable;
 import java.io.IOException;
 import java.net.MalformedURLException;
-import java.net.ProxySelector;
 import java.net.URL;
 import java.time.Duration;
 import java.util.ArrayList;
@@ -410,7 +409,10 @@ public final class ApacheHttpClientChannels {
                     .setConnectionTimeToLive(idleConnectionTimeout)
                     .setDefaultSocketConfig(SocketConfig.custom()
                             .setSoKeepAlive(true)
-                            .setSoTimeout(socketTimeout)
+                            // The default socket configuration socket timeout only applies prior to request execution.
+                            // By using the connect timeout here, we apply it to the handshake in addition to the
+                            // socket.connect call.
+                            .setSoTimeout(connectTimeout)
                             .build())
                     .setMaxConnPerRoute(Integer.MAX_VALUE)
                     .setMaxConnTotal(Integer.MAX_VALUE)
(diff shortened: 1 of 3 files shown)


dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientChannels.java
```diff
@@ -473,12 +465,29 @@ public CloseableClient build() {
...
+                    "Read and write timeouts do not match, The value of the readTimeout {} will be used and write "
+                            + "timeout {} will be ignored.",
+                    SafeArg.of("readTimeout", conf.readTimeout()),
+                    SafeArg.of("writeTimeout", conf.writeTimeout()));
+        }
+        if (socketTimeoutMillis == 0) {
+            // https://issues.apache.org/jira/browse/HTTPCLIENT-2099
+            log.debug("Working around HTTPCLIENT-2099 by using a 1 day socket timeout instead of zero (unlimited)");
+            socketTimeoutMillis = Duration.ofDays(1).toMillis();
+        }
```
carterkozak, line 487: these 5 lines are the only real change, the rest is shuffling away from numbers to proper Timeout values

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientChannels.java
```diff
@@ -425,21 +420,18 @@ public CloseableClient build() {
...
+            boolean proxyConfigured = !ProxySelector.getDefault().equals(conf.proxy());
             HttpClientBuilder builder = HttpClients.custom()
                     .setDefaultRequestConfig(RequestConfig.custom()
-                            // HTTPCLIENT-1478: Work around the connection timeout being used to negotiate proxy
-                            // traffic.
-                            .setConnectTimeout(
-                                    ProxySelector.getDefault().equals(conf.proxy())
-                                            ? connectTimeout
-                                            : Timeout.ofMilliseconds(socketTimeoutMillis))
+                            // HTTPCLIENT-2091: Work around the connection timeout being applied to proxy traffic.
```
carterkozak: Fixed the ticket number here, I pasted the wrong link originally...

carterkozak: and removed entirely, no need to work around this at all when we set a RequestConfig.responseTimeout.

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientChannels.java
carterkozak: no longer the case, this became a large change which will have a positive impact in several scenarios, particularly for the timelock case.

iamdanfox (approved):
Woah there is a lot of subtlety to apache's timeouts!

PR #908, opened 2020-07-16
Fix massive synchronization and gc cost of jaxrs clients

carterkozak:
Do not attempt to magically bind reloadable channels for a
one-shot invocation.

==COMMIT_MSG==
Fix massive synchronization and gc cost of jaxrs clients
==COMMIT_MSG==

## Before

```
Benchmark                         Mode  Cnt       Score       Error  Units
Benchmarks.jaxrsDialogueClient    thrpt    4   19571.205 ±  2295.951  ops/s
Benchmarks.pureDialogueClient     thrpt    4  103466.692 ± 16021.078  ops/s
```

## After

```
Benchmark                         Mode  Cnt       Score       Error  Units
Benchmarks.jaxrsDialogueClient    thrpt    4   76504.935 ± 14078.987  ops/s
Benchmarks.pureDialogueClient     thrpt    4  104787.366 ± 25389.212  ops/s
```

### DIFF
--- a/changelog/@unreleased/pr-908.v2.yml
+++ b/changelog/@unreleased/pr-908.v2.yml
@@ -1,0 +1,5 @@
+type: fix
+fix:
+  description: Fix massive synchronization and gc cost of jaxrs clients
+  links:
+  - https://github.com/palantir/dialogue/pull/908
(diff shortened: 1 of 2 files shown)


robert3005 (approved):

jeremyk-91 (approved):
This would fix what I saw in the profiles.

iamdanfox (commented):
Oof, I've already found the EndpointChannel thing to be a bit of a foot gun... shall we push EndpointChannelFactory into the conjure-java codegen so that it becomes a first-class way to create things? Then hopefully we can do less of the instanceof magic

carterkozak:
Strong +1 on passing an EndpointChannelFactory through :-)

PR #912, opened 2020-07-17
Fix UnsupportedOperationException with old dialouge bindings

carterkozak:
==COMMIT_MSG==
Fix UnsupportedOperationException with old dialouge bindings
==COMMIT_MSG==

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/StickyEndpointChannels.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/StickyEndpointChannels.java
@@ -96,9 +96,8 @@ public final class StickyEndpointChannels implements Supplier<Channel> {
          */
         @Deprecated
         @Override
-        public ListenableFuture<Response> execute(Endpoint _endpoint, Request _request) {
-            // TODO(dfox): remove this by adding client factory method taking a 'EndpointChannelFactory' not a 'Channel'
-            throw new UnsupportedOperationException("Not implemented");
+        public ListenableFuture<Response> execute(Endpoint endpoint, Request request) {
+            return endpoint(endpoint).execute(request);
         }
 
         @Override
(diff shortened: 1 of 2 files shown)


dialogue-core/src/main/java/com/palantir/dialogue/core/StickyEndpointChannels.java
```diff
@@ -96,9 +96,8 @@ public EndpointChannel endpoint(Endpoint endpoint) {
          */
         @Deprecated
         @Override
-        public ListenableFuture<Response> execute(Endpoint _endpoint, Request _request) {
-            // TODO(dfox): remove this by adding client factory method taking a 'EndpointChannelFactory' not a 'Channel'
-            throw new UnsupportedOperationException("Not implemented");
+        public ListenableFuture<Response> execute(Endpoint endpoint, Request request) {
+            return endpoint(endpoint).execute(request);
```
j-baker, line 100: Is there a test that could/should have caught this?

carterkozak: The original implementation was built around external assumptions that turned out to be incorrect, the goal was not to support old generated paths with the new sticky channel feature. We could write a test that fails without this change and passes with it, but I'm not sure that would provide value.

iamdanfox: Yep - I didn't want people to go through this codepath originally and I still don't (thought this was loud enough to prevent that happening), but apparently some upgrades can race and cause users to hit this anyway.

Long term solution is to just work towards deleting this method entirely (https://github.com/palantir/conjure-java/pull/1015)

ferozco (approved):

carterkozak:
:robot:

PR #913, opened 2020-07-18
Clients respect the server Keep-Alive timeout when present

carterkozak:
Pooled connections are no longer limited to a reusable lifespan of 50 seconds.

The client no longer uses a fixed timeout, allowing the server
to describe itself using a `Keep-Alive: timeout=<seconds>` header.

==COMMIT_MSG==
Clients respect the server Keep-Alive timeout when present
==COMMIT_MSG==

These things are difficult to test outside of end to end verification
because connection metadata is not exposed through our APIs.

### DIFF
--- a/dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientChannels.java
+++ b/dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientChannels.java
@@ -403,7 +403,6 @@ public final class ApacheHttpClientChannels {
                     .setPoolConcurrencyPolicy(PoolConcurrencyPolicy.LAX)
                     // Allow unnecessary connections to time out reducing system load.
                     .setConnPoolPolicy(PoolReusePolicy.LIFO)
-                    .setConnectionTimeToLive(IDLE_CONNECTION_TIMEOUT)
                     .setDefaultSocketConfig(SocketConfig.custom()
                             .setSoKeepAlive(true)
                             // The default socket configuration socket timeout only applies prior to request execution.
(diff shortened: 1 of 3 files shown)


dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ScheduledIdleConnectionEvictor.java
```diff
@@ -54,21 +53,16 @@
...
-            ConnPoolControl<?> connectionManager,
-            Duration delayBetweenChecks,
-            Duration maxIdleTime,
-            ScheduledExecutorService scheduler) {
+            ConnPoolControl<?> connectionManager, Duration delayBetweenChecks, ScheduledExecutorService scheduler) {
         return scheduler.scheduleWithFixedDelay(
                 () -> {
                     try {
                         connectionManager.closeExpired();
-                        connectionManager.closeIdle(TimeValue.ofMilliseconds(maxIdleTime.toMillis()));
```
carterkozak, line 71: This acted as an upper bound, our connections already use the value provided by the server, or our default (50 seconds) if the server doesn't send `Keep-Alive: timeout=`.

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientChannels.java
```diff
@@ -403,7 +403,6 @@ public CloseableClient build() {
                     .setPoolConcurrencyPolicy(PoolConcurrencyPolicy.LAX)
                     // Allow unnecessary connections to time out reducing system load.
                     .setConnPoolPolicy(PoolReusePolicy.LIFO)
-                    .setConnectionTimeToLive(IDLE_CONNECTION_TIMEOUT)
```
carterkozak, line 406: This one is the :hot_pepper:, it's a maximum total timeout for each connection. After a connections lifespan has exceeded 50 seconds it will be closed and never reused.

iamdanfox (approved):

carterkozak:
:robot:

PR #915, opened 2020-07-19
The validateAfterInactivity duration may be modified via a sysprop

carterkozak:
==COMMIT_MSG==
The validateAfterInactivity duration may be modified via a sysprop.
The `dialogue.experimental.inactivity.check.threshold.millis` property
can be used to modify the inactivity timeout for connection health
checks. This is not meant to be a permanent fixture, but to allow
us to assess the tradeoff between our current four second value and
a higher value for less frequent hiccups.
==COMMIT_MSG==

### DIFF
--- a/dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientChannels.java
+++ b/dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientChannels.java
@@ -329,7 +329,8 @@ public final class ApacheHttpClientChannels {
         // dropwizard: 30 seconds (see idleTimeout in the linked docs)
         // https://www.dropwizard.io/en/latest/manual/configuration.html#Connectors
         // wc: 60 seconds (internal)
-        private static final TimeValue CONNECTION_INACTIVITY_CHECK = TimeValue.ofSeconds(4);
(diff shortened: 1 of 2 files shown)


dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientChannels.java
```diff
@@ -329,7 +329,8 @@ public static ClientBuilder clientBuilder() {
         // dropwizard: 30 seconds (see idleTimeout in the linked docs)
         // https://www.dropwizard.io/en/latest/manual/configuration.html#Connectors
         // wc: 60 seconds (internal)
-        private static final TimeValue CONNECTION_INACTIVITY_CHECK = TimeValue.ofSeconds(4);
+        private static final TimeValue CONNECTION_INACTIVITY_CHECK = TimeValue.ofMilliseconds(
```
iamdanfox, line 332: @carterkozak are you intending to vary this at particular services, or just in in one-off test harnesses?

carterkozak: Goal is to iterate with timelock reverting this to the original 20 second value in order to determine if/how much we hurt performance with #887. Hopefully there's no meaningful change and we remove the property.

iamdanfox (approved):

carterkozak:
:robot:

PR #917, opened 2020-07-21
Disable the response-out-of-order check

carterkozak:
This removes protection added to the apache client 5, which
did not exist in okhttp or httpclient 4.x.
https://issues.apache.org/jira/browse/HTTPCLIENT-2093

Our services are built to withstand this sort of behavior, and
CJR wasn't negatively impacted.

I've created a ticket upstream to propose a way to configure this without subclassing:
https://issues.apache.org/jira/browse/HTTPCORE-639

==COMMIT_MSG==
Disable the response-out-of-order check introduced by httpclient 5
==COMMIT_MSG==

### DIFF
--- a/dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/DialogueConnectionFactory.java
+++ b/dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/DialogueConnectionFactory.java
@@ -1,0 +1,55 @@
+/*
+ * (c) Copyright 2020 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.dialogue.hc5;
+
+import java.io.IOException;
+import java.net.Socket;
+import java.util.concurrent.atomic.AtomicLong;
+import org.apache.hc.client5.http.impl.io.DefaultHttpResponseParserFactory;
+import org.apache.hc.client5.http.io.ManagedHttpClientConnection;
+import org.apache.hc.core5.http.config.Http1Config;
+import org.apache.hc.core5.http.impl.DefaultContentLengthStrategy;
+import org.apache.hc.core5.http.impl.io.DefaultHttpRequestWriterFactory;
+import org.apache.hc.core5.http.io.HttpConnectionFactory;
+
+/**
+ * This {@link HttpConnectionFactory} is equivalent to the default factory, but provides
+ * {@link DialogueManagedHttpClientConnection} connections in order to provide the functionality
+ * proposed in <a href="https://issues.apache.org/jira/browse/HTTPCORE-639">HTTPCORE-639</a>.
+ */
+enum DialogueConnectionFactory implements HttpConnectionFactory<ManagedHttpClientConnection> {
+    INSTANCE;
+
+    private static final AtomicLong COUNTER = new AtomicLong();
+
+    @Override
+    public ManagedHttpClientConnection createConnection(Socket socket) throws IOException {
+        ManagedHttpClientConnection conn = new DialogueManagedHttpClientConnection(
+                "http-outgoing-" + COUNTER.getAndIncrement(),
+                null,
+                null,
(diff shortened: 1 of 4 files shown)


dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/TracedManagedHttpClientConnection.java
```diff
@@ -0,0 +1,181 @@
...
+        try (CloseableTracer ignored = CloseableTracer.startSpan("Dialogue: Connection.close")) {
+            delegate.close(closeMode);
+        }
+    }
+
+    @Override
+    public String toString() {
+        return "TracedManagedHttpClientConnection{" + delegate + '}';
+    }
+}
```
carterkozak: pulling this into a separate PR

carterkozak:
This should be good to go.

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/DialogueConnectionFactory.java
```diff
@@ -0,0 +1,55 @@
...
+    INSTANCE;
+
+    private static final AtomicLong COUNTER = new AtomicLong();
+
+    @Override
+    public ManagedHttpClientConnection createConnection(Socket socket) throws IOException {
+        ManagedHttpClientConnection conn = new DialogueManagedHttpClientConnection(
+                "http-outgoing-" + COUNTER.getAndIncrement(),
+                null,
+                null,
```
iamdanfox, line 44: Reading through ManagedHttpClientConnectionFactory there's a lot of configurability in their version, but since we just used ManagedHttpClientConnectionFactory.INSTANCE we weren't using any of it. Happy to use defaults and keep it simple!

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/DialogueManagedHttpClientConnection.java
```diff
@@ -0,0 +1,174 @@
...
+/**
+ * Connection which produces behavior matching the default managed connection from httpcore while
+ * providing the functionality proposed in
+ * <a href="https://issues.apache.org/jira/browse/HTTPCORE-639">HTTPCORE-639</a>.
+ */
+final class DialogueManagedHttpClientConnection extends DefaultBHttpClientConnection
+        implements ManagedHttpClientConnection, Identifiable {
+
+    private final String id;
+    private final AtomicBoolean closed = new AtomicBoolean();
```
iamdanfox, line 55: ~why is it important to have this extra close AtomicBoolean? I thought the intention was to just override the isDataAvailable thing?~

EDIT nvm I see we're just copying stuff from DefaultManagedHttpClientConnection as this is a final class, so we have to extend _DefaultBHttpClientConnection_ and re-add things.

carterkozak: :+1: exactly right.

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/DialogueManagedHttpClientConnection.java
```diff
@@ -0,0 +1,174 @@
...
+        this.id = id;
+    }
+
+    @Override
+    public String getId() {
+        return id;
+    }
+
+    @Override
+    public void bind(SocketHolder socketHolder) throws IOException {
```
iamdanfox, line 89: I think I'd find it reassuring to have little comments or something on these methods:

```
/** Identical to {@link DefaultManagedHttpClientConnection#bind}. */
```

So that readers/maintainers can clearly identify which are the bits we've added

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/DialogueManagedHttpClientConnection.java
```diff
@@ -0,0 +1,174 @@
...
+        if (!closed.getAndSet(true)) {
+            super.close(closeMode);
+        }
+    }
+
+    @Override
+    protected void onResponseReceived(ClassicHttpResponse _response) {}
+
+    @Override
+    protected void onRequestSubmitted(ClassicHttpRequest _request) {}
```
carterkozak: These two stubs can be removed the methods they extend are also no-ops.

iamdanfox:
Not gonna lie this one makes me a _teeny_ bit nervous... using inheritance to tweak behaviour is always gonna be a bit fragile (especially when we copy and paste stuff from `DefaultManagedHttpClientConnection`). There's the danger that extra functionality/fixes are added to that class and we forget to copy them into our version.

carterkozak:
That's why we're fixing it upstream ;-)

iamdanfox (approved):
This matches the old world of hc4 and okhttp, so I think it's OK to keep this around until 5.1 is released when we can hopefully switch to using a first-class API

PR #918, opened 2020-07-21
PDS-127702: Dialogue Apahce Httpclient connections produce tracing spans

carterkozak:
==COMMIT_MSG==
Dialogue Apache Httpclient connections produce tracing spans
==COMMIT_MSG==
2020-07-21_1431x332_screen


iamdanfox (approved):
<3 more spans.

Yes it'll be noisy if you're fully zoomed in, but our trace-log-viewer has an option to hide spans that are very short, so most people won't need to worry about these.

carterkozak:
:robot:

ferozco:
For posterity here is the before:

image

after:
image

markelliot:
@ferozco what's happening between `Dialogue ConnectionManager.lease` and `Dialogue ConnectionManager.connect` in that sample?

carterkozak:
Lease is a bit odd, it returns a future that does work when `get()` is called, so the lease span creates the future and may look up existing connections.
Connect is creating and initializing a new connection, via the future returned from lease iirc. We could probably improve the tracing in that area.

In most cases we won't see this many spans because it requires an intersection between new connection creation as opposed to using an existing pooled connection, and rolling a sampled trace.

markelliot:
the other thing here is that in the before picture there's a pretty big gap between dialogue-blocking-channel's start and `Dialogue ExecChain.execute`, which is similar to a what we observed problem we've been trying to chase down. Since whatever you're doing to generate the spans seems to exhibit this frequently enough for it to show up in your sampling, wondering if there's potential to instrument enough to figure out what happens in those ~25ms shown in the before picture, or the ~4ms shown in the after picture.

carterkozak:
This code is all interpreted in these spans, time spent between `dialogue-blocking-channel` and `Dialogue ExecChain.execute` corresponds with this code: https://github.com/palantir/dialogue/blob/d444559a1618146897ca8ac46aac53a923d1dee3/dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientBlockingChannel.java#L72-L91 where we do a bit of object allocation to convert from dialogue types into apache httpclient types. That section becomes faster as the VM warms up.

markelliot:
I think there's something kind of fishy here since in the spans we're tracking down that are slow it's in the same spot and in prod is taking closer to 100ms.

PR #920, opened 2020-07-23
Connect timeouts are retried

carterkozak:
==COMMIT_MSG==
Connect timeouts are retried
==COMMIT_MSG==

ConnectTimeoutExcetion implements SocketTimeoutException, which is not retried by default. It should be wrapped with a retryable exception..
https://github.com/palantir/dialogue/blob/6bea1de7df48d2c930a53551d8c4eeadfc256baa/dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java#L325-L338

### DIFF
--- a/changelog/@unreleased/pr-920.v2.yml
+++ b/changelog/@unreleased/pr-920.v2.yml
@@ -1,0 +1,5 @@
+type: fix
+fix:
+  description: Connect timeouts are retried
+  links:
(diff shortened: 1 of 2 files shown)


ferozco (approved):

ferozco (approved):

carterkozak:
Implementing test coverage separately, unblocking folks with broken services first

PR #924, opened 2020-07-24
Fix response leaks on Future.cancel

carterkozak:
The guava methods `Futures.transform`, `Futures.transformAsync`,
`Futures.catchAsync`, and similar allow completion to race with
`Future.cancel`. This is especially dangerous when callers rely
on `SettableFuture.set` methods to transfer resource ownership.
Using the guava methods, it's possible successfully hand off
a result while the future ignores the value due to a cancel.

## Before this PR
leak.

## After this PR
==COMMIT_MSG==
Fix a race condition that can result in response leaks when dialogue requests are canceled
==COMMIT_MSG==

## Possible downsides?
Concurrency is hard, this is another utility we must maintain. Compared to the cost of a resource leak this is easily worthwhile.

This introduces a new module so we can share `dialogue-futures` between modules. dialogue-futures is exposed only as an implementation dependency, but there's risk that it will be used directly.
There's an open question around how we expect others to interact with `ListenableFuture<InputStream>`.


dialogue-futures/src/main/java/com/palantir/dialogue/futures/DialogueDirectTransformationFuture.java
```diff
@@ -0,0 +1,102 @@
...
+            output.set(function.apply(result));
+        } catch (Throwable t) {
+            output.setException(t);
+        }
+    }
+
+    @Override
+    public void onFailure(Throwable throwable) {
+        if (input.isCancelled()) {
+            output.cancel(false);
```
iamdanfox, line 117: Something kinda interesting here is that by calling `.cancel(false)` we lose the stacktrace information of who originally called the .cancel method on the original future, resulting in this very minimal stacktrace:

```
java.util.concurrent.CancellationException: Task was cancelled.

	at com.google.common.util.concurrent.AbstractFuture.cancellationExceptionWithCause(AbstractFuture.java:1386)
	at com.google.common.util.concurrent.AbstractFuture.getDoneValue(AbstractFuture.java:562)
	at com.google.common.util.concurrent.AbstractFuture.get(AbstractFuture.java:525)
	at com.google.common.util.concurrent.AbstractFuture$TrustedFuture.get(AbstractFuture.java:102)
	at com.palantir.dialogue.futures.DialogueDirectTransformationFuture.get(DialogueDirectTransformationFuture.java:75)
	at com.palantir.dialogue.futures.DialogueFuturesTest.testTransform_resultCancel(DialogueFuturesTest.java:54)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
```

Compared to:
```


java.util.concurrent.ExecutionException: java.util.concurrent.CancellationException: Task was cancelled.

	at com.google.common.util.concurrent.AbstractFuture.getDoneValue(AbstractFuture.java:564)
	at com.google.common.util.concurrent.AbstractFuture.get(AbstractFuture.java:525)
	at com.google.common.util.concurrent.AbstractFuture$TrustedFuture.get(AbstractFuture.java:102)
	at com.palantir.dialogue.futures.DialogueDirectTransformationFuture.get(DialogueDirectTransformationFuture.java:75)
	at com.palantir.dialogue.futures.DialogueFuturesTest.testTransform_resultCancel(DialogueFuturesTest.java:54)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
  ...
Caused by: java.util.concurrent.CancellationException: Task was cancelled.
	at com.google.common.util.concurrent.AbstractFuture.cancellationExceptionWithCause(AbstractFuture.java:1386)
	at com.google.common.util.concurrent.AbstractFuture.getDoneValue(AbstractFuture.java:562)
	at com.google.common.util.concurrent.AbstractFuture.get(AbstractFuture.java:525)
	at com.google.common.util.concurrent.AbstractFuture$TrustedFuture.get(AbstractFuture.java:102)
	at com.google.common.util.concurrent.Uninterruptibles.getUninterruptibly(Uninterruptibles.java:237)
	at com.google.common.util.concurrent.Futures.getDone(Futures.java:1132)
	at com.google.common.util.concurrent.Futures$CallbackListener.run(Futures.java:1081)
	at com.google.common.util.concurrent.DirectExecutor.execute(DirectExecutor.java:30)
	at com.google.common.util.concurrent.AbstractFuture.executeListener(AbstractFuture.java:1174)
	at com.google.common.util.concurrent.AbstractFuture.complete(AbstractFuture.java:969)
	at com.google.common.util.concurrent.AbstractFuture.cancel(AbstractFuture.java:623)
	at com.google.common.util.concurrent.AbstractFuture$TrustedFuture.cancel(AbstractFuture.java:130)
⭐️⭐️⭐️at com.palantir.dialogue.futures.DialogueDirectTransformationFuture.cancel(DialogueDirectTransformationFuture.java:59)
	at com.palantir.dialogue.futures.DialogueFuturesTest.testTransform_resultCancel(DialogueFuturesTest.java:52)
	... 63 more

```

dialogue-futures/src/main/java/com/palantir/dialogue/futures/DialogueDirectTransformationFuture.java
```diff
@@ -0,0 +1,102 @@
...
+    public O get(long timeout, TimeUnit unit) throws InterruptedException, ExecutionException, TimeoutException {
+        return output.get(timeout, unit);
+    }
+
+    // Future callback methods update the output future based on the input.
+
+    @Override
+    public void onSuccess(I result) {
+        try {
+            output.set(function.apply(result));
```
iamdanfox:
```suggestion
            O newValue = function.apply(result);
            output.set(newValue);
```

mild preference to split this onto two lines so that it's trivial to set a breakpoint if you ever need to hunt through this in a debugger

dialogue-futures/src/main/java/com/palantir/dialogue/futures/DialogueFutures.java
```diff
@@ -0,0 +1,77 @@
...
+import com.google.common.util.concurrent.AsyncFunction;
+import com.google.common.util.concurrent.FutureCallback;
+import com.google.common.util.concurrent.Futures;
+import com.google.common.util.concurrent.ListenableFuture;
+import com.google.common.util.concurrent.MoreExecutors;
+import com.google.errorprone.annotations.CanIgnoreReturnValue;
+import java.util.function.Consumer;
+import java.util.function.Function;
+
+/** Internal utility functionality used by Dialogue modules. */
```
iamdanfox: Might be worth putting a little sentence in here to say we rolled our own because we found leaks in the guava ones when stuff is interrupted? I know people can look at git blame when they have this repo open, but if they're just browsing dialogue-futures-sources.jar it might help reassure people?

dialogue-futures/src/main/java/com/palantir/dialogue/futures/DialogueDirectAsyncTransformationFuture.java
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    @Override
+    public void addListener(Runnable listener, Executor executor) {
+        output.addListener(listener, executor);
+    }
+
+    @Override
+    public boolean cancel(boolean mayInterruptIfRunning) {
+        ListenableFuture<?> snapshot = currentFuture;
```
iamdanfox, line 63: I assume there's some JVM reason why doing a read of the volatile field into a method-local variable is slightly more efficient??

carterkozak:
The race occurs any time `transformed.cancel` is invoked while `transform.apply` is running. It's unclear whether a change would be accepted upstream due to expectations around future.cancel having immediate results for other waiters.

carterkozak:
I've filed a ticket upstream, but I'm not confident there will be a change beyond perhaps documentation: https://github.com/google/guava/issues/3975

dialogue-futures/src/test/java/com/palantir/dialogue/futures/DialogueFuturesTest.java
```diff
@@ -0,0 +1,216 @@
...
+    }
+
+    @Test
+    void testTransform_resultCancel() {
+        SettableFuture<String> original = SettableFuture.create();
+        ListenableFuture<String> doubled = DialogueFutures.transform(original, value -> value + value);
+        assertThat(doubled.cancel(false)).isTrue();
+        assertThat(doubled).isCancelled();
+        assertThatThrownBy(doubled::get).isInstanceOf(CancellationException.class);
+    }
```
iamdanfox, line 63: the implementation makes it pretty obvious, but I think it'd be nice to also assert `original.isCancelled()` just for completeness?

dialogue-futures/src/test/java/com/palantir/dialogue/futures/DialogueFuturesTest.java
```diff
@@ -0,0 +1,216 @@
...
+        ListenableFuture<String> doubled = DialogueFutures.transform(original, value -> value + value);
+        original.set("a");
+        assertThat(doubled.get()).isEqualTo("aa");
+    }
+
+    @Test
+    void testTransform_originalCancel() {
+        SettableFuture<String> original = SettableFuture.create();
+        ListenableFuture<String> doubled = DialogueFutures.transform(original, value -> value + value);
+        assertThat(doubled.cancel(false)).isTrue();
```
iamdanfox: think this should be `original.cancel(false)` right? (otherwise it's identical to the one below)

dialogue-futures/src/test/java/com/palantir/dialogue/futures/DialogueFuturesTest.java
iamdanfox, line 63: might be also nice to asset that `oriignal.isCancelled`, just to prove to people it definitely marks both as done?

dialogue-futures/src/test/java/com/palantir/dialogue/futures/DialogueFuturesTest.java
```diff
@@ -0,0 +1,216 @@
...
+    void testTransformAsync_originalCancel() {
+        SettableFuture<String> one = SettableFuture.create();
+        SettableFuture<String> two = SettableFuture.create();
+        ListenableFuture<String> transformed = DialogueFutures.transformAsync(one, _value -> two);
+        assertThat(transformed).isNotDone();
+        assertThat(one.cancel(false)).isTrue();
+        assertThat(transformed).isCancelled();
+        assertThat(transformed).isDone();
+        assertThatThrownBy(transformed::get).isInstanceOf(CancellationException.class);
+        // The intermediate future is not impacted
```
iamdanfox: i'd pop helpful little explanations like this into a `.describedAs()` on the assertion below

dialogue-futures/src/test/java/com/palantir/dialogue/futures/DialogueFuturesTest.java
iamdanfox: Actually could this behaviour be kinda spicy if someone decided to make a future and then do two _independent_ transformations based on it? Cancelling one of them would have the effect of cancelling the root one which ends up cancelling both?

dialogue-futures/src/test/java/com/palantir/dialogue/futures/DialogueFuturesTest.java
```diff
@@ -0,0 +1,216 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+import com.google.common.util.concurrent.Futures;
+import com.google.common.util.concurrent.ListenableFuture;
+import com.google.common.util.concurrent.SettableFuture;
+import java.util.concurrent.CancellationException;
+import org.junit.jupiter.api.Test;
+
+class DialogueFuturesTest {
```
iamdanfox, line 33: I know it's a bit of a faff, but what do you think about making this test class parameterized over {DIALOGUE, GUAVA} so that we could run the same assertions against both and reassure people we're not diverging (seems like testCatchingAllAsync_originalCancel is the only one that currently fails with guava)?

iamdanfox (approved):
Awesome find Carter - will be very nice to see those unbounded connection growth graphs drop down to sensible numbers :)

dialogue-futures/src/main/java/com/palantir/dialogue/futures/DialogueDirectAsyncTransformationFuture.java
carterkozak: I structured this method to prevent future foot-guns, we wouldn't want this method to be updated to accidentally load `currentFuture` twice. This could just as easily use one-liner `currentFuture.cancel(mayInterruptIfRunning);`

dialogue-futures/src/main/java/com/palantir/dialogue/futures/DialogueDirectTransformationFuture.java
carterkozak: That's correct. Cancellation in this direction is less common than the other direction.

dialogue-futures/src/test/java/com/palantir/dialogue/futures/DialogueFuturesTest.java
carterkozak: That's the desired behavior, it matches the guava method as well. When that is undesirable we use `Futures.nonCancellationPropagating`

dialogue-futures/src/test/java/com/palantir/dialogue/futures/DialogueFuturesTest.java
carterkozak: Added the parameterized tests. testCatchingAllAsync isn't quite a 1-1 comparison because using guava with Throwable directly will catch the CancellationException, so I've added a comment and not parameterized it.

PR #925, opened 2020-07-24
QueuedChannel logs cancellation exceptions at debug

carterkozak:
No need to produce noise in the logs in this case -- we don't log
a stack trace in the onSuccess path when the future has been
canceled.

==COMMIT_MSG==
QueuedChannel logs cancellation exceptions at debug to reduce noise.
==COMMIT_MSG==


iamdanfox (approved):

iamdanfox:
🐙

PR #928, opened 2020-07-27
Empty HttpEntity which doesn't attempt a write or flush

carterkozak:
There's no need to attempt an empty write or additional flush on
empty POST and PUT invocations.

==COMMIT_MSG==
Simpler empty HttpEntity implementation which doesn't attempt a write an empty array or flush the socket stream unnecessarily.
==COMMIT_MSG==


ferozco:
👍

PR #930, opened 2020-07-28
Handshake time is no longer bounded by the connect timeout

carterkozak:
Large garbage collections cause handshakes to fail. Ideally we
would not take more than a few hundred milliseconds to handshake,
but that's clearly not the current state of the world.

Currently our retry metrics show `safeioexception` which isn't
terribly informative. This change will replace those with the
self-describing `safeconnecttimeoutexception`.

==COMMIT_MSG==
Handshake time is no longer bounded by the connect timeout
==COMMIT_MSG==


dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientChannels.java
```diff
@@ -406,10 +406,7 @@ public CloseableClient build() {
                     .setConnPoolPolicy(PoolReusePolicy.LIFO)
                     .setDefaultSocketConfig(SocketConfig.custom()
                             .setSoKeepAlive(true)
-                            // The default socket configuration socket timeout only applies prior to request execution.
-                            // By using the connect timeout here, we apply it to the handshake in addition to the
-                            // socket.connect call.
-                            .setSoTimeout(connectTimeout)
+                            .setSoTimeout(socketTimeout)
```
ferozco, line 412: slap a bit of a comment explaining that we explicitly do not want to consider handashake time towards the timeout

dialogue-test-common/src/main/java/com/palantir/dialogue/AbstractHandshakeTimeoutTest.java
```diff
@@ -0,0 +1,203 @@
...
+
+    private static int getPort(Undertow undertow) {
+        return ((InetSocketAddress)
+                        Iterables.getOnlyElement(undertow.getListenerInfo()).getAddress())
+                .getPort();
+    }
+
+    @Test
+    public void testHandshakeTimeout() throws Exception {
+        SSLContext sslContext = SslSocketFactories.createSslContext(TestConfigurations.SSL_CONFIG);
```
ferozco: can we factor out boilerplate into a separate method or into `BeforeEach`?

dialogue-test-common/src/main/java/com/palantir/dialogue/AbstractHandshakeTimeoutTest.java
```diff
@@ -0,0 +1,203 @@
...
+            executor.delayNextTask(Duration.ofSeconds(1));
+
+            assertThatThrownBy(noRetryChannel.execute(TestEndpoint.POST, request)::get)
+                    .getCause()
+                    .satisfies(cause -> assertThat(cause)
+                            .isInstanceOf(SafeLoggable.class)
+                            .isInstanceOf(IOException.class)
+                            .isNotInstanceOf(SocketTimeoutException.class)
+                            .extracting(value -> value.getClass().getSimpleName())
+                            .isEqualTo("SafeConnectTimeoutException"));
```
ferozco: need to do all of this since `SafeConnectTimeoutException` is packagePrivate

carterkozak: good call, originally it was nested+private!

dialogue-test-common/src/main/java/com/palantir/dialogue/AbstractHandshakeTimeoutTest.java
```diff
@@ -0,0 +1,203 @@
...
+
+            ClientConfiguration retryConfig = ClientConfiguration.builder()
```
ferozco: mind pulling this out into a separate test case so that we have one that verifies the exception that is thrown and another which verifies the retry behaviour? I think having them be named test cases will make it clearer for future readers

dialogue-test-common/src/main/java/com/palantir/dialogue/AbstractHandshakeTimeoutTest.java
carterkozak: eh disregard, these are in different modules. Since we only really care about hc at this point I can move the test and get rid of the abstract-ion.

dialogue-test-common/src/main/java/com/palantir/dialogue/AbstractHandshakeTimeoutTest.java
carterkozak: :+1:

dialogue-apache-hc5-client/src/test/java/com/palantir/dialogue/hc5/ApacheHandshakeTimeoutTest.java
```diff
@@ -0,0 +1,212 @@
...
+    }
+
+    @BeforeEach
+    public void beforeEach() {
+        SSLContext sslContext = SslSocketFactories.createSslContext(TestConfigurations.SSL_CONFIG);
+        Xnio xnio = Xnio.getInstance(Undertow.class.getClassLoader());
+        // This test operates based on the understanding that handshakes complete by executing
+        // the Runnable from SSLEngine.getDelegatedTask on the worker executor. Each handshake
+        // executes at least one of these tasks, and we can prolong the handshake by introducing
+        // a delay.
```
carterkozak, line 97: Added notes describing how and why this test works.

carterkozak:
Should be good to go.

ferozco (approved):
Awesome thanks @carterkozak!

PR #944, opened 2020-08-06
Response leak detection is always active

carterkozak:
## Before this PR
Response leak detection caused poor performance due to the use of finalizers, so it was not enabled by default. Now Dialogue uses a cleaner on supported runtimes to implement this functionality without the overhead.

## After this PR
Our internal benchmark results don't appear to be impacted, results change from 112k ops/sec to 
110k ops/sec which is well within the margin of error.

I considered only wrapping streaming binary responses, others cannot be leaked externally. Given such a low cost of detection for all responses I don't think that's necessary, and the confidence we gain is worth a small performance cost (if one exists).

==COMMIT_MSG==
Response leak detection is always active. Leaked responses are automatically closed on GC.
==COMMIT_MSG==

## Possible downsides?
There may be performance impacts that we don't expect. We haven't used cleaner much due to the java 9 requirement.
The cleaner is invoked reflectively in order to avoid requiring all dialogue users run on 9+, those on jre8 do no report leak metrics, nor are leaked resources automatically returned. I considered a multi-release jar, however they're difficult to test and complicate the build system.


dialogue-apache-hc5-client/src/test/java/com/palantir/dialogue/hc5/CleanerSupportTest.java
```diff
@@ -0,0 +1,58 @@
...
+import java.time.Duration;
+import java.util.concurrent.atomic.AtomicInteger;
+import org.awaitility.Awaitility;
+import org.junit.jupiter.api.Test;
+
+class CleanerSupportTest {
+
+    @Test
+    void testEnabled() {
+        boolean isJava8 = "1.8".equals(System.getProperty("java.specification.version"));
```
ferozco, line 34: Junit exposes `EnabledForJreRange` to manage which JREs test should run on

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/CleanerSupport.java
```diff
@@ -0,0 +1,78 @@
...
+import com.palantir.logsafe.exceptions.SafeRuntimeException;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.util.function.BiConsumer;
+import javax.annotation.Nullable;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** Reflective shim to allow consumers on new runtime versions to take advantage of java.lang.ref.Cleaner. */
+final class CleanerSupport {
```
ferozco: Why do we have this duplicate definition?

carterkozak: my reset didn't hit new files apparently!

dialogue-apache-hc5-client/src/test/java/com/palantir/dialogue/hc5/CleanerSupportTest.java
carterkozak: We're both testing that it's enabled on java9+ and that it's disabled on 11. I think we can use EnabledForJreRange for `testCleaner` though.

ferozco:
👍

PR #946, opened 2020-08-07
Report the delta between client and server call time

carterkozak:
This also introduces a meter for connection validity testing to
provide additional datapoints for the validateAfterInactivity
value.

==COMMIT_MSG==
Report the delta between client and server call time
==COMMIT_MSG==

### DIFF
--- a/dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientBlockingChannel.java
+++ b/dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientBlockingChannel.java
@@ -15,6 +15,7 @@
  */
 package com.palantir.dialogue.hc5;
 
+import com.codahale.metrics.Timer;
 import com.google.common.collect.ListMultimap;
 import com.google.common.collect.MultimapBuilder;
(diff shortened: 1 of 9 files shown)


dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientBlockingChannel.java
```diff
@@ -88,12 +93,15 @@ public Response execute(Endpoint endpoint, Request request) throws IOException {
         } else if (requiresEmptyBody(endpoint)) {
             builder.setEntity(EmptyHttpEntity.INSTANCE);
         }
+        ClassicHttpRequest classicHttpRequest = builder.build();
+        long startTimeNanos = System.nanoTime();
         try {
-            CloseableHttpResponse httpClientResponse = client.apacheClient().execute(builder.build());
+            CloseableHttpResponse httpClientResponse = client.apacheClient().execute(classicHttpRequest);
```
carterkozak, line 99: this includes handshake and idle connection validation time, so the p99s will be relatively large. I can look into starting the timer later in the request lifecycle.

carterkozak:
Closing in favor of #951

PR #948, opened 2020-08-10
Dialogue clients use a sympathetic validation interval

carterkozak:
This doesn't impact clients unless a Keep-Alive response header
is sent with a timeout value. When the server specifies a timeout,
clients trust the server to maintain connections as was foretold.
This design centralizes configuration on the server so clients
needn't be updated to match server configuration.

==COMMIT_MSG==
Dialogue clients use a sympathetic validation interval
==COMMIT_MSG==

## Possible downsides?
If unexpected servers (proxies) set a response `Keep-Alive` header, but fail to uphold the contract this could result in degraded performance.

### DIFF
--- a/changelog/@unreleased/pr-948.v2.yml
+++ b/changelog/@unreleased/pr-948.v2.yml
@@ -1,5 +1,6 @@
 type: improvement
 improvement:
-  description: Dialogue clients use a sympathetic validation interval
+  description: Dialogue clients use a sympathetic validation interval when servers
+    explicitly set a `Keep-Alive`  response header including a timeout duration.
   links:
   - https://github.com/palantir/dialogue/pull/948
--- a/changelog/@unreleased/pr-948.v2.yml
+++ b/changelog/@unreleased/pr-948.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Dialogue clients use a sympathetic validation interval
+  links:
+  - https://github.com/palantir/dialogue/pull/948
--- a/changelog/@unreleased/pr-948.v2.yml
+++ b/changelog/@unreleased/pr-948.v2.yml
@@ -1,6 +1,6 @@
 type: improvement
 improvement:
   description: Dialogue clients use a sympathetic validation interval when servers
-    explicitly set a `Keep-Alive`  response header including a timeout duration.
+    explicitly set a `Keep-Alive` response header including a timeout duration.
   links:
   - https://github.com/palantir/dialogue/pull/948
--- a/dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/InactivityValidationAwareConnectionKeepAliveStrategy.java
+++ b/dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/InactivityValidationAwareConnectionKeepAliveStrategy.java
@@ -43,6 +43,7 @@ import org.slf4j.LoggerFactory;
 final class InactivityValidationAwareConnectionKeepAliveStrategy implements ConnectionKeepAliveStrategy {
     private static final Logger log =
             LoggerFactory.getLogger(InactivityValidationAwareConnectionKeepAliveStrategy.class);
+    private static final String TIMEOUT_ELEMENT = "timeout";
 
     private final PoolingHttpClientConnectionManager connectionManager;
     private final String clientName;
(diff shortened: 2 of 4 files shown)


changelog/@unreleased/pr-948.v2.yml
```diff
@@ -0,0 +1,6 @@
+type: improvement
+improvement:
+  description: Dialogue clients use a sympathetic validation interval when servers
+    explicitly set a `Keep-Alive`  response header including a timeout duration.
```
carterkozak:
```suggestion
    explicitly set a `Keep-Alive` response header including a timeout duration.
```

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/InactivityValidationAwareConnectionKeepAliveStrategy.java
```diff
@@ -0,0 +1,88 @@
...
+    public TimeValue getKeepAliveDuration(HttpResponse response, HttpContext context) {
+        TimeValue result = DELEGATE.getKeepAliveDuration(response, context);
+        if (result != null
+                // Only use keep-alive values from 2xx responses
+                && response.getCode() / 100 == 2) {
+            TimeValue newInterval =
+                    containsKeepAliveHeaderWithTimeout(response) ? result : defaultValidateAfterInactivity;
+            TimeValue previousInterval = currentValidationInterval.getAndSet(newInterval);
+            if (!Objects.equals(previousInterval, newInterval)) {
+                log.info(
```
ferozco, line 98: might want to push this down to debug otherwise it could get noisy

changelog/@unreleased/pr-948.v2.yml
```diff
@@ -0,0 +1,6 @@
+type: improvement
+improvement:
+  description: Dialogue clients use a sympathetic validation interval when servers
+    explicitly set a `Keep-Alive` response header including a timeout duration.
```
iamdanfox, line 4: :+1: this functionality is available in WC 0.289.4 https://internal-github/foundry/witchcraft/pull/2640, but servers will havee to opt in using a install.server config option

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/InactivityValidationAwareConnectionKeepAliveStrategy.java
carterkozak: I added the `currentValidationInterval` getAndSet to avoid noise when the values are equal. I don't expect this to trigger very frequently. Rolling upgrades across versions can cause a little churn, but I don't expect it to be terrible.

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/InactivityValidationAwareConnectionKeepAliveStrategy.java
```diff
@@ -0,0 +1,88 @@
...
+        this.clientName = clientName;
+        // Store the initial inactivity interval to restore if responses re received without
+        // keep-alive headers.
+        this.defaultValidateAfterInactivity = connectionManager.getValidateAfterInactivity();
+        this.currentValidationInterval = new AtomicReference<>(defaultValidateAfterInactivity);
+    }
+
+    @Override
+    public TimeValue getKeepAliveDuration(HttpResponse response, HttpContext context) {
+        TimeValue result = DELEGATE.getKeepAliveDuration(response, context);
```
iamdanfox: the behaviour of falling back to return the `requestConfig.getConnectionKeepAlive();` is quite strange here... I wonder if it would be clearer to just copy and paste the relevant lines of this method?

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/InactivityValidationAwareConnectionKeepAliveStrategy.java
```diff
@@ -0,0 +1,88 @@
...
+            if (!Objects.equals(previousInterval, newInterval)) {
+                log.info(
+                        "Updating the validation interval for {} from {} to {}",
+                        SafeArg.of("client", clientName),
+                        SafeArg.of("previousInterval", previousInterval),
+                        SafeArg.of("newInterval", newInterval));
+            }
+            // Simple volatile write, no need to protect this in the getAndSet check. The getAndSet may race this call
+            // so it's best to completely decouple the two.
+            connectionManager.setValidateAfterInactivity(newInterval);
```
iamdanfox, line 106: Do we want to do any sanity checking of the range here? like what happens if some bizarre proxy/server sends timeout=-1, timeout=0, timeout=999999 etc?

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/InactivityValidationAwareConnectionKeepAliveStrategy.java
carterkozak: I went back and forth on this. I think you're right that it's easier to follow if we duplicate the implementation, even if it means potential future upstream changes won't be reflected. I don't think there will be changes to the default implementation because newer http protocols don't use keep-alive headers.

iamdanfox: :+1: I'm fine with this, it makes it fully obvious what's going on, and gives us an opporunity to point out with a comment that it's in _seconds_!

iamdanfox (approved):
I'm intrigued to see if this codepath gets triggered by any existing proxies out there in the wild - it should just magically give us better re-use (unless the proxy is buggy as you mention)... in that case, our retries should save us 🤷 .

Intention is to switch the WC-side flag on in timelock right?

carterkozak:
> Intention is to switch the WC-side flag on in timelock right?

That's correct

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/InactivityValidationAwareConnectionKeepAliveStrategy.java
carterkozak: non-positive values should be omitted since those values as timeouts are unclear, some components use negative and zero values as infinite. If a server wants to advertise an exceedingly long timeout, I'm inclined to trust it for now. We could add a maximum later. In the worst case, when a server goes down, our retries will get us into a good state.

PR #951, opened 2020-08-10
Report the difference between client and server request duration

carterkozak:
This approach leverages the connection `receiveResponseHeader`
method which is as close to the socket as we can get, but has
the drawback that it cannot account for headers sent earlier
with large request entities.
In many cases the request header and entity can fit into a single
packet, so this is a reasonable trade-off, and captures data for
our most performance sensitive endpoints.

==COMMIT_MSG==
Report the difference between client and server request duration
==COMMIT_MSG==

### DIFF
--- a/dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/InstrumentedManagedHttpClientConnection.java
+++ b/dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/InstrumentedManagedHttpClientConnection.java
@@ -16,27 +16,33 @@
 
 package com.palantir.dialogue.hc5;
 
+import com.codahale.metrics.Timer;
+import com.google.common.net.HttpHeaders;
 import com.palantir.tracing.CloseableTracer;
 import java.io.IOException;
 import java.net.Socket;
 import java.net.SocketAddress;
+import java.util.concurrent.TimeUnit;
 import javax.net.ssl.SSLSession;
 import org.apache.hc.client5.http.io.ManagedHttpClientConnection;
 import org.apache.hc.core5.http.ClassicHttpRequest;
 import org.apache.hc.core5.http.ClassicHttpResponse;
 import org.apache.hc.core5.http.EndpointDetails;
+import org.apache.hc.core5.http.Header;
 import org.apache.hc.core5.http.HttpException;
 import org.apache.hc.core5.http.ProtocolVersion;
 import org.apache.hc.core5.io.CloseMode;
 import org.apache.hc.core5.util.Timeout;
 
-/** A simple wrapper around a {@link ManagedHttpClientConnection} which provides tracing information. */
-final class TracedManagedHttpClientConnection implements ManagedHttpClientConnection {
+/** A simple wrapper around a {@link ManagedHttpClientConnection} which provides instrumentation. */
+final class InstrumentedManagedHttpClientConnection implements ManagedHttpClientConnection {
 
     private final ManagedHttpClientConnection delegate;
+    private final Timer responseDeltaTimer;
 
-    TracedManagedHttpClientConnection(ManagedHttpClientConnection delegate) {
+    InstrumentedManagedHttpClientConnection(ManagedHttpClientConnection delegate, Timer responseDeltaTimer) {
         this.delegate = delegate;
+        this.responseDeltaTimer = responseDeltaTimer;
     }
 
     @Override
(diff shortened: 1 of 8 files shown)


dialogue-apache-hc5-client/src/test/java/com/palantir/dialogue/hc5/ServerTimingParserTest.java
```diff
@@ -0,0 +1,50 @@
...
+
+    @Test
+    void testMissingValue() {
+        assertThat(ServerTimingParser.getServerDurationNanos("missedCache", "missedCache"))
+                .isEqualTo(ServerTimingParser.UNKNOWN);
+    }
+
+    @Test
+    void testSingleMetricWithValue() {
+        assertThat(ServerTimingParser.getServerDurationNanos("cpu;dur=2.4", "cpu"))
```
carterkozak, line 33: Our servers don't report floating point dur values, but I'd like to include more precise timing at some point.

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ServerTimingParser.java
```diff
@@ -0,0 +1,83 @@
...
+                    // Ignore any extraneous fields, search for `dur`
+                    int durationStart = segment.indexOf(DURATION);
+                    if (durationStart < 0) {
+                        return UNKNOWN;
+                    }
+                    int durationValueStart = durationStart + DURATION.length();
+                    int durationValueEnd = findNumericSequenceEnd(segment, durationValueStart);
+                    String durationValue = segment.substring(durationValueStart, durationValueEnd);
+                    double durationMilliseconds = Double.parseDouble(durationValue);
+                    return (long) (NANOSECONDS_PER_MILLISECOND * durationMilliseconds);
```
iamdanfox, line 64: Interesting that https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Server-Timing doesn't seem to specify the unit that `dur` should be in, and seems like it's a chrome convention to make it milliseconds??  (which we've complied with in WC)

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/InstrumentedManagedHttpClientConnection.java
```diff
@@ -95,7 +101,29 @@ public void sendRequestEntity(ClassicHttpRequest request) throws HttpException,
     @Override
     public ClassicHttpResponse receiveResponseHeader() throws HttpException, IOException {
```
iamdanfox, line 102: Looking for callers of this, it seems like it's only HttpRequestExecutor#execute that calls this, but as it's in a while loop, this could get a little trippy for status codes < 200 (i.e. 100 continue, 101 switching protocols)... i guess that's fine as those are presumably very rare?

dialogue-clients/metrics.md
```diff
@@ -7,6 +7,7 @@
 ### dialogue.client
 Dialogue client response metrics provided by the Apache client channel.
 - `dialogue.client.response.leak` tagged `client-name`, `service-name`, `endpoint` (meter): Rate that responses are garbage collected without being closed. This should only occur in the case of a programming error.
+- `dialogue.client.response.delta` tagged `client-name` (timer): Difference in request time reported by the client and by the server. This metric is only reported when the remote server provides a `Server-Timing` response header with a timing value for `server`.
```
iamdanfox: naming is hard, but maybe "networkoverhead" "servertimingoverhead" or "servertimingdelta" ?

Might also be worth reminding people that the timing value for `server` needs to be in milliseconds?

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/InstrumentedManagedHttpClientConnection.java
carterkozak: That's correct. We could limit this to the 2xx range, but even if we encounter 1xx responses they won't include the `Server-Timing` header.

iamdanfox (approved):
I'm pretty interested to see what values this ends up showing!

dialogue-clients/metrics.md
carterkozak: servertimingoverhead sounds much better, thanks!

> Might also be worth reminding people that the timing value for server needs to be in milliseconds?

I don't think we should mention units here since all our timers report values normalized to microseconds, I don't want to confuse folks.

iamdanfox: I meant in the header string, in `Server-Timing: server;dur=50` the 50 represents milliseconds

PR #955, opened 2020-08-11
[experimental] balanced channel factors in concurrency limiter's `limit` info

iamdanfox:
## Before this PR
The theory here is that our BalancedScoreTracker is watching out for 4xxs and 5xxs and trying to infer which upstream is 'best' and therefore which we should send requests too, but the ConcurrencyLimitedChannel is _also_ watching out for 4xxs and 5xxs to infer whether one upstream is healthy or not. Maybe we should stop tracking this in two places and combine them?

## After this PR
==COMMIT_MSG==

==COMMIT_MSG==


dialogue-core/src/main/java/com/palantir/dialogue/core/BalancedScoreTracker2.java
```diff
@@ -0,0 +1,343 @@
...
+/**
+ * Chooses nodes based on stats about each channel, i.e. how many requests are currently
+ * being served, how many failures have been seen in the last few seconds.
+ *
+ * This is intended to be a strict improvement over Round Robin and Random Selection which can leave fast servers
+ * underutilized, as it sends the same number to both a slow and fast node. It is *not* appropriate for transactional
+ * workloads (where n requests must all land on the same server) or scenarios where cache warming is very important.
+ * {@link PinUntilErrorNodeSelectionStrategyChannel} remains the best choice for these.
+ */
+final class BalancedScoreTracker2 {
```
iamdanfox: I had to duplicate the BalancedScoreTracker because the way it was wired into the sticky endpoint channels makes refactoring a giant ballache - dw I don't consider the PR mergeable in this state.

carterkozak: It's difficult to read the diff, I'm not sure what's changing.

dialogue-core/src/main/java/com/palantir/dialogue/core/BalancedScoreTracker2.java
```diff
@@ -0,0 +1,343 @@
...
+        /** A low score is considered desirable (and they can be negative). */
+        private ScoreSnapshot computeScoreSnapshot() {
+            int requestsInflight = concurrencyLimitedChannel.getInflight();
+            int limit = (int) concurrencyLimitedChannel.getMax().orElse(0);
+
+            double failureReservoir = recentFailuresReservoir.get();
+
+            // it's important that scores are integers because if we kept the full double precision, then a single 4xx
+            // would end up influencing host selection long beyond its intended lifespan in the absence of other data.
+            int score = requestsInflight - limit + Ints.saturatedCast(Math.round(failureReservoir));
```
iamdanfox: Here's the interesting change:
```diff
-int score = requestsInflight + Ints.saturatedCast(Math.round(failureReservoir));
+int score = requestsInflight - limit + Ints.saturatedCast(Math.round(failureReservoir));
```

I just decided to linearly factor the concurrencylimiter's upper limit in... it's a bit confusing because low score = good, so many inflight = bad, but we want a channel with high available capacity to have a low score, so I just subtracted it :shrug:

dialogue-core/src/main/java/com/palantir/dialogue/core/ConcurrencyLimitedChannelImpl.java
```diff
@@ -0,0 +1,187 @@
...
+import java.util.function.Function;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * A channel that monitors the successes and failures of requests in order to determine the number of concurrent
+ * requests allowed to a particular channel. If the channel's concurrency limit has been reached, the {@link
+ * #maybeExecute} method returns empty.
+ */
+final class ConcurrencyLimitedChannelImpl implements ConcurrencyLimitedChannel {
```
iamdanfox, line 45: this file hasn't changed.

iamdanfox:
Honestly having written this I'm very unconvinced that this is useful.

dialogue-core/src/main/java/com/palantir/dialogue/core/BalancedScoreTracker2.java
carterkozak: Assuming some base rate of failures (qos, network blips, etc) the in-flight requests and limit should be roughly equivalent under load.

carterkozak:
Want to vtc through the idea?

PR #963, opened 2020-08-21
[proof-of-concept] ConcurrencyLimiter per <uri, endpoint>, not per <uri>

iamdanfox:
## Before this PR

Probably the heaviest internal users of 429s is our atlasdb proxy, which has different read and write ratelimiters, which sometimes differ by >10x. Up until now, Dialogue has set up a single concurrency limiter for each _host_, so if if there are 20 endpoints in a service and 9 hosts, we'd just have 9 concurrency limiters. This would end up inferring a single number to protect all the endpoints of that server. 

This has actually worked completely fine at the vast majority of cases, and defends very effectively in the scenario where there's a server-level problem (e.g. GC / resource contention / some other slowdown) 

However, we've seen issues from one specific deployment (p) where it seems atlasdb-proxy clients are either being throttled too much or not enough, resulting in either too much queueing or blasting through retries and experiencing failures.

## After this PR

So what changed:
- there's now a concurrency limiter for every <host,endpoint> pair, so in the example above where there are 9 hosts each serving the same 20 endpoints, we'd have 180 ConcurrencyLimitedChannels floating around
- instead of one queue for the entire client, we now have a queue per endpoint (so 20 of these)
- the original 'Channel' abstraction doesn't work nicely anymore, because we really need to know how many endpoints there are _upfront_, and Channel lets users people in new endpoints whenever they like...

The result is _most_ simulations are unchanged, but we have got the desired effect of differentiating between endpoints:

image

_Note: This PR is _not_ a serious contender to merge, I mainly wanted to explore the implications of this change and see what the simulations look like. As such, I haven't spent the time fleshing out implementations of:_
- full logging/observability to the TaggedMetricRegistry (although the simple approach would _dramatically_ expand our DD $ cost)
- anything that understands the server-driven Node-Selection-Strategy header

==COMMIT_MSG==

==COMMIT_MSG==

## Possible downsides?
ooo boy
- Dialogue will probably behave quite differently depending on the number of endpoints in your API, e.g. if you talk to a service with hundreds (or even thousands of endpoints) but one rate limiter to span across many of them, then things are going to be pretty non-sensical.
- observability will be much harder. going from 1 queue to O(endpoints) queues, and multiplying the number of concurrency limiters will be a bit nuts.
- it's _obviously_ not safe to just make the `DialogueChannel#execute(Endpoint,Request)` method throw an UnsupportedOperationException - we'd likely have to introduce some kind of cache for this.
- we're creating at least an order of magnitude more objects per client.


carterkozak:
> The result is most simulations are unchanged

How many simulations use N > 1 endpoints?

iamdanfox:
I think it's literally just the `one_endpoint_dies_on_each_server` case that uses multiple endpoints.  We do actually have a dump of request logs now though, so I think we could make a simulation case that has the same number of endpoints as the atlasdb-proxy has, with a similarly proportional workload?

carterkozak:
I'm confident that the atlas proxy will work better if we model dialogue after it. I'm concerned about regressing everything else.

PR #971, opened 2020-08-26
Dialogue uses a safe direct executor which catches and logs failures

carterkozak:
This resolves an issue where dialogue state could be corrupted
when Errors are thrown. In general it's not a good idea to handle
Error, however leaving Dialogue in a degraded state leads to
investigation in the wrong areas.

==COMMIT_MSG==
Dialogue uses a safe direct executor which catches and logs failures
==COMMIT_MSG==

### DIFF
--- a/dialogue-futures/src/test/java/com/palantir/dialogue/futures/DialogueFuturesTest.java
+++ b/dialogue-futures/src/test/java/com/palantir/dialogue/futures/DialogueFuturesTest.java
@@ -17,14 +17,16 @@
 package com.palantir.dialogue.futures;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatCode;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
 import com.google.common.util.concurrent.AsyncFunction;
 import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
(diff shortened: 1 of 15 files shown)


dialogue-futures/src/test/java/com/palantir/dialogue/futures/DialogueFuturesTest.java
```diff
@@ -235,6 +237,36 @@ void testCatchingAllAsync_resultCancel_completed(Transformer transformer) throws
...
+        AtomicInteger invocations = new AtomicInteger();
+        Runnable listener = () -> {
+            invocations.incrementAndGet();
+            throw new Error();
+        };
+        DialogueFutures.addDirectListener(future, listener);
+        DialogueFutures.addDirectListener(future, listener);
+        future.set("value");
+        assertThat(invocations).hasValue(2);
+    }
```
carterkozak, line 268: This one is :hot_pepper: without the fix. `future.set` throws the Error and invocations value is 1.

iamdanfox (approved):
Definitely in favour of the extra safety provided by this, but do you think this would actually save the day in PDS-132792? 

I couldn't figure out what Error / non-RuntimeExceptions would be blasting through the AbstractFuture#executeListener's `catch RuntimeException` block??

carterkozak:
Unfortunately we don't have data to be confident either way -- I am confident that this _could_ cause the behavior we observed.

PR #973, opened 2020-08-26
Dialogue reports concurrency limiter metrics for in-flight requests

carterkozak:
==COMMIT_MSG==
Dialogue reports concurrency limiter metrics for in-flight requests
==COMMIT_MSG==

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/ConcurrencyLimitedChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/ConcurrencyLimitedChannel.java
@@ -16,7 +16,6 @@
 
 package com.palantir.dialogue.core;
 
-import com.codahale.metrics.Gauge;
 import com.codahale.metrics.Meter;
 import com.google.common.util.concurrent.ListenableFuture;
 import com.palantir.conjure.java.client.config.ClientConfiguration;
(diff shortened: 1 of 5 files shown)


dialogue-core/src/main/java/com/palantir/dialogue/core/ConcurrencyLimitedChannel.java
```diff
@@ -76,15 +73,25 @@ static LimitedChannel create(Config cf, LimitedChannel channel, int uriIndex) {
...
+                metrics.max()
                         .channelName(channelName)
                         .hostIndex(Integer.toString(uriIndex))
                         .buildMetricName(),
-                this,
-                ConcurrencyLimitedChannel::getMax);
+                ConcurrencyLimitedChannel::getMax,
+                doubleStream -> doubleStream.max().orElse(0D),
+                this);
+        DialogueInternalWeakReducingGauge.getOrCreate(
```
iamdanfox, line 86: Would it be crazy to wrap this in a `log.isDebugEnabled()` flag? I know it would only enable at startup, just that every metric costs $ and we previously deleted some similar metrics (I think we had active requests, active connections and concurrency limiter inflight, which all pretty much showed the same number)

iamdanfox (approved):
This would certainly confirm/disprove our hypothesis about concurrency limiter permits somehow not being released in PDS-132792

dialogue-core/src/main/java/com/palantir/dialogue/core/ConcurrencyLimitedChannel.java
```diff
@@ -76,15 +73,25 @@ static LimitedChannel create(Config cf, LimitedChannel channel, int uriIndex) {
...
-                DialogueConcurrencylimiterMetrics.of(taggedMetrics)
-                        .max()
+                metrics.max()
                         .channelName(channelName)
                         .hostIndex(Integer.toString(uriIndex))
                         .buildMetricName(),
-                this,
-                ConcurrencyLimitedChannel::getMax);
+                ConcurrencyLimitedChannel::getMax,
+                doubleStream -> doubleStream.max().orElse(0D),
```
carterkozak: Or do we want min here?

dialogue-core/src/main/java/com/palantir/dialogue/core/ConcurrencyLimitedChannel.java
carterkozak: I'd rather have data available for incidents that have already occurred. Days of engineer time are expensive.

PR #974, opened 2020-08-26
[pure refactor] Emphasise usage of NeverThrowChannel

iamdanfox:
## Before this PR

While investigating #pds-132792, we had a hypothesis that the concurrency limiter's `inflight` AtomicInteger was being incremented, but then an exception thrown when calling the delegate, leading to it never being decremented again.

On closer reading, I don't actually think this is possible because we were already using a `NeverThrowChannel`, but I didn't spot this on the first read through, so figured it might be worth rejigging for clarity.

## After this PR
==COMMIT_MSG==
Emphasis usage of NeverThrowChannel in cases where correctness relies on a listener being run (e.g. ConcurrencyLimitedChannel, QueuedChannel and TraceEnrichingChannel).
==COMMIT_MSG==

_This PR is mostly cosmetic tbh_

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/ConcurrencyLimitedChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/ConcurrencyLimitedChannel.java
@@ -20,6 +20,7 @@ import com.codahale.metrics.Gauge;
 import com.codahale.metrics.Meter;
 import com.google.common.util.concurrent.ListenableFuture;
 import com.palantir.conjure.java.client.config.ClientConfiguration;
+import com.palantir.dialogue.Channel;
 import com.palantir.dialogue.Endpoint;
 import com.palantir.dialogue.Request;
 import com.palantir.dialogue.Response;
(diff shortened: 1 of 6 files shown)


dialogue-core/src/main/java/com/palantir/dialogue/core/ConcurrencyLimitedChannel.java
```diff
@@ -97,13 +98,9 @@ static CautiousIncreaseAggressiveDecreaseConcurrencyLimiter createLimiter() {
         if (maybePermit.isPresent()) {
             CautiousIncreaseAggressiveDecreaseConcurrencyLimiter.Permit permit = maybePermit.get();
             logPermitAcquired();
-            Optional<ListenableFuture<Response>> result = delegate.maybeExecute(endpoint, request);
-            if (result.isPresent()) {
-                DialogueFutures.addDirectCallback(result.get(), permit);
-            } else {
-                permit.ignore();
-            }
```
iamdanfox, line 105: this branching was also pointless - we'd literally never end up in this `else` case because the adapter always returned Optional.of

carterkozak (approved):

carterkozak:
> While investigating #pds-132792, we had a hypothesis that the concurrency limiter's inflight AtomicInteger was being incremented, but then an exception thrown when calling the delegate, leading to it never being decremented again.

I thought our theory was Errors thrown from ListenableFuture listeners prevented other listeners from running, not necessarily channels themselves

iamdanfox:
Yep I think that's the main one-  this PR was just one of the things I wondered about when reading the code early on, and I've almost pushed up this PR a few times so figured I might as well.

carterkozak:
Makes sense :+1:

PR #991, opened 2020-09-15
Chill out log line

iamdanfox:
## Before this PR

I've been seeing a few log lines around startup saying `Multiple ({}) objects contribute to the same gauge, taking the average (beware this may be misleading) {} {}`.  

This is _kinda_ interesting, but ultimately a false positive because it helpfully showed me the array of values was all zero: `values: "[0, 0]"`, so averaging zeros is totally fine.

## After this PR
==COMMIT_MSG==
False positive log lines warning about `Multiple ({}) objects contribute to the same gauge, taking the average (beware this may be misleading) {} {}` will no longer appear - these will only show up if the different constituent values are _actually_ different.
==COMMIT_MSG==

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/BalancedScoreTracker.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/BalancedScoreTracker.java
@@ -36,6 +36,7 @@ import java.util.List;
 import java.util.Random;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.stream.IntStream;
+import java.util.stream.LongStream;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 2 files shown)


carterkozak (approved):

dialogue-core/src/main/java/com/palantir/dialogue/core/BalancedScoreTracker.java
```diff
@@ -250,7 +251,9 @@ private static void registerGauges(
                     c -> c.computeScoreSnapshot().getScore(),
                     longStream -> {
                         long[] longs = longStream.toArray();
```
carterkozak, line 253: stream -> array -> stream

iamdanfox: I know, it's a big funky, but you can't consume them twice!

carterkozak: I wonder if we should filter out zero from channels created and never used. If we have an empty stream, we can be confident the value is zero.

carterkozak: (just musing, happy merging as is)

PR #992, opened 2020-09-16
Server errors resulting from PUT requests are retried

carterkozak:
PUT requests are considered idempotent based on rfc7231, several
internal projects already assume PUT requests are retried (in fact
the proxy does retry them). Outside of those projects, PUT is not
utilized, thus there's minimal risk.

==COMMIT_MSG==
Server errors resulting from PUT requests are retried
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-992.v2.yml
+++ b/changelog/@unreleased/pr-992.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Server errors resulting from PUT requests are retried
+  links:
+  - https://github.com/palantir/dialogue/pull/992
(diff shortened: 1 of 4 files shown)


iamdanfox:
How confident are you that the current PUT requests out there in the wild are idempotent? I guess we could look at the big ol' request log dataset in papaya to sanity check _exactly_ who is using these?

sourcegraph shows ~280 in foundry alone: https://sourcegraph.p.b/search?q=http.+PUT+lang:yaml+file:conjure+count:1000+repo:foundry/&patternType=regexp&case=yes#67

carterkozak:
not stale

PR #993, opened 2020-09-18
429 applies to endpoints, not entire hosts

carterkozak:
It's queues all the way down. The host-level queue and concurrency
limiter ignore 429 where a new per-endpoint queue will queue requests
based on these responses.

Drawbacks:
The per-endpoint queue is not capable of switching hosts, much
like CJR. Once a host is selected for a request, endpoint-level
limits can only prevent early attempts to make requests.


dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueChannel.java
```diff
@@ -127,12 +130,27 @@ public DialogueChannel build() {
...
                 channel = HostMetricsChannel.create(cf, channel, uri);
-                channel = new TraceEnrichingChannel(channel);
+                Channel tracingChannel = new TraceEnrichingChannel(channel);
+                channel = cf.clientConf().clientQoS() == ClientQoS.ENABLED
+                        ? new ChannelToEndpointChannel(() -> {
+                            Config withoutMetrics = ImmutableConfig.builder()
+                                    .from(cf)
+                                    .rawConfig(ClientConfiguration.builder()
+                                            .from(cf.rawConfig())
+                                            .taggedMetricRegistry(new DefaultTaggedMetricRegistry())
```
iamdanfox: hahaha

dialogue-core/src/main/java/com/palantir/dialogue/core/ChannelToEndpointChannel.java
```diff
@@ -0,0 +1,59 @@
...
+import com.palantir.dialogue.Response;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Function;
+import java.util.function.Supplier;
+
+final class ChannelToEndpointChannel implements Channel {
+
+    private final Function<Object, Channel> adapter;
+    private final Map<Object, Channel> cache;
```
iamdanfox, line 31: mr I don't like caches

carterkozak: Ya, I didn't see a cleaner way that doesn't break feign

carterkozak:
~still needs docs/ascii-art~

dialogue-core/src/main/java/com/palantir/dialogue/core/CautiousIncreaseAggressiveDecreaseConcurrencyLimiter.java
```diff
@@ -76,7 +82,59 @@
...
     }
 
-    final class Permit implements FutureCallback<Response> {
+    enum Behavior {
+        HOST_LEVEL() {
+            @Override
+            void onSuccess(Response result, PermitControl control) {
+                if ((Responses.isQosStatus(result) && !Responses.isTooManyRequests(result))) {
+                    control.dropped();
+                } else if (Responses.isServerError(result) || Responses.isTooManyRequests(result)) {
```
carterkozak: What do you think about promoting non-500 server errors to the host level, and only using 500 status server errors for per-endpoint limits?

carterkozak: 502 and 504 are relatively common in some environments, and they apply to the entire channel.

iamdanfox: I'd be OK with carving out 502 and 504 as we do commonly see them from nginx / proxies, but personally I'd actually keep 500 as an endpoint-level concern tbh, because I think it's very possible that someone writes a bug in their FooResource#myEndpoint method that wouldn't affect any other endpoints

dialogue-core/src/main/java/com/palantir/dialogue/core/CautiousIncreaseAggressiveDecreaseConcurrencyLimiter.java
```diff
@@ -76,7 +82,59 @@
...
+                    control.dropped();
+                } else {
+                    control.ignore();
+                }
+            }
+        },
+        ENDPOINT_LEVEL() {
+            @Override
+            void onSuccess(Response result, PermitControl control) {
+                if (Responses.isTooManyRequests(result) || Responses.isServerError(result)) {
```
iamdanfox: could we rename these methods to make em slightly more readable (e.g. isServerError to is5xx or isServerError5xx)? I had assumed that isServerError meant ==500 when it's actually _any_ 5xx...

dialogue-core/src/main/java/com/palantir/dialogue/core/CautiousIncreaseAggressiveDecreaseConcurrencyLimiter.java
carterkozak: Yep, I think that's exactly what I'm suggesting, we apply the `500` status to individual endpoints, but other 5xx status codes (501 through 599 inclusive) to the host.

dialogue-core/src/main/java/com/palantir/dialogue/core/ChannelToEndpointChannel.java
```diff
@@ -0,0 +1,58 @@
...
+
+    private Channel channelFor(Endpoint endpoint) {
+        return cache.computeIfAbsent(key(endpoint), _key -> adapter.apply(endpoint));
+    }
+
+    private static boolean isConstant(Endpoint endpoint) {
+        return endpoint instanceof Enum;
+    }
+
+    private static Object key(Endpoint endpoint) {
```
iamdanfox, line 62: I think the use of Object in class deserves a little comment actually, cos people might well read through generated code and think _everything_ will be enums, whereas we're actually saying that people can hand-roll Endpoints (and also the feign ones create em on the fly)

dialogue-core/src/main/java/com/palantir/dialogue/core/ChannelToEndpointChannel.java
```diff
@@ -0,0 +1,58 @@
...
+    @Override
+    public ListenableFuture<Response> execute(Endpoint endpoint, Request request) {
+        return channelFor(endpoint).execute(endpoint, request);
+    }
+
+    private Channel channelFor(Endpoint endpoint) {
+        return cache.computeIfAbsent(key(endpoint), _key -> adapter.apply(endpoint));
+    }
+
+    private static boolean isConstant(Endpoint endpoint) {
```
iamdanfox, line 51: only one usage of this dude, might be more readable inlined?

dialogue-core/src/main/java/com/palantir/dialogue/core/CautiousIncreaseAggressiveDecreaseConcurrencyLimiter.java
carterkozak: Most places call the entire 5xx range "server error" where the 500 status specifically is an _internal_ server error. Definitely some ambiguity in the spec... perhaps we add a `Range` suffix to the method name?

dialogue-core/src/main/java/com/palantir/dialogue/core/ConcurrencyLimitedChannel.java
```diff
@@ -41,16 +41,42 @@
...
+            Endpoint endpoint,
+            ClientConfiguration.ClientQoS clientQoS) {
+        switch (clientQoS) {
+            case ENABLED:
+                return new ConcurrencyLimitedChannel(
+                        channel,
+                        createLimiter(Behavior.ENDPOINT_LEVEL),
+                        new EndpointConcurrencyLimitedChannelInstrumentation(channelName, uriIndex, endpoint));
             case DANGEROUS_DISABLE_SYMPATHETIC_CLIENT_QOS:
                 return new ChannelToLimitedChannelAdapter(channel);
```
iamdanfox: if a user builds a client with ClientQoS disabled, ideally we want to omit the endpoint-level queue _and_ this adapter with it's concurrent hashmap right? 

I think this is implemented in DialogueChannel#L138, but doesn't that mean this branch will never be reached?

simulation/src/test/resources/report.md
```diff
@@ -1,8 +1,8 @@
 # Report
 <!-- Run SimulationTest to regenerate this report. -->
 ```
-                all_nodes_500[CONCURRENCY_LIMITER_PIN_UNTIL_ERROR].txt:	success=79.3%	client_mean=PT6.15542S     	server_cpu=PT20M          	client_received=2000/2000	server_resps=2000	codes={200=1586, 500=414}
-                    all_nodes_500[CONCURRENCY_LIMITER_ROUND_ROBIN].txt:	success=73.7%	client_mean=PT3.269745S    	server_cpu=PT20M          	client_received=2000/2000	server_resps=2000	codes={200=1474, 500=526}
+                all_nodes_500[CONCURRENCY_LIMITER_PIN_UNTIL_ERROR].txt:	success=73.9%	client_mean=PT4.24575S     	server_cpu=PT20M          	client_received=2000/2000	server_resps=2000	codes={200=1478, 500=522}
```
iamdanfox, line 4: do you know what's going on with double-peak recovery graph here? client_mean seems to have improved, but overall client-perceived success rate decreased... does this mean we're burning through retries faster?

image

dialogue-core/src/main/java/com/palantir/dialogue/core/ConcurrencyLimitedChannel.java
carterkozak: :+1: yep, this can be simplified

simulation/src/test/resources/report.md
```diff
@@ -25,23 +25,23 @@
...
                               one_big_spike[UNLIMITED_ROUND_ROBIN].txt:	success=99.9%	client_mean=PT1.000523583S 	server_cpu=PT7M37.35S     	client_received=1000/1000	server_resps=3049	codes={200=999, 429=1}
-one_endpoint_dies_on_each_server[CONCURRENCY_LIMITER_PIN_UNTIL_ERROR].txt:	success=63.9%	client_mean=PT9.2169728S   	server_cpu=PT25M          	client_received=2500/2500	server_resps=2500	codes={200=1597, 500=903}
- one_endpoint_dies_on_each_server[CONCURRENCY_LIMITER_ROUND_ROBIN].txt:	success=64.1%	client_mean=PT4.6030464S   	server_cpu=PT25M          	client_received=2500/2500	server_resps=2500	codes={200=1602, 500=898}
+one_endpoint_dies_on_each_server[CONCURRENCY_LIMITER_PIN_UNTIL_ERROR].txt:	success=66.0%	client_mean=PT2.0950112S   	server_cpu=PT25M          	client_received=2500/2500	server_resps=2500	codes={200=1649, 500=851}
+ one_endpoint_dies_on_each_server[CONCURRENCY_LIMITER_ROUND_ROBIN].txt:	success=66.2%	client_mean=PT2.0278288S   	server_cpu=PT25M          	client_received=2500/2500	server_resps=2500	codes={200=1655, 500=845}
```
iamdanfox, line 29: Oh this is pretty 😍 - requests keep flowing to the working endpoints and are properly throttled to the broken ones!

image

carterkozak: It works better -- it's not the ideal design for this scenario because we don't take the endpoint into account when we make a node selection decision, but I think it will save us some headache :-)

simulation/src/test/resources/report.md
carterkozak: I think this is an artifact of requests sitting in per-endpoint queues being considered in-flight.

dialogue-core/src/main/java/com/palantir/dialogue/core/ChannelToEndpointChannel.java
carterkozak: I added comments, I like the separate function because it makes the meaning clear, the additional javadoc should help as well.

dialogue-core/src/main/java/com/palantir/dialogue/core/CautiousIncreaseAggressiveDecreaseConcurrencyLimiter.java
```diff
@@ -76,7 +82,62 @@
...
+    }
+
+    interface PermitControl {
+
+        void ignore();
+
+        void dropped();
+
+        void success();
+    }
```
iamdanfox, line 138: I know this terminology matches the netflix lib, but given how much indirection we already have, could we name them `unchangedLimit / decreaseLimit / increaseLimit` ?

carterkozak: That sound reasonable, but I’m not sure I’d like to rename methods in this refactor as it’s already fairly large. This only extracts an interface from existing code.

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
```diff
@@ -69,24 +69,44 @@
...
+    QueuedChannel(LimitedChannel delegate, String channelName, QueuedChannelInstrumentation metrics, int maxQueueSize) {
         this.delegate = new NeverThrowLimitedChannel(delegate);
         this.channelName = channelName;
         // Do _not_ call size on a ConcurrentLinkedDeque. Unlike other collections, size is an O(n) operation.
         this.queuedCalls = new ProtectedConcurrentLinkedDeque<>();
         this.maxQueueSize = maxQueueSize;
-        this.queueSizeCounter = DialogueClientMetrics.of(metrics).requestsQueued(channelName);
-        this.queuedTime = DialogueClientMetrics.of(metrics).requestQueuedTime(channelName);
+        // Lazily create the counter. Unlike meters, timers, and histograms, counters cannot be ignored when they have
+        // zero interactions because they support both increment and decrement operations.
```
iamdanfox, line 86: have opened https://github.com/palantir/tritium/pull/872 as a possible alternative to this... not sure whether the $ saving would be significant 🤷

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
```diff
@@ -118,6 +140,8 @@ static QueuedChannel create(Config cf, LimitedChannel delegate) {
             return Optional.empty();
         }
 
+        queueUsed = true;
```
iamdanfox: I think i'd call this `shouldRecordQueuedTimeMetric` just to emphasise that this is purely kept around for observability purposes and shouldn't affect behaviour

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
```diff
@@ -315,4 +339,47 @@ public int size() {
...
+
+            @Override
+            public Timer requestQueuedTime() {
+                return metrics.requestQueuedTime(channelName);
+            }
+        };
+    }
+
+    static QueuedChannelInstrumentation endpointInstrumentation(
+            DialogueClientMetrics metrics, String channelName, String service, String endpoint) {
```
iamdanfox, line 366: Three strings in a row makes me nervous, what about just passing in the entire `com.palantir.dialogue.Endpoint` instead of the last two args, and then pulling out endpoint.servicveName() and endpoint.endpointName() as necessary below?

carterkozak: We can do that, I wanted to be careful not to hold an instance of the endpoint and potentially leak temporary instances.

dialogue-core/src/main/java/com/palantir/dialogue/core/BalancedScoreTracker.java
```diff
@@ -151,10 +151,10 @@ public void undoStartRequest() {
         public void onSuccess(Response response) {
             inflight.decrementAndGet();
 
-            if (Responses.isQosStatus(response) || Responses.isServerError(response)) {
+            if (isGlobalQosStatus(response) || Responses.isServerErrorRange(response)) {
```
iamdanfox, line 154: this change should have a pretty big impact on node-selection decisions for our internal ~ski product~ atlasdb proxy :+1:

iamdanfox (approved):
Given we've validated this at a heavy internal user (foundry/apollo), I think we can get this rolling and do trivial tweaks later

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
carterkozak: In favor of the general solution, I think QueuedChannel would still need to track state to guard `queuedTime` which is updated with `0` when the queue is bypassed, we avoid throwing zeros into it until it's first used.

carterkozak:
:+1:

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
carterkozak: Updated to `shouldRecordQueueMetrics` since it applies to the time and counter

PR #1009, opened 2020-10-07
Downgrade jackson to 2.11.1 to prevent consumers from hitting bugs

carterkozak:
==COMMIT_MSG==
Downgrade Jackson to 2.11.1 to prevent consumers taking buggy Jackson releases
==COMMIT_MSG==


carterkozak:
changelog?

PR #1012, opened 2020-10-08
Fix server_side_rate_limits simulation

carterkozak:
Dropwizard Meter rates are only updated every five seconds, resulting
in misleading results.
By multiplying our values by a thousand we can tick on each increment
without package hacks to access a guava limiter with a custom clock,
or reinventing the limiter ourselves. Ultimately rate limiters are
a poor choice over concurrency based limiting as there's no guarantee
that calls will take a uniform amount of time.

==COMMIT_MSG==
Fix server_side_rate_limits simulation
==COMMIT_MSG==


iamdanfox:
Oof, good spot! I've certainly found the actual graphs pretty non-sensical to look at, but I think this provides the reassurance that disabling concurrencylimiters results in 82.7% client-perceived success vs 99.7% with concurrency limiters enabled.

Somewhat concerning that only PIN_UNTIL_ERROR hits 100% though!!

iamdanfox:
> Ultimately rate limiters are a poor choice over concurrency based limiting as there's no guarantee that calls will take a uniform amount of time.

💯

iamdanfox (approved):

carterkozak:
We're likely decaying too fast to produce representative data, but I'm not sure the rate limit is worth spending more time on...

PR #1067, opened 2020-11-29
Reduce the client TLS session cache size

carterkozak:
## Before this PR
We've seen heap dumps with ~500mb of tls session cache despite an
expectation that connections are reused. These large session caches seem
to show that we are creating new sessions anyhow, so there's little
point in caching old data.

## After this PR
==COMMIT_MSG==
Reduce the client TLS session cache size
==COMMIT_MSG==

## Possible downsides?
if the cache is needed in some scenario, we may see a lower hit rate.

### DIFF
--- a/dialogue-clients/src/main/java/com/palantir/dialogue/clients/AugmentClientConfig.java
+++ b/dialogue-clients/src/main/java/com/palantir/dialogue/clients/AugmentClientConfig.java
@@ -27,6 +27,7 @@ import com.palantir.tritium.metrics.registry.SharedTaggedMetricRegistries;
 import com.palantir.tritium.metrics.registry.TaggedMetricRegistry;
 import java.security.Provider;
 import java.util.Optional;
+import javax.net.ssl.SSLContext;
 import org.immutables.value.Value;
 
 /**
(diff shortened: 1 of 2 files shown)


dialogue-clients/src/main/java/com/palantir/dialogue/clients/AugmentClientConfig.java
```diff
@@ -64,13 +65,19 @@ static ClientConfiguration getClientConf(ServiceConfiguration serviceConfig, Aug
         ClientConfiguration.Builder builder =
                 ClientConfiguration.builder().from(ClientConfigurations.of(serviceConfig));
 
+        SSLContext context = augment.securityProvider()
+                .map(provider -> SslSocketFactories.createSslContext(serviceConfig.security(), provider))
+                .orElseGet(() -> SslSocketFactories.createSslContext(serviceConfig.security()));
+        // Reduce the session cache size for clients. We expect TLS connections to be reused, thus the cache isn't
+        // terribly important.
+        context.getClientSessionContext().setSessionCacheSize(100);
```
iamdanfox, line 73: down from a default of 20480 by the looks of things

```java
    /**
     * Returns the size of the cache used for storing <code>SSLSession</code>
     * objects grouped under this <code>SSLSessionContext</code>.
     *
     * @implNote The JDK implementation returns the cache size as set by
     *           the {@code setSessionCacheSize} method, or if not set, the
     *           value of the {@systemProperty javax.net.ssl.sessionCacheSize}
     *           system property.  If neither is set, it returns a default
     *           value of 20480.
     *
     * @return size of the session cache; zero means there is no size limit.
     *
     * @see #setSessionCacheSize
     */
    public int getSessionCacheSize();
```

iamdanfox (approved):

PR #1071, opened 2020-12-02
Upgrade Jackson to 2.12.0 (from 2.11.1)

carterkozak:
==COMMIT_MSG==
Upgrade Jackson to 2.12.0 (from 2.11.1)
==COMMIT_MSG==


iamdanfox:
```
runTestCase[receiveOptionalIntegerExample({"value":"1234"}) -> should succeed false] - com.palantir.verification.AutoDeserializeTest
java.lang.AssertionError: [The test passed but the test case was ignored - remove this from ignored-test-cases.yml] 
Expecting Optional to contain a value but it was empty.
```

fawind (approved):

PR #1077, opened 2020-12-09
Fix: Downgrade jackson to 2.11.1

ferozco:
## Before this PR
We encountered an issue with de serializing doubles as BigDecimal which caused services to fail to accept valid requests 

## After this PR
==COMMIT_MSG==
Downgrade Jackson to 2.11.1
==COMMIT_MSG==


carterkozak:
Can we add a test case which fails on 2.12?

carterkozak:
👍 pending the test fixup/removal

PR #1087, opened 2021-01-12
Tag response metrics with endpoint

jkozlowski:
## After this PR
==COMMIT_MSG==

==COMMIT_MSG==

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/TimingEndpointChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/TimingEndpointChannel.java
@@ -48,12 +48,12 @@ final class TimingEndpointChannel implements EndpointChannel {
         this.responseTimer = metrics.response()
                 .channelName(channelName)
                 .serviceName(endpoint.serviceName())
-                .endpointName(endpoint.endpointName())
+                .endpoint(endpoint.endpointName())
                 .build();
         this.ioExceptionMeter = metrics.responseError()
                 .channelName(channelName)
                 .serviceName(endpoint.serviceName())
-                .endpointName(endpoint.endpointName())
+                .endpoint(endpoint.endpointName())
                 .reason("IOException")
                 .build();
     }
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/TimingEndpointChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/TimingEndpointChannel.java
@@ -48,10 +48,12 @@ final class TimingEndpointChannel implements EndpointChannel {
         this.responseTimer = metrics.response()
                 .channelName(channelName)
                 .serviceName(endpoint.serviceName())
+                .endpointName(endpoint.endpointName())
                 .build();
         this.ioExceptionMeter = metrics.responseError()
                 .channelName(channelName)
                 .serviceName(endpoint.serviceName())
+                .endpointName(endpoint.endpointName())
                 .reason("IOException")
                 .build();
     }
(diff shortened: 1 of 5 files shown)


dialogue-core/src/main/java/com/palantir/dialogue/core/TimingEndpointChannel.java
```diff
@@ -48,10 +48,12 @@
...
                 .channelName(channelName)
                 .serviceName(endpoint.serviceName())
+                .endpoint(endpoint.endpointName())
                 .build();
         this.ioExceptionMeter = metrics.responseError()
                 .channelName(channelName)
                 .serviceName(endpoint.serviceName())
+                .endpoint(endpoint.endpointName())
                 .reason("IOException")
                 .build();
```
carterkozak, line 58: I think we should report frequency of failures in some way without ignoring non-io exceptions. On a per-endpoint basis I'd like to know the ratio of client-perceived failures regardless(ish[1]) of the failure type.

1. We'd need to exclude cancelation exceptions since it's a failure type that the client requests.

simulation/src/main/java/com/palantir/dialogue/core/Benchmark.java
```diff
@@ -262,6 +262,7 @@ public void onFailure(Throwable throwable) {
                                     .response()
                                     .channelName(SimulationUtils.CHANNEL_NAME)
                                     .serviceName(SimulationUtils.SERVICE_NAME)
+                                    .endpoint(DEFAULT_ENDPOINT.endpointName())
```
jkozlowski, line 265: Ok, it looks like there are tests that have multiple endpoints. So need to figure out how to get the total here.

jkozlowski: Options I can see: @carterkozak 

1. Looks like we only look at mean, I can calculate that.
2. Make the test files be endpoint aware?
3. (For now) I can just fudge the TimingChannel to not respect the endpoint which should make the metrics come out as if there wasn't an endpoint tag.

carterkozak: I defer to your judgement -- In some places we've added benchmark-specific metrics, so that's another option if it makes things easier.

jkozlowski: Ok what I have works, so I think that's fine for now. We have unit tests in TimingChannel that actually test the metrics; I'd maybe consider just returning the mean on the BenchmarkResult, instead of a whole metric snapshot, but not too fussed.

dialogue-core/src/main/java/com/palantir/dialogue/core/TimingEndpointChannel.java
jkozlowski: Do you want to add more exceptions in there as reason, or just a new counter maybe that counts everything together, except for the cancellation exceptions?

jkozlowski: basically, I don't have context on what is an exhaustive list of exceptions if we want to list them (can probably figure it out, but if you can gimme a pointer that would be appreciated); could just look at the exception type, but then I'd need a cache for metric names keyed by that, which feels overkill; why do you think knowing per-reason type is useful?

carterkozak: I don't have a strong preference about the reason, I think we want to get a sense for the calls that failed unexpectedly. ServiceException/RemoteException are a little bit murky, I'm not sure how we want to measure those because it's possible infra changes can cause them, but in most cases they're the result of misuse. Happy to defer any changes around those to another PR :-)

jkozlowski: Alright, let's merge as is and chat when you're online.

jkozlowski:
Dang you changelog bot, likely doesn't like my fork PR. Gonna reraise to see if that helps.

PR #1088, opened 2021-01-13
Improve response metrics

jkozlowski:
## Before this PR
Request and error metrics accumulate all requests together, regardless of endpoint.

## After this PR
==COMMIT_MSG==
Response metrics are tagged by endpoint and filtered into 3 different statuses.
==COMMIT_MSG==


dialogue-core/src/main/java/com/palantir/dialogue/core/TimingEndpointChannel.java
```diff
@@ -68,24 +79,39 @@ static EndpointChannel create(Config cf, EndpointChannel delegate, Endpoint endp
 
         return DialogueFutures.addDirectCallback(response, new FutureCallback<Response>() {
             @Override
-            public void onSuccess(Response _result) {
-                updateResponseTimer();
+            public void onSuccess(Response response) {
```
jkozlowski, line 81: The code here mostly simplifies/replicates the retry decisions in RetryingChannel; I think it's ok to start this way, but once we determine that these metrics are the right ones, we could either move this timing down to that channel, or possibly have a field on the response that RetryingChannel sets?

dialogue-core/src/main/metrics/dialogue-core-metrics.yml
```diff
@@ -8,12 +8,21 @@ namespaces:
     metrics:
       response:
         type: timer
-        tags: [channel-name, service-name]
-        docs: Request time, note that this does not include time spent reading the response body.
-      response.error:
-        type: meter
-        tags: [channel-name, service-name, reason]
-        docs: Rate of errors received by reason and service-name. Currently only errors with reason `IOException` are reported.
+        tags: [channel-name, service-name, endpoint, status]
```
jkozlowski, line 11: Happy to bikeshed the names here

dialogue-core/src/main/metrics/dialogue-core-metrics.yml
```diff
@@ -8,12 +8,21 @@ namespaces:
...
+            which we successfully routed, maybe retried and returned to the user.
+            This does not mean that the ultimate user might not see an exception, e.g. 404, 401 will still potentially
+            result in a user visible failure.
+          * preventable_failure:
+            - QoS failures (429, 308, 503)
+            - 500 requests for endpoints that are retryable.
+            - IO related failures that we retry: e.g. UnknownHostException, connect timeouts, SSLException.
+          * other_failure: things we can't do anything about, e.g.
+            - 500 requests which we cannot retry
+            - read/write timeouts.
```
carterkozak: I'm not sure about categorizing `preventable_failures` differently from `other_failures`, we may get diminishing returns for the metric-name fanout at that point.

dialogue-core/src/main/metrics/dialogue-core-metrics.yml
```diff
@@ -8,12 +8,21 @@ namespaces:
...
-        type: meter
-        tags: [channel-name, service-name, reason]
-        docs: Rate of errors received by reason and service-name. Currently only errors with reason `IOException` are reported.
+        tags: [channel-name, service-name, endpoint, status]
+        docs: |
+          Request time split by status. These metrics and their definitions are mostly useful to dialogue developers,
+          to measure the success of routing and retry strategies. Possible status values are:
+          * success: always excludes time spent reading the response body. This classifies requests,
+            which we successfully routed, maybe retried and returned to the user.
+            This does not mean that the ultimate user might not see an exception, e.g. 404, 401 will still potentially
```
carterkozak: I think we should filter out non-2xx responses from this case to avoid poisoning the data -- if needs be we can add a `success-with-errors` type tag later on. What do you think?

dialogue-core/src/main/metrics/dialogue-core-metrics.yml
carterkozak: I like `status`.

dialogue-core/src/main/metrics/dialogue-core-metrics.yml
jkozlowski: Alright so we want literally:

1. success -> 2xx
2. failure -> QoS, 500, IO Exception

Everything else filter?

simulation/src/main/java/com/palantir/dialogue/core/BenchmarkTimingEndpointChannel.java
```diff
@@ -0,0 +1,74 @@
...
+
+            @Override
+            public void onFailure(Throwable _throwable) {
+                updateResponseTimer();
+            }
+
+            private void updateResponseTimer() {
+                responseTimer.update(ticker.read() - beforeNanos, TimeUnit.NANOSECONDS);
+            }
+        });
```
carterkozak: style (no action required):

I think this method can be simplified to use a listener (Runnable) because the action is run upon completion regardless of the result:

```java
return DialogueFutures.addDirectListener(
    delegate.execute(request),
    responseTimer.time()::stop);
```

dialogue-core/src/main/metrics/dialogue-core-metrics.yml
jkozlowski: Done

dialogue-core/src/main/metrics/dialogue-core-metrics.yml
jkozlowski: Done

simulation/src/main/java/com/palantir/dialogue/core/BenchmarkTimingEndpointChannel.java
jkozlowski: Done. Kept the ticker, cause I assume that's what allows the benchmark to mock time.

dialogue-core/src/main/java/com/palantir/dialogue/core/Responses.java
```diff
@@ -44,6 +44,10 @@ static boolean isInternalServerError(Response response) {
         return response.code() == 500;
     }
 
+    static boolean isClientSuccess(Response response) {
```
carterkozak: pedantic nit:
```suggestion
    static boolean isSuccess(Response response) {
```

dialogue-core/src/main/java/com/palantir/dialogue/core/TimingEndpointChannel.java
```diff
@@ -68,20 +70,23 @@ static EndpointChannel create(Config cf, EndpointChannel delegate, Endpoint endp
...
+                }
             }
 
             @Override
             public void onFailure(Throwable throwable) {
-                updateResponseTimer();
                 if (throwable instanceof IOException) {
-                    ioExceptionMeter.mark();
+                    updateTimer(failureTimer);
                 }
```
carterkozak, line 100: Thoughts on adding an else here safe-logging the exception class name at `info` level? That way we can verify we aren't missing non-ioexceptions that should be counted toward failure.

carterkozak (approved):
lgtm, thanks!

dialogue-core/src/main/java/com/palantir/dialogue/core/TimingEndpointChannel.java
jkozlowski: are non IOExceptions even allowed here? I will add the logging, do you think if should be rate limited, in case we are actually missing a lot of stuff here?

simulation/src/main/java/com/palantir/dialogue/core/BenchmarkTimingEndpointChannel.java
carterkozak: I think we pass the same ticker to our metric-registry so either way is probably fine -- avoiding the context allocation as you've written this likely makes the simulations run slightly more efficiently.

dialogue-core/src/main/java/com/palantir/dialogue/core/TimingEndpointChannel.java
carterkozak: > are non IOExceptions even allowed here?

That's what I'd like to find out :-)

> do you think if should be rate limited, in case we are actually missing a lot of stuff here?

I think logging the exception type rather than the stack trace cuts down enough potential cruft that rate limiting shouldn't be necessary. Any exception thrown here will be logged elsewhere, so +1 log line shouldn't be an issue, but I've been wrong before.

jkozlowski: I'll add it to log once a second, if it turns up anything after 2 weeks, we can up it or just drive it down.

PR #1090, opened 2021-01-15
Add CallingThreadExecutor to allow for running blocking calls on the …

jkozlowski:
…running thread.

## After this PR
==COMMIT_MSG==

==COMMIT_MSG==


dialogue-target/src/main/java/com/palantir/dialogue/Request.java
```diff
@@ -82,6 +83,10 @@ private Request(Builder builder) {
         return body;
     }
 
+    public Optional<CallingThreadExecutor> getCallingThreadExecutor() {
```
jkozlowski: A bit sucky this is here, but it's good enough. The problem is that queuing is cooperative, so when we finish a request, we use that thread to schedule the next request in the queue to run. This means we need to be able to then jump back to the calling thread for each request.

dialogue-blocking-channels/src/main/java/com/palantir/dialogue/blocking/BlockingChannelAdapter.java
```diff
@@ -80,7 +80,12 @@ private BlockingChannelAdapter() {}
             BlockingChannelAdapterTask runnable =
                     new BlockingChannelAdapterTask(delegate, endpoint, request, settableFuture);
             try {
-                Future<?> future = executor.submit(runnable);
+                Future<?> future;
```
jkozlowski, line 85: I know there's the backoff scheduler, but I think that can be shared, it's just 1 thread?

dialogue-target/src/main/java/com/palantir/dialogue/DefaultCallingThreadExecutor.java
```diff
@@ -0,0 +1,129 @@
...
+
+    private static final Logger log = LoggerFactory.getLogger(DefaultCallingThreadExecutor.class);
+    private final long threadId = Thread.currentThread().getId();
+    private final Queue queue = new Queue();
+
+    /** Notification when main future completes. */
+    private final Runnable notifier = queue::poison;
+
+    @Override
+    public synchronized Future<?> submit(Runnable task) {
```
carterkozak: The Queue handles concurrency
```suggestion
    public Future<?> submit(Runnable task) {
```

dialogue-target/src/main/java/com/palantir/dialogue/DefaultCallingThreadExecutor.java
```diff
@@ -0,0 +1,129 @@
...
+                    public void onSuccess(Object _result) {
+                        queue.submit(notifier);
+                    }
+
+                    @Override
+                    @SuppressWarnings("FutureReturnValueIgnored")
+                    public void onFailure(Throwable _throwable) {
+                        queue.submit(notifier);
+                    }
+                },
```
carterkozak: This callback can be a listener (Futures.addCallback converts the callback object into a Listener runnable which does a `get` on the completed future to call onSuccess/onFailure).
The Listener object might as well be created when DefaultCallingThreadExecutor is initialized, much like the notifier rather than per-invocation.

dialogue-target/src/main/java/com/palantir/dialogue/DefaultCallingThreadExecutor.java
```diff
@@ -0,0 +1,129 @@
...
+
+        RunnableFuture<?> toRun;
+        while ((toRun = queue.getWork()) != null) {
+            toRun.run();
+        }
+    }
+
+    private static final class Queue {
+        private boolean poisoned = false;
+        private final BlockingDeque<RunnableFuture<?>> queue = new LinkedBlockingDeque<>();
```
carterkozak: I think we can use a BlockingQueue rather than BlockingDequeue based on our access pattern.

Unsure how LinkedBlockingDequeue and LinkedBlockingQueue performance differs (no change requested, just thinking. Also wonder about LinkedTransferQueue, but that one spins before parking and eats some CPU)

In terms of state tracking and concurrency, I think we can remove synchronization entirely if we replace the `poisoned` access with a volatile boolean for `submit`, but add a sentinel value to the queue to signal `getWork` callers so they never read the `poisoned` field.

dialogue-target/src/main/java/com/palantir/dialogue/DefaultCallingThreadExecutor.java
jkozlowski: These executors are created per Request, so only ever one listener will be created, but yeah should probably be just the whole thing.

jkozlowski: Done

dialogue-target/src/main/java/com/palantir/dialogue/DefaultCallingThreadExecutor.java
jkozlowski: Switched to BlockingQueue, I didn't even think twice about that. There's also ArrayListBlockingQueue: frankly that is probably the best for us, there shouldn't be more than 1 elements at any given time here right? Otherwise it's a bug. We can discuss

jkozlowski: re: removing synchronization: consider this ordering:


There's a blocking task in the queue to run the request.

Thread 1:
* calls #submit, checks volatile boolean, sleeps.

Thread 2:
* calls #executeQueue with the future.
* runs the task that is already in the queue, that completes the future, adds the notifier to the queue (or sentinel task).
* runs the notifier task, that sets the volatile boolean.
* there's no more tasks, the future is finished, so it exits.

Thread 1:
* wakes up, adds the task.

We have a hanging future?

Arguably, that is not really possible with the way the rest of dialogue works, but I'd rather write this in a bulletproof way.

jkozlowski: Ultimately, there should be zero contention for this lock, at least in the current dialogue implementation: is it so bad to write it this way?

If we add some sort of racing/speculative execution of requests, then by definition we'll need to abandon this approach.

carterkozak: Ya, the synchronization is probably fine, it should just be a cas to acquire the lock.

dialogue-target/src/main/java/com/palantir/dialogue/DefaultCallingThreadExecutor.java
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    @Override
+    public void executeQueue(ListenableFuture<?> await) {
+        Preconditions.checkState(Thread.currentThread().getId() == threadId, "Executing queue on different thread");
+        Futures.addCallback(await, callback, DialogueFutures.safeDirectExecutor());
+
+        RunnableFuture<?> toRun;
+        while ((toRun = queue.getWork()) != null) {
+            toRun.run();
```
jkozlowski: from sync: this needs to cancel the passed in future on an interrupt, otherwise we might break things.

jkozlowski:
#1107

PR #1095, opened 2021-01-19
Fix Response leak in `EndToEndBenchmark.rawApacheBlocking`

carterkozak:
non-production code change.

==COMMIT_MSG==
Fix Response leak in `EndToEndBenchmark.rawApacheBlocking`
==COMMIT_MSG==

### DIFF
--- a/dialogue-jmh/src/main/java/com/palantir/dialogue/core/EndToEndBenchmark.java
+++ b/dialogue-jmh/src/main/java/com/palantir/dialogue/core/EndToEndBenchmark.java
@@ -160,7 +160,7 @@ public class EndToEndBenchmark {
     @Threads(4)
     @Benchmark
     public void rawApacheBlocking() {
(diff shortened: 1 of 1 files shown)


dialogue-jmh/src/main/java/com/palantir/dialogue/core/EndToEndBenchmark.java
```diff
@@ -160,7 +160,7 @@ public void setZeroNetworkDialogue() {
     @Threads(4)
     @Benchmark
     public void rawApacheBlocking() {
-        clientUtils.block(apacheChannel.execute(TestEndpoint.GET, request));
+        clientUtils.block(apacheChannel.execute(TestEndpoint.GET, request)).close();
```
jkozlowski, line 163: try-with-resources?

jkozlowski (approved):

PR #1105, opened 2021-01-27
Implement Clients#callBlocking.

jkozlowski:
## Before this PR

Blocking code calls through async code and calls Clients#block on the returned future. Stepping stone towards https://github.com/palantir/dialogue/pull/1090.

## After this PR
==COMMIT_MSG==
Implement new API for blocking requests.
==COMMIT_MSG==


carterkozak (approved):

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/DefaultClients.java
```diff
@@ -67,6 +67,11 @@
         return DialogueFutures.transform(response, deserializer::deserialize);
     }
 
+    @Override
+    public <T> T callBlocking(EndpointChannel channel, Request request, Deserializer<T> deserializer) {
```
jkozlowski, line 71: @carterkozak / @CRogers for capturing the discussion: we're worried that if someone has the new version of dialogue-target, but old version of dialogue-serde, they could get a runtime failure.

PR #1106, opened 2021-01-28
Clients.callBlocking provides a default implementation

carterkozak:
This provides safety in case a newer target is used than implementation,
which can occur when a project pulls in generated bindings with a
newer target dep.
Long term we should consider using constraints to avoid version
slip between jars produced by the same repository.

==COMMIT_MSG==
Clients.callBlocking provides a default implementation for cross-version compatibility
==COMMIT_MSG==


jkozlowski (approved):
Let's unblock the rest of the work, and we can come back to this. It's scary

PR #1107, opened 2021-01-29
Introduce attachments to Request API.

jkozlowski:
## Before this PR

There is not way to attach extra internal state that the channels can share and pass along the request.

## After this PR
==COMMIT_MSG==
Request.attachments() allows channels to attach additional state.
==COMMIT_MSG==

### Why is this mutable you ask?

Well, mutable is not-in vogue, but with the current design of channels, it might be difficult to associate any sort of mutable state per-request. For example, say when we retry, we wanted to make sure we first tried on hosts that we didn't try before, we couldn't really do that with mutable RequestAttachments, as channels get passed a request, and return a response: if I channel makes changes to the request, this will not be visible to upstream channels.


dialogue-target/src/main/java/com/palantir/dialogue/Request.java
```diff
@@ -139,16 +148,21 @@ public static Builder builder() {
 
         private Optional<RequestBody> body = Optional.empty();
 
+        private RequestAttachments attachments;
+
         private int mutableCollectionsBitSet = 0;
 
-        private Builder() {}
+        private Builder() {
+            attachments = DefaultRequestAttachments.create();
```
carterkozak: nit: let's keep this null and initialize in `build()`. That way `Builder.from(Request)` doesn't create objects to throw away.

dialogue-target/src/main/java/com/palantir/dialogue/RequestAttachments.java
```diff
@@ -0,0 +1,26 @@
...
+
+package com.palantir.dialogue;
+
+import javax.annotation.Nullable;
+
+public interface RequestAttachments {
+
+    <V> V put(RequestAttachmentKey<V> key, V value);
+
+    <V> V getOrDefault(RequestAttachmentKey<V> key, @Nullable V defaultValue);
```
carterkozak: Should both methods be annotated `@Nullable`?

dialogue-target/src/main/java/com/palantir/dialogue/Request.java
jkozlowski: ah, misread the code, for some reason I took from as an overloaded constructor.

jkozlowski: DOne

dialogue-target/src/main/java/com/palantir/dialogue/RequestAttachments.java
jkozlowski: I don't think we want to support putting null, unless you want to reuse this api for remove?

carterkozak: If there's no existing value, won't `put` return null?

jkozlowski: oh I thought you wanted to annotate the value; yeah I should annotate the return, you're right.

jkozlowski: Done

dialogue-target/src/main/java/com/palantir/dialogue/Request.java
jkozlowski: Null analysis was failing, so I had to initialize in the Request constructor.

jkozlowski:
Need to loose the bogus APIs, revapi is getting weird: https://github.com/palantir/dialogue/pull/1107

dialogue-target/src/main/java/com/palantir/dialogue/RequestAttachmentKey.java
```diff
@@ -0,0 +1,37 @@
...
+
+    private RequestAttachmentKey(Class<V> valueClazz) {
+        this.valueClazz = valueClazz;
+    }
+
+    void checkIsInstance(V value) {
+        assert valueClazz.isInstance(value) : "Value not instance of class " + valueClazz;
+    }
+
+    public static <T> RequestAttachmentKey<T> create(Class<T> valueClazz) {
```
carterkozak: Do we want `Class<? super T>` here for flexibility with generics?

jkozlowski: I couldn't get that to work, but maybe it was just late... Lemme try this again

jkozlowski: ```java: incompatible types: java.lang.Class<capture#1 of ? super T> cannot be converted to java.lang.Class<T>```

jkozlowski: Alright, I think we got it.

dialogue-target/src/main/java/com/palantir/dialogue/Request.java
```diff
@@ -82,6 +85,14 @@ private Request(Builder builder) {
         return body;
     }
 
+    /**
+     * The mutable request attachments for this request. Attachments will be propagated when this request is mutated
+     * through the builder
+     */
+    public RequestAttachments attachments() {
```
carterkozak, line 94: May want to add a `@Beta` annotation here and on the `RequestAttachments` type for now.

carterkozak (approved):

dialogue-target/src/main/java/com/palantir/dialogue/Request.java
jkozlowski: Ack

PR #1109, opened 2021-02-01
Clients#callBlocking now uses the calling thread for network calls.

jkozlowski:
## Before this PR

* TODO: add stacktrace from a failure when this feature is enabled, so that we can demonstrate that we now get useable stacktraces!

## After this PR
==COMMIT_MSG==

==COMMIT_MSG==


dialogue-blocking-channels/src/main/java/com/palantir/dialogue/blocking/DefaultCallingThreadExecutor.java
```diff
@@ -0,0 +1,150 @@
...
+            @Override
+            @SuppressWarnings("FutureReturnValueIgnored")
+            public void onSuccess(Object _result) {
+                queue.submit(notifier);
+            }
+
+            @Override
+            @SuppressWarnings("FutureReturnValueIgnored")
+            public void onFailure(Throwable _throwable) {
+                queue.submit(notifier);
```
carterkozak: nit: this can be a listener because both success and failure do the same thing.

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/UseCallingThreadExecutor.java
```diff
@@ -0,0 +1,50 @@
...
+import com.palantir.logsafe.Preconditions;
+import com.palantir.random.SafeThreadLocalRandom;
+
+/**
+ * Package private API for enabling
+ * {@link com.palantir.dialogue.blocking.CallingThreadExecutor} optimization.
+ */
+final class UseCallingThreadExecutor {
+
+    private static final AtomicDouble PROBABILITY = new AtomicDouble(0.0);
```
carterkozak: nit: volatile since we're not doing any sort of CAS where we need the padding. float type should be sufficient, I don't imagine we expect a ton of decimal points.

dialogue-blocking-channels/src/main/java/com/palantir/dialogue/blocking/DefaultCallingThreadExecutor.java
```diff
@@ -0,0 +1,150 @@
...
+        }
+
+        public RunnableFuture<?> poll() {
+            return queue.poll();
+        }
+
+        private synchronized void checkNotPoisoned() {
+            if (poisoned) {
+                log.info("Submitted task after queue is closed");
+                throw new RejectedExecutionException("Queue closed");
```
carterkozak, line 108: Is this always logged after an interruption? abortQueue calls queue.poison followed by future.cancel, which sets the future state to failed/canceled and executes the listener to submit the notifier to the poisoned queue?

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/UseCallingThreadExecutor.java
```diff
@@ -0,0 +1,50 @@
...
+
+    private UseCallingThreadExecutor() {}
+
+    static boolean shouldUseCallingThreadExecutor() {
+        return SafeThreadLocalRandom.get().nextDouble() < PROBABILITY.doubleValue();
```
carterkozak: Might want to do one volatile read and compare with 1 and zero prior to random number generation (much like we do in tracing). I imagine zero and one will be common values.

dialogue-blocking-channels/src/main/java/com/palantir/dialogue/blocking/DefaultCallingThreadExecutor.java
jkozlowski: Indeed!

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/UseCallingThreadExecutor.java
jkozlowski: done

dialogue-blocking-channels/src/main/java/com/palantir/dialogue/blocking/DefaultCallingThreadExecutor.java
jkozlowski: Good point, I'm not that fussed, but should probably solve that somehow.

jkozlowski: Done

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/UseCallingThreadExecutor.java
jkozlowski: Done

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/UseCallingThreadExecutor.java
```diff
@@ -0,0 +1,58 @@
...
+    private UseCallingThreadExecutor() {}
+
+    static boolean shouldUseCallingThreadExecutor() {
+        float currentProbability = probability;
+        if (currentProbability == DISABLED) {
+            return false;
+        } else if (currentProbability == ENABLED) {
+            return true;
+        }
+        return SafeThreadLocalRandom.get().nextDouble() < currentProbability;
```
carterkozak:
```suggestion
        return SafeThreadLocalRandom.get().nextFloat() < currentProbability;
```

dialogue-blocking-channels/src/main/java/com/palantir/dialogue/blocking/CallingThreadExecutor.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+package com.palantir.dialogue.blocking;
+
+import com.google.common.util.concurrent.ListenableFuture;
+import com.palantir.dialogue.Request;
+
+public interface CallingThreadExecutor {
+
+    ListenableFuture<?> submit(Runnable task);
```
carterkozak: I think we only use the result to cancel the submitted task, we can probably get away with `Future<?>` or even a simpler disposer/runnable since the value is only used to provide input (cancel with interruption) and not output (`future.get()`).

```suggestion
    Future<?> submit(Runnable task);
```

dialogue-blocking-channels/src/main/java/com/palantir/dialogue/blocking/DefaultCallingThreadExecutor.java
```diff
@@ -0,0 +1,143 @@
...
+            await.cancel(DO_NOT_INTERRUPT);
+        }
+    }
+
+    private void abortQueue() {
+        queue.poison();
+
+        RunnableFuture<?> toRun;
+        while ((toRun = queue.poll()) != null) {
+            toRun.cancel(DO_NOT_INTERRUPT);
```
carterkozak: Note: I'm not sure this does anything because we wrap runnables with FutureTask, and never use the result except to call `cancel()`. Leaving this as is makes the future returned from SameThreadExecutor make a little more sense if we change how we use it, but thar be dragons.

No change requested, just a case that took me a bit to reason through.

carterkozak (commented):
Generally looks good to me, I'm going to make a final pass over this tomorrow morning and rerun our internal benchmarks.

dialogue-blocking-channels/src/main/java/com/palantir/dialogue/blocking/CallingThreadExecutor.java
jkozlowski: I had it as future before, but writing tests required some gymnastics, since I'm not that familiar with the CompletableFuture API as much. Since we use ListenableFuture everywhere else, I'd rather we keep it as is.

dialogue-blocking-channels/src/main/java/com/palantir/dialogue/blocking/DefaultCallingThreadExecutor.java
jkozlowski: Ah, I think I understand your previous concern a little bit more now, but still a bit hazy. The futures returned from #submit get transformed further down the line right? so it's very important we;re able to cancel them, so we don't leave resources that are tied to these futures (like concurrency limiters) leaked?

carterkozak: I think the futures returned from our executor are exclusively used for cancellation, and never transformed. The adapter submits a runnable which completes a SettableFuture, otherwise we risk leaking responses in cancellation races. There's unfortunately a lot of code in DialogueFutures to prevent resource leaks, which we may be able to replace with CloseableFutures from recent guava versions which didn't exist when this was written.

dialogue-blocking-channels/src/main/java/com/palantir/dialogue/blocking/BlockingChannelAdapter.java
```diff
@@ -67,11 +69,11 @@ private BlockingChannelAdapter() {}
     private static final class BlockingChannelAdapterChannel implements Channel {
 
         private final BlockingChannel delegate;
-        private final ExecutorService executor;
+        private final ListeningExecutorService executor;
 
         BlockingChannelAdapterChannel(BlockingChannel delegate, ExecutorService executor) {
             this.delegate = delegate;
-            this.executor = executor;
+            this.executor = MoreExecutors.listeningDecorator(executor);
```
carterkozak: Isn't this still unnecessary?

jkozlowski: Indeed, fixed.

carterkozak:
👍

PR #1112, opened 2021-02-02
Simplifications from a read-through

carterkozak:
==COMMIT_MSG==
Possible simplifications from a read-through
==COMMIT_MSG==

### DIFF
--- a/dialogue-blocking-channels/src/main/java/com/palantir/dialogue/blocking/DefaultCallingThreadExecutor.java
+++ b/dialogue-blocking-channels/src/main/java/com/palantir/dialogue/blocking/DefaultCallingThreadExecutor.java
@@ -18,14 +18,12 @@ package com.palantir.dialogue.blocking;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.util.concurrent.ListenableFuture;
-import com.google.common.util.concurrent.ListenableFutureTask;
 import com.palantir.dialogue.RequestAttachmentKey;
 import com.palantir.dialogue.futures.DialogueFutures;
 import com.palantir.logsafe.Preconditions;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.RejectedExecutionException;
-import java.util.concurrent.RunnableFuture;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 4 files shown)


dialogue-blocking-channels/src/main/java/com/palantir/dialogue/blocking/DefaultCallingThreadExecutor.java
```diff
@@ -54,49 +52,44 @@
...
+                    toRun.run();
+                } catch (Throwable t) {
+                    log.error("Failed to execute runnable {}", toRun, t);
+                }
             }
         } catch (InterruptedException e) {
             Thread.currentThread().interrupt();
 
-            abortQueue();
+            queue.poison();
```
jkozlowski, line 70: Why do you think queue aborts are not necessary? You reckon aborting the main future will propagate everything correctly?

jkozlowski: I think that's right, I haven't thought about that.

dialogue-blocking-channels/src/main/java/com/palantir/dialogue/blocking/DefaultCallingThreadExecutor.java
```diff
@@ -54,49 +52,44 @@
...
+        DialogueFutures.addDirectListener(await, () -> queue.submitNotifier(queue::poison));
         try {
-            RunnableFuture<?> toRun;
+            Runnable toRun;
             while ((toRun = queue.getWork()) != null) {
-                toRun.run();
+                try {
+                    toRun.run();
+                } catch (Throwable t) {
+                    log.error("Failed to execute runnable {}", toRun, t);
```
jkozlowski: This should basically not happen, so I think we can have this here.

dialogue-blocking-channels/src/test/java/com/palantir/dialogue/blocking/DefaultCallingThreadExecutorTest.java
```diff
@@ -78,111 +81,140 @@ public void testThrowsWhenPoisoned() {
...
+                    return true;
+                } catch (ExecutionException e) {
+                    throw new UncheckedExecutionException(e);
+                }
+            });
 
-        assertThat(Futures.getUnchecked(queueExecuted)).isTrue();
+            assertThat(Futures.getUnchecked(queueExecuted)).isTrue();
+        } finally {
+            shutdown(queueExecutor);
```
jkozlowski, line 102: I've never bothered about shutting these in tests, but let's do it.

dialogue-blocking-channels/src/main/java/com/palantir/dialogue/blocking/DefaultCallingThreadExecutor.java
carterkozak: Exactly, we submit a task which uses a SettableFuture as output rather than throwing exceptions outside of the Runnable.

jkozlowski (approved):

dialogue-blocking-channels/src/test/java/com/palantir/dialogue/blocking/DefaultCallingThreadExecutorTest.java
```diff
@@ -78,111 +79,135 @@ public void testThrowsWhenPoisoned() {
...
-                throw new UncheckedExecutionException(e);
-            }
-        });
+        try {
+            CallingThreadExecutor executorToUse =
+                    Futures.getUnchecked(queueExecutor.submit(DefaultCallingThreadExecutor::new));
+
+            ListenableFuture<Boolean> queueExecuted = queueExecutor.submit(() -> {
+                try {
+                    Thread.currentThread().interrupt();
```
jkozlowski, line 88: Ah, I totally missed this simpler way of doing this!

dialogue-blocking-channels/src/main/java/com/palantir/dialogue/blocking/DefaultCallingThreadExecutor.java
carterkozak: Calling cancel on the FutureTask instances is equivalent to ignoring the queue. We can get away with it here because we're on the only thread that can drain the queue. I'm planning to add comments here describing the constraints.

PR #1113, opened 2021-02-03
Enable CallingThreadExecutor for 1% of traffic. 

jkozlowski:
## After this PR
==COMMIT_MSG==
Enable CallingThreadExecutor for 1% of traffic. 
==COMMIT_MSG==


dialogue-serde/build.gradle
```diff
@@ -22,6 +25,7 @@ dependencies {
     implementation 'com.palantir.conjure.java:conjure-lib'
     implementation 'com.palantir.safe-logging:safe-logging'
     implementation 'jakarta.ws.rs:jakarta.ws.rs-api'
+    compileOnly 'org.immutables:value::annotations'
```
jkozlowski, line 28: I feel like metric-schema should add this somehow?

carterkozak: metric-schema doesn't use this, I think it may come from tritium though

jkozlowski: oh interesting

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/UseCallingThreadExecutor.java
```diff
@@ -16,29 +16,58 @@
...
+    private final Consumer<RuntimeException> failureConsumer;
+    private final DialogueFeatureFlagsMetrics metrics;
+    private volatile float probability = DEFAULT_PROBABILITY;
+
+    @VisibleForTesting
+    UseCallingThreadExecutor(Random random, TaggedMetricRegistry taggedMetricRegistry) {
+        this.random = random;
+        this.metrics = DialogueFeatureFlagsMetrics.of(taggedMetricRegistry);
+        failureConsumer = runtimeException -> {
+            log.info("Unexpected exception from calling thread executor. Disabling this feature", runtimeException);
```
carterkozak: I'd call this an error, wouldn't you?

I'm not sure how valuable this check is since it's not the most interesting way this could fail (for some definition of interesting), and I think there's a race condition where the caller is interrupted as a retry (or queued request) is submitted which results in RejectedExecutionException, but the system is still working as expected.

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/UseCallingThreadExecutor.java
```diff
@@ -16,29 +16,58 @@
...
+            metrics.callingThreadExecutorFailure().mark();
```
carterkozak: I tend to prefer hoisting the `Meter` object to a field so that it's loaded from the TaggedMetricRegistry map once. We could optimize the generated `*Metrics` classes, but that would only apply to metrics with no tags.
This case is expected to be rare, no change requested.

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/UseCallingThreadExecutor.java
jkozlowski: aye, going to do as we discussed.

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/UseCallingThreadExecutor.java
jkozlowski: Done

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/UseCallingThreadExecutor.java
```diff
@@ -49,10 +79,43 @@ static boolean shouldUseCallingThreadExecutor() {
...
+    }
+
+    @VisibleForTesting
+    float getCurrentProbability() {
+        return probability;
+    }
+
+    static UseCallingThreadExecutor get() {
+        return INSTANCE;
+    }
```
carterkozak: This `get` method looks a little odd, I'd probably try to avoid changing the code in `callBlocking`. Perhaps we split the implementation into a package-private-for-testing class which we can instantiate with custom values, and leave this class in place as a static adapter to a singleton instance of that class?

carterkozak (approved):
lgtm, thanks!

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/UseCallingThreadExecutor.java
jkozlowski: Done

carterkozak:
oof, hc.apache.org isn't responding, so our builds are red...

carterkozak:
should be back up, I'll make sure this gets rolled out in a WC release in the next hour or so.

fawind (approved):

PR #1135, opened 2021-02-26
Annotation based Dialogue clients API.

jkozlowski:
## Before this PR

Non-conjure clients have to be written manually.

## After this PR

Non-conjure clients have to be written manually. But...

==COMMIT_MSG==
Annotation based Dialogue client API + manual test implementation.
==COMMIT_MSG==


dialogue-annotations-processor/src/test/java/com/palantir/myservice/service/MyService.java
```diff
@@ -0,0 +1,68 @@
...
+    @Request(method = HttpMethod.POST, path = "/greet")
+    String greet(
+            // Idea is that Json.class provides the encoder to transform the
+            // greeting string into application/json. Another hand-written
+            // CustomJson.class may provide a custom ObjectMapper, but we
+            // should make it relatively easy to implement this sort of thing.
+            @Request.Body String greeting);
+
+    // Support blocking and listenablefuture based on the return type
+    @Request(method = HttpMethod.GET, path = "/greeting", accept = CustomStringDeserializer.class)
```
carterkozak, line 40: I don't think we'll want to annotate directly with deserializer classes, rather with some kind of DeserializerFactory. This way we get some type safety while retaining the ability to reuse the jackson ObjectMapper using base-type Object.

This differs slightly from conjure, were we have a factory for all our types which we know are json serializable, in this case we want to safely support custom per-type deserializers.

```java
@Request(method = HttpMethod.GET, path = "/greeting", accept = CustomStringDeserializer.class)
String getGreetingAsync();

interface DeserializerFactory<T> {
    <U extends T> Deserializer<U> deserializerFor(TypeMarker<U> type);
}

class CustomStringDeserializer implements DeserializerFactory<String> {
   Deserializer<String> deserializerFor(TypeMarker<String> type);
}
```

dialogue-annotations-processor/src/test/java/com/palantir/myservice/service/MyService.java
```diff
@@ -0,0 +1,68 @@
...
+import com.google.common.util.concurrent.ListenableFuture;
+import com.palantir.dialogue.HttpMethod;
+import com.palantir.dialogue.RequestBody;
+import com.palantir.dialogue.Response;
+import com.palantir.dialogue.annotations.DialogueClient;
+import com.palantir.dialogue.annotations.Request;
+import java.util.UUID;
+
+@DialogueClient
+public interface MyService {
```
ferozco, line 29: Out of curiosity, how is the interface expected to be used since its missing the static factory method typical of Dialogue interfaces? Will the client factory be aware of the annotations and the generated classes?

carterkozak: We'd discussed requiring the generated class ref `value` of the `@DialogueClient` annotation, and teaching the factory to follow links. I don't think we can get away with requiring folks to pass the generated class to the client factory because it's too easy to get wrong, and fails at runtime.

dialogue-annotations-processor/src/test/java/com/palantir/myservice/service/MyService.java
ferozco: @carterkozak I'm not sure I understand how users would implement the `DeserializerFactory<T>`. There is a fair bit of logic in https://github.com/palantir/dialogue/blob/develop/dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ConjureBodySerDe.java#L234-L265 for handling errors and various status codes and that users would conceivably want to take advantage of but would not necessarily have access to without the current clientFactory's `ConjureRuntime`

dialogue-annotations-processor/src/test/java/com/palantir/myservice/service/MyService.java
carterkozak: Very much open to suggestions if you have ideas :-)

dialogue-annotations-processor/src/test/java/com/palantir/myservice/service/MyService.java
carterkozak: Yep, we will need to provide some basic building blocks (which likely delegate to our default conjureruntime) to make custom encodings easier, however I don't want to prevent folks from using custom encodings that are capable of supporting non-conjure use-cases like mapping a 404 response to an empty optional.

dialogue-annotations-processor/src/test/java/com/palantir/myservice/service/MyService.java
ferozco: I think it would actually be kinda nice to expect users to do something like:

```java
wc.client(DialogueMyService.class, "my-service").get();
```

supposing we generated something like:

```java
interface DialogueMyService extends MyService {
  static DialogueMyService of(Channel channel, ConjureRuntime runtime) {
    ...
  }
}
```

I worry that if we allowed:
```java
wc.client(MyService.class, "my-service").get();
```

that it would get a bit confusing for users what was going on under the covers and that you could end up in a split state where sometimes the generated interface is used over the declared interface.

carterkozak: Right, that's exactly what I'm worried about because the type system isn't giving us any safety, and based on all our existing clients it feels like the right thing to do.

dialogue-annotations-processor/src/test/java/com/palantir/myservice/service/MyService.java
ferozco: I think that if you passed the clientFactories `ConjureRuntime` into the deserializerFactory then you could relatively cleanly implement what you want. I would stay wasay from using the default conjure runtime since it would prevent us from taking advantage of any additional configuration injected by WC

carterkozak: That's a good point. I like the idea of sharing customization and monitoring with the wc clientfactory, but we'll have to be careful not to couple too closely with conjure, otherwise we may inadvertently modify behavior of custom clients without realizing it. For example, the ConjureRuntime doesn't provide any guarantee that data will be encoded to JSON. We want to allow folks to bring their own ObjectMapper for custom type handling which isn't something I'd be thrilled about using with proper conjure dialogue clients.

dialogue-annotations-processor/src/test/java/com/palantir/myservice/service/MyService.java
jkozlowski: We'll codegen it eventually, right now just spitballing. Of course I forgot to add the "draft" PR thing... 🤦

jkozlowski: So if I understand it, your worries are:

1. People will sometimes pass the generated interface, and sometimes the interface they hand annotate.
2. If we pass in the hand-annotated one, it's possible they haven't actually run the generator, and so if witchcraft were to go looking for the generated class, it would not be found?

Doing:

```
MyService myService = wc.client(DialogueMyService.class, "my-service").get();
```

feels unnecessary to me and is not what we do for conjure generated ones: you pass in the DialogueClientBlocking or DialogueClientAsync right? So there's no asymmetry like is being proposed here.

Personally, I'd rather we actually make a gradle plugin or something lightweight that people apply, and that thing correctly sets up the annotation processor and the correct compileOnly dependency, so it's not like with immutables were most people forget about the compileOnly dependency and get weird warnings.

But maybe I'm misunderstanding the concern?

dialogue-annotations-processor/src/test/java/com/palantir/myservice/service/MyService.java
jkozlowski: Ok, I think I understand the concern: you'd like to be able to write one deserializer for Object that can handle anything it wants. So we need the TypeMarker so the information can be retained at runtime.

jkozlowski: I went for DeserializerFactory without the param, as I couldn't write a simple StdSerializer without it, but it might be that I suck at generics.

jkozlowski: Hmm, that doesn't quite work either. Let me see what I can do

carterkozak: I think it's the opposite case where we want the parameter: If I need to send an encoded `Foo` object to a remote server, I don't need to support serializing every type. In that case we'd want an error if I tried to serialize an `Object` or `Bar`, but work properly if I serialize a `Foo`.

jkozlowski: Yeah, I think the thing that was problematic was the ``<U extends T>``. I tried it without and it seems to work nicer, but I'm yet to try to implement the Json serializer/deserializer to see if that breaks something.

dialogue-annotations-processor/src/test/java/com/palantir/myservice/service/MyService.java
carterkozak: If we require the generated interface to be passed to the conjure factory, forgetting to do so and passing the raw interface (which is what we do for conjure-generated interfaces) will compile locally correctly, and fail at runtime. We generally want to fail at compilation rather than runtime.

So we need a way to map from `MyService.class` to `DialogueMyService.class` safely so we can teach the generator to look for factory methods on the latter. One of the easiest ways is to include a direct reference to the generated class from our `DialogueClient` annotation, this will both show the factory which interface to construct, as well as enforce at compile time that the processor has been applied, otherwise the DialogueMyService.class cannot be located.

Alternatively we could do something clever with a new service-loadable factory of the form:
```java
interface DialogueServiceFactory<MyService> {
  MyService create(EndpointChannelFactory channel, ConjureRuntime runtime);
}
```
I like the idea of avoiding the reflection hackery we currently use, but I think this solution could be overkill.

dialogue-annotations-processor/src/test/java/com/palantir/myservice/service/MyService.java
jkozlowski: The ``ConjureRuntime`` piece: I don't feel we want to pass that in, because it's got the ability to do actual RPC - so a Serializer/Deserializer should not have access to that? Maybe BodySerDe? But I don't see any customization of that in Witchcraft, so I can instantiate that directly in my Json implementation.

dialogue-annotations-processor/src/test/java/com/palantir/myservice/service/MyService.java
jkozlowski: We can do the "include direct reference to the generated class", but I feel that a small gradle plugin that sets up dependencies correctly, and then generates the class in the same package as the annotated class works? Then the Witchcraft factory can assume that it should just load the ``"Dialogue" + clazz.getFullyQualifiedClass()``.

The custom annotation works for me, I don't have a strong preference other than the workflow of "add plugin and write your custom interface" feels less work than "correctly setup dependencies -> write interface with annotations -> annotate the interface".

carterkozak: imo gradle plugins are a pain, and specific to gradle. They can break in their own odd set of ways, so I'm not convinced a plugin is the right way to tackle this problem. That takes us from setting up 2 depenedncies, to setting up 3 dependencies, one of which may set up the other two if it works properly ;-)

The more I think about this, the better I like a more generic solution that could simplify our existing conjure-generated clients as well. I think our magical static method lookups were a mistake, we don't need to further compound upon them especially given newer jres reflection restrictions.

jkozlowski: Alright, so:

1. DialogueMyService is generated by the annotation processor.
2. We annotate MyService with @DialogueClient(DialogueMyService.class)
3. DialogueMyService implements DialogueServiceFactory<MyService>
4. Clients use this: ``MyService myService = wc.client(MyService.class, "my-service").get();``

Did I miss something?

carterkozak: 1. :heavy_check_mark: DialogueMyService is generated by the annotation processor.
2. :heavy_check_mark: We annotate MyService with @DialogueClient(DialogueMyService.class)
3. :yellow_circle: DialogueMyService implements DialogueServiceFactory: This one is somewhat nebulous. Either our annotation points to a standard dialogue service, or we reimplement the way dialogue services are created. I'd like to improve how all dialogue services are created, so I'll try to prototype something this evening. Failing that, the annotaiton will point to the "concrete" dialogue interface which our client factory will follow and create.
4. :x: Clients use this: `MyService myService = wc.client(DialogueMyService.class, "my-service").get()`: This is dangerous because the type system allows folks to pass MyService.class without failing compilation. We should hide the generated code as much as possible at this point.

carterkozak: For initial prototyping we can use `factory.client(DialogueMyService.class, "my-service")` and improve the system in parallel, but I wouldn't want anyone to use it that way.

jkozlowski: https://github.com/palantir/dialogue/pull/1138

dialogue-annotations/src/main/java/com/palantir/dialogue/annotations/DeserializerFactory.java
```diff
@@ -0,0 +1,24 @@
...
+ */
+
+package com.palantir.dialogue.annotations;
+
+import com.palantir.dialogue.Deserializer;
+import com.palantir.dialogue.TypeMarker;
+
+public interface DeserializerFactory {
+    <T> Deserializer<T> deserializerFor(TypeMarker<T> type);
+}
```
carterkozak, line 24: I think we may want to bound the factory to reduce ambiguity for deserializers that apply to a single type, while allowing the default json deserializer to apply to Object:

```java
public interface DeserializerFactory<T> {
    <U extends T> Deserializer<U> deserializerFor(TypeMarker<U> type);
}
```
I think the same applies to serializers.

dialogue-annotations/src/main/java/com/palantir/dialogue/annotations/Json.java
```diff
@@ -0,0 +1,136 @@
...
+            };
+        }
+
+        @Override
+        public final <T> Deserializer<T> deserializer(TypeMarker<T> type) {
+            ObjectReader reader = mapper.readerFor(mapper.constructType(type.getType()));
+            return input -> {
+                try (InputStream inputStream = input) {
+                    T value = reader.readValue(inputStream);
+                    // Bad input should result in a 4XX response status, throw IAE rather than NPE.
```
carterkozak: I expect we copied this from conjure-undertow, Dialogue should cause a 5xx if this happens. We can fix the core dialoge implementation separately.

carterkozak: That fix: #1148

jkozlowski: Alright, I'll recopy the code.

jkozlowski: Done

dialogue-annotations-processor/src/test/java/com/palantir/myservice/service/MyServiceDialogue.java
```diff
@@ -0,0 +1,279 @@
...
+import com.palantir.dialogue.TypeMarker;
+import com.palantir.dialogue.UrlBuilder;
+import com.palantir.dialogue.annotations.Json;
+import com.palantir.dialogue.annotations.RequestDeserializer;
+import java.util.Map;
+import java.util.Optional;
+import java.util.OptionalInt;
+import java.util.UUID;
+
+interface MyServiceDialogue extends DialogueServiceFactory<MyService> {
```
jkozlowski: Given autocomplete in IDEs, maybe this should be the nice and long *MyServiceDialogueServiceFactory*?

carterkozak: Sounds good to me!

changelog/@unreleased/pr-1135.v2.yml
```diff
@@ -0,0 +1,5 @@
+type: improvement
+improvement:
+  description: <!-- User-facing outcomes this PR delivers -->
```
carterkozak: huh, I think the robot has been automatically committing these lately

dialogue-annotations/src/main/java/com/palantir/dialogue/annotations/RequestDeserializer.java
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    @Override
+    public Optional<String> accepts() {
+        return Optional.empty();
+    }
+
+    @Override
+    public String toString() {
+        return "EmptyBodyDeserializer{}";
```
carterkozak:
```suggestion
        return "ResponseDeserializer{}";
```

changelog/@unreleased/pr-1135.v2.yml
jkozlowski: Nah, I think I just didn't fill this stuff out 🤦

dialogue-annotations/src/main/java/com/palantir/dialogue/annotations/RequestDeserializer.java
```diff
@@ -0,0 +1,40 @@
...
+    INSTANCE;
+
+    @Override
+    public Response deserialize(Response response) {
+        return response;
+    }
+
+    @Override
+    public Optional<String> accepts() {
+        return Optional.empty();
```
carterkozak: Perhaps something along these lines?
```suggestion
        return Optional.of("*/*");
```

dialogue-annotations/src/main/java/com/palantir/dialogue/annotations/RequestDeserializer.java
```diff
@@ -0,0 +1,40 @@
...
+ * limitations under the License.
+ */
+
+package com.palantir.dialogue.annotations;
+
+import com.palantir.dialogue.Deserializer;
+import com.palantir.dialogue.Response;
+import java.util.Optional;
+
+public enum RequestDeserializer implements Deserializer<Response> {
```
carterkozak: s/Request/Response

dialogue-annotations/src/main/java/com/palantir/dialogue/annotations/RequestDeserializer.java
jkozlowski: Sorry clicked commit too quickly. This is an interesting question I believe: I can imagine that maybe there might be some usecase where you'd like to be able to affect this? And therefore should perhaps be not a special thing here, but you need to implement the deserializer yourself?

jkozlowski: Actually, this is perhaps an OK default, but we should always allow the escape hatch of:

```
    @Request(method = HttpMethod.PUT, path = "/custom/request1", accept = MyCustomResponseDeserializer.class)
    Response customResponse();
```

dialogue-annotations/src/main/java/com/palantir/dialogue/annotations/RequestDeserializer.java
jkozlowski: Done

dialogue-annotations/src/main/java/com/palantir/dialogue/annotations/RequestDeserializer.java
carterkozak: For what it's worth, I think we could leave this deserializer out for the time being, there's some additional ambiguity around whether users will expect it to succeed for non-200 responses, so it may be best to provide developers with the tools to do the right thing. In most cases I think it's preferable if they return a custom type instead of the Response object anyhow.

carterkozak: To be clear, I don't have a strong opinion on this and defer to your judgement. Only pointing out how a standard implementation might turn out to be a foot-gun.

jkozlowski: That's exactly what I meant: this can be covered by a custom deserializer that anyone can implement, so we should take this out. It was in the example on the Quip, so I impl'ed it, but it doesn't really make sense. Or rather, I'll have it as a custom one in the example, just to show it off, but it won't be in the -annotations project.

jkozlowski: Alright, I moved it around and added some docs to Json class to point out that it's conjure style json, including special error behavior.

carterkozak (approved):

jkozlowski:
👍

PR #1138, opened 2021-03-01
Annotation based solution to avoid static method string matching

carterkozak:
This approach uses the old implemnetation as a fallback, but allows
us to simplify creation using an intermediate factory that must
provide a public no-arg constructor.

This will help us unify the annotation processor and conjure generated
services.

==COMMIT_MSG==
Annotation based solution to avoid static method string matching
==COMMIT_MSG==

### DIFF
--- a/dialogue-target/src/main/java/com/palantir/dialogue/DialogueService.java
+++ b/dialogue-target/src/main/java/com/palantir/dialogue/DialogueService.java
@@ -1,0 +1,30 @@
+/*
+ * (c) Copyright 2021 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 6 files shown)


dialogue-target/src/main/java/com/palantir/dialogue/DialogueService.java
```diff
@@ -0,0 +1,30 @@
...
+package com.palantir.dialogue;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Target(ElementType.TYPE)
+@Retention(RetentionPolicy.RUNTIME)
+public @interface DialogueService {
```
carterkozak, line 26: I think I'd prefer `DialogueClient`, but the factory interface naming would overlap with the `dialogue-clients` factory components. Thoughts/ideas/concerns?

jkozlowski: No strong preference, Service kinda stuck with me for whatever reasons. But if DialogueClientFactory is clashing, then it's clashing.

The generated classes from conjure are "AutoDeserializeConfirmServiceBlocking", so I think DialogueService it is.

jkozlowski (approved):

PR #1145, opened 2021-03-04
Use SOURCE retention for immutables style annotations

pkoenig10:
See https://github.com/immutables/immutables/issues/291.

This eliminates compilation warnings for consumers that do not directly depend on immutables.

Requires upgrades to pull in:
- https://github.com/palantir/conjure-java-runtime/pull/1897
- https://github.com/palantir/tritium/pull/1006


versions.props
```diff
@@ -3,13 +3,14 @@ com.google.guava:guava = 30.1-jre
 com.palantir.conjure:conjure = 4.14.2
 com.palantir.conjure.java:* = 5.45.0
 com.palantir.conjure.java.api:* = 2.16.2
+com.palantir.conjure.java.runtime:* = 6.14.0
```
pkoenig10, line 6: conjure-java-runtime 6.0.0 was released 9 months ago and the only breaking change was switching JAX-RS clients to Dialogue.
- Most consumers have already upgraded to conjure-java-runtime 6.0.0.
- This change should not affect the vast majority of consumers.
-  If you're depending on Dialogue, you probably are okay using Dialogue for your JAX-RS clients.

https://github.com/palantir/conjure-java-runtime/releases/tag/6.0.0

carterkozak: :+1:

carterkozak (approved):

PR #1150, opened 2021-03-05
Fix #1149: Network failures while writing request bytes are retried

### LINKED ISSUE #1149: Dialogue unnecessarily wraps IOExceptions from ser/de in AbstractJacksonEncoding
## What happened?

Some failures are not retried because they appear to be unrecoverable RuntimeExceptions, despite being `IOException("broken pipe")` due to a network hiccup.

https://github.com/palantir/dialogue/blob/1617106236d22d6cc235cb152f4039c410bff070/dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/Encodings.java#L56-L63

`Encoding.Serializer.serialize(T,OutputStream)` should throw IOException. As should `T Encoding.Deserializer.deserialize(InputStream)`

carterkozak:
## Before this PR
Some calls weren't retried when we expected them to be

## After this PR
Dialogue encodings no longer wrap IOExceptions in RuntimeExceptions,
allowing the retryer to identify and retry socket exceptions.
==COMMIT_MSG==
Network failures while writing request bytes are retried
==COMMIT_MSG==

## Possible downsides?
Encoding serializers and deserializers are implemented within
the serde module, same as all callers.

The modified interface is documented:
> This interface is considered internal API and may
> change without a major version rev.

If external parties build against these interfaces, there's limited
risk of new dialogue calling old compiled code because it's perfectly
reasonable to implement an interface and narrow/remove exception
types. Potential failure is UndeclaredThrowableException if custom
code calls into one of these interfaces. That shouldn't occur
unless something else has already failed.

## Alternatives
I also implemented a fix without any API changes which inspected
the entire causal chain, rather than the top level exception.
This scared me, I'm confident it would have resulted in a subtle
P0.


.palantir/revapi.yml
```diff
@@ -157,6 +157,21 @@ acceptedBreaks:
     - code: "java.method.addedToInterface"
       new: "method void com.palantir.dialogue.RequestBody::close()"
       justification: "internal interface not for extension"
+  "1.104.0":
+    com.palantir.dialogue:dialogue-serde:
+    - code: "java.method.exception.checkedAdded"
+      old: "method T com.palantir.conjure.java.dialogue.serde.Encoding.Deserializer<T>::deserialize(java.io.InputStream)"
+      new: "method T com.palantir.conjure.java.dialogue.serde.Encoding.Deserializer<T>::deserialize(java.io.InputStream)\
```
jkozlowski, line 164: The Encoding interface is public though and there are non-internal impls: internal/search?q=%22com.palantir.conjure.java.dialogue.serde.Encoding%22&type=Code, hyperion, container-vuln-scanner and assets2 have all copied the same snippet of code. I guess this break is fine, they'll just get a compile break right?

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ConjureBodySerDe.java
```diff
@@ -258,6 +258,12 @@ public T deserialize(Response response) {
                 // deserializer has taken on responsibility for closing the response body
                 closeResponse = false;
                 return deserialized;
+            } catch (IOException e) {
+                throw new SafeRuntimeException(
```
jkozlowski, line 262: For posterity: this does not get retried, it will just fail to the client. It's just that for consistency we're wrapping the exception here instead of Deserializer impls.

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ConjureBodySerDe.java
```diff
@@ -172,7 +172,7 @@ public RequestBody serialize(T value) {
             return new RequestBody() {
 
                 @Override
-                public void writeTo(OutputStream output) {
+                public void writeTo(OutputStream output) throws IOException {
                     encoding.serializer.serialize(value, output);
```
jkozlowski, line 176: For documenting: the actual retries here will now properly happen in RetryingChannel#shouldAttemptToRetry, because we only retry IOExceptions. Previously the underlying IOException was wrapped, and thus not retried.

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ConjureBodySerDe.java
jkozlowski: Fixing this would require a lot more work.

jkozlowski (approved):

.palantir/revapi.yml
carterkozak: These fall into the category that won't get a compile break, it's legal to override a method which declares checked exceptions with an implementation that does not declare exceptions. The risk of compile breakage is only for code that delegates to another Encoding.Serializer/Deserializer.

PR #1155, opened 2021-03-08
Use more specific generic type arguments to bound serializer factories

carterkozak:
==COMMIT_MSG==
Use more specific generic type arguments to bound serializer factories
==COMMIT_MSG==

### DIFF
--- a/dialogue-annotations/src/main/java/com/palantir/dialogue/annotations/StdSerializer.java
+++ b/dialogue-annotations/src/main/java/com/palantir/dialogue/annotations/StdSerializer.java
@@ -20,10 +20,10 @@ import com.palantir.dialogue.Serializer;
 import com.palantir.dialogue.TypeMarker;
 
 @SuppressWarnings({"RawTypes", "unchecked"})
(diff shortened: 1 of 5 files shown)


dialogue-annotations/src/main/java/com/palantir/dialogue/annotations/StdSerializer.java
```diff
@@ -20,10 +20,10 @@
...
 
 @SuppressWarnings({"RawTypes", "unchecked"})
-public abstract class StdSerializer<T> implements SerializerFactory, Serializer<T> {
+public abstract class StdSerializer<T> implements SerializerFactory<T>, Serializer<T> {
 
     @Override
-    public final <T1> Serializer<T1> serializerFor(TypeMarker<T1> _type) {
-        return (Serializer) this;
+    public final <T1 extends T> Serializer<T1> serializerFor(TypeMarker<T1> _type) {
+        return (Serializer<T1>) this;
```
jkozlowski, line 27: Hmm I think this cast was getting me when I tried to impl this. Glad this works!

jkozlowski (approved):

PR #1170, opened 2021-03-25
Wiring annotation processor and javapoet.

jkozlowski:
## After this PR

Types of testing:

### Invoking the compiler manually
* There is a golden-file test in DialogueAnnotationsProcessorTest that checks generated code against a reference; this reuses the MyService defined in the testCases test-suite.
* There is a validation failure test in DialogueAnnotationsProcessorTest: this basically tries to compile RequestAnnotatedClass to check that we do no allow annotating classes.

### using annotationProcessor dependency in gradle
* The standard unit-test suite is wired up with the annotation processor, as a user would; that has an integration test from previous PRs (which currently disables all cases) which tries to instantiate the client and checks we get the UnsupportedOperationException that the generated code throws.

==COMMIT_MSG==
Wire annotation processor and javapoet.
==COMMIT_MSG==

## Possible downsides?

* Technically (I think) the annotation processor is invoked twice in the case of the DialogueAnnotationsProcessorTest, but nothing seems to mind so I'm going to let this slide.


dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/DialogueAnnotationsProcessor.java
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    @Override
+    public Set<String> getSupportedAnnotationTypes() {
+        return ImmutableSet.of(Request.class.getCanonicalName());
+    }
+
+    @Override
+    public boolean process(Set<? extends TypeElement> _annotations, RoundEnvironment roundEnv) {
+        Set<Element> elements = roundEnv.getElementsAnnotatedWith(Request.class).stream()
```
jkozlowski, line 66: This will likely evolve a lot, because it currently doesn't actually look at the annotated methods, so I wouldn't review this deeply. As I learn and go further down the codegen path, it might turn out that some pieces here are not quite right.

dialogue-annotations-processor/src/test/java/com/palantir/myservice/service/MyServiceDialogueServiceFactory.java
```diff
@@ -1,278 +0,0 @@
...
-import com.palantir.dialogue.Serializer;
-import com.palantir.dialogue.TypeMarker;
-import com.palantir.dialogue.UrlBuilder;
-import com.palantir.dialogue.annotations.Json;
-import java.util.Map;
-import java.util.Optional;
-import java.util.OptionalInt;
-import java.util.UUID;
-
-public final class MyServiceDialogueServiceFactory implements DialogueServiceFactory<MyService> {
```
jkozlowski, line 42: Deleting as this was an initial test just to check everything gets wired properly. In order to still keep it I'd probably need to have the MyService interface duplicated, since @DialogueClient annotation cannot be repeated. So I opted to just nuke it, and I'll enable tests when I have the implementations of the methods.

dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/DialogueAnnotationsProcessor.java
```diff
@@ -0,0 +1,115 @@
...
+            return false;
+        }
+
+        ClassName outputClass = ClassName.get(
+                MoreElements.getPackage(annotatedInterface).getQualifiedName().toString(),
+                annotatedInterface.getSimpleName().toString());
+
+        TypeSpec generatedClass = new DialogueServiceFactoryGenerator(outputClass).generate();
+        try {
+            JavaFile.builder(outputClass.packageName(), generatedClass).build().writeTo(filer);
```
jkozlowski: Forgot to format the code with our style, will add that shortly.

tetigi (approved):
Jakub and I chatted through the detail on this - seems like a good bouncing point for starting to work on the actual code generation. Some small unknowns about exactly how to handle the gradle part, but both agreed that that can be followed up on later.

PR #1179, opened 2021-03-29
Fix testing setup for annotation processor.

jkozlowski:
## Before this PR

Cannot really write code in IntelliJ and run tests.

## After this PR
==COMMIT_MSG==
Annotation processor should work in IntelliJ.
==COMMIT_MSG==

## Possible downsides?

Separate subproject is a bit meh, since I'm duplicating all the test code. But I suspect over time there'll be more test code than example code, so probably not a big deal.


dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/DialogueRequestAnnotationsProcessor.java
```diff
@@ -0,0 +1,185 @@
...
+import javax.annotation.processing.SupportedSourceVersion;
+import javax.lang.model.SourceVersion;
+import javax.lang.model.element.Element;
+import javax.lang.model.element.ElementKind;
+import javax.lang.model.element.TypeElement;
+import javax.tools.Diagnostic.Kind;
+
+@SupportedSourceVersion(SourceVersion.RELEASE_8)
+@AutoService(Processor.class)
+public final class DialogueRequestAnnotationsProcessor extends AbstractProcessor {
```
jkozlowski, line 54: Likely enough changed that git thinks this is a new file :( I snuck a few things I have learned were a bit broken about my previous attempt.

1. The #process method can actually get a bunch of classes with annotations, not one call per class.
2. Tried to restructure the code so I can throw exceptions for control flow, not ```return false``` everywhere because that got boring rather quickly.
3. Added ability to report multiple errors and then abort from a "step" if there were any errors reported. Will mostly be useful for error reporting when validating the annotations later.

carterkozak (approved):

PR #1184, opened 2021-03-30
Generate Endpoint implementation and instantiate channels.

jkozlowski:
## After this PR
==COMMIT_MSG==
Create endpoint channels.
==COMMIT_MSG==

### DIFF
--- a/dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/data/EndpointDefinition.java
+++ b/dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/data/EndpointDefinition.java
@@ -34,4 +34,9 @@ public interface EndpointDefinition {
     List<ArgumentDefinition> arguments();
 
     TypeName returns();
+
+    @Value.Derived
+    default String channelFieldName() {
+        return endpointName().get() + "Channel";
+    }
 }
--- a/dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/data/EndpointDefinitions.java
+++ b/dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/data/EndpointDefinitions.java
@@ -18,6 +18,7 @@ package com.palantir.dialogue.annotations.processor.data;
 
 import com.google.auto.common.MoreElements;
 import com.google.common.base.Predicates;
+import com.google.common.base.Splitter;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Iterables;
 import com.palantir.common.streams.KeyedStream;
@@ -94,7 +95,24 @@ public final class EndpointDefinitions {
     private Optional<HttpPath> getHttpPath(Element context, Request requestAnnotation) {
         try {
             UriTemplateParser uriTemplateParser = new UriTemplateParser(requestAnnotation.path());
-            return Optional.of(ImmutableHttpPath.of(uriTemplateParser.getNormalizedTemplate()));
+
+            Splitter splitter = Splitter.on('/');
+            Iterable<String> rawSegments = splitter.split(uriTemplateParser.getNormalizedTemplate());
+
+            List<HttpPathSegment> pathSegments = new ArrayList<>();
+            for (String segment : rawSegments) {
+                if (segment.isEmpty()) {
+                    continue; // avoid empty segments; typically the first segment is empty
+                }
+
+                if (segment.startsWith("{") && segment.endsWith("}")) {
+                    pathSegments.add(HttpPathSegments.variable(segment.substring(1, segment.length() - 1)));
+                } else {
+                    pathSegments.add(HttpPathSegments.fixed(segment));
+                }
+            }
+
+            return Optional.of(ImmutableHttpPath.of(pathSegments));
         } catch (IllegalArgumentException e) {
             errorContext.reportError("Failed to parse http path", context, e);
             return Optional.empty();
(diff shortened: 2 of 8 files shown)


dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/data/EndpointDefinition.java
```diff
@@ -34,4 +34,9 @@
     List<ArgumentDefinition> arguments();
 
     TypeName returns();
+
+    @Value.Derived
+    default String channelFieldName() {
+        return endpointName().get() + "Channel";
```
jkozlowski, line 40: I rather like this way of solving the "I need this value in a bunch of generators" rather than static util methods scattered around everywhere.

dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/generate/EndpointsEnumGenerator.java
```diff
@@ -0,0 +1,142 @@
...
+import com.squareup.javapoet.FieldSpec;
+import com.squareup.javapoet.MethodSpec;
+import com.squareup.javapoet.ParameterizedTypeName;
+import com.squareup.javapoet.TypeName;
+import com.squareup.javapoet.TypeSpec;
+import java.util.Map;
+import java.util.Optional;
+import javax.lang.model.element.Modifier;
+
+public final class EndpointsEnumGenerator {
```
jkozlowski, line 39: Pretty much copied from conjure, except I eagerly parse out HttpPath into a bunch of segments. Validation will also happen ahead of time.

dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/data/EndpointDefinitions.java
```diff
@@ -94,7 +95,24 @@ public EndpointDefinitions(ErrorContext errorContext, Elements elements, Types t
...
+            Iterable<String> rawSegments = splitter.split(uriTemplateParser.getNormalizedTemplate());
+
+            List<HttpPathSegment> pathSegments = new ArrayList<>();
+            for (String segment : rawSegments) {
+                if (segment.isEmpty()) {
+                    continue; // avoid empty segments; typically the first segment is empty
+                }
+
+                if (segment.startsWith("{") && segment.endsWith("}")) {
+                    pathSegments.add(HttpPathSegments.variable(segment.substring(1, segment.length() - 1)));
```
carterkozak, line 109: same as the other PR, I imagine we'll want to validate that this doesn't use any of the funky regex options e.g. `/foo/{var:.*}`
No change requested to move this PR forward though.

dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/generate/EndpointsEnumGenerator.java
```diff
@@ -0,0 +1,142 @@
...
+                .initializer(
+                        "$T.ofNullable($T.class.getPackage().getImplementationVersion()).orElse(\"0.0.0\")",
+                        TypeName.get(Optional.class),
+                        serviceDefinition.serviceInterface())
+                .build());
+
+        return enumBuilder.build();
+    }
+
+    private TypeSpec endpointField(EndpointDefinition def, String serviceName, Optional<String> apiVersion) {
```
carterkozak: Can we remove `Optional<String> apiVersion` entirely? This will always use the package implementation version since there's no external definition.

dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/generate/EndpointsEnumGenerator.java
```diff
@@ -0,0 +1,142 @@
...
+                        .build())
+                .addMethod(MethodSpec.methodBuilder("version")
+                        .addAnnotation(Override.class)
+                        .addModifiers(Modifier.PUBLIC)
+                        .returns(String.class)
+                        .addCode(CodeBlock.builder()
+                                .add(apiVersion
+                                        .map(s -> CodeBlock.of("return $S;", s))
+                                        .orElseGet(() -> CodeBlock.of("return VERSION;")))
+                                .build())
```
carterkozak: I think this becomes
```suggestion
                        .addCode(CodeBlock.of("return VERSION;"))
```

dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/data/EndpointDefinition.java
carterkozak: It's what they recommend in "Effective Java" these days :-)

dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/generate/EndpointsEnumGenerator.java
jkozlowski: yeah, that's what I thought

dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/data/EndpointDefinitions.java
jkozlowski: Ack, added to tracking doc.

carterkozak (approved):

jkozlowski:
+1

carterkozak:
Ahh I was too late removing autorelease.

PR #1187, opened 2021-03-30
Build Requests and serialize parameters and body

jkozlowski:
## After this PR
==COMMIT_MSG==
Build Requests and serialize parameters and body
==COMMIT_MSG==

### DIFF
--- a/dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/data/ArgumentTypesResolver.java
+++ b/dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/data/ArgumentTypesResolver.java
@@ -51,7 +51,7 @@ import javax.lang.model.util.Types;
 
 public final class ArgumentTypesResolver {
 
-    private static final String VALUE_OF_METHOD_NAME = "valueOf";
+    private static final String AS_STRING_METHOD_NAME = "asString";
 
     /**
      * Why not generate this by inspecting the {@link com.palantir.dialogue.PlainSerDe} interface?
@@ -121,7 +121,7 @@ public final class ArgumentTypesResolver {
             // TODO(12345): We only want to go one level down: don't allow Optional<Optional<Type>>.
             return Optional.of(ArgumentTypes.optional(typeName, optionalType.get()));
         } else if (isCustomAndHasValueOfMethod(actualTypeMirror)) {
-            return Optional.of(ArgumentTypes.customTypeWithValueOf(typeName, VALUE_OF_METHOD_NAME));
+            return Optional.of(ArgumentTypes.customTypeWithValueOf(typeName, AS_STRING_METHOD_NAME));
         } else {
             return Optional.of(ArgumentTypes.customType(typeName));
         }
@@ -189,7 +189,7 @@ public final class ArgumentTypesResolver {
         Visibility visibility = Visibility.ofElement(executableElement);
         Set<Modifier> modifiers = executableElement.getModifiers();
         // TODO(12345): Actually give the user good feedback on what exactly they need to add to make this type comply.
-        return executableElement.getSimpleName().toString().equals(VALUE_OF_METHOD_NAME)
+        return executableElement.getSimpleName().toString().equals(AS_STRING_METHOD_NAME)
                 && executableElement.getParameters().isEmpty()
                 && executableElement.getTypeParameters().isEmpty()
                 && types.isSameType(executableElement.getReturnType(), stringType)
--- a/dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/data/ArgumentTypesResolver.java
+++ b/dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/data/ArgumentTypesResolver.java
@@ -46,8 +46,6 @@ import javax.lang.model.util.Types;
 
 public final class ArgumentTypesResolver {
 
-    private static final String TO_STRING_PARAM_VALUE_METHOD = "toStringParamValue";
-
     /**
      * Why not generate this by inspecting the {@link com.palantir.dialogue.PlainSerDe} interface?
      * If we suddenly decide to provide special serialization for a widely used type, we could break wire
--- a/dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/data/ArgumentTypesResolver.java
+++ b/dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/data/ArgumentTypesResolver.java
@@ -1,0 +1,139 @@
+/*
+ * (c) Copyright 2021 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.dialogue.annotations.processor.data;
+
+import com.google.auto.common.MoreElements;
+import com.google.common.collect.ImmutableMap;
+import com.palantir.common.streams.KeyedStream;
+import com.palantir.conjure.java.lib.SafeLong;
+import com.palantir.dialogue.RequestBody;
+import com.palantir.dialogue.annotations.processor.ArgumentType;
+import com.palantir.dialogue.annotations.processor.ArgumentTypes;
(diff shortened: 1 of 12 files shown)


dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/data/EndpointDefinitions.java
```diff
@@ -94,7 +95,24 @@ public EndpointDefinitions(ErrorContext errorContext, Elements elements, Types t
...
+            Iterable<String> rawSegments = splitter.split(uriTemplateParser.getNormalizedTemplate());
+
+            List<HttpPathSegment> pathSegments = new ArrayList<>();
+            for (String segment : rawSegments) {
+                if (segment.isEmpty()) {
+                    continue; // avoid empty segments; typically the first segment is empty
+                }
+
+                if (segment.startsWith("{") && segment.endsWith("}")) {
+                    pathSegments.add(HttpPathSegments.variable(segment.substring(1, segment.length() - 1)));
```
carterkozak: We may want to validate that the more obscure features of UriTemplateParser aren't used, e.g. `/foo/{bar:.*}`. Apologies if we've already validated the template elsewhere.

jkozlowski: Nope, I haven't. Just copying conjure code for now: I think as you say, we'll want to reimpl this.

dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/generate/ServiceImplementationGenerator.java
```diff
@@ -77,4 +106,198 @@ private FieldSpec bindEndpointChannel(EndpointDefinition endpoint) {
...
+    //             return "getAsDouble";
+    //         } else if (primitive.equals(PrimitiveType.INTEGER)) {
+    //             return "getAsInt";
+    //         }
+    //     }
+    //     return "get";
+    // }
+
+    private static final ImmutableMap<TypeName, String> PRIMITIVE_TO_TYPE_NAME =
+            ImmutableMap.copyOf(KeyedStream.stream(new ImmutableMap.Builder<Class<?>, String>()
```
jkozlowski: I feel like it would be good to generate that... But then again, we probably don't want to change people's code, if we suddenly change serialization for a type, that they rely on to be serialized using e.g. Jackson.

jkozlowski: Actually, this is not Jackson, this is for path/query/header params, so if someone is relying on a type that we were to later add here, they have likely wrapped it: it must've come from a library, and the library probably does not have the #valueOf -> String method that we would generate a binding to. Soo, I think it's ok to generate this

dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/data/HttpPathParser.java
```diff
@@ -0,0 +1,62 @@
...
+
+public final class HttpPathParser {
+
+    private final ErrorContext errorContext;
+
+    public HttpPathParser(ErrorContext errorContext) {
+        this.errorContext = errorContext;
+    }
+
+    public Optional<HttpPath> getHttpPath(Element context, Request requestAnnotation) {
```
jkozlowski, line 36: Extracted

dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/data/EndpointDefinitions.java
```diff
@@ -92,129 +69,18 @@ public EndpointDefinitions(ErrorContext errorContext, Elements elements, Types t
...
-                    continue; // avoid empty segments; typically the first segment is empty
-                }
-
-                if (segment.startsWith("{") && segment.endsWith("}")) {
-                    pathSegments.add(HttpPathSegments.variable(segment.substring(1, segment.length() - 1)));
-                } else {
-                    pathSegments.add(HttpPathSegments.fixed(segment));
-                }
-            }
+    private Optional<ArgumentDefinition> getArgumentDefinition(VariableElement param) {
```
jkozlowski: Just extracted code into smaller classes.

dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/data/ParamTypesResolver.java
```diff
@@ -0,0 +1,151 @@
...
+import javax.lang.model.type.TypeMirror;
+import javax.lang.model.util.Elements;
+import javax.lang.model.util.Types;
+import org.immutables.value.Value;
+
+public final class ParamTypesResolver {
+
+    private static final ImmutableSet<Class<?>> PARAM_ANNOTATION_CLASSES = ImmutableSet.of(
+            Request.Body.class, Request.PathParam.class, Request.QueryParam.class, Request.Header.class);
+    private static final ImmutableSet<String> PARAM_ANNOTATIONS =
```
jkozlowski, line 49: All extracted.

dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/data/ArgumentTypesResolver.java
```diff
@@ -0,0 +1,200 @@
...
+public final class ArgumentTypesResolver {
+
+    private static final String VALUE_OF_METHOD_NAME = "valueOf";
+
+    /**
+     * Why not generate this by inspecting the {@link com.palantir.dialogue.PlainSerDe} interface?
+     * If we suddenly decide to provide special serialization for a widely used type, we could break wire
+     * compatibility.
+     */
+    private static final ImmutableMap<TypeName, String> PLAIN_SER_DE_TYPES =
```
jkozlowski, line 54: as a FLUP I'll switch this to dialogue-annotations specific interface, that handles more types, so we're not tied to the conjure one.

dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/data/ArgumentTypesResolver.java
```diff
@@ -0,0 +1,200 @@
...
+        TypeElement typeElement = (TypeElement) element;
+        return MoreElements.getAllMethods(typeElement, types, elements).stream().anyMatch(this::isValueOfMethod);
+    }
+
+    private boolean isValueOfMethod(ExecutableElement executableElement) {
+        Visibility visibility = Visibility.ofElement(executableElement);
+        Set<Modifier> modifiers = executableElement.getModifiers();
+        // TODO(12345): Actually give the user good feedback on what exactly they need to add to make this type comply.
+        return executableElement.getSimpleName().toString().equals(VALUE_OF_METHOD_NAME)
+                && executableElement.getParameters().isEmpty()
```
carterkozak: Shouldn't this check for size 1 of type String?

carterkozak: I think I see, this is backwards -- `valueOf` is used to produce an object from a string, I think we may want toString here, or to require a custom type that can handle serialization.

e.g. 
`Long parsed = Long.valueOf("123");`
`String serialized = parsed.toString();`

jkozlowski: I opted for instance method? so no args, return type String?

jkozlowski: Oh I see what you mean, you're objecting to the method name. How about asString?

jkozlowski: I definitely don't think we should really use #toString, that feels too footgunny; https://stackoverflow.com/questions/29207562/how-to-use-custom-type-for-pathparam is some prior art.

jkozlowski: #asString feels like the least of ceremony, you can just wrap types you don't own (if you really want to do that) and impl the method. It's definitely magic, so maybe all annotations could get an Optional<Class<? extends ParamConverter>>, but that is defo more work

carterkozak: Not a hug fan of wrapping types, especially with magic instance methods. An arg-wrapper type could be better, but the client API would feel dirty to use, even if it was easy to write. I'd much rather use `Class<? extends ParamConverter>` for unknown types with validation at compile time.

carterkozak: I think it's reasonable to exclude custom plain parameter handling entirely for this PR given the potential complexity.

carterkozak (approved):
Looks great, thanks!

PR #1194, opened 2021-04-07
Dialogue allows the default port to be used based on the input URI

carterkozak:
This behavior is based on the configured URI -- URIs which provide
a port result in a `Host` header containing a port, and those which
rely on the implicit default (e.g. http -> 80, https -> 443) do
not include the port in their `Host` header.

## Before this PR
It turns out some servers require that the default port is used,
and _not_ included in URIs. Until we have a better way to test this
I've opted not to exclude default ports everywhere, only when
the provided URI doesn't include a port. Testing is difficult because
permissions are required to run a server on port 80 or 443 to verify,
however I have validated that this behaves in the way we expect
locally.

## After this PR
==COMMIT_MSG==
Dialogue allows the default port to be used based on the input URI
==COMMIT_MSG==

## Possible downsides?
Difficult to test as described above

### DIFF
--- a/changelog/@unreleased/pr-1194.v2.yml
+++ b/changelog/@unreleased/pr-1194.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Dialogue allows the default port to be used based on the input URI
+  links:
+  - https://github.com/palantir/dialogue/pull/1194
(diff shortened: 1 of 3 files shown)


jkozlowski (approved):

carterkozak:
:+1:

PR #1197, opened 2021-04-08
Refactor ParameterSerializer.

jkozlowski:
## After this PR
==COMMIT_MSG==

==COMMIT_MSG==

### DIFF
--- a/dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/data/ArgumentTypesResolver.java
+++ b/dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/data/ArgumentTypesResolver.java
@@ -16,46 +16,53 @@
 
 package com.palantir.dialogue.annotations.processor.data;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableMap;
-import com.google.common.primitives.Primitives;
+import com.palantir.common.streams.KeyedStream;
+import com.palantir.conjure.java.lib.SafeLong;
 import com.palantir.dialogue.RequestBody;
-import com.palantir.dialogue.annotations.ParameterSerializer;
 import com.palantir.dialogue.annotations.processor.ArgumentType;
 import com.palantir.dialogue.annotations.processor.ArgumentType.OptionalType;
 import com.palantir.dialogue.annotations.processor.ArgumentTypes;
 import com.palantir.dialogue.annotations.processor.ImmutableOptionalType;
 import com.palantir.logsafe.Preconditions;
-import com.palantir.logsafe.SafeArg;
+import com.palantir.ri.ResourceIdentifier;
+import com.palantir.tokens.auth.AuthHeader;
+import com.palantir.tokens.auth.BearerToken;
 import com.squareup.javapoet.ClassName;
 import com.squareup.javapoet.TypeName;
-import java.lang.reflect.Method;
+import java.time.OffsetDateTime;
 import java.util.Optional;
 import java.util.OptionalInt;
+import java.util.UUID;
+import java.util.function.Function;
 import javax.lang.model.element.Element;
 import javax.lang.model.element.VariableElement;
 import javax.lang.model.type.TypeMirror;
 
 public final class ArgumentTypesResolver {
 
-    private static final ImmutableMap<TypeName, String> PARAMETER_SERIALIZER_TYPES;
-
-    static {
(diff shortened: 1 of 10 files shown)


dialogue-annotations/src/test/java/com/palantir/dialogue/annotations/ParameterSerializerTest.java
```diff
@@ -0,0 +1,146 @@
...
+        for (Pair<Character, String> inOut : Arrays.asList(Pair.of('c', "c"), Pair.of('\n', "\n"))) {
+            char in = inOut.getLeft();
+            String out = inOut.getRight();
+            assertThat(PLAIN.serializeChar(in)).isEqualTo(out);
+        }
+    }
+
+    @Test
+    public void testByte() {
+        for (Pair<Byte, String> inOut : Arrays.asList(Pair.of((byte) 0x0, "0"), Pair.of((byte) 0xde, "-34"))) {
```
jkozlowski: I am a bit torn on the byte here... @carterkozak what do you think?

dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/data/ArgumentTypesResolver.java
```diff
@@ -17,49 +17,45 @@
...
+        ImmutableMap.Builder<TypeName, String> builder = ImmutableMap.builder();
+        for (Method method : ParameterSerializer.class.getDeclaredMethods()) {
+            Preconditions.checkArgument(
+                    method.getReturnType().equals(String.class),
+                    "Return type is not String",
+                    SafeArg.of("method", method));
+            Preconditions.checkArgument(
+                    method.getParameterCount() == 1,
+                    "Serializer methods should have a single " + "arg",
+                    SafeArg.of("method", method));
```
carterkozak: This feels brittle, I'd instead use a unit test to verify, and if/else here to avoid compilation failures if our assumptions are incorrect in future jvms.

I think we should ignore methods that are non-public, non-static, return non-string, or don't have exactly one parameter.

dialogue-annotations/src/test/java/com/palantir/dialogue/annotations/ParameterSerializerTest.java
carterkozak: I might even consider leaving them out entirely, using the `int` method, and relying on implicit casting.

dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/data/ArgumentTypesResolver.java
jkozlowski: Sure, can you elaborate on compilation failures if incorrect in future jvms? like if something like default methods get added?

carterkozak: Yep, exactly. There's also some weirdness where lambdas can become special methods on the encapsulating class I encountered here: https://github.com/FasterXML/jackson-databind/pull/2928

jkozlowski: Also "ParameterSerializer" is an interface we own and in this codebase, so probably no-non public, non-static, non-string; I would kinda prefer this to keep as is, it will be in a test so happy to break tests.

dialogue-annotations/src/test/java/com/palantir/dialogue/annotations/ParameterSerializerTest.java
jkozlowski: you mean like force people to (int) their bytes? yeah, I meant to leave off bytes basically.

jkozlowski: Removed entirely.

dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/data/ArgumentTypesResolver.java
jkozlowski: Have a look now, I've made the prod class list them explicitly, and there's a consistency test.

carterkozak (approved):

PR #1198, opened 2021-04-09
Make AuthHeader special.

jkozlowski:
## Before this PR
User's have to annotate AuthHeader with @Request.Header("Authorization") for it to do properly. This is obviously error-prone, while we know that this type is supposed to be a special header (hint is in the name).

## After this PR
==COMMIT_MSG==
AuthHeader does not need to be annotated.
==COMMIT_MSG==


dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/data/ParamTypesResolver.java
```diff
@@ -93,8 +96,6 @@ public ParamTypesResolver(ResolverContext context) {
                     TypeName.get(customSerializer.orElseGet(() -> context.getTypeMirror(Json.class))), serializerName));
         } else if (annotationReflector.isAnnotation(Request.Header.class)) {
             return Optional.of(ParameterTypes.header(annotationReflector.getStringValueField()));
-        } else if (annotationReflector.isAnnotation(Request.Header.class)) {
```
jkozlowski, line 96: How did this sneak in?

carterkozak (approved):

PR #1200, opened 2021-04-09
Handle custom param types.

jkozlowski:
## After this PR
==COMMIT_MSG==
dialogue-annotations supports custom parameter types.
==COMMIT_MSG==


dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/data/InstanceVariables.java
```diff
@@ -0,0 +1,40 @@
...
+
+    private InstanceVariables() {}
+
+    static String joinCamelCase(String... segments) {
+        return Joiner.on("")
+                .join(IntStream.range(0, segments.length)
+                        .mapToObj(i -> {
+                            String segment = segments[i];
+                            if (i == 0) {
+                                return CaseFormat.LOWER_CAMEL.to(CaseFormat.LOWER_CAMEL, segment);
```
carterkozak: This should be a no-op, no?

dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/data/InstanceVariables.java
```diff
@@ -0,0 +1,40 @@
...
+                            }
+                            return CaseFormat.UPPER_CAMEL.to(CaseFormat.UPPER_CAMEL, segment);
```
carterkozak: and this?

dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/data/InstanceVariables.java
carterkozak: This metric-schema PR might be relevant here: https://github.com/palantir/metric-schema/pull/434

dialogue-annotations/src/main/java/com/palantir/dialogue/annotations/HeaderParamEncoder.java
```diff
@@ -0,0 +1,23 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.dialogue.annotations;
+
+import java.util.List;
+
+public interface HeaderParamEncoder<T> {
+    List<String> toHeaderParamValues(T value);
```
carterkozak: actually, wouldn't this apply to query parameters and header parameters? It's paths that are slightly different because there are no lists or optionals.

dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/data/InstanceVariables.java
jkozlowski: yeah, I noticed this bit doesn't quite work. I need to reread how to handle this properly

dialogue-annotations/src/main/java/com/palantir/dialogue/annotations/HeaderParamEncoder.java
jkozlowski: good point!

dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/data/InstanceVariables.java
jkozlowski: Ah, you already did, I'mma copy pasta.

dialogue-annotations/src/main/java/com/palantir/dialogue/annotations/HeaderParamEncoder.java
jkozlowski: Done

dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/data/CaseFormats.java
```diff
@@ -0,0 +1,69 @@
...
+
+import com.google.common.base.CaseFormat;
+import com.google.common.base.CharMatcher;
+import com.palantir.logsafe.Preconditions;
+import java.util.Optional;
+
+/**
+ * Utility functionality for handling guava {@link CaseFormat} values.
+ */
+final class CaseFormats {
```
carterkozak, line 27: I've written or fixed this code in a few projects (conjure-java, metric-schema, an internal config annotation processor). We should probably add a new oss library specifically around case formats and conversion between them... not today though.

jkozlowski: Not today, not today... maybe you fancy pushing this to guava?

dialogue-annotations-example/src/test/java/com/palantir/myservice/example/MyServiceIntegrationTest.java
```diff
@@ -151,13 +150,11 @@ public void testCustomResponse500() {
     }
 
     @Test
-    @Disabled
```
jkozlowski, line 154: 🍾

dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/data/CaseFormats.java
carterkozak: ha it has been suggested by others and shut down in guava ;-)

dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/data/ReturnTypesResolver.java
```diff
@@ -44,7 +44,7 @@ public ReturnTypesResolver(ResolverContext context) {
                 .deserializerFactory(maybeAcceptDeserializerFactory
                         .map(TypeName::get)
                         .orElseGet(() -> context.getTypeName(Json.class)))
-                .deserializerFieldName(endpointName.get() + "Deserializer")
+                .deserializerFieldName(InstanceVariables.joinCamelCase(endpointName.get(), "Deserializer"))
```
carterkozak, line 47: Out of curiosity, is `endpointName + "Deserializer"` ever incorrect?

dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/data/InstanceVariables.java
```diff
@@ -0,0 +1,42 @@
...
+
+    private InstanceVariables() {}
+
+    static String joinCamelCase(String... segments) {
+        return Joiner.on("")
+                .join(IntStream.range(0, segments.length)
+                        .mapToObj(i -> {
+                            String segment = segments[i];
+                            CaseFormat caseFormat =
+                                    CaseFormats.estimate(segment).get();
```
carterkozak, line 34: If estimation fails, should we return the input segment?

Could convert each segment to upper camel and convert the first letter of the result to lower. Doesn't make much difference.

dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/data/ReturnTypesResolver.java
jkozlowski: No. But for completeness, might as well have nice names. I don't think I validate endpoint names right now, I guess we have palantir-java-format so it should be guaranteed? Doesn't hurt to do this.

dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/data/InstanceVariables.java
jkozlowski: *I hope* that palantir-java-format saves us here? I guess this will be a pain to debug if someone actually hits this... I think I would probably prefer to validate this in the parsing stage, and throw exceptions here, as is?

dialogue-annotations/src/main/java/com/palantir/dialogue/annotations/ParamEncoder.java
```diff
@@ -0,0 +1,21 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.dialogue.annotations;
+
+public interface ParamEncoder<T> {
```
carterkozak, line 19: I wonder if this should extend ListParamEncoder with a default implementation that always returns one value?

dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/data/InstanceVariables.java
carterkozak: Heh, not everyone uses pjf yet, but I think most projects have some kind of style guidelines.
We'd need agreement between checkstyle, pjf, and this validation to stay somewhat in sync -- probably safe, but it would be painful if we discovered an edgecase.

carterkozak (approved):

dialogue-annotations/src/main/java/com/palantir/dialogue/annotations/ParamEncoder.java
jkozlowski: Do you mind throwing a quick PR together of what you mean? or we can pair

PR #1208, opened 2021-04-14
Dialogue uses tracing tags

carterkozak:
Depends on https://github.com/palantir/tracing-java/pull/711

## Before this PR
No tracing tags

## After this PR
==COMMIT_MSG==
Dialogue uses tracing tags
==COMMIT_MSG==

## Possible downsides?
More complex in the sampled case


dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueTracing.java
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    static TagTranslator<Throwable> failureTranslator(ImmutableMap<String, String> tags) {
+        return new TagTranslator<Throwable>() {
+
+            @Override
+            public <T> void translate(TagAdapter<T> sink, T target, Throwable response) {
+                sink.tag(target, tags);
+                sink.tag(target, "outcome", "failure");
+                sink.tag(target, "cause", response.getClass().getSimpleName());
```
jkozlowski, line 72: nice and small hopefully

dialogue-core/src/main/java/com/palantir/dialogue/core/TracedChannel.java
```diff
@@ -24,31 +26,37 @@
...
+    private static ImmutableMap<String, String> tracingTags(Config cf, Endpoint endpoint) {
+        return ImmutableMap.<String, String>builder()
+                .putAll(DialogueTracing.tracingTags(endpoint))
+                .putAll(DialogueTracing.tracingTags(cf))
+                .build();
     }
 
-    private static String meshSuffix(MeshMode meshMode) {
-        return meshMode == MeshMode.USE_EXTERNAL_MESH ? " (Mesh)" : "";
+    static EndpointChannel requestAttempt(Config cf, EndpointChannel delegate, Endpoint endpoint) {
```
jkozlowski, line 58: These are instantiated per request? Maybe we can drop to lower-level?

jkozlowski (commented):
One comment to address

dialogue-core/src/main/java/com/palantir/dialogue/core/TracedChannel.java
carterkozak: EndpointChannels are created once and reused.

feign recreates them per request since it dynamically builds endpoints, but most things are using full dialogue at this point.

jkozlowski (approved):

PR #1216, opened 2021-04-21
Dialogue uses standard tracing tags

carterkozak:
Unfortunately dialogue doesn't make it easy to discover the
endpoint path template, we may wish to expose it in the future.

==COMMIT_MSG==
Dialogue uses standard tracing tags
==COMMIT_MSG==


dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueTracing.java
```diff
@@ -57,7 +59,7 @@
                 sink.tag(target, tags);
                 int status = response.code();
                 sink.tag(target, "outcome", status / 100 == 2 ? "success" : "failure");
-                sink.tag(target, "status", Integer.toString(status));
+                sink.tag(target, TraceTags.HTTP_STATUS_CODE, Integer.toString(status));
```
jkozlowski, line 62: You could probably also centralize the int -> String for status codes? I think you have that multi-if-statement in abunch of places.

jkozlowski (approved):

PR #1217, opened 2021-04-21
Drop exceptional log level in ChannelCache under normal conditions

mglazer:
## Before this PR

An exception was logged all the time, even under healthy conditions.

## After this PR
==COMMIT_MSG==
Drop exceptional log level in ChannelCache under normal conditions
==COMMIT_MSG==

### DIFF
--- a/dialogue-clients/src/main/java/com/palantir/dialogue/clients/ChannelCache.java
+++ b/dialogue-clients/src/main/java/com/palantir/dialogue/clients/ChannelCache.java
@@ -80,14 +80,14 @@ final class ChannelCache {
                     "Created ChannelCache instance #{} ({} alive): {}",
                     SafeArg.of("instanceNumber", newCache.instanceNumber),
                     SafeArg.of("totalAliveNow", numLiveInstances),
-                    SafeArg.of("newCache", newCache),
-                    new SafeRuntimeException("ChannelCache constructed here"));
+                    SafeArg.of("newCache", newCache));
         } else if (numLiveInstances >= 1 && log.isDebugEnabled()) {
             log.debug(
                     "Created ChannelCache instance #{} ({} alive): {}",
                     SafeArg.of("instanceNumber", newCache.instanceNumber),
                     SafeArg.of("totalAliveNow", numLiveInstances),
-                    SafeArg.of("newCache", newCache));
+                    SafeArg.of("newCache", newCache),
+                    new SafeRuntimeException("ChannelCache constructed here"));
         }
 
         return newCache;
--- a/dialogue-clients/src/main/java/com/palantir/dialogue/clients/ChannelCache.java
+++ b/dialogue-clients/src/main/java/com/palantir/dialogue/clients/ChannelCache.java
@@ -75,13 +75,19 @@ final class ChannelCache {
         LIVE_INSTANCES.add(newCache);
 
         int numLiveInstances = LIVE_INSTANCES.size();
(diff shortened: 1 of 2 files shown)


dialogue-clients/src/main/java/com/palantir/dialogue/clients/ChannelCache.java
```diff
@@ -75,13 +75,19 @@ static ChannelCache createEmptyCache() {
...
                     SafeArg.of("instanceNumber", newCache.instanceNumber),
                     SafeArg.of("totalAliveNow", numLiveInstances),
                     SafeArg.of("newCache", newCache),
                     new SafeRuntimeException("ChannelCache constructed here"));
+        } else if (numLiveInstances >= 1 && log.isDebugEnabled()) {
+            log.debug(
+                    "Created ChannelCache instance #{} ({} alive): {}",
+                    SafeArg.of("instanceNumber", newCache.instanceNumber),
+                    SafeArg.of("totalAliveNow", numLiveInstances),
+                    SafeArg.of("newCache", newCache));
```
carterkozak: I think we still want the stack trace if debug is enabled

mglazer: I did the complete opposite of what I said I would do.

dialogue-clients/src/main/java/com/palantir/dialogue/clients/ChannelCache.java
```diff
@@ -75,8 +75,14 @@ static ChannelCache createEmptyCache() {
         LIVE_INSTANCES.add(newCache);
 
         int numLiveInstances = LIVE_INSTANCES.size();
-        if ((numLiveInstances > 1 && log.isInfoEnabled()) || log.isDebugEnabled()) {
+        if ((numLiveInstances >= 10 && log.isInfoEnabled())) {
             log.info(
+                    "Created ChannelCache instance #{} ({} alive): {}",
+                    SafeArg.of("instanceNumber", newCache.instanceNumber),
+                    SafeArg.of("totalAliveNow", numLiveInstances),
+                    SafeArg.of("newCache", newCache));
```
carterkozak: I think we still need a stack trace in this path, otherwise it's hard to figure out where these things are allocated. Also, if we have >=10 instances, updating the log level from INFO to DEBUG won't have any impact.

I think if debug is enabled, we record a stack trace in all cases, regardless of the `numLiveInstances` (just like before).

dialogue-clients/src/main/java/com/palantir/dialogue/clients/ChannelCache.java
```diff
@@ -75,8 +75,14 @@ static ChannelCache createEmptyCache() {
         LIVE_INSTANCES.add(newCache);
 
         int numLiveInstances = LIVE_INSTANCES.size();
-        if ((numLiveInstances > 1 && log.isInfoEnabled()) || log.isDebugEnabled()) {
```
carterkozak, line 78: I think we want to emulate replacing `numLiveInstances > 1` with `numLiveInstances >= 10` here, but improving it to log at debug level when debug is enabled and events wouldn't be recorded at info. Totally up to you if you want to include that part, always logging at info, but using `isDebugEnabled` as a sentinel (as before) is fine with me as well.

mglazer: How about this update?

carterkozak (approved):

PR #1222, opened 2021-04-23
dialogue-annotations supports MultipartRequestBody.

jkozlowski:
## Before this PR
No built-in support for multipart.

## After this PR
==COMMIT_MSG==
dialogue-annotations provides support for MultipartRequestBody.
==COMMIT_MSG==

## Possible downsides?

* Implementation just wraps apache-httpclient things. I do not feel like we should invest the effort to reimplement the serialization.


dialogue-annotations/src/main/java/com/palantir/dialogue/annotations/HttpEntityBodyRequestBodyAdapter.java
```diff
@@ -0,0 +1,64 @@
...
+import java.io.OutputStream;
+import org.apache.hc.core5.http.HttpEntity;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Adapts {@link HttpEntity} to a {@link RequestBody}.
+ */
+// Salt needs MIXED; other places use multipartBuilder.setType(MultipartBody.FORM);
+abstract class HttpEntityBodyRequestBodyAdapter implements RequestBody {
```
jkozlowski: I'm thinking we should do the same trick with FormUrlEncoded?

carterkozak: Don't we already have a pure form-encoded body implementation in WC?

dialogue-annotations/src/main/java/com/palantir/dialogue/annotations/MultipartRequestBody.java
```diff
@@ -0,0 +1,97 @@
...
+        public MultipartRequestBody build() {
+            return new MultipartRequestBody(builder.build());
+        }
+    }
+
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    public static RequestBodyPartBuilder requestBodyPartBuilder(RequestBody unsafeRequestBody) {
```
jkozlowski: Not convinced RequestBody is the right interface here; one client provides InputStreams (roughly), another Files. We should think if supporting Content-Length makes much sense

dialogue-annotations/src/main/java/com/palantir/dialogue/annotations/HttpEntityBodyRequestBodyAdapter.java
jkozlowski: yeah, that is defo a possibility. But thought might as well use the same trick for both of these. No preference though.

dialogue-annotations/src/main/java/com/palantir/dialogue/annotations/MultipartRequestBody.java
jkozlowski: In apache land, the repeatability is inferred by ```#getContentLength() != -1```

dialogue-annotations/src/main/java/com/palantir/dialogue/annotations/HttpEntityBodyRequestBodyAdapter.java
```diff
@@ -0,0 +1,63 @@
...
+import java.io.IOException;
+import java.io.OutputStream;
+import org.apache.hc.core5.http.HttpEntity;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Adapts {@link HttpEntity} to a {@link RequestBody}.
+ */
+abstract class HttpEntityBodyRequestBodyAdapter implements RequestBody {
```
jkozlowski: Thought's on separate subproject for these? I guess Json is already here...

dialogue-annotations/build.gradle
```diff
@@ -12,6 +12,12 @@ dependencies {
     implementation 'com.google.code.findbugs:jsr305'
     implementation 'com.palantir.safe-logging:preconditions'
     implementation 'com.palantir.safe-logging:safe-logging'
+    implementation 'com.google.guava:guava'
```
jkozlowski, line 15: Dunno why this needs to be here? I didn't add any guava things only to tests...

carterkozak:
Thoughts on a separate PR to update RequestBody with something like this? I considered it while working on the feign shim, but it was one-off enough that it wasn't worthwhile, but I think we have enough reason at this point.

```java
iface RequestBody {
default OptionalLong contentLength() {
  return OptionalLong.empty();
}
```

dialogue-annotations/src/main/java/com/palantir/dialogue/annotations/HttpEntityBodyRequestBodyAdapter.java
carterkozak: I don't have a strong opinion either way -- I suppose the question is whether we'd expect these to ever be used without the dialogue annotation processor? I'd hope not, so this is a reasonable enough place for them to live.

dialogue-annotations/build.gradle
carterkozak: huh, I'm not sure.

carterkozak: huh, I'm not sure.

dialogue-annotations/src/main/java/com/palantir/dialogue/annotations/HttpEntityBodyRequestBodyAdapter.java
jkozlowski: I was going to call it dialogue-annotation-processor-extras or something, just to lighten the dependency. But then I realized that people already depend on apache, so it's a moot point. 

>  I'd hope not, so this is a reasonable enough place for them to live.

No, we do not want anyone to use this outside of annotation processor.

dialogue-annotations/src/test/java/com/palantir/dialogue/annotations/MultipartRequestBodyTest.java
```diff
@@ -0,0 +1,333 @@
...
+        String bucket();
+
+        String value();
+
+        Map<String, String> keyValues();
+
+        String contentType();
+    }
+
+    private RequestBody unknownLengthRequestBody(byte[] value, okhttp3.MediaType contentType) {
```
jkozlowski, line 327: So all this ceremony is because apache does not set Content-Length header for each individual part. as discussed with @carterkozak it doesn't feel like a problem, so we'll have it on our ContentBody interface, in case it's supported in the future, but unlikely to be a big deal for anyone that has to use this.

dialogue-annotations/src/main/java/com/palantir/dialogue/annotations/ContentBody.java
```diff
@@ -0,0 +1,42 @@
...
+import java.io.OutputStream;
+import java.util.OptionalLong;
+
+public interface ContentBody extends Closeable {
+
+    /** The content of this request body, possibly empty. */
+    void writeTo(OutputStream output) throws IOException;
+
+    /** A HTTP content type (e.g., "application/json") indicating the type of content. */
+    String contentType();
```
jkozlowski, line 32: Ok, so another quirk of apache implementation of this, is that #isRepeatable is based on getContentLength() != 1; interestingly we have a an internal client that deals with InputStreams that it may arbitrarily stream to disk, to make sure it can get retries. But I think that might be ameanable to providing the contentLength of those values, to get repeatability.

carterkozak: We don't currently provide repeatability for `BinaryRequestBody` binary requests, it's reasonable not to support it in this case when lengths aren't provided -- I would not want to record contents to disk, as disk performance can be finicky, and space may not be available.

jkozlowski: Yeah, I agree it's reasonable. Internal client likely was written because it was dealing with finnicky upstreams, so the way it deals with that is write to disk first, then send to itself. Anyway, feels like we can support what they need, so shouldn't be a big deal.

dialogue-annotations/src/main/java/com/palantir/dialogue/annotations/ContentBody.java
```diff
@@ -0,0 +1,63 @@
...
+     * preferring to catch and log.
+     */
+    @Override
+    void close();
+
+    static ContentBody file(String contentType, Path filePath) {
+        return new com.palantir.dialogue.annotations.ContentBody() {
+            @Override
+            public void writeTo(OutputStream output) throws IOException {
+                Files.copy(filePath, output);
```
carterkozak: `contentLength`?

carterkozak: Probably worth a similar factory for InputStream (non-repeatable)

dialogue-annotations/src/main/java/com/palantir/dialogue/annotations/ContentBody.java
```diff
@@ -0,0 +1,63 @@
...
+package com.palantir.dialogue.annotations;
+
+import java.io.Closeable;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.OptionalLong;
+
+public interface ContentBody extends Closeable {
```
carterkozak, line 26: We might as well add a `repeatable` flag while we're here rather than relying on contentLength as a sentinel.

dialogue-annotations/src/main/java/com/palantir/dialogue/annotations/MultipartRequestBody.java
```diff
@@ -0,0 +1,153 @@
...
+        private final ContentBody unsafeRequestBody;
+
+        @Nullable
+        private String fileName;
+
+        private ContentBodyAdapter(ContentBody requestBody) {
+            super(Preconditions.checkNotNull(
+                    ContentType.parse(requestBody.contentType()),
+                    "Invalid content type",
+                    UnsafeArg.of("contentType", requestBody.contentType())));
```
carterkozak: content-type should be safe, even if invalid.

carterkozak:
generally lgtm, just a couple comments

dialogue-annotations/src/main/java/com/palantir/dialogue/annotations/ContentBody.java
jkozlowski: I thought about it, but it felt racey? But I think there's enough precent to suggest this is fine.

dialogue-annotations/src/main/java/com/palantir/dialogue/annotations/ContentBody.java
```diff
@@ -0,0 +1,63 @@
...
+    default OptionalLong contentLength() {
+        return OptionalLong.empty();
+    }
+
+    /**
+     * Closes this {@link ContentBody} and releases all resources. Calling {@link #close()} should never throw,
+     * preferring to catch and log.
+     */
+    @Override
+    void close();
```
jkozlowski, line 46: I didn't want to break ranks here for reasons of consistency, but this feels weird, to have the logging in a bunch of places, as opposed to 1. Yes, the log message would be more generic, but the stacktrace should already capture everything you need to know about the type of resource, so not sure what we're getting by pushing this onto impls.

dialogue-annotations/src/main/java/com/palantir/dialogue/annotations/MultipartRequestBody.java
```diff
@@ -0,0 +1,153 @@
...
+
+        @Override
+        @Nullable
+        public String getFilename() {
+            return fileName;
+        }
+
+        @Override
+        public void writeTo(OutputStream out) throws IOException {
+            try (ContentBody requestBody = unsafeRequestBody) {
```
jkozlowski: Just realized this isn't going to work so well, if we want repeatability.

dialogue-annotations/src/main/java/com/palantir/dialogue/annotations/MultipartRequestBody.java
jkozlowski: Fixed

dialogue-annotations/src/main/java/com/palantir/dialogue/annotations/ContentBody.java
jkozlowski: Added and contentLength is pulled out of the File.

dialogue-annotations/src/main/java/com/palantir/dialogue/annotations/ContentBody.java
jkozlowski: Done

dialogue-annotations/src/main/java/com/palantir/dialogue/annotations/MultipartRequestBody.java
jkozlowski: Done

dialogue-annotations/src/main/java/com/palantir/dialogue/annotations/MultipartRequestBody.java
jkozlowski: Fixed.

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientBlockingChannel.java
```diff
@@ -177,24 +176,7 @@ private static boolean requiresEmptyBody(Endpoint endpoint) {
...
-            String contentLengthValue = contentLengthHeader.getValue();
-            try {
-                return OptionalLong.of(Long.parseLong(contentLengthValue));
-            } catch (NumberFormatException nfe) {
-                log.warn(
-                        "Failed to parse content-length value '{}'",
-                        SafeArg.of(HttpHeaders.CONTENT_LENGTH, contentLengthValue),
-                        nfe);
-            }
-        }
```
carterkozak: I think the feign shim relies on this logic, we should support Content-Length in both places at least for a few releases

jkozlowski: Ack, I shall split a separate PR as discussed.

carterkozak (approved):

PR #1224, opened 2021-04-26
Bring back unique tracing operation names per endpoint

carterkozak:
These were lost in #1208 when we introduced tracing tags, however
they the UI doesn't provide tags in trace graphs, making the data
less valuable.

==COMMIT_MSG==
Bring back unique tracing operation names per endpoint
==COMMIT_MSG==


dialogue-core/src/main/java/com/palantir/dialogue/core/TracedChannel.java
```diff
@@ -45,7 +45,13 @@ private TracedChannel(EndpointChannel delegate, String operationName, ImmutableM
     }
 
     static EndpointChannel create(Config cf, EndpointChannel delegate, Endpoint endpoint) {
-        return new TracedChannel(delegate, "Dialogue: request", tracingTags(cf, endpoint));
+        String operationName =
+                "Dialogue: request " + endpoint.serviceName() + "#" + endpoint.endpointName() + meshSuffix(cf.mesh());
```
smoorpal, line 49: What is the point of the `meshSuffix`?

carterkozak: Makes it obvious when we're running on the service mesh, since the behavior is likely to be very different if retries are handled by an intermediate proxy.

smoorpal (approved):

jkozlowski (approved):

PR #1226, opened 2021-04-27
Dialogue ErrorDecoder provides additional response diagnostic info

carterkozak:
==COMMIT_MSG==
Dialogue ErrorDecoder provides additional response diagnostic info
==COMMIT_MSG==

### DIFF
--- a/dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ErrorDecoder.java
+++ b/dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ErrorDecoder.java
@@ -17,7 +17,9 @@
 package com.palantir.conjure.java.dialogue.serde;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.common.collect.ImmutableList;
 import com.google.common.io.CharStreams;
+import com.google.common.net.HttpHeaders;
 import com.google.common.primitives.Longs;
 import com.palantir.conjure.java.api.errors.QosException;
 import com.palantir.conjure.java.api.errors.RemoteException;
@@ -25,7 +27,11 @@ import com.palantir.conjure.java.api.errors.SerializableError;
 import com.palantir.conjure.java.api.errors.UnknownRemoteException;
 import com.palantir.conjure.java.serialization.ObjectMappers;
 import com.palantir.dialogue.Response;
+import com.palantir.logsafe.Arg;
+import com.palantir.logsafe.SafeArg;
+import com.palantir.logsafe.SafeLoggable;
 import com.palantir.logsafe.UnsafeArg;
+import com.palantir.logsafe.exceptions.SafeExceptions;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
@@ -34,8 +40,8 @@ import java.net.MalformedURLException;
 import java.net.URL;
 import java.nio.charset.StandardCharsets;
 import java.time.Duration;
+import java.util.List;
 import java.util.Optional;
-import javax.ws.rs.core.HttpHeaders;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -56,6 +62,15 @@ enum ErrorDecoder {
     }
 
     RuntimeException decode(Response response) {
+        if (log.isDebugEnabled()) {
+            log.debug("Received an error response", diagnosticArgs(response).toArray(new Object[0]));
+        }
+        RuntimeException result = decodeInternal(response);
+        result.addSuppressed(diagnostic(response));
+        return result;
+    }
+
+    private RuntimeException decodeInternal(Response response) {
         // TODO(rfink): What about HTTP/101 switching protocols?
         // TODO(rfink): What about HEAD requests?
 
@@ -96,6 +111,7 @@ enum ErrorDecoder {
         } catch (NullPointerException | IOException e) {
             UnknownRemoteException exception = new UnknownRemoteException(code, "<unparseable>");
             exception.initCause(e);
+            exception.addSuppressed(diagnostic(response));
             return exception;
         }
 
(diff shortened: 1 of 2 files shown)


dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ErrorDecoder.java
```diff
@@ -117,4 +133,54 @@ private static String toString(InputStream body) throws IOException {
...
+        recordHeader("x-envoy-response-flags", response, args);
+        recordHeader("x-envoy-response-code-details", response, args);
+        return args.build();
+    }
+
+    private static void recordHeader(String header, Response response, ImmutableList.Builder<Arg<?>> args) {
+        response.getFirstHeader(header).ifPresent(server -> args.add(SafeArg.of(header, server)));
+    }
+
+    private static final class ResponseDiagnostic extends RuntimeException implements SafeLoggable {
```
jkozlowski, line 156: Looks like second time we're using this trick, shall we extract a little util thingie

carterkozak: They're used in separate modules, I don't think we want to expose public API for something this simple.

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ErrorDecoder.java
```diff
@@ -117,4 +133,54 @@ private static String toString(InputStream body) throws IOException {
...
+
+    private static ImmutableList<Arg<?>> diagnosticArgs(Response response) {
+        ImmutableList.Builder<Arg<?>> args = ImmutableList.<Arg<?>>builder().add(SafeArg.of("status", response.code()));
+        recordHeader(HttpHeaders.SERVER, response, args);
+        recordHeader(HttpHeaders.CONTENT_TYPE, response, args);
+        recordHeader(HttpHeaders.CONTENT_LENGTH, response, args);
+        recordHeader(HttpHeaders.CONNECTION, response, args);
+        recordHeader(HttpHeaders.DATE, response, args);
+        recordHeader("x-envoy-response-flags", response, args);
+        recordHeader("x-envoy-response-code-details", response, args);
```
carterkozak, line 148: Pending better header names, but we can ship these for now given they're already in use.

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ErrorDecoder.java
```diff
@@ -96,6 +111,7 @@ RuntimeException decode(Response response) {
         } catch (NullPointerException | IOException e) {
```
jkozlowski: What about on 308/429 etc.? Those don't send those headers?

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ErrorDecoder.java
```diff
@@ -56,6 +62,15 @@ boolean isError(Response response) {
     }
 
     RuntimeException decode(Response response) {
+        if (log.isDebugEnabled()) {
+            log.debug("Received an error response", diagnosticArgs(response).toArray(new Object[0]));
```
jkozlowski, line 66: Why the double logging on debug?

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ErrorDecoder.java
```diff
@@ -117,4 +133,54 @@ private static String toString(InputStream body) throws IOException {
...
         }
     }
+
+    private static ResponseDiagnostic diagnostic(Response response) {
+        return new ResponseDiagnostic(diagnosticArgs(response));
+    }
+
+    private static ImmutableList<Arg<?>> diagnosticArgs(Response response) {
+        ImmutableList.Builder<Arg<?>> args = ImmutableList.<Arg<?>>builder().add(SafeArg.of("status", response.code()));
+        recordHeader(HttpHeaders.SERVER, response, args);
```
jkozlowski, line 142: Is this safe?

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ErrorDecoder.java
carterkozak: This gives us a single place where we can enable logging instead of checking all possible places that can catch an exception.

jkozlowski (approved):
Just some questions

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ErrorDecoder.java
```diff
@@ -96,6 +111,7 @@ RuntimeException decode(Response response) {
         } catch (NullPointerException | IOException e) {
             UnknownRemoteException exception = new UnknownRemoteException(code, "<unparseable>");
             exception.initCause(e);
+            exception.addSuppressed(diagnostic(response));
```
carterkozak: woops, this is duplicated above
```suggestion
```

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ErrorDecoder.java
carterkozak: Yep, just the server name, WC uses the local service UserAgent, nginx uses `nginx`.

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ErrorDecoder.java
carterkozak: We're adding the diagnostic info above in `decode` which delegates to this method, `decodeInternal`. Diagnostic data is applied to all exceptions returned by this method.

PR #1227, opened 2021-04-28
Support RequestBody#contentLength.

jkozlowski:
## Before this PR

No way to specify content length of body directly.

## After this PR
==COMMIT_MSG==
RequestBody supports optional content length.
==COMMIT_MSG==

## Possible downsides?

More API. It's kinda already API, for feign replacement, so likely makes sense to support directly. Also useful for https://github.com/palantir/dialogue/pull/1222


carterkozak (approved):

dialogue-core/src/test/resources/tracing/DialogueChannelTest/traces_on_retries.log
```diff
@@ -1,30 +1,31 @@
...
+{"traceId":"3519228122fee79e","parentSpanId":"b9eede02cfc9a230","spanId":"fc414565b6a32ddd","type":"LOCAL","operation":"Dialogue-request-attempt initial","startTimeMicroSeconds":1619611357982666,"durationNanoSeconds":1428971,"metadata":{}}
+{"traceId":"3519228122fee79e","parentSpanId":"76a3f96c2bd32ebb","spanId":"257e95ea327f70b4","type":"LOCAL","operation":"Dialogue-RetryingChannel","startTimeMicroSeconds":1619611357020816,"durationNanoSeconds":964010756,"metadata":{"serviceName":"service","endpointName":"endpoint","failures":"5","channel":"my-channel"}}
+{"traceId":"3519228122fee79e","parentSpanId":null,"spanId":"4791d408917316f4","type":"LOCAL","operation":"Dialogue: request service#endpoint","startTimeMicroSeconds":1619611357020205,"durationNanoSeconds":970171438,"metadata":{"endpointService":"service","endpointName":"endpoint","http.method":"POST","channel":"my-channel","mesh":"false","outcome":"failure","http.status_code":"429"}}
+{"traceId":"3519228122fee79e","parentSpanId":"866a99f10ab0c9b9","spanId":"cff8c71cc0a4de65","type":"LOCAL","operation":"dialogue-RetryingChannel-scheduler","startTimeMicroSeconds":1619611357982457,"durationNanoSeconds":8481191,"metadata":{}}
```
jkozlowski, line 31: Thoughts on what happened?

jkozlowski: Feels like this should've been there from the start?

carterkozak: My guess is that the scheduler runnable is still executing while the test completes, which allows the test to record spans up to, but not including the final `dialogue-RetryingChannel-scheduler`.
In both before and after, we have the correct number of `Dialogue-request-attempt` and `Dialogue-request-attempt initial` spans.

PR #1228, opened 2021-04-29
Support application/x-www-form-urlencoded.

jkozlowski:
## After this PR
==COMMIT_MSG==
Support application/x-www-form-urlencoded.
==COMMIT_MSG==

### DIFF
--- a/dialogue-annotations/src/main/java/com/palantir/dialogue/annotations/FormUrlEncodedSerializerFactory.java
+++ b/dialogue-annotations/src/main/java/com/palantir/dialogue/annotations/FormUrlEncodedSerializerFactory.java
@@ -26,6 +26,7 @@ import java.util.Map;
  */
 public abstract class FormUrlEncodedSerializerFactory<T> extends StdSerializer<T> {
 
+    @Override
     public final RequestBody serialize(T value) {
         return new FormUrlEncodedRequestBody(extractFields(value));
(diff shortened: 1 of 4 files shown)


dialogue-annotations/src/main/java/com/palantir/dialogue/annotations/FormUrlEncodedSerializerFactory.java
```diff
@@ -0,0 +1,36 @@
...
+
+import com.palantir.dialogue.RequestBody;
+import java.util.Map;
+
+/**
+ * Extend this class in order to serialize {@link T} to {@code application/x-www-form-urlencoded} request body.
+ *
+ * @param <T> type of value to serialize
+ */
+public abstract class FormUrlEncodedSerializerFactory<T> extends StdSerializer<T> {
```
carterkozak, line 27: I’d prefer composition over inheritance for this sort of thing, but the abstract class is probably most obvious for consumers.

jkozlowski: I initially just pushed the FormUrlEncodedRequestBody, but then realized that everyone will need this snippet so might as well.

carterkozak (approved):

PR #1229, opened 2021-04-30
[Experiment] Use Request#attachments to associate requests coming from same sticky channel instance.

jkozlowski:
## Before this PR

Clients using sticky channels have no way of enforcing fairness between "transactions" (that's what we use sticky channels to approximate).

## After this PR
==COMMIT_MSG==
StickyChannels provide fairness.
==COMMIT_MSG==

## Possible downsides?

* Queue implementation uses synchronized, so not good, but probably given enough brain cycles can come up with something non-blocking.
* Need to think how to write a test that this actually does what I expect.

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
@@ -393,6 +393,8 @@ final class QueuedChannel implements Channel {
     // WARNING: Not final implementation, proof of concept, hence using global locking.
     private static final class DeferredQueueImpl implements DeferredQueue {
 
+        private static final UUID GLOBAL_QUEUE = UUID.randomUUID();
+
         private final Map<UUID, Deque<DeferredCall>> queues = new LinkedHashMap<>();
 
         @Override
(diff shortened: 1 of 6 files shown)


dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
```diff
@@ -334,14 +338,6 @@ static Builder builder() {
         }
     }
 
-    private static final class ProtectedConcurrentLinkedDeque<T> extends ConcurrentLinkedDeque<T> {
```
jkozlowski, line 337: Would argue that this is actually a nice refactoring in and of itself!

jkozlowski:
Discussed, will try to summarise it here:

* Limiting key attachment feels like a good mechanism for propagating that.
* Carter has a preference towards the queuing implementation where we give each transaction a max num of concurrent requests (this is basically a scaled number based on the current concurrency limiter value), as opposed to building up a large queue if a single thread kicks off 10k request. I am on board with that.
* Then we'd actually probably want to either hide the transaction id generation here based on sticky channels, OR preference would be to make this an explicit feature of the async channels: where they would have an extra method:

```
// Name to be bikeshedded.
MyServiceAsync createScopedClient();
```

When async methods are called on an instance of this client, they will be treated as being related, and therefore as a key for fairness consideration.

The actual blocker here is the fact that sticky and non-sticky channels recreate per-host channels therefore there is 0 information sharing between them: meaning that even if we were to have fairness implemented, a fair and non-fair client could still railroad each other.

This should be the first thing we should attempt to fix, whilst concurrently looking into the actual implementation of the feature.

jkozlowski:
This approach is one solution to the puzzle, closing for now.

PR #1231, opened 2021-05-03
Enable incremental annotation processing

pkoenig10:
## Before this PR
The Dialogue annotation processor does not support incremental compilation.

## After this PR
The Dialogue annotation processor does support incremental compilation.


dialogue-annotations-processor/build.gradle
```diff
@@ -30,13 +30,16 @@ dependencies {
...
+    annotationProcessor 'net.ltgt.gradle.incap:incap-processor'
+    compileOnly 'net.ltgt.gradle.incap:incap'
     annotationProcessor 'org.immutables:value'
     compileOnly 'org.immutables:value::annotations'
     annotationProcessor 'org.derive4j:derive4j'
     compileOnly 'org.derive4j:derive4j-annotation'
 
     testAnnotationProcessor 'org.immutables:value'
     testCompileOnly 'org.immutables:value::annotations'
+    testCompileOnly 'net.ltgt.gradle.incap:incap'
```
pkoenig10, line 42: Necessary because the `IncrementalAnnotationProcessor` annotation has class retention. This is acceptable because downstream consumers do not compile against the annotation processor, see https://github.com/tbroyer/gradle-incap-helper#usage.

carterkozak (approved):

jkozlowski:
Thanks! How does this actually work, and where was this causing issues?

PR #1233, opened 2021-05-04
Support custom encoders for primitive types

pkoenig10:
## Before this PR
Using a custom encoder with a primitive types fails with `NoSuchElementException`:
```
at java.base/java.util.Optional.orElseThrow(Optional.java:375)
  	at com.palantir.dialogue.annotations.processor.generate.ServiceImplementationGenerator.underlyingCustomType(ServiceImplementationGenerator.java:200)
  	at com.palantir.dialogue.annotations.processor.generate.ServiceImplementationGenerator.encoder(ServiceImplementationGenerator.java:188)
  	at com.palantir.dialogue.annotations.processor.generate.ServiceImplementationGenerator.lambda$generate$1(ServiceImplementationGenerator.java:78)
  	at java.base/java.util.Optional.map(Optional.java:258)
  	at com.palantir.dialogue.annotations.processor.generate.ServiceImplementationGenerator.lambda$generate$2(ServiceImplementationGenerator.java:78)
  	at com.palantir.dialogue.annotations.processor.data.ParameterTypes$LambdaCases.header(ParameterTypes.java:228)
  	at com.palantir.dialogue.annotations.processor.data.ParameterTypes$Header.match(ParameterTypes.java:280)
  	at com.palantir.dialogue.annotations.processor.data.ParameterTypes$CaseOfMatchers$PartialMatcher.otherwise(ParameterTypes.java:739)
  	at com.palantir.dialogue.annotations.processor.data.ParameterTypes$CaseOfMatchers$PartialMatcher.otherwise_(ParameterTypes.java:743)
  	at com.palantir.dialogue.annotations.processor.generate.ServiceImplementationGenerator.lambda$generate$7(ServiceImplementationGenerator.java:81)
  	at java.base/java.util.stream.ReferencePipeline$7$1.accept(ReferencePipeline.java:271)
  	at java.base/java.util.Spliterators$ArraySpliterator.forEachRemaining(Spliterators.java:948)
  	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:484)
  	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:474)
  	at java.base/java.util.stream.ForEachOps$ForEachOp.evaluateSequential(ForEachOps.java:150)
  	at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.evaluateSequential(ForEachOps.java:173)
  	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
  	at java.base/java.util.stream.ReferencePipeline.forEach(ReferencePipeline.java:497)
  	at com.palantir.dialogue.annotations.processor.generate.ServiceImplementationGenerator.lambda$generate$8(ServiceImplementationGenerator.java:83)
  	at com.google.common.collect.ImmutableList.forEach(ImmutableList.java:406)
  	at com.palantir.dialogue.annotations.processor.generate.ServiceImplementationGenerator.generate(ServiceImplementationGenerator.java:73)
```

For context, I'm using Dialogue annotations to define a service interface for testing and want to define something like:
```java
@Request(method = HttpMethod.GET, path = "/foo")
void foo(
    @Header(value = "Cookie", encoder = CookieTokenEncoder.class) BearerToken token,
```
```java
final class CookieTokenEncoder implements ListParamEncoder<BearerToken> {
    @Override
    public List<String> toParamValues(BearerToken token) {
        return List.of("TOKEN=" + token);
    }
}
```

I could wrap it in a custom class, but there doesn't seem to be much benefit to forcing consumers to do that.

## After this PR
Custom encoders can be used with primitive types.

## Possible downsides?
Is there a reason we explicitly did not want to support this? Is there a benefit to only allowing one serialization format for primitive types, even when manually using the Dialogue annotations?

### DIFF
--- a/dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/generate/ServiceImplementationGenerator.java
+++ b/dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/generate/ServiceImplementationGenerator.java
@@ -194,6 +194,7 @@ public final class ServiceImplementationGenerator {
 
     private TypeName underlyingCustomType(ArgumentType argumentType) {
         return ArgumentTypes.caseOf(argumentType)
+                .primitive((javaTypeName, _unused) -> javaTypeName)
                 .optional((_optionalJavaType, optionalType) -> underlyingCustomType(optionalType.underlyingType()))
                 .customType(Function.identity())
                 .otherwiseEmpty()
@@ -277,14 +278,16 @@ public final class ServiceImplementationGenerator {
         return type.match(new ArgumentType.Cases<>() {
             @Override
             public Optional<CodeBlock> primitive(TypeName _unused, String parameterSerializerMethodName) {
-                return Optional.of(CodeBlock.of(
-                        "$L.$L($S, $L.$L($L));",
-                        REQUEST,
-                        singleValueMethod,
-                        key,
-                        PARAMETER_SERIALIZER,
-                        parameterSerializerMethodName,
-                        argName));
+                return Optional.of(maybeParameterEncoderType
+                        .map(this::customType)
+                        .orElseGet(() -> CodeBlock.of(
+                                "$L.$L($S, $L.$L($L));",
+                                REQUEST,
+                                singleValueMethod,
+                                key,
+                                PARAMETER_SERIALIZER,
+                                parameterSerializerMethodName,
+                                argName)));
             }
 
             @Override
@@ -310,37 +313,35 @@ public final class ServiceImplementationGenerator {
 
             @Override
             public Optional<CodeBlock> customType(TypeName _typeName) {
-                if (maybeParameterEncoderType.isEmpty()) {
-                    return Optional.empty();
-                }
+                return maybeParameterEncoderType.map(this::customType);
+            }
 
-                return maybeParameterEncoderType.flatMap(parameterEncoderType -> parameterEncoderType
-                        .type()
-                        .match(new EncoderType.Cases<Optional<CodeBlock>>() {
-                            @Override
-                            public Optional<CodeBlock> param() {
-                                return Optional.of(CodeBlock.of(
-                                        "$L.$L($S, $L.$L($L));",
-                                        REQUEST,
-                                        singleValueMethod,
-                                        key,
-                                        parameterEncoderType.encoderFieldName(),
-                                        parameterEncoderType.encoderMethodName(),
-                                        argName));
-                            }
+            private CodeBlock customType(ParameterEncoderType parameterEncoderType) {
+                return parameterEncoderType.type().match(new EncoderType.Cases<>() {
+                    @Override
+                    public CodeBlock param() {
+                        return CodeBlock.of(
+                                "$L.$L($S, $L.$L($L));",
+                                REQUEST,
+                                singleValueMethod,
+                                key,
+                                parameterEncoderType.encoderFieldName(),
+                                parameterEncoderType.encoderMethodName(),
+                                argName);
+                    }
 
-                            @Override
-                            public Optional<CodeBlock> listParam() {
-                                return Optional.of(CodeBlock.of(
-                                        "$L.$L($S, $L.$L($L));",
-                                        REQUEST,
-                                        multiValueMethod,
-                                        key,
-                                        parameterEncoderType.encoderFieldName(),
-                                        parameterEncoderType.encoderMethodName(),
-                                        argName));
-                            }
-                        }));
+                    @Override
+                    public CodeBlock listParam() {
+                        return CodeBlock.of(
+                                "$L.$L($S, $L.$L($L));",
+                                REQUEST,
+                                multiValueMethod,
+                                key,
+                                parameterEncoderType.encoderFieldName(),
+                                parameterEncoderType.encoderMethodName(),
+                                argName);
+                    }
+                });
             }
         });
     }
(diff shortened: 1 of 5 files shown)


dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/generate/ServiceImplementationGenerator.java
```diff
@@ -310,37 +313,35 @@ private TypeName underlyingCustomType(ArgumentType argumentType) {
...
-                                        "$L.$L($S, $L.$L($L));",
-                                        REQUEST,
-                                        multiValueMethod,
-                                        key,
-                                        parameterEncoderType.encoderFieldName(),
-                                        parameterEncoderType.encoderMethodName(),
-                                        argName));
-                            }
-                        }));
+                return maybeParameterEncoderType.map(this::customType);
```
pkoenig10, line 316: Is there a reason `generatePlainSerializer` returns an optional? The only empty return value is here and it seems like it would be better to throw if a parameter encoder is not defined for a custom type.

jkozlowski: Yes, so that I could build this incrementally, and ignore stuff that wasn't handled :) I think at this point the Optional could be removed.

dialogue-annotations-processor/src/test/java/com/palantir/myservice/service/MyCustomStringParameterEncoder.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+package com.palantir.myservice.service;
+
+import com.palantir.dialogue.annotations.ListParamEncoder;
+import com.palantir.dialogue.annotations.ParamEncoder;
+import java.util.Collections;
+import java.util.List;
+
+public final class MyCustomStringParameterEncoder implements ParamEncoder<String>, ListParamEncoder<String> {
```
jkozlowski, line 24: So my first gut reaction is that we should not do this, the wrapping for a custom type costs basically nothing. Certainly for Strings it just feels wrong.

Maybe for something like BearerToken that might make sense, but maybe that just means it's not a primitive/out of the box type?

jkozlowski:
@pkoenig10 perhaps with your example, actual out of the box support for cookies is what we want?

pkoenig10:
Maybe? Given that cookies are just headers though, I'm not sure we should special-case them in that way.

If we do add support cookies, I think we'd want to keep the cookie -> header conversation code inside the annotation processor.

pkoenig10:
But there's probably not a valid use case for this outside of cookies, so adding cookie support is probably the right approach.

I'll put up a PR to see what this would look like.

jkozlowski:
Thanks! My knowledge of cookies is rather rusty, circa 2000, when trying to write some web apps with php. So not clear if that is a good idea or not, but if the serialization format is fixed for cookies, I think out of the box support might be nice.

carterkozak:
I had thought we supported cookie auth in generated dialogue clients, but a quick grep doesn't look like it was ever implemented.
Custom serializers for conjure primitives sound reasonable to me, I could see a string->string case in which we normalize values in some way (`.toUpperCase(Locale.ENGLISH)`).
First-class cookie support sounds great as well, but I'd prefer to start by adding support for cookie auth based on conjure, then implementing it here to avoid duplication.

carterkozak:
Actually, we do support cookie auth out of the box, see https://github.com/palantir/conjure-java/blob/9d0e994aaf5b923bd264923f21ed3b14b523757f/conjure-java-core/src/integrationInput/java/com/palantir/product/CookieServiceBlocking.java#L45

pkoenig10:
@carterkozak Do you have a preference between cookie params and custom encoder?

Given that was already support cookie params in Conjure, that seems like the more appropriate approach here. Because we don't, for example, let you define custom encoders in Conjure.

jkozlowski:
> Custom serializers for conjure primitives sound reasonable to me, I could see a string->string case in which we normalize values in some way (.toUpperCase(Locale.ENGLISH)).

See that's where I feel like a custom alias is a better solution, you're doing some sort of operation on the Strings, so it should probably be specifically identified in your code. Also, why would you need to uppercase your strings?

carterkozak:
> Given that was already support cookie params in Conjure, that seems like the more appropriate approach here. Because we don't, for example, let you define custom encoders in Conjure.

I think we only support cookie auth BearerToken values in conjure, but they use `Cookie: <NAME>=<BEARER_TOKEN>`. Cookies can get difficult if we want to support multiple cookies at once because they must be merged into a single header value.

>See that's where I feel like a custom alias is a better solution, you're doing some sort of operation on the Strings

I'm not sure I agree with this for the annotation processor, the point is not to lock users into the conjure opinions -- we shouldn't force types to be wrapped in order to provide custom encoding, that may expose implementation details to those who consume the client (not necessarily the same folks who wrote the client).

> so it should probably be specifically identified in your code.

Isn't this what the annotations encoder value does?

> Also, why would you need to uppercase your strings?

A contrived example, but it's not so much different from the auth cookie bearertoken.

carterkozak:
(accidental close -- reopened)

pkoenig10:
> Cookies can get difficult if we want to support multiple cookies at once because they must be merged into a single header value.

I was planning to only support a single value. I don't think there are many legitimate use cases for sending multiple cookie values and limiting to a single value better aligns with the cookie support in Conjure.

Unless you think there's a good chance we'd regret that limitation?

carterkozak:
> Do you have a preference between cookie params and custom encoder?

I think we want both, custom encoders are likely easiest to do first and unblock.

jkozlowski:
@pkoenig10 for posterity, can you link where you need this for btw?

jkozlowski (approved):

carterkozak:
> Unless you think there's a good chance we'd regret that limitation?

I imagine we may end up with both cookie auth and a second cookie header, no?

pkoenig10:
Yeah I realized this doesn't quite work. You either need to limit endpoints to 1 cookie parameter, which feels a bit limiting, or add the logic to merge cookies, which is fairly complex. Going to punt this for now.

PR #1240, opened 2021-05-06
Provide originating element for generated service factory

pkoenig10:
https://github.com/palantir/dialogue/pull/1231 didn't quite work to enable incremental compilation because Gradle requires that each generated file has exactly one originating element.
https://docs.gradle.org/current/userguide/java_plugin.html#sec:incremental_annotation_processing
```
> Task :compileJava
Full recompilation is required because the generated type 'com.palantir.foo.api.FooServiceDialogueServiceFactory' must have exactly one originating element, but had 0.
```
This PR updates the Dialogue annotation processor to provide the enclosing interface elements as the originating element to the `Filer` API.

### DIFF
--- a/dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/DialogueRequestAnnotationsProcessor.java
+++ b/dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/DialogueRequestAnnotationsProcessor.java
@@ -113,7 +113,7 @@ public final class DialogueRequestAnnotationsProcessor extends AbstractProcessor
                     }
 
                     try {
-                        Goethe.formatAndEmit(javaFile, filer);
+                        Goethe.formatAndEmit(javaFile, filer, interfaceElement);
                     } catch (FilerException e) {
                         // Happens when same file is written twice. Dunno why this is a problem
                     } catch (IOException e) {
--- a/dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/format/Goethe.java
+++ b/dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/format/Goethe.java
@@ -28,6 +28,7 @@ import java.io.IOException;
 import java.io.Writer;
 import java.util.List;
 import javax.annotation.processing.Filer;
+import javax.lang.model.element.Element;
 import javax.tools.JavaFileObject;
 
 public final class Goethe {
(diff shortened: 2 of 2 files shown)


pkoenig10:
I've verified locally that this works as expected.

jkozlowski (approved):
Thanks

jkozlowski:
For reference: what does this incremental compilation actually help us with? do we cache java build results and restore them in Circle? Or is the win here in IntelliJ and it's somehow plumbed fully through there?

jkozlowski:
@pkoenig10 dunno if this is related, but I just tried to update to develop on my local clone, and now immutables is not being invoked in the dialogue-annotations-processor project in IntelliJ, so I'm getting build failures. Going to try a fresh clone

jkozlowski:
Nope, no give; locally removing the incap processor fixes the problem: any ideas?

pkoenig10:
> For reference: what does this incremental compilation actually help us with?

Incremental compilation allows Gradle to avoid recompiling every file in a project when only some files have changed. For example, if I change some other code unrelated to my annotated service definition, I would be nice if we could avoid needing to invoke the Dialogue annotation processor.

> Or is the win here in IntelliJ and it's somehow plumbed fully through there?

If you're using the Intellij Gradle integration, then you get the wins here because Intellij delegates build tasks to Gradle.

If you're using the built-in Intellij compiler, it has it's own incremental compilation logic that bypasses all of this. It mostly works, although I have run into issues where the Intellij incremental compiler makes faulty assumptions about the behavior of annotation processors causing failures and requiring a manual rebuild.

Most notably, when using Immutables with a `package-info.class` which contains a custom style annotation in a parent package, the Intellij incremental compiler will rerun the Immutables annotation processor but it won't take the `package-info.class` into account leading to incorrect generated code.

pkoenig10:
Looking at Intellij build debug logs I think I see the issue.

Without the incap processor, the build succeeds and I see:
```
DEBUG - s.incremental.java.JavaBuilder - Running processors for chunk [dialogue-annotations-processor] with options: ...
DEBUG - s.incremental.java.JavaBuilder - Compiling chunk [dialogue-annotations-processor] with options: ...
```

With the incap processor, the build fails and I see:
```
DEBUG - s.incremental.java.JavaBuilder - Running processors for chunk [dialogue-annotations-processor] with options: ...
DEBUG - s.incremental.java.JavaBuilder - java:ERROR:cannot find symbol
  symbol:   class ImmutableServiceDefinition
  location: package com.palantir.dialogue.annotations.processor.data; file: /Volumes/git/_dialogue2/dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/DialogueRequestAnnotationsProcessor.java at (29:56)
DEBUG - s.incremental.java.JavaBuilder - java:ERROR:cannot find symbol
  symbol:   class ArgumentTypes
  location: package com.palantir.dialogue.annotations.processor.data; file: /Volumes/git/_dialogue2/dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/generate/ServiceImplementationGenerator.java at (31:56)
DEBUG - s.incremental.java.JavaBuilder - java:ERROR:cannot find symbol
  symbol:   class ParameterTypes
  location: package com.palantir.dialogue.annotations.processor.data; file: /Volumes/git/_dialogue2/dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/generate/ServiceImplementationGenerator.java at (36:56)
```

Notice that the failure happens while running the annotation processors, _not_ while compiling. And it fails because it can’t find the symbols for the immutables types which are also generated during annotation processing

But we shouldn’t be trying to resolve symbols when processing annotations - we only need the AST. I’m guessing this is some unusual behavior of the built-in Intellij compiler where it incorrectly resolves symbols during annotation processing.

PR #1245, opened 2021-05-07
Fair request execution + per-host/sticky refactoring.

jkozlowski:
## Before this PR

1. Per-host/sticky channels do not share underlying queue nor host concurrency-limiters with balanced channels.
2. per-host/sticky channels exist outside of DialogueChannel class: they have their special logic implemented in ReloadingClientFactory.
3. Sticky channel routing decision is done upfront: if that request happens to hit a node that just became degraded, that's where it will stay.
4. Asynchronous requests can overwhelm blocking requests, there is no backpressure for asynchronous requests.

## After this PR
==COMMIT_MSG==
If maybe does not have these problems?
==COMMIT_MSG==


dialogue-target/src/main/java/com/palantir/dialogue/Response.java
```diff
@@ -22,19 +22,23 @@
 import java.util.List;
 import java.util.Optional;
 
-public interface Response extends Closeable {
+// TODO(1234): probs not want abstract
+public abstract class Response implements Closeable {
```
carterkozak, line 26: This can't merge, it would be an ABI break that causes consumers to fail at runtime if they compiled against the interface.

carterkozak: We can safely require that all implementations implement a `ResponseAttachments attachments()` method though, since we don't expect external implementations.

dialogue-core/src/main/java/com/palantir/dialogue/core/NodeSelectionStrategyChannel.java
```diff
@@ -93,6 +93,12 @@ static LimitedChannel create(Config cf, ImmutableList<LimitedChannel> channels)
 
     @Override
     public Optional<ListenableFuture<Response>> maybeExecute(Endpoint endpoint, Request request) {
+
+        Integer hostId = request.attachments().getOrDefault(RoutingAttachments.HOST_KEY, null);
+        if (hostId != null) {
+            return channels.get(hostId).maybeExecute(endpoint, request);
```
carterkozak: we should verify that this works as expected across reloads. What if we use the delegate `LimitedChannel` itself as a key?

dialogue-core/src/main/java/com/palantir/dialogue/core/HostIndexResponseMarkingChannel.java
```diff
@@ -0,0 +1,45 @@
...
+        this.delegate = delegate;
+    }
+
+    @Override
+    public ListenableFuture<Response> execute(Endpoint endpoint, Request request) {
+        return DialogueFutures.transform(delegate.execute(endpoint, request), this::addHostIndex);
+    }
+
+    Response addHostIndex(Response response) {
+        response.attachments().put(RoutingAttachments.HOST_KEY, hostIndex);
```
carterkozak, line 44: Should we use a cooperative design here, in which the request must provide an attachment to opt into getting a response attachment describing the host?

dialogue-target/src/main/java/com/palantir/dialogue/Response.java
jkozlowski: Absolutely, just was quickest way to compile. Default method won't really work, cause there isn't really a good implementation here, since interfaces can't have instance variables.

dialogue-core/src/main/java/com/palantir/dialogue/core/NodeSelectionStrategyChannel.java
jkozlowski: How would per-host channel bind them then? all it knows are URLs and host indexes

dialogue-core/src/main/java/com/palantir/dialogue/core/HostIndexResponseMarkingChannel.java
jkozlowski: I don't think this would work: the first request in a sticky channel does not know the host to route to, but wants to know where the request went to be able to route following requests.

carterkozak: Right, I meant something along the lines of `request.attachments().put(RoutingAttachments.HOST_PLEASE, Boolean.TRUE)`
That way this channel can only set the host data on the response when it's requested.

dialogue-core/src/main/java/com/palantir/dialogue/core/NodeSelectionStrategyChannel.java
carterkozak: I think we can figure something else out for the per-host channel, it's a little too special -- I wouldn't worry about attaching it to the sticky implementation. Realistically I don't think it should have ever been made public.

dialogue-core/src/main/java/com/palantir/dialogue/core/HostIndexResponseMarkingChannel.java
jkozlowski: I don't quite understand the concern here though? is it about allocating the attachments map? I also don't mind doing what you propose, IF we don't need to support per-host.

dialogue-core/src/main/java/com/palantir/dialogue/core/NodeSelectionStrategyChannel.java
jkozlowski: Cool, let's leave this for when you come back. I will focus on request scheduling, this was all prep work.

jkozlowski:
Possible API: https://github.com/palantir/conjure-java/pull/1354

jkozlowski:
Abandoning this approach as too radical: trying a less invasive approach here: https://github.com/palantir/dialogue/pull/1299

PR #1257, opened 2021-05-14
Add methods to create channels/clients from ClientConfiguration.

jkozlowski:
## After this PR
==COMMIT_MSG==

==COMMIT_MSG==


dialogue-clients/src/main/java/com/palantir/dialogue/clients/ReloadingClientFactory.java
```diff
@@ -94,6 +95,22 @@ default ConjureRuntime runtime() {
...
+                .clientConfiguration(clientConf)
+                .clientName(channelName);
+        params.blockingExecutor().ifPresent(clientBuilder::executor);
+        ApacheHttpClientChannels.CloseableClient apacheClient = clientBuilder.build();
+        Channel channel = DialogueChannel.builder()
+                .channelName(channelName)
+                .clientConfiguration(clientConf)
+                .channelFactory(uri -> ApacheHttpClientChannels.createSingleUri(uri, apacheClient))
+                .build();
+        return Reflection.callStaticFactoryMethod(clazz, channel, params.runtime());
```
carterkozak, line 116: Generally in favor of this, I imagine we'd be better off splitting it into a `Channel getNonReloadingChannel(ClientConfig)` + the current method, that way we could reuse it from CJR jaxrsclient.create: https://github.com/palantir/conjure-java-runtime/blob/27be7573c5b0e9d09e5b0a4ed10e882f969330a9/conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/AbstractFeignJaxRsClientBuilder.java#L78-L94

I think many folks use ClientConfiguration in cases where they're blocked from converting due to ServiceConfiguration requiring certificate material on disk, we may want to provide a non-deprecated API to set an SSLContext in code.

jkozlowski: > I think many folks use ClientConfiguration in cases where they're blocked from converting due to ServiceConfiguration requiring certificate material on disk, we may want to provide a non-deprecated API to set an SSLContext in code.

I only looked at it briefly, but I got stuck on how to amend SslConfiguration to be EITHER set of on disk paths OR SSLContext, without potential breaks. I think, since we already have this split between ON DISK (ServicesBlock, PartialServiceConfiguration, ServiceConfiguration) and IN MEMORY (ClientConfiguration), I think I'd rather go with what I have in this PR.

+1 on splitting 2 methods

jkozlowski: Done

dialogue-clients/src/main/java/com/palantir/dialogue/clients/DialogueClients.java
```diff
@@ -66,6 +67,20 @@ public static ReloadingFactory create(Refreshable<ServicesConfigBlock> scb) {
         Channel getChannel(String serviceName);
     }
 
+    public interface NonReloadingChannelFactory {
+        Channel getNonReloadingChannel(String channelName, ClientConfiguration clientConfiguration);
+    }
+
+    public interface ClientConfigurationNonReloadingClientFactory {
```
carterkozak, line 80: Wasn't this deprecated before? I think we want to keep that around so folks don't use ClientConfiguration when they have other options.

dialogue-clients/src/main/java/com/palantir/dialogue/clients/ReloadingClientFactory.java
carterkozak: > I only looked at it briefly, but I got stuck on how to amend SslConfiguration to be EITHER set of on disk paths OR SSLContext, without potential breaks. I think, since we already have this split between ON DISK (ServicesBlock, PartialServiceConfiguration, ServiceConfiguration) and IN MEMORY (ClientConfiguration), I think I'd rather go with what I have in this PR.

I don't think it's an either/or, ClientConfiguration is the legacy CJR configuration which isn't the best fit for dialogue at this point, but enough folks still use it that we're better off making it convenient to use with dialogue.
I think we'd need some new way to provide an SSLContext to the existing ClientBuilder API without ClientConfiguration, but I haven't through through what that would look like, but I'm not in any rush to implement that.

carterkozak (commented):
lgtm pending method deprecation

dialogue-clients/src/main/java/com/palantir/dialogue/clients/DialogueClients.java
jkozlowski: Will readd.

jkozlowski: Done hopefully

carterkozak: Sorry, one more nit: Should we annotate the method itself instead of the type? We should probably annotate the channel factory as well.

carterkozak (approved):

PR #1260, opened 2021-05-17
Remove the 'UseCallingThreadExecutor' feature flag, feature is always enabled

carterkozak:
We haven't found any issues with this feature, it should be used
everywhere. This also removes the feature-flag metric which no
longer provides any real value.

==COMMIT_MSG==
Remove the 'UseCallingThreadExecutor' feature flag, feature is always enabled
==COMMIT_MSG==


dialogue-serde/src/test/java/com/palantir/conjure/java/dialogue/serde/DefaultClientsTest.java
```diff
@@ -225,18 +223,6 @@ public void testCallClosesRequestOnCompletion_failure(CallType callType) {
     @Test
     @Execution(ExecutionMode.SAME_THREAD)
     public void testCallBlockingPropagatesCallingThreadExecutor_whenEnabled() {
```
tpetracca: nit: remove the `_whenEnabled` suffix

tpetracca (approved):

dialogue-serde/src/test/java/com/palantir/conjure/java/dialogue/serde/DefaultClientsTest.java
carterkozak:
```suggestion
    public void testCallBlockingPropagatesCallingThreadExecutor() {
```

PR #1267, opened 2021-05-20
Provide default error handling strategy.

jkozlowski:
## Before this PR

By default, users need to do the right thing on errors.

## After this PR
==COMMIT_MSG==
Provide default error decoding.
==COMMIT_MSG==


dialogue-annotations/src/main/java/com/palantir/dialogue/annotations/Request.java
```diff
@@ -53,6 +55,8 @@
      */
     Class<? extends DeserializerFactory> accept() default Json.class;
 
+    Class<? extends ErrorDecoder> errorDecoder() default ConjureErrorDecoder.class;
```
jkozlowski, line 62: @carterkozak I think this is what you described?

dialogue-annotations/src/main/java/com/palantir/dialogue/annotations/ErrorHandlingDeserializerFactory.java
```diff
@@ -0,0 +1,64 @@
...
+            @Override
+            public T1 deserialize(Response response) {
+                boolean closeResponse = true;
+                try {
+                    if (errorDecoder.isError(response)) {
+                        throw errorDecoder.decode(response);
+                    } else {
+                        // delegate has taken on responsibility for closing the response body
+                        closeResponse = false;
+                        return delegateDeserializer.deserialize(response);
```
carterkozak: Perhaps we should close this if the result is not closeable?

jkozlowski: For now just copied code without thinking about it too much; wanted to make sure we were on the same page with what you were describing

dialogue-annotations/src/main/java/com/palantir/dialogue/annotations/Request.java
jkozlowski: so flows would be:
1. Say you hate exceptions, you'd impl ErrorDecoder and always return false from isError, and then all the handling would happen in DeserializerFactory.
2. Say you don't like conjure exceptions, you'd just impl your own ErrorDecoder.
3. Most people don't care about errors, so they don't need to do anything special.
4. There's some fun extra handling in ConjureBodySerDe around 204s, which might be useful to port over here?

carterkozak: Yep, I like it!

> There's some fun extra handling in ConjureBodySerDe around 204s, which might be useful to port over here?

I think that's handled by individual Deserializer implementations, wouldn't worry about it here.

dialogue-serde-core/src/main/java/com/palantir/conjure/java/dialogue/serde/core/ConjureErrorDecoder.java
```diff
@@ -51,17 +52,18 @@
  * {@link RemoteException}) if a {@link RemoteException} could not be extracted, e.g., when the given {@link
  * Response} does not adhere to an expected format.
  */
-enum ErrorDecoder {
-    INSTANCE;
+public final class ConjureErrorDecoder implements ErrorDecoder {
```
carterkozak: Instead of implementing an interface that's only used by the annotation processor, could the processor provide this interface, and an adapter from the default implementation into the annotaiton-processor error handler? That way we're free to add data from the Runtime object to the default dialogue ErrorDecoder as long as we can provide sane default args from the annotation processor shim.

jkozlowski: Yeah, just need to make the enum public and that should work.

dialogue-annotations/src/main/java/com/palantir/dialogue/annotations/ErrorHandlingDeserializerFactory.java
jkozlowski: I feel like the strategy you're proposing is not bullet-proof though: it would be great if everyone wrote code like this, but I fear that not everyone does. On the flip side, if they write code that returns something that is supposed to keep the InputStream open, and we close it for them, that might be a good nudge, but will likely be difficult to discover?

jkozlowski: I implemented it, let's see what happens.

dialogue-serde-core/src/main/java/com/palantir/conjure/java/dialogue/serde/core/ConjureErrorDecoder.java
jkozlowski: Done

dialogue-annotations-example/src/main/java/com/palantir/myservice/example/IHateExceptionsErrorDecoder.java
```diff
@@ -0,0 +1,32 @@
...
+
+public final class IHateExceptionsErrorDecoder implements ErrorDecoder {
+    @Override
+    public boolean isError(Response _response) {
+        return false;
+    }
+
+    @Override
+    public RuntimeException decode(Response _response) {
+        throw new UnsupportedOperationException("Exceptions were a mistake");
```
carterkozak, line 30: doth protest too much, methinks

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ErrorDecoder.java
```diff
@@ -51,17 +51,17 @@
  * {@link RemoteException}) if a {@link RemoteException} could not be extracted, e.g., when the given {@link
  * Response} does not adhere to an expected format.
  */
-enum ErrorDecoder {
+public enum ErrorDecoder {
```
carterkozak, line 54: This becomes public API, potential downside is that it's harder to add additional info (metrics?) to this type in the future, but I think it's more important that we allow everything to parse/throw the correct exceptions. Worst case we can always gut this enum and move the logic elsewhere, leaving this class for backcompat.

carterkozak (approved):
lgtm!

pkoenig10:
Just want to flag that this was a breaking behavior change for anyone who was relying on handling things like 3xx redirect responses with an annotated Dialogue client.

 Couple thoughts:
- Does it really make sense for an annotation Dialogue client, which provides annotations to create generic HTTP clients, to use a default error decoder that expects Conjure `SerializableError`?
- Should we really consider 3xx responses as errors? I know we use 308 to retry another node, but we probably should not consider other 3xx responses as errors. Using 308 here is actually a bit weird since 308 signals a permanent redirect.

At the very least, it feels like we should make the error decoding failable so we can fallback and let the client handle the response in that case, instead of just giving up and throwing an exception if the body isn't a `SerializableError`.
```java
interface ErrorDecoder {
    public Optional<RuntimeException> decode(Response response);
}
```

Thoughts?

dialogue-annotations-example/src/main/java/com/palantir/myservice/example/IHateExceptionsErrorDecoder.java
```diff
@@ -0,0 +1,32 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.myservice.example;
+
+import com.palantir.dialogue.Response;
+import com.palantir.dialogue.annotations.ErrorDecoder;
+
+public final class IHateExceptionsErrorDecoder implements ErrorDecoder {
```
pkoenig10, line 22: This is a pretty crude name, can we change it to something more appropriate for a public open-source library?

jkozlowski:
> Just want to flag that this was a breaking behavior change for anyone who was relying on handling things like 3xx redirect responses with an annotated Dialogue client.

Ack, I haven't seen anyone who picked it up do any special handling yet.

> At the very least, it feels like we should make the error decoding failable so we can fallback and let the client handle the response in that case, instead of just giving up and throwing an exception if the body isn't a SerializableError.

I think the whole point of this is that YOU CAN just implement your own, but by default people generally don't care/shouldn't care: if you're not handling those exceptions anyway, what difference does it make whether you'll get a nice stacktrace or UnknownRemoteException? The only thing I can see is that IF your upstream sends back "application/json", you'll always end up throwing and catching an exception. But it's an error case, so if that's hot code you probably have a bigger problem. The happy case in the Json class is also Conjure deserialization.

I understand the points on being opinionated here, but it looks to me like there are always escape hatches, and the thinking is the default handling is sensible enough to cover most cases.

dialogue-annotations-example/src/main/java/com/palantir/myservice/example/IHateExceptionsErrorDecoder.java
jkozlowski: Ack

jkozlowski: https://github.com/palantir/dialogue/pull/1269

pkoenig10:
> if you're not handling those exceptions anyway, what difference does it make whether you'll get a nice stacktrace or UnknownRemoteException?

Ah right, this makes sense to me.

PR #1270, opened 2021-05-24
Expose DisabledErrorDecoder to consumers

pkoenig10:
## Before this PR
Consumers have to implement their own `ErrorDecoder` if they don't want any error handling.

## After this PR
Dialogue provides a `DisabledErrorDecoder` to disable error handling.


dialogue-annotations/src/main/java/com/palantir/dialogue/annotations/DisabledErrorDecoder.java
```diff
@@ -14,12 +14,12 @@
  * limitations under the License.
  */
 
-package com.palantir.myservice.example;
+package com.palantir.dialogue.annotations;
 
 import com.palantir.dialogue.Response;
-import com.palantir.dialogue.annotations.ErrorDecoder;
 
 public final class DisabledErrorDecoder implements ErrorDecoder {
```
carterkozak: Perhaps `ErrorDecoder.None.class` for discoverability?

carterkozak: Can you describe how you'd use this (internally if necessary)?

pkoenig10: Should we do something similar with `ConjureErrorDecoder`?

carterkozak: > Should we do something similar with ConjureErrorDecoder?

Since it's the default, moving it would be a breaking change, also less important for it to be discoverable since it's the default. A `Conjure.class` alias within ErrorDecoder would be reasonable if you like, we could deprecate the existing `ConjureErrorDecoder`.

pkoenig10: Nah, let's just leave it for now. `ConjureErrorDecoder` is a specific implementation so it's not unreasonable for it to be separate. It's also the default so probably doesn't need to be as discoverable.

jkozlowski (approved):

dialogue-annotations/src/main/java/com/palantir/dialogue/annotations/DisabledErrorDecoder.java
pkoenig10: > Can you describe how you'd use this (internally if necessary)?

I'm just using this as a test interface to test an [OAuth2 authorization code flow](https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.1). The authorization response is a redirect and I need to grab the `code` query parameter from the redirect URI.

I could always just use the Java `HttpClient`, but by using Dialogue I'm able to take advantage of the discovery features of our integration testing library.

Even though this use case is fairly niche, I think it's reasonable for Dialogue to give users a way to opt-out of the opinionated Conjure error handling.

PR #1272, opened 2021-05-25
 Dialogue requires Java 11

carterkozak:
==COMMIT_MSG==
 Dialogue requires Java 11
==COMMIT_MSG==

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -60,7 +60,7 @@ subprojects {
     apply plugin: 'java-library'
     apply plugin: 'org.inferred.processors'
 
-    sourceCompatibility = 1.8
+    sourceCompatibility = 11
 
     tasks.withType(JavaCompile) {
         options.compilerArgs += ['-Werror']
--- a/dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientChannels.java
+++ b/dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientChannels.java
@@ -122,10 +122,10 @@ public final class ApacheHttpClientChannels {
     }
 
     public static Channel createSingleUri(String uri, CloseableClient client) {
-        BlockingChannel blockingChannel = new ApacheHttpClientBlockingChannel(client, url(uri), client.leakDetector);
-        return client.executor == null
+        BlockingChannel blockingChannel = new ApacheHttpClientBlockingChannel(client, url(uri), client.leakDetector());
+        return client.executor() == null
                 ? BlockingChannelAdapter.of(blockingChannel)
-                : BlockingChannelAdapter.of(blockingChannel, client.executor);
+                : BlockingChannelAdapter.of(blockingChannel, client.executor());
     }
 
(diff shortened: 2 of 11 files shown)


dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientChannels.java
```diff
@@ -235,21 +199,132 @@ static CloseableClient wrap(
                     .clientType(CLIENT_TYPE)
                     .build();
             createMeter.mark();
-            return newInstance;
+            CloseableClient wrapper = new CloseableClientWrapper(newInstance);
+            CleanerSupport.register(wrapper, newInstance::closeApacheClient);
+            return wrapper;
```
carterkozak, line 204: refactor exists for these three lines -- we must create a delegate which can be strongly referenced by the cleaner without blocking the sentinel reference (the wrapper) from being garbage collected. When the outer class is garbage collected, our cleaner is able to execute.

build.gradle
```diff
@@ -60,7 +60,7 @@ subprojects {
     apply plugin: 'java-library'
     apply plugin: 'org.inferred.processors'
 
-    sourceCompatibility = 1.8
+    sourceCompatibility = 11
```
jkozlowski, line 63: Can you remove overrides in dialogue-annotations-example and dialogue-annotations-processor?

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ResponseLeakDetector.java
```diff
@@ -71,9 +72,6 @@ Response wrap(Response input, Endpoint endpoint) {
     }
 
     private boolean shouldApplyLeakDetection() {
```
jkozlowski: While you're here, can you remove the whole shouldApplyLeakDetection? probability is 1 now, so all this code can probably go away?

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientChannels.java
jkozlowski: Discussed: I found this kinda hard to reason about, the cleaner stuff always is fun. We discussed instead having private static final InternalCloseableClient and public static final class CloseableClient that takes InternalCloseableClient and delegates. I find that easier to understand, and Carter doesn't so let's just ship this.

jkozlowski (commented):
I think there is 1 action and maybe another action (but can be in a FLUP PR)

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ResponseLeakDetector.java
carterkozak: good catch!

jkozlowski (approved):
approving to unblock, since I'm going offline shortly

build.gradle
carterkozak: Also removed the `-source 11` lines from the test javac runner since it's the new default.

PR #1275, opened 2021-06-01
Handle NoHttpResponseException for non-idempotent endpoints

natgabb:
## Before this PR
NoHttpResponseException can happen when the server has actually received a request and started processing it, usually when it is shut down afterwards.

## After this PR
==COMMIT_MSG==
Handle NoHttpResponseException for non-idempotent endpoints
==COMMIT_MSG==

### DIFF
--- a/dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientBlockingChannel.java
+++ b/dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientBlockingChannel.java
@@ -26,6 +26,7 @@ import com.palantir.dialogue.RequestBody;
 import com.palantir.dialogue.Response;
 import com.palantir.dialogue.blocking.BlockingChannel;
 import com.palantir.dialogue.core.BaseUrl;
+import com.palantir.dialogue.core.SafeToRetryForIdempotentEndpoints;
 import com.palantir.logsafe.Arg;
 import com.palantir.logsafe.Preconditions;
 import com.palantir.logsafe.SafeArg;
@@ -39,6 +40,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.net.URL;
+import java.time.Duration;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.Iterator;
@@ -53,6 +55,7 @@ import org.apache.hc.client5.http.impl.classic.CloseableHttpResponse;
 import org.apache.hc.core5.function.Supplier;
 import org.apache.hc.core5.http.Header;
 import org.apache.hc.core5.http.HttpEntity;
+import org.apache.hc.core5.http.NoHttpResponseException;
 import org.apache.hc.core5.http.io.support.ClassicRequestBuilder;
 import org.apache.hc.core5.http.message.BasicHeader;
 import org.slf4j.Logger;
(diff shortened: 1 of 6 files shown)


dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -353,6 +354,11 @@ private boolean shouldAttemptToRetry(Throwable throwable) {
                             // String matches CJR RemotingOkHttpCall.shouldRetry
                             && socketTimeout.getMessage().contains("connect timed out");
                 }
+                if (throwable instanceof NoHttpResponseException) {
```
natgabb: might be leaking implementation details (apache client) here, open to suggestions.

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientBlockingChannel.java
```diff
@@ -115,6 +118,17 @@ public Response execute(Endpoint endpoint, Request request) throws IOException {
             // not retried by default, so ours implements SafeLoggable and retains the simple-name for
             // cleaner metrics.
             throw new SafeConnectTimeoutException(e, failureDiagnosticArgs(endpoint, request, startTime));
+        } catch (NoHttpResponseException e) {
+            // This should not be retried on non-idempotent endpoints since it can happen when the server
+            // already processed the request, however this can also happen if a persistent connection is timed out and
+            // closed by the server or intermediate proxy, in that case it's important that we do retry. Therefore, we
+            // only allow retries if less than 5 seconds have passed.
+            if (Duration.ofNanos(System.nanoTime() - startTime).getSeconds() <= 5) {
```
jkozlowski, line 126: Is there any other way we can detect the connection time out? Where does the 5 seconds come from?

dialogue-core/src/main/java/com/palantir/dialogue/core/OnlySafeToRetryForIdempotentEndpoints.java
```diff
@@ -0,0 +1,54 @@
...
+import com.google.errorprone.annotations.CompileTimeConstant;
+import com.palantir.logsafe.Arg;
+import com.palantir.logsafe.SafeLoggable;
+import com.palantir.logsafe.exceptions.SafeExceptions;
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+
+public final class OnlySafeToRetryForIdempotentEndpoints extends IOException implements SafeLoggable {
```
jkozlowski, line 28: PersistentConnectionTimeout?

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientBlockingChannel.java
```diff
@@ -115,6 +118,17 @@ public Response execute(Endpoint endpoint, Request request) throws IOException {
             // not retried by default, so ours implements SafeLoggable and retains the simple-name for
             // cleaner metrics.
             throw new SafeConnectTimeoutException(e, failureDiagnosticArgs(endpoint, request, startTime));
+        } catch (NoHttpResponseException e) {
```
jkozlowski, line 121: Are there any other exceptions which we should check for specifically? I did not realize that we retry most exceptions for non-idempotent requests.

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientBlockingChannel.java
```diff
@@ -115,6 +118,17 @@ public Response execute(Endpoint endpoint, Request request) throws IOException {
...
+            // This should not be retried on non-idempotent endpoints since it can happen when the server
+            // already processed the request, however this can also happen if a persistent connection is timed out and
+            // closed by the server or intermediate proxy, in that case it's important that we do retry. Therefore, we
+            // only allow retries if less than 5 seconds have passed.
+            if (Duration.ofNanos(System.nanoTime() - startTime).getSeconds() <= 5) {
+                throw e;
+            }
+            throw new OnlySafeToRetryForIdempotentEndpoints(
+                    "No http response exception occurred after more than 5 seconds have passed", e);
```
jkozlowski, line 130: failureDiagnosticArgs probably?

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientBlockingChannel.java
jkozlowski, line 126: Reading a bit more into this: https://github.com/palantir/dialogue/blob/develop/dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientChannels.java#L410 sets our inactivity check value. If anything we should probably tie the value to this.

So apache ends up calling BHttpConnection#isStale which tries to do a read. So we're worried about the race condition between that check and finishing sending the response? How often do we hit this condition? What if we just not retry at all, regardless of whether the request is idempotent?

natgabb:
Closing in favour of https://github.com/palantir/dialogue/pull/1276

PR #1276, opened 2021-06-02
Avoid retrying NoHttpResponseException for long-running requests

carterkozak:
Previously a peer could time out and close an idle TCP connection
while a server is actively processing the request, resulting in
an unexpected retry. No we avoid retrying these exceptions after
a grace period has elapsed.

## After this PR
==COMMIT_MSG==
Avoid retrying NoHttpResponseException for long-running requests
==COMMIT_MSG==

## Possible downsides?
* It's possible, though unlikely, that connections can be closed quickly after the target has begun processing a request. Those cases are no different from the previous behavior, and have have no evidence this occurs in practice.
* It's possible connections have been closed by proxies and retried in cases it's beneficial to do so. Our stance has been that it's okay to retry all requests when we're reasonably confident they haven't been processed by the target server, so this aligns with what we though was happening, and prevents duplicate data from being created/posted when non-idempotent requests are attempted.

### DIFF
--- a/dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientBlockingChannel.java
+++ b/dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientBlockingChannel.java
@@ -53,6 +53,7 @@ import org.apache.hc.client5.http.impl.classic.CloseableHttpResponse;
 import org.apache.hc.core5.function.Supplier;
 import org.apache.hc.core5.http.Header;
 import org.apache.hc.core5.http.HttpEntity;
+import org.apache.hc.core5.http.NoHttpResponseException;
 import org.apache.hc.core5.http.io.support.ClassicRequestBuilder;
 import org.apache.hc.core5.http.message.BasicHeader;
 import org.slf4j.Logger;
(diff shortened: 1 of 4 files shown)


dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientBlockingChannel.java
```diff
@@ -115,6 +116,23 @@ public Response execute(Endpoint endpoint, Request request) throws IOException {
...
+            // The other possibility is that a remote server or proxy may time out an active request due
+            // to inactivity and close the connection without a response, in this case the request mustn't
+            // be retried.
+            // We attempt to differentiate these two cases based on request duration, we expect most of
+            // the prior case to occur within a couple milliseconds, however we must use a larger value
+            // to account for large garbage collections.
+            long durationNanos = System.nanoTime() - startTime;
+            Arg<?>[] diagnosticArgs = failureDiagnosticArgs(endpoint, request, startTime);
+            if (durationNanos < TimeUnit.SECONDS.toNanos(5)) {
+                e.addSuppressed(new Diagnostic(diagnosticArgs));
```
jkozlowski, line 132: So if it's below 5 seconds, RetryingChannel will retry the request.
If it's longer than 5 seconds, it goes through the SocketTimeoutException codepath, and that only retries "connect timed out" exceptions.

jkozlowski (approved):

PR #1277, opened 2021-06-03
Use error deserializer with void return type

gmaretic:


jkozlowski:
Will re-raise with tests

PR #1282, opened 2021-06-09
Support Map<String, String> in queries

mglazer:
I had a use case for supporting the Feign like QueryMap, where
I might have an indeterminate amount of potentially user supplied
query parameters which aren't known up front when calling a 3rd party
API.

This PR adds a very rudimentary QueryMap capability which supports
both Map and Multimap.

I still likely have to do a fair bit of cleanup here, by adding tests, documentation, etc. But wanted to ensure this wouldn't be flat out rejected before going forward.

## After this PR
==COMMIT_MSG==
Support Map<String, String> in queries
==COMMIT_MSG==

### DIFF
--- a/dialogue-annotations/src/main/java/com/palantir/dialogue/annotations/Request.java
+++ b/dialogue-annotations/src/main/java/com/palantir/dialogue/annotations/Request.java
@@ -101,5 +101,7 @@ public @interface Request {
      */
     @Retention(RetentionPolicy.SOURCE)
     @Target(ElementType.PARAMETER)
-    @interface QueryMap {}
+    @interface QueryMap {
+        Class<? extends MultimapParamEncoder<?>> encoder() default DefaultMultimapParamEncoder.class;
+    }
 }
--- a/dialogue-annotations/src/main/java/com/palantir/dialogue/annotations/Request.java
+++ b/dialogue-annotations/src/main/java/com/palantir/dialogue/annotations/Request.java
@@ -101,6 +101,5 @@ public @interface Request {
      */
     @Retention(RetentionPolicy.SOURCE)
     @Target(ElementType.PARAMETER)
-    @interface QueryMap {
-    }
+    @interface QueryMap {}
 }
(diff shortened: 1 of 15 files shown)


dialogue-annotations/src/main/java/com/palantir/dialogue/annotations/Request.java
```diff
@@ -95,4 +95,13 @@
...
         Class<? extends ListParamEncoder<?>> encoder() default DefaultListParamEncoder.class;
     }
+
+    /**
+     * Similar to the Feign QueryMap serialization type. Only supports simple string -> string maps.
+     */
+    @Retention(RetentionPolicy.SOURCE)
+    @Target(ElementType.PARAMETER)
+    @interface QueryMap {
+        Class<? extends MultimapParamEncoder<?>> encoder() default DefaultMultimapParamEncoder.class;
```
carterkozak, line 105: alternatively we could require a value (no default) and nix the Default since I'm not sure how frequently that will be used. No strong preference though.

dialogue-annotations/src/main/java/com/palantir/dialogue/annotations/Request.java
```diff
@@ -95,4 +95,13 @@
 
         Class<? extends ListParamEncoder<?>> encoder() default DefaultListParamEncoder.class;
     }
+
+    /**
+     * Similar to the Feign QueryMap serialization type. Only supports simple string -> string maps.
```
carterkozak, line 100: Mind updating the docs to describe the updated functionality? I think this has diverged from the feign implementation a bit, but in a way that allows greater flexibility.

dialogue-annotations/src/main/java/com/palantir/dialogue/annotations/Request.java
mglazer: I do like that this is somewhat self documenting (even if duplicative) since it does show you what the presumed default is.

dialogue-annotations/src/main/java/com/palantir/dialogue/annotations/Request.java
mglazer: Done.

carterkozak (approved):

PR #1291, opened 2021-06-23
Validate RetryOther responses that have Location header

jkozlowski:
## Before this PR

No handling for Location headers.

## After this PR
==COMMIT_MSG==
Validate RetryOther responses that have Location header.
==COMMIT_MSG==

## Possible downsides?

* Extra logging on validation failure: I rate limited the logging to 1/s, so shouldn't be a problem.


dialogue-core/src/main/java/com/palantir/dialogue/core/PinUntilErrorNodeSelectionStrategyChannel.java
```diff
@@ -117,6 +117,7 @@ static PinUntilErrorNodeSelectionStrategyChannel of(
             return new PinUntilErrorNodeSelectionStrategyChannel(shuffling, initialPin, metrics, channelName);
         } else if (strategy == DialogueNodeSelectionStrategy.PIN_UNTIL_ERROR_WITHOUT_RESHUFFLE) {
             NodeList constant = new ConstantNodeList(initialShuffle);
+            log.info("Created PinUntilErrorNodeSelectionStrategyChannel", SafeArg.of("numHosts", constant.size()));
```
jkozlowski: I thought about logging in ChannelCache instead, but I didn't want to add extra logging volume on startup for everyone, so maybe this is a good enough of a compromise? this should be pretty much the one single client we care about?

carterkozak: I think this may fill in the missing pieces of the puzzle: #1292

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryOtherValidatingChannel.java
```diff
@@ -0,0 +1,75 @@
...
+                }
+            }
+
+            return response;
+        });
+    }
+
+    private boolean isKnown(String retryOtherUri) {
+        for (int i = 0; i < uris.size(); i++) {
+            if (uris.get(i).equals(retryOtherUri)) {
```
carterkozak: Is the retry-other uri always equivalent to the configured URI? I suppose we'll know if this fires.

Mostly curious if the values may include additional path segments, or potentially missing the `/api` prefix. URI parsing can be a little bit tricky as the same value may be encoded a couple different ways. Perhaps we should extract the `host` component of the URI and compare those?

dialogue-core/src/main/java/com/palantir/dialogue/core/PinUntilErrorNodeSelectionStrategyChannel.java
jkozlowski: Yeah

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryOtherValidatingChannel.java
jkozlowski: I wanted to dig up what conjure-okhttp used to do, but I'm pretty sure it would've just used the same method as here.

jkozlowski: I started by trying to parse the urls, but decided this should probably be the contract IF we were to every implement this: it needs to match what we know about, otherwise it's a little wonky. Like as a strawman, I'm talking to a service using uris IT HAS published: it would make sense to me that IF that service sends me a retryOther, it should be the same uri format as it would've published them to me to use.

jkozlowski: Hmm maybe it was doing some more checks: https://github.com/palantir/conjure-java-runtime/blob/develop/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/UrlSelectorImpl.java

jkozlowski: I would still maintain that this is slightly weird API, they should be exchanging equivalent uris, so the client can just do a simple search.

carterkozak: Given that this isn't connected to correctness or security, rather validation and debugging, I think the question we'd like to answer is "Is this URI definitely wrong?" as opposed to "Is this URI exactly right?". Not worth thinking about the subtle edge cases we don't need to.

jkozlowski: ok, in that case host is probably all we want

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryOtherValidatingChannel.java
```diff
@@ -0,0 +1,131 @@
...
+
+    RetryOtherValidatingChannel(Channel delegate, List<String> hosts) {
+        this(delegate, hosts, RetryOtherValidatingChannel.failureReporter(hosts));
+    }
+
+    @VisibleForTesting
+    RetryOtherValidatingChannel(Channel delegate, List<String> hosts, Consumer<String> failureReporter) {
+        this.delegate = delegate;
+        this.hosts =
+                hosts.stream().map(RetryOtherValidatingChannel::strictParseHost).collect(Collectors.toList());
```
jkozlowski: I feel like we're missing parsing String -> URL (or some other type) somewhere earlier, this parsing is done in a bunch of places now. But probably not for this PR

carterkozak: java URL and URI types aren't very good about keeping up with standards (with good reason I imagine, since changes are likely to break users) so we don't necessarily standardize on them.

Should we use a set to avoid unnecessary iteration? for-loop over strings is probably faster for small values, but we don't need to optimize the 308 path.
```suggestion
                hosts.stream().map(RetryOtherValidatingChannel::strictParseHost).collect(ImmutableSet.toImmutableSet());
```

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryOtherValidatingChannel.java
```diff
@@ -0,0 +1,131 @@
...
+    @Override
+    @SuppressWarnings("Finally")
+    public ListenableFuture<Response> execute(Endpoint endpoint, Request request) {
+        return DialogueFutures.transform(delegate.execute(endpoint, request), response -> {
+            try {
+                validateRetryOther(response);
+                return response;
+            } finally {
+                return response;
+            }
```
carterkozak: Transform is unnecessary here, I think we want a simple callback instead -- that instance can be eagerly initialized in the channel constructor rather than per-call.

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryOtherValidatingChannel.java
```diff
@@ -0,0 +1,131 @@
...
+        this.failureReporter = failureReporter;
+    }
+
+    @Override
+    @SuppressWarnings("Finally")
+    public ListenableFuture<Response> execute(Endpoint endpoint, Request request) {
+        return DialogueFutures.transform(delegate.execute(endpoint, request), response -> {
+            try {
+                validateRetryOther(response);
+                return response;
```
carterkozak: I don't think this line is necessary because the `finally` is always guaranteed to execute. That said, we don't expect validateRetryOther to throw, do we?
```suggestion
```

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryOtherValidatingChannel.java
```diff
@@ -0,0 +1,131 @@
...
+        return (maybeHost != null) && isKnownHost(maybeHost);
+    }
+
+    private boolean isKnownHost(String host) {
+        for (int i = 0; i < hosts.size(); i++) {
+            if (hosts.get(i).equals(host)) {
+                return true;
+            }
+        }
+        return false;
```
carterkozak: If we use a set, this can be simplified to `hosts.contains(host)`

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryOtherValidatingChannel.java
jkozlowski: Done

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryOtherValidatingChannel.java
jkozlowski: Done

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryOtherValidatingChannel.java
jkozlowski: Done

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryOtherValidatingChannel.java
jkozlowski: Done

carterkozak (approved):

jkozlowski:
Audit logs from graal:

```
{"type":"audit.2","time":"2021-06-24T13:16:48.487769Z","uid":"7191bd18-4d8c-4810-b3cb-f835ac87a890","sid":"f2330131-3a70-4db1-8230-cff67e2f9ef8","traceId":"f3b6f2e2d9e62c37","otherUids":[],"origin":"192.168.224.1","name":"CREATE_LINK","result":"SUCCESS","requestParams":{"link":{"enabled":true,"owner":"7191bd18-4d8c-4810-b3cb-f835ac87a890","resource":"ri.share-link.main.resource.dfb86959-f05b-4327-9639-d02084350205","forwardingPath":"/my/forward","grants":[]},"_sourceOrigin":"192.168.224.1","_category":{"type":"dataShareCreate","dataShareCreate":{"target":[{"type":"foundry","foundry":{"id":"ri.share-link.main.resource.dfb86959-f05b-4327-9639-d02084350205"}}]}},"_userAgent":"DevEnvClients/2.91.0 ShareLinkService/0.246.0-rc1-1-g115286e dialogue/2.3.0","_auditEventId":"ce16efe7-4bd0-4125-884f-f89f8ad4866a"},"resultParams":{"linkId":"iktutqfheb7co"}}
{"type":"audit.2","time":"2021-06-24T13:16:48.865651Z","uid":"7191bd18-4d8c-4810-b3cb-f835ac87a890","sid":"f2330131-3a70-4db1-8230-cff67e2f9ef8","traceId":"f3b6f2e2d9e62c37","otherUids":[],"origin":"192.168.224.1","name":"VISIT_LINK","result":"ERROR","requestParams":{"_sourceOrigin":"192.168.224.1","_auditEventId":"ec7c6bbe-694d-42e3-b5f8-2c72c8d844b2","_category":{"type":"dataShare","dataShare":{"id":"iktutqfheb7co","reason":"visited share link","target":[{"type":"foundry","foundry":{"id":"ri.share-link.main.resource.dfb86959-f05b-4327-9639-d02084350205"}}]}},"_userAgent":"DevEnvClients/2.91.0 ShareLinkService/0.246.0-rc1-1-g115286e dialogue/2.3.0"},"resultParams":{"result":"redirect"}}
{"type":"audit.2","time":"2021-06-24T13:16:49.00278Z","uid":"7191bd18-4d8c-4810-b3cb-f835ac87a890","sid":"f2330131-3a70-4db1-8230-cff67e2f9ef8","traceId":"f3b6f2e2d9e62c37","otherUids":[],"origin":"192.168.224.1","name":"VISIT_LINK","result":"ERROR","requestParams":{"_sourceOrigin":"192.168.224.1","_auditEventId":"356972fb-36a4-44e6-a0e4-8ce7e5e4ded8","_category":{"type":"dataLoad","dataLoad":{}},"_userAgent":"DevEnvClients/2.91.0 TestFollowService/0.0.0 dialogue/2.3.0"},"resultParams":{"result":"redirect"}}
{"type":"audit.2","time":"2021-06-24T13:16:49.159912Z","uid":"7191bd18-4d8c-4810-b3cb-f835ac87a890","sid":"f2330131-3a70-4db1-8230-cff67e2f9ef8","traceId":"f3b6f2e2d9e62c37","otherUids":[],"origin":"192.168.224.1","name":"VISIT_LINK","result":"ERROR","requestParams":{"_sourceOrigin":"192.168.224.1","_auditEventId":"c1792478-d69e-454c-b0ca-41aeee2166e3","_category":{"type":"dataLoad","dataLoad":{}},"_userAgent":"DevEnvClients/2.91.0 TestFollowService/0.0.0 dialogue/2.3.0"},"resultParams":{"result":"redirect"}}
{"type":"audit.2","time":"2021-06-24T13:16:49.331578Z","uid":"7191bd18-4d8c-4810-b3cb-f835ac87a890","sid":"f2330131-3a70-4db1-8230-cff67e2f9ef8","traceId":"f3b6f2e2d9e62c37","otherUids":[],"origin":"192.168.224.1","name":"VISIT_LINK","result":"ERROR","requestParams":{"_sourceOrigin":"192.168.224.1","_auditEventId":"49d5acd3-aef0-4a27-ad91-a49c97b28474","_category":{"type":"dataLoad","dataLoad":{}},"_userAgent":"DevEnvClients/2.91.0 TestFollowService/0.0.0 dialogue/2.3.0"},"resultParams":{"result":"redirect"}}
{"type":"audit.2","time":"2021-06-24T13:16:49.451086Z","uid":"7191bd18-4d8c-4810-b3cb-f835ac87a890","sid":"f2330131-3a70-4db1-8230-cff67e2f9ef8","traceId":"f3b6f2e2d9e62c37","otherUids":[],"origin":"192.168.224.1","name":"VISIT_LINK","result":"ERROR","requestParams":{"_sourceOrigin":"192.168.224.1","_auditEventId":"a68632b9-896a-4caa-801c-0a517fb0b761","_category":{"type":"dataLoad","dataLoad":{}},"_userAgent":"DevEnvClients/2.91.0 TestFollowService/0.0.0 dialogue/2.3.0"},"resultParams":{"result":"redirect"}}
```

audit logs from develop:
```

```

PR #1293, opened 2021-06-23
Plumb through hostIndex to individual channels for debugging

carterkozak:
This was an initial attempt to surface per-host information for individual failures. #1292 should give us the data we want for the specific case we're investigating, but this may be more helpful in the general case (outside of pin-until-error) because we can track which node each retry was executed on.

It hits a lot of code. Worth the churn? ¯\_(ツ)_/¯

## After this PR
==COMMIT_MSG==
Plumb through hostIndex to individual channels for debugging
==COMMIT_MSG==

### DIFF
--- a/dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientBlockingChannel.java
+++ b/dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientBlockingChannel.java
@@ -44,6 +44,7 @@ import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Optional;
+import java.util.OptionalInt;
 import java.util.OptionalLong;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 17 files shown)


dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientBlockingChannel.java
```diff
@@ -65,12 +66,17 @@
...
     private final BaseUrl baseUrl;
     private final ResponseLeakDetector responseLeakDetector;
+    private final OptionalInt uriIndexForInstrumentation;
 
     ApacheHttpClientBlockingChannel(
-            ApacheHttpClientChannels.CloseableClient client, URL baseUrl, ResponseLeakDetector responseLeakDetector) {
+            ApacheHttpClientChannels.CloseableClient client,
+            URL baseUrl,
+            ResponseLeakDetector responseLeakDetector,
+            OptionalInt uriIndexForInstrumentation) {
```
jkozlowski, line 75: Would Map<String,String> be a bit more generic and mean we could pass more data to this layer later should we wish to?

jkozlowski: What if this was encapsulated in an attachment on the request that is set/unset by another channel before this channel and then this channel pulls it out from the attachment and attaches as args to the exception here? Maybe then you wouldn't even need to change the creation API here.

jkozlowski (commented):
Overall as a goal makes total sense, we definintely want this data in the individual failure logs; suggested some other strategies on getting this data through

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientBlockingChannel.java
carterkozak: The args object provided by the new channel factory gives us just this, in a strongly typed way. We can always add additional optional fields to it.

I'd prefer not to use request attachments because a request instance may be reused, and I'm sure I'd eventually break the code which unsets data, and confuse myself into oblivion :-)

jkozlowski (approved):

PR #1297, opened 2021-06-25
Sticky client fairness simulation.

jkozlowski:
## Before this PR

For transactional usecases, a single transaction that sends instantly sends a large number of asynchronous requests can starve out other transactions that are playing nicely and sending small number of requests. We would like to simulate this situation so that we can track improvements towards enforcing fairness between transactions.

## After this PR
==COMMIT_MSG==
Simulation for bursty vs steady client utilizing same channel
==COMMIT_MSG==


dialogue-core/src/main/java/com/palantir/dialogue/core/BalancedScoreTracker.java
```diff
@@ -62,7 +61,7 @@
 
     BalancedScoreTracker(
             int channelCount, Random random, Ticker ticker, TaggedMetricRegistry taggedMetrics, String channelName) {
-        Preconditions.checkState(channelCount >= 2, "At least two channels required");
+        //        Preconditions.checkState(channelCount >= 2, "At least two channels required");
```
jkozlowski: @carterkozak thoughts on this? I'd like to simulate it based on the BalancedScoreTracker, but need to make sure that both request flows land on the same host. So I just made the simulation have a single host, but then this piece complains...

carterkozak: I don't think that check was ever really necessary, nothing wrong with using the mechanism for a single channel even if it doesn't always make sense.

simulation/src/test/resources/txt/uncommon_flakes(Strategy)[CONCURRENCY_LIMITER_PIN_UNTIL_ERROR].txt
```diff
@@ -0,0 +1,11 @@
+success=99.0%	client_mean=PT0.000001S    	server_cpu=PT0.01S        	client_received=10000/10000	server_resps=10000	codes={200=9900, 500=100}
```
Comment (carterkozak) [inline simulation/src/test/resources/txt/uncommon_flakes(Strategy)[CONCURRENCY_LIMITER_PIN_UNTIL_ERROR].txt]: Can we avoid renaming these? It looks like the names are identical, but we've just added `(Strategy)` to the filename.

simulation/src/test/resources/report.md
```diff
@@ -1,277 +1,682 @@
...
-                  uncommon_flakes[CONCURRENCY_LIMITER_ROUND_ROBIN].txt:	success=99.0%	client_mean=PT0.000001S    	server_cpu=PT0.01S        	client_received=10000/10000	server_resps=10000	codes={200=9900, 500=100}
-                            uncommon_flakes[UNLIMITED_ROUND_ROBIN].txt:	success=99.0%	client_mean=PT0.000001S    	server_cpu=PT0.01S        	client_received=10000/10000	server_resps=10000	codes={200=9900, 500=100}
-```
-
-
-
-## `all_nodes_500[CONCURRENCY_LIMITER_PIN_UNTIL_ERROR]`
-<table><tr><th>develop</th><th>current</th></tr>
-<tr><td><image width=400 src="https://media.githubusercontent.com/media/palantir/dialogue/develop/simulation/src/test/resources/all_nodes_500[CONCURRENCY_LIMITER_PIN_UNTIL_ERROR].png" /></td><td><image width=400 src="all_nodes_500[CONCURRENCY_LIMITER_PIN_UNTIL_ERROR].png" /></td></tr></table>
+      all_nodes_500(Strategy)[CONCURRENCY_LIMITER_PIN_UNTIL_ERROR].txt:	success=73.9%	client_mean=PT4.24575S     	server_cpu=PT20M          	client_received=2000/2000	server_resps=2000	codes={200=1478, 500=522}
```
carterkozak: These as well, I'd prefer if we could keep the diff cleaner, otherwise it's much harder to verify the simulation results are identical.

carterkozak:
Looks reasonable to me, would prefer to avoid renaming all the existing data though.

simulation/src/test/resources/txt/uncommon_flakes(Strategy)[CONCURRENCY_LIMITER_PIN_UNTIL_ERROR].txt
Comment (jkozlowski) [reply simulation/src/test/resources/txt/uncommon_flakes(Strategy)[CONCURRENCY_LIMITER_PIN_UNTIL_ERROR].txt]: This test suite was written assuming you want to run all the tests for all the strategies, therefore all the tests get a Strategy and then have to set it as part of the test body, so that all the #after methods can use it for rendering. It felt like the sticky stuff doesn't make sense for all the routing strategies, but I guess I can just stay in the mould here?

Otherwise, If I don't do that, I need to make up a fake Strategy for that one test.

I'll push a commit that undoes this and runs the test for all strategies.

Comment (jkozlowski) [reply simulation/src/test/resources/txt/uncommon_flakes(Strategy)[CONCURRENCY_LIMITER_PIN_UNTIL_ERROR].txt]: Done

carterkozak (approved):

PR #1299, opened 2021-06-30
[WIP] Fair asynchronous requests

jkozlowski:
## Before this PR

Clients that use asynchronous requests can easily overwhelm the global queue with requests that take a while to drain. If a normally behaved client (blocking or asynchronous) comes along and tries to send a request through, their requests may end up having to wait in the queue for all the queued up requests to drain.

This PR is a PoC for solving that problem, as well as a few associated issues in the following way:

1. RoutingKey (currently just a UUID) is a grouping key that allows related requests (e.g. in a transactional usecase, coming from the same transaction) to be grouped logically and scheduled fairly against other request flows, mitigating the bursty client issue.
2. Sticky channels now are implemented in a way that allows them to share underlying host channels with non-sticky clients: this helps with host problem discovery in situations where a clients have sticky as well as balanced usecases that are talking to the same service.
3. Sticky channels by default set the RoutingKey to a new value for each instantiation: this means they're fair by default.
4. We solve ahead of time route selection from sticky channels: we now simply pass first request through to be routed through, while we queue up any subsequent requests behind it. Once the first request succeeds, it carries an appropriate attachment which can be used to route following requests to the same host. This solves a frequent problem we experience where if a host goes down after it has been selected for a sticky channel, the request ends up having to retry on the same host until it runs out of retries and fails to the client.
5. This PR only implements the RoutingKey setting for sticky channels: it feels like this should be an API that is available in general, so that non-sticky clients that use asynchronous requests can use this. I think this PR should likely merge as is with StickyChannels2 as new API that we can remove once we prove it out. Then we could replace the implementation of StickyChannels with StickyChannels2, and move clients to set the RoutingKey with a more general API.

#### Before

It takes the slow and steady client about 25 seconds to be able to send any requests through, and that likely only happens because some bursty client's requests end up failing so they go to the end of the queue.

server_side_rate_limits_with_sticky_clients_stready_vs_bursty_client CONCURRENCY_LIMITER_ROUND_ROBIN

#### After

Requests are going out at a steady pace.

server_side_rate_limits_with_sticky2_clients_stready_vs_bursty_client CONCURRENCY_LIMITER_ROUND_ROBIN

## After this PR
==COMMIT_MSG==
Asynchronous requests can be associated together for more fair scheduling.
==COMMIT_MSG==


dialogue-core/src/main/java/com/palantir/dialogue/core/NodeSelectionStrategyChannel.java
```diff
@@ -93,6 +95,11 @@ static LimitedChannel create(Config cf, ImmutableList<LimitedChannel> channels)
 
     @Override
     public Optional<ListenableFuture<Response>> maybeExecute(Endpoint endpoint, Request request) {
+        RoutingKey routingKey = request.attachments().getOrDefault(RoutingAttachments.ROUTING_KEY, null);
+        if (routingKey != null) {
+            return channels.get(routingKey.hostId().value()).maybeExecute(endpoint, request);
```
carterkozak: On config update we may land on another node, but that's probably fine. We don't necessarily want to use an outdated config target either.

jkozlowski: Yeah, I am punting that. I think there needs to be an expectation that if a host stops being in the set of URLs, this sticky channel is just broken and there's not much we can do about it.

There's something to be said though about just indexes changing underneath somehow? Like if we received a config update that somehow shuffled the URLs, we don't want to start routing to a random different node? So probably in order for this to be stable is we need to actually choose based on URLs.

I think you suggested we could just directly send a reference to the per host channel? I think that suffers from the problem that those will be re-created when the set of URLs changes (I think we recreate all of channels then?), so even if some URLs stayed the same, the in-flight transactions would just do something weird (possible send requests to stale channels?)

jkozlowski: Either way I'm going to punt since this should be a minor point, as long as we solve the routing and refreshing.

dialogue-core/src/main/java/com/palantir/dialogue/core/NodeSelectionStrategyChannel.java
```diff
@@ -93,8 +95,16 @@ static LimitedChannel create(Config cf, ImmutableList<LimitedChannel> channels)
 
     @Override
     public Optional<ListenableFuture<Response>> maybeExecute(Endpoint endpoint, Request request) {
-        Optional<ListenableFuture<Response>> maybe = delegate.maybeExecute(endpoint, request);
-        if (!maybe.isPresent()) {
+        HostId hostId = request.attachments().getOrDefault(RoutingAttachments.EXECUTE_ON_HOST_ID_KEY, null);
```
jkozlowski, line 97: I'm afraid this might skip some metrics gathering and updating, so likely all the strategies need to be HostId aware and do sensible things

dialogue-core/src/main/java/com/palantir/dialogue/core/StickyConcurrencyLimitedChannel.java
```diff
@@ -0,0 +1,92 @@
...
+        this.channelNameForLogging = channelNameForLogging;
+    }
+
+    @Override
+    public Optional<ListenableFuture<Response>> maybeExecute(Endpoint endpoint, Request request) {
+        Optional<CautiousIncreaseAggressiveDecreaseConcurrencyLimiter.Permit> maybePermit = limiter.acquire();
+        if (maybePermit.isPresent()) {
+            CautiousIncreaseAggressiveDecreaseConcurrencyLimiter.Permit permit = maybePermit.get();
+            logPermitAcquired();
+            Optional<ListenableFuture<Response>> result = delegate.maybeExecute(endpoint, request);
```
jkozlowski: Somewhere in here I need to stick the "if my current in flight is zero, use the force the request through so it gets queued regardless of downstream concurrency limit".

dialogue-clients/src/main/java/com/palantir/dialogue/clients/ReloadingClientFactory.java
```diff
@@ -245,6 +248,32 @@ public String toString() {
         };
     }
 
+    @Override
+    public StickyChannels getStickyChannels2(String serviceName) {
+        Channel channel = getChannel(serviceName);
+        Supplier<Supplier<Channel>> endpointChannelFactorySupplier = StickyEndpointChannels2.create(
```
carterkozak, line 254: doubly nested generics should generally be an interface.

dialogue-clients/src/main/java/com/palantir/dialogue/clients/ReloadingClientFactory.java
```diff
@@ -245,6 +248,32 @@ public String toString() {
...
+                    public <T> T create(Class<T> clientInterface) {
+                        return Reflection.callStaticFactoryMethod(
+                                clientInterface, channelSupplier.get(), params.runtime());
+                    }
+                };
+            }
+
+            @Override
+            public String toString() {
+                return "StickyChannels{}";
```
carterkozak, line 272: I'd either add the relevant data from the delegate channel/factory, or remove the toString.

dialogue-core/src/main/java/com/palantir/dialogue/core/CautiousIncreaseAggressiveDecreaseConcurrencyLimiter.java
```diff
@@ -68,18 +72,25 @@
...
+            // of
+            // 1, we'll still hand out a permit
+            if (force || currentInFlight <= getLimit() - 1) {
+                int inFlightSnapshot = currentInFlight + 1;
+                if (inFlight.compareAndSet(currentInFlight, inFlightSnapshot)) {
+                    return Optional.of(new Permit(inFlightSnapshot));
+                }
+            }
+
+            return Optional.empty();
```
carterkozak: nit: I'd prefer to make this change in a separate PR.

dialogue-core/src/main/java/com/palantir/dialogue/core/CautiousIncreaseAggressiveDecreaseConcurrencyLimiter.java
```diff
@@ -68,18 +72,25 @@
...
+        while (true) {
+            int currentInFlight = getInflight();
+
+            // We don't want to hand out a permit if there are 4 inflight and a limit of 4.1, as this will immediately
+            // send
+            // our inflight number to 5, which is clearly above the limit.  Instead, we wait until there is capacity for
+            // one whole request before handing out a permit. In the worst-case scenario with zero inflight and a limit
+            // of
+            // 1, we'll still hand out a permit
+            if (force || currentInFlight <= getLimit() - 1) {
```
carterkozak: This can be simplified
```suggestion
            if (force || currentInFlight < getLimit()) {
```

dialogue-core/src/main/java/com/palantir/dialogue/core/HostIndexResponseMarkingChannel.java
```diff
@@ -0,0 +1,56 @@
...
+
+    private HostIndexResponseMarkingChannel(int hostIndex, Channel delegate) {
+        this.hostIndex = HostId.of(hostIndex);
+        this.delegate = delegate;
+    }
+
+    @Override
+    public ListenableFuture<Response> execute(Endpoint endpoint, Request request) {
+        Boolean attachHostId = request.attachments().getOrDefault(RoutingAttachments.ATTACH_HOST_ID, false);
+        if (attachHostId != null && attachHostId) {
```
carterkozak: May be slightly easier to follow as:
```suggestion
        Boolean attachHostId = request.attachments().getOrDefault(RoutingAttachments.ATTACH_HOST_ID, Boolean.FALSE);
        if (Boolean.TRUE.equals(attachHostId)) {
```

dialogue-core/src/main/java/com/palantir/dialogue/core/PinUntilErrorNodeSelectionStrategyChannel.java
```diff
@@ -161,6 +162,12 @@ public void onFailure(Throwable throwable) {
         return maybeResponse;
     }
 
+    @Override
+    public Optional<ListenableFuture<Response>> maybeExecuteOnHost(
+            HostId _hostId, Endpoint _endpoint, Request _request) {
+        return Optional.empty();
+    }
```
carterkozak, line 169: todo?

dialogue-core/src/main/java/com/palantir/dialogue/core/StickyChannelRejectedException.java
```diff
@@ -0,0 +1,38 @@
...
+import com.palantir.logsafe.exceptions.SafeRuntimeException;
+
+@SuppressWarnings("ImmutableEnumChecker")
+enum StickyChannelRejectedException {
+    INSTANCE;
+
+    private final RuntimeException exception;
+
+    StickyChannelRejectedException() {
+        this.exception = new SafeRuntimeException("Rejected");
```
carterkozak: Stack trace will always point to the StickyChannelRejectedException class initialization location.
Potential memory leak if anyone calls `addSuppressed(Throwable)`. We can define a new SafeLoggable exception which overrides `fillInStackTrace()` and `addSuppressed(Throwable exception)` as no-op to prevent these issues.

dialogue-core/src/main/java/com/palantir/dialogue/core/NodeSelectionStrategyChannel.java
```diff
@@ -92,9 +94,18 @@ static LimitedChannel create(Config cf, ImmutableList<LimitedChannel> channels)
...
-    public Optional<ListenableFuture<Response>> maybeExecute(Endpoint endpoint, Request request) {
-        Optional<ListenableFuture<Response>> maybe = delegate.maybeExecute(endpoint, request);
-        if (!maybe.isPresent()) {
+    public Optional<ListenableFuture<Response>> maybeExecute(Endpoint endpoint, Request request, boolean force) {
+        HostId hostId = request.attachments().getOrDefault(RoutingAttachments.EXECUTE_ON_HOST_ID_KEY, null);
+
+        // TODO(12345): This is broken for stats gathering, make sure to propagate this to strategies.
+        final Optional<ListenableFuture<Response>> maybe;
+        if (hostId != null) {
+            maybe = channels.get(hostId.value()).maybeExecute(endpoint, request, force);
```
carterkozak, line 102: I wonder if it would be safer/easier to attach the `NodeSelectionStrategyLimitedChannel` instance instead of a HostID? Worst case we replace the client in a rebuild, but it would be impossible to accidentally switch nodes.

dialogue-core/src/main/java/com/palantir/dialogue/core/StickyEndpointChannels2.java
```diff
@@ -0,0 +1,302 @@
...
+    }
+
+    // This single-in-flight-call has been implemented a few times, this impl is likely a bit buggy,
+    // but shows the approach here.
+    // Initial implementation just blocked all calls until we have hostId, but that means simulations were not
+    // working.
+    private static final class DefaultStickyRouter implements StickyRouter {
+
+        private static final ThreadLocal<Integer> stackDepth = ThreadLocal.withInitial(() -> 0);
+        private static final ThreadLocal<Integer> numDeferred = ThreadLocal.withInitial(() -> 0);
```
carterkozak: Why are we using threadlocals?

dialogue-core/src/main/java/com/palantir/dialogue/core/StickyEndpointChannels2.java
```diff
@@ -0,0 +1,302 @@
...
+            request.attachments().put(RoutingAttachments.ROUTING_KEY, routingKey);
+            return stickyRouter.execute(request, delegate);
+        }
+
+        @Override
+        public String toString() {
+            return "ScoreTrackingEndpointChannel{" + "delegate=" + delegate + '}';
+        }
+    }
+}
```
carterkozak, line 214: for the most part I haven't reviewed this class

carterkozak:
Please update the PR description with some background, bonus points if it includes pictures/graphs.

dialogue-core/src/main/java/com/palantir/dialogue/core/PinUntilErrorNodeSelectionStrategyChannel.java
jkozlowski: yeah, I need to implement this in all the strategies, but opted for a quicker route of just putting this in the NodeSelectionStrategyChannel

.palantir/revapi.yml
```diff
@@ -178,6 +178,12 @@ acceptedBreaks:
       old: "method void com.palantir.dialogue.annotations.StdDeserializer<T>::<init>(java.lang.String)"
       new: "method void com.palantir.dialogue.annotations.StdDeserializer<T>::<init>(java.lang.String)"
       justification: "no functional difference"
+  "1.128.0":
+    com.palantir.dialogue:dialogue-target:
+    - code: "java.class.kindChanged"
+      old: "interface com.palantir.dialogue.Response"
+      new: "class com.palantir.dialogue.Response"
```
carterkozak: Per our discussion on the last PR, this is a temporary hack that will be reverted before the feature can be merged

dialogue-core/src/main/java/com/palantir/dialogue/core/StickyEndpointChannels2.java
jkozlowski: Discussed offline: I just tried to implement this in a way to allow me to make progress, it's rather janky. What this is supposed to do is allow you to execute first request, and queue up other requests behind that one completing successfully, so they can peel out the host and route next requests to the same host. @carterkozak suggested using future callbacks to form a natural queue.

carterkozak: The retrying channel is a great resource/template for this sort of thing, the trasnformAsync function works wonders. The retrying channel effectively queues retries after the initial request, conditionally upon remaining retries and the result of the request (success/failure).

dialogue-core/src/main/java/com/palantir/dialogue/core/CautiousIncreaseAggressiveDecreaseConcurrencyLimiter.java
jkozlowski: the #compareAndSet race condition right? Yep, we discussed that, I just had to touch this codepath so did it here as well.

dialogue-core/src/main/java/com/palantir/dialogue/core/NodeSelectionStrategyChannel.java
jkozlowski: See the discussion above:

> I think you suggested we could just directly send a reference to the per host channel? I think that suffers from the problem that those will be re-created when the set of URLs changes (I think we recreate all of channels then?), so even if some URLs stayed the same, the in-flight transactions would just do something weird (possible send requests to stale channels?)

I think this is the crux of the issue that is worth discussing: on a configuration change (say you add a host), you'll end up keeping a reference to the old channel, while the particular sticky channel is alive. I think your suggestion is better in terms of not screwing up the routing ever, and I don't see many downsides (we'll have a bit of a split brain for a little bit, maybe that's not the end of the world? The only problem is that these sticky channels can potentially be quite long lived (current most prevalent client can hold onto one for upwards of 5 minutes).

Curious of your thoughts

jkozlowski: Also second reason why I originally (on the previous PR) went with indexes: we actually have a user of #perHost channels, and I'd really like to move it over to this architecture as well (although I haven't tackled that in this PR yet), because the same user that uses sticky channels, also uses perHost for some heartbeating that needs to address nodes in the cluster.

jkozlowski:
For those interested, I have updated the description with pretty pictures.

dialogue-core/src/main/java/com/palantir/dialogue/core/StickyRoutingChannel.java
```diff
@@ -0,0 +1,107 @@
...
+        if (routingKey == null) {
+            return globalQueue.execute(endpoint, request);
+        } else {
+            return stickyQueues.execute(ImmutableStickyRoutingChannel.QueueKey.of(routingKey), endpoint, request);
+        }
+    }
+
+    private static final class StickyQueues {
+
+        private final LoadingCache<QueueKey, ExpiringQueue> queues;
```
carterkozak: I think this cache can be avoided if the routing key is/has the sticky session queue. That way work is front-loaded on sticky session creation, lifecycle is well bounded to the scope of the sticky channel reference, and we don't have to worry about half of the hardest problems in CS.

dialogue-core/src/main/java/com/palantir/dialogue/core/StickyChannelRejectedException.java
jkozlowski: This was just silly Jakub, didn't realize I can just call permit#dropped directly, instead of having to wire this into the Behavior.

dialogue-core/src/main/java/com/palantir/dialogue/core/StickyRoutingChannel.java
jkozlowski: Uuuu, I think that would be nice, lemme see if that works.

dialogue-core/src/main/java/com/palantir/dialogue/core/StickyEndpointChannels2.java
```diff
@@ -0,0 +1,212 @@
...
+            }
+
+            synchronized (this) {
+                if (hostId != null) {
+                    request.attachments().put(RoutingAttachments.EXECUTE_ON_HOST_ID_KEY, hostId);
+                    return endpointChannel.execute(request);
+                }
+
+                if (callInFlight == null) {
+                    callInFlight = DialogueFutures.addDirectCallback(
```
jkozlowski: *I think* this should be a transformAsync right? so that all the queued up future's callbacks execute AFTER this current one, since futures don't guarantee the ordering? I feel like it's not particularly safe otherwise.

dialogue-core/src/main/java/com/palantir/dialogue/core/StickyEndpointChannels2.java
```diff
@@ -0,0 +1,212 @@
...
+                return executeWithHostId(hostId, request, endpointChannel);
+            }
+
+            synchronized (this) {
+                if (hostId != null) {
+                    request.attachments().put(RoutingAttachments.EXECUTE_ON_HOST_ID_KEY, hostId);
+                    return endpointChannel.execute(request);
+                }
+
+                if (callInFlight == null) {
```
carterkozak: I find these easier to read if we do a single volatile read at the top before our conditional
```suggestion
                ListenableFuture<Response> callInFlightSnapshot = callInFlight;
                if (callInFlightSnapshot == null) {
```

dialogue-core/src/main/java/com/palantir/dialogue/core/StickyEndpointChannels2.java
```diff
@@ -0,0 +1,212 @@
...
+        @Override
+        public ListenableFuture<Response> execute(Request request, EndpointChannel endpointChannel) {
+            if (hostId != null) {
+                return executeWithHostId(hostId, request, endpointChannel);
+            }
+
+            synchronized (this) {
+                if (hostId != null) {
+                    request.attachments().put(RoutingAttachments.EXECUTE_ON_HOST_ID_KEY, hostId);
+                    return endpointChannel.execute(request);
```
carterkozak: Should simplify, this way it's more obvious that we're double-check locking. It threw me off that the code in the synchronized check wasn't identical to the initial check.
```suggestion
                    return executeWithHostId(hostId, request, endpointChannel);
```

dialogue-core/src/main/java/com/palantir/dialogue/core/StickyEndpointChannels2.java
carterkozak: Same goes for the `hostId` reference as well. It can be refreshed within the synchronized block.

dialogue-core/src/main/java/com/palantir/dialogue/core/StickyEndpointChannels2.java
```diff
@@ -0,0 +1,212 @@
...
+                }
+
+                if (callInFlight == null) {
+                    callInFlight = DialogueFutures.addDirectCallback(
+                            executeWithAttachHostId(request, endpointChannel), callback);
+                    return callInFlight;
+                } else {
+                    ListenableFuture<Response> result = callInFlight;
+                    result = DialogueFutures.transformAsync(result, _input -> execute(request, endpointChannel));
```
carterkozak, line 136: we need to wrap callInFlight initially to prevent `cancel` on queued calls from causing collateral damage. That's probably best to do around setting `callInFlight` initially rather than on each subsequent request.

dialogue-core/src/main/java/com/palantir/dialogue/core/StickyEndpointChannels2.java
```diff
@@ -0,0 +1,212 @@
...
+
+    private static final class DefaultStickyRouter implements StickyRouter {
+
+        private final InFlightCallCallback callback = new InFlightCallCallback();
+
+        @Nullable
+        private volatile HostId hostId;
+
+        @Nullable
+        private volatile ListenableFuture<Response> callInFlight;
```
carterkozak: Let's help error-prone help us:
```suggestion
        @Nullable
        @GuardedBy("this")
        private volatile ListenableFuture<Response> callInFlight;
```

dialogue-core/src/main/java/com/palantir/dialogue/core/StickyRoutingChannel.java
```diff
@@ -0,0 +1,107 @@
...
+import com.palantir.dialogue.RoutingAttachments;
+import com.palantir.dialogue.RoutingAttachments.RoutingKey;
+import java.time.Duration;
+import javax.annotation.CheckForNull;
+import org.immutables.value.Value;
+
+@Value.Enclosing
+final class StickyRoutingChannel implements Channel {
+
+    private final Channel globalQueue;
```
carterkozak: Might be easier to understand if we call this `delegate`. It doesn't strictly speaking have to be a queue at all.

simulation/src/test/java/com/palantir/dialogue/core/SimulationTest.java
```diff
@@ -714,18 +747,13 @@ public void after(TestInfo testInfo) throws IOException {
...
-                    totalRequestsPerClientPerEndpoint
-                    // simulation.metrics().chart(Pattern.compile("(responseClose|globalResponses)"))
-                    );
+            charts.add(simulation.metricsReporter().chart(MetricNames.serverRequestMeterPattern()));
+
+            charts.addAll(simulation.metricsReporter().charts(MetricNames.perClientEndpointResponseTimerPattern()));
+
+            // charts.add(simulation.metrics().chart(Pattern.compile("(responseClose|globalResponses)")));
+
+            SimulationMetricsReporter.png(pngPath, charts);
```
carterkozak: Let's break out any changes to the simulation output into a separate stacked/parallel PR, that way the simulation changes are more obviously coupled to improvements we make.

simulation/src/test/java/com/palantir/dialogue/core/SimulationTest.java
```diff
@@ -544,6 +546,33 @@ void server_side_rate_limits(Strategy strategy) {
...
+                    .build();
+        };
+        server_side_rate_limits_with_sticky_clients_stready_vs_bursty_client_impl(
+                Benchmark.builder(), strategy, factory);
+    }
+
+    @SimulationCase
+    void server_side_rate_limits_with_sticky2_clients_stready_vs_bursty_client(Strategy strategy) {
+        // Ignore that one, because it's currently failing.
+        Assumptions.assumeTrue(strategy != Strategy.UNLIMITED_ROUND_ROBIN);
```
carterkozak, line 563: Is this due to the unimplemented stub method?

dialogue-core/src/main/java/com/palantir/dialogue/core/StickyConcurrencyLimitedChannel.java
```diff
@@ -0,0 +1,94 @@
...
+
+    @Override
+    public Optional<ListenableFuture<Response>> maybeExecute(Endpoint endpoint, Request request, boolean force) {
+        Optional<CautiousIncreaseAggressiveDecreaseConcurrencyLimiter.Permit> maybePermit = limiter.acquire(force);
+        if (maybePermit.isPresent()) {
+            CautiousIncreaseAggressiveDecreaseConcurrencyLimiter.Permit permit = maybePermit.get();
+            logPermitAcquired();
+
+            Optional<ListenableFuture<Response>> result =
+                    delegate.maybeExecute(endpoint, request, permit.isOnlyInFlight());
```
carterkozak: bypassing other concurrency limiters when `isOnlyInFlight` is critical to this working without a busy spin loop, but it's non-obvious. We should add a comment explaining the importance.

carterkozak: I think this needs to be 
```suggestion
                    delegate.maybeExecute(endpoint, request, force || permit.isOnlyInFlight());
```

dialogue-core/src/main/java/com/palantir/dialogue/core/StickyEndpointChannels2.java
```diff
@@ -0,0 +1,212 @@
...
+    }
+
+    @Override
+    public Supplier<Channel> get() {
+        return new StickySessionSupplier(delegate);
+    }
+
+    @Override
+    public String toString() {
+        return "StickyEndpointChannels2{}";
```
carterkozak:
```suggestion
        return "StickyEndpointChannels2{" + delegate + "}";
```

dialogue-core/src/main/java/com/palantir/dialogue/core/StickyEndpointChannels2.java
```diff
@@ -0,0 +1,212 @@
...
+         */
+        @Deprecated
+        @Override
+        public ListenableFuture<Response> execute(Endpoint endpoint, Request request) {
+            return endpoint(endpoint).execute(request);
+        }
+
+        @Override
+        public String toString() {
+            return "Sticky{" + channelFactory + '}';
```
carterkozak, line 99: should include the routing key

dialogue-core/src/main/java/com/palantir/dialogue/core/StickyEndpointChannels2.java
```diff
@@ -0,0 +1,212 @@
...
+
+        @Override
+        public ListenableFuture<Response> execute(Request request) {
+            request.attachments().put(RoutingAttachments.ROUTING_KEY, routingKey);
+            return stickyRouter.execute(request, delegate);
+        }
+
+        @Override
+        public String toString() {
+            return "ScoreTrackingEndpointChannel{" + "delegate=" + delegate + '}';
```
carterkozak: +routingKey

carterkozak (commented):
Let's try to break off small, well-encapsulated pieces to merge separately so we can keep the diff reasonably small in this PR. Separating out response attachments should simplify this a bit, and extracting out simulations changes will drop the file count substantially.

simulation/src/test/java/com/palantir/dialogue/core/SimulationTest.java
jkozlowski: Absolutely

dialogue-core/src/main/java/com/palantir/dialogue/core/StickyEndpointChannels2.java
jkozlowski: Grr, I swear I had this in, but must've rushed it.

jkozlowski: I don't think callInFlight works, because then cancel on the returned future, would not cancel the actual work future?

dialogue-core/src/main/java/com/palantir/dialogue/core/StickyRoutingChannel.java
jkozlowski: Have a look, I've done it this way.

jkozlowski:
Closing as close to merging in smaller pieces.

PR #1307, opened 2021-07-08
Concurrency limiter uses #compareAndSet.

jkozlowski:
## Before this PR

Not using #compareAndSet can lead to releasing more permits than limit under load.

## After this PR
==COMMIT_MSG==
Use #compareAndSet when checking limits.
==COMMIT_MSG==


dialogue-core/src/main/java/com/palantir/dialogue/core/CautiousIncreaseAggressiveDecreaseConcurrencyLimiter.java
```diff
@@ -69,17 +69,28 @@
...
+        AtomicInteger localInFlight = inFlight;
+        AtomicDouble localLimit = limit;
+        while (true) {
+
+            // We don't want to hand out a permit if there are 4 inflight and a limit of 4.1, as this will immediately
+            // send our inflight number to 5, which is clearly above the limit.
+            // Instead, we wait until there is capacity for one whole request before handing out a permit.
+            // In the worst-case scenario with zero inflight and a limit of 1, we'll still hand out a permit.
+            int currentInFlight = localInFlight.get();
+            if (currentInFlight >= localLimit.get()) {
```
carterkozak: I think we should snapshot the limit value (including the cast to an integer) before we enter this loop. Either way there's an allowable data-race because the limit may change between the time it's read, and when we verify currentInFight, but that's not problematic.

carterkozak: This comparison isn't the inverse of the original `currentInFlight <= getLimit() - 1` because values are converted to `double` rather than `integer`.

jkozlowski: I guess it's a slightly larger window for inconsistency this way, but yeah why not.

dialogue-core/src/main/java/com/palantir/dialogue/core/CautiousIncreaseAggressiveDecreaseConcurrencyLimiter.java
```diff
@@ -69,17 +70,28 @@
...
+        // Capture the limit field reference once to avoid work in a tight loop. The JIT cannot
+        // reliably optimize out references to final fields due to the potential for reflective
+        // modification.
+        AtomicInteger localInFlight = inFlight;
+
+        // We don't want to hand out a permit if there are 4 inflight and a limit of 4.1, as this will immediately
+        // send our inflight number to 5, which is clearly above the limit.
+        // Instead, we wait until there is capacity for one whole request before handing out a permit.
+        // In the worst-case scenario with zero inflight and a limit of 1, we'll still hand out a permit.
+        int currentLimit = (int) (getLimit() - 1);
```
carterkozak: The offset doesn’t look right, that could give us a limit of zero in the worst case, which is irrecoverable. Casting to int should be sufficient to truncate the fractional components, right?
```suggestion
        int currentLimit = (int) getLimit();
```

dialogue-core/src/main/java/com/palantir/dialogue/core/CautiousIncreaseAggressiveDecreaseConcurrencyLimiter.java
```diff
@@ -69,17 +70,28 @@
...
+        while (true) {
+            int currentInFlight = localInFlight.get();
+            if (currentInFlight > currentLimit) {
```
carterkozak: Proposed change above requires:
```suggestion
            if (currentInFlight >= currentLimit) {
```

dialogue-core/src/test/java/com/palantir/dialogue/core/CautiousIncreaseAggressiveDecreaseConcurrencyLimiterTest.java
```diff
@@ -211,4 +244,51 @@ public void onFailure_ignoresForNonIoExceptions(Behavior behavior) {
         limiter.acquire().get().onFailure(exception);
         assertThat(limiter.getLimit()).isEqualTo(max);
     }
+
+    @Test
+    public void acquire_doesNotReleaseMorePermitsThanLimit() throws ExecutionException, InterruptedException {
```
jkozlowski, line 249: Test fails before this change, takes 40ms on my laptop.

carterkozak (approved):

PR #1310, opened 2021-07-09
More updates to simulations

jkozlowski:
## Before this PR

Bursty sticky client simulation does not clearly show the slow client making progress. To remedy that, we also render per client requests in separate charts.

## After this PR
==COMMIT_MSG==
Small improvements to simulations.
==COMMIT_MSG==


simulation/src/main/java/com/palantir/dialogue/core/Benchmark.java
```diff
@@ -86,6 +86,11 @@ static Benchmark builder() {
         return new Benchmark();
     }
 
+    public Benchmark mutableRequests() {
+        requestSupplier = Benchmark::constructNewRequest;
+        return this;
+    }
```
carterkozak: Is this used?

jkozlowski: Not yet, thought I'd reduce the diff. Can remove if you want, but this is necessary for the other sticky implementation that pushes attachments around.

simulation/src/main/java/com/palantir/dialogue/core/SimulationMetricsReporter.java
```diff
@@ -116,17 +135,39 @@ public XYChart chart(Pattern metricNameRegex) {
         chart.getStyler().setToolTipsEnabled(true);
         chart.getStyler().setToolTipsAlwaysVisible(true);
 
+        if (!simulation.events().getEvents().isEmpty()) {
+            double[] eventXs = simulation.events().getEvents().keySet().stream()
+                    .mapToDouble(nanos -> TimeUnit.NANOSECONDS.toMillis(nanos) / 1000d)
```
carterkozak: Is this equivalent to:
```suggestion
                    .mapToDouble(nanos -> nanos / 1_000_000_000D)
```
It took me a moment to parse that we were converting nanos to seconds with decimals.

jkozlowski: This is copy pasta from what was already there, I did not even read this code, just shuffled things around.

simulation/src/main/java/com/palantir/dialogue/core/Benchmark.java
carterkozak: I see, perhaps since we've already added request attachments, we should always use mutable request objects to prevent leaking data between requests?

simulation/src/main/java/com/palantir/dialogue/core/SimulationMetricsReporter.java
jkozlowski: Wait, isn't this just trying to convert to seconds? but just like get fractional seconds.

jkozlowski: Seems to not change the pngs, so let's do it. I also extracted a method to give this a name.

simulation/src/main/java/com/palantir/dialogue/core/Benchmark.java
jkozlowski: That would work, although I suspected this was done because the simulations were melting laptops. Lemme try with this by default.

jkozlowski: Alright, laptop didn't explode let's see CI.

carterkozak (approved):
lgtm :+1:

PR #1312, opened 2021-07-09
Response attachments.

jkozlowski:
## Before this PR

No way to attach mutable state to responses. As with requests, this turns out to be quite a nice and useful capability, so we'd like to extend this to responses.

## After this PR
==COMMIT_MSG==
Responses can have attachments.
==COMMIT_MSG==

## Possible downsides?

* Moar API.
* A lot of copy pasta: we could consider a nice AbstractResponse and shove some shared code there, but I didn't want to start with that because I know this is a contentious topic around here.

### DIFF
--- a/.palantir/revapi.yml
+++ b/.palantir/revapi.yml
@@ -249,3 +249,8 @@ acceptedBreaks:
       new: "method <T> T com.palantir.dialogue.Clients::callBlocking(com.palantir.dialogue.EndpointChannel,\
         \ com.palantir.dialogue.Request, com.palantir.dialogue.Deserializer<T>)"
       justification: "New API for blocking calls."
+  "2.10.0":
+    com.palantir.dialogue:dialogue-target:
+    - code: "java.method.addedToInterface"
+      new: "method com.palantir.dialogue.ResponseAttachments com.palantir.dialogue.Response::attachments()"
+      justification: "New state in Response."
--- a/dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ResponseLeakDetector.java
+++ b/dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ResponseLeakDetector.java
@@ -19,6 +19,7 @@ package com.palantir.dialogue.hc5;
 import com.google.common.collect.ListMultimap;
 import com.palantir.dialogue.Endpoint;
 import com.palantir.dialogue.Response;
+import com.palantir.dialogue.ResponseAttachments;
 import com.palantir.logsafe.SafeArg;
 import com.palantir.logsafe.exceptions.SafeRuntimeException;
 import com.palantir.tritium.metrics.registry.TaggedMetricRegistry;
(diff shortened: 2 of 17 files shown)


.palantir/revapi.yml
```diff
@@ -249,3 +249,8 @@ acceptedBreaks:
       new: "method <T> T com.palantir.dialogue.Clients::callBlocking(com.palantir.dialogue.EndpointChannel,\
         \ com.palantir.dialogue.Request, com.palantir.dialogue.Deserializer<T>)"
       justification: "New API for blocking calls."
+  "2.10.0":
```
jkozlowski, line 252: Technically if someone published a library that implemented Response, and then someone else pushed used that, they'd get a runtime error.

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ResponseLeakDetector.java
```diff
@@ -182,6 +183,11 @@ public int code() {
             return delegate.getFirstHeader(header);
         }
 
+        @Override
+        public ResponseAttachments attachments() {
+            return delegate.attachments();
```
jkozlowski, line 188: I think the delegation thing kinda precludes AbstractResponse.

.palantir/revapi.yml
carterkozak: Yep, that's fine. We should mention it in the changelog, but I don't expect any external Response implementations. We could call this a `break` and major rev because numbers are cheap.

carterkozak:
lgtm, needs a changelog

.palantir/revapi.yml
jkozlowski: alright calling this a break

.palantir/revapi.yml
```diff
@@ -249,3 +249,8 @@ acceptedBreaks:
...
+    com.palantir.dialogue:dialogue-target:
```
jkozlowski, line 253: Assuming this version needs to be bumped to the release version?

dialogue-target/src/main/java/com/palantir/dialogue/ResponseAttachments.java
```diff
@@ -0,0 +1,39 @@
...
+package com.palantir.dialogue;
+
+import javax.annotation.Nullable;
+
+public final class ResponseAttachments {
+    private final Attachments attachments = Attachments.create();
+
+    private ResponseAttachments() {}
+
+    public static ResponseAttachments create() {
```
jkozlowski, line 26: This is public because Response implementations are scattered all over.

jkozlowski: Would be nice if Response was also a concrete type, not sure what we're gaining from it being an interface. If it just took Supplier<InputStream>, code and headers, looks like it should be possible to have a concrete class?

Then this wouldn't have had to be a break I guess. No strong preference though.

.palantir/revapi.yml
carterkozak: It’s a bit of a race, the version isn’t revalidated and that won’t break anything.

dialogue-target/src/main/java/com/palantir/dialogue/ResponseAttachments.java
carterkozak: I think that would result in a subtler break in the case someone wrapped an existing response object, since it would compile but the attachments would be reset.

carterkozak (approved):

PR #1316, opened 2021-07-12
LimitedChannel supports disabling limit enforcement.

jkozlowski:
## Before this PR

In order to support fair queues, we need the ability to skip limits for certain (small number) of requests.

## After this PR
==COMMIT_MSG==
LimitedChannel supports skipping limits.
==COMMIT_MSG==

## Possible downsides?

Need to add tests still.

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/ConcurrencyLimitedChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/ConcurrencyLimitedChannel.java
@@ -17,7 +17,6 @@
 package com.palantir.dialogue.core;
 
 import com.google.common.util.concurrent.ListenableFuture;
-import com.palantir.conjure.java.client.config.ClientConfiguration;
 import com.palantir.dialogue.Channel;
 import com.palantir.dialogue.Endpoint;
 import com.palantir.dialogue.Request;
@@ -26,7 +25,6 @@ import com.palantir.dialogue.core.CautiousIncreaseAggressiveDecreaseConcurrencyL
 import com.palantir.dialogue.futures.DialogueFutures;
 import com.palantir.logsafe.Preconditions;
 import com.palantir.logsafe.SafeArg;
-import com.palantir.logsafe.exceptions.SafeIllegalStateException;
 import com.palantir.tritium.metrics.registry.TaggedMetricRegistry;
 import java.util.Optional;
 import java.util.stream.LongStream;
@@ -36,7 +34,7 @@ import org.slf4j.LoggerFactory;
 /**
  * A channel that monitors the successes and failures of requests in order to determine the number of concurrent
  * requests allowed to a particular channel. If the channel's concurrency limit has been reached, the
- * {@link #maybeExecute} method returns empty.
+ * {@link LimitedChannel#maybeExecute} method returns empty.
  */
 final class ConcurrencyLimitedChannel implements LimitedChannel {
     private static final Logger log = LoggerFactory.getLogger(ConcurrencyLimitedChannel.class);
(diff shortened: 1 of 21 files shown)


dialogue-core/src/main/java/com/palantir/dialogue/core/LimitedChannel.java
```diff
@@ -29,7 +29,13 @@
...
+ * Implementations should support skipping the limit checks, while still tracking usage appropriately.
  */
 interface LimitedChannel {
-    Optional<ListenableFuture<Response>> maybeExecute(Endpoint endpoint, Request request);
+    Optional<ListenableFuture<Response>> maybeExecute(Endpoint endpoint, Request request, SkipLimits skipLimits);
+
+    enum SkipLimits {
+        Yes,
+        No
+    }
```
carterkozak: Perhaps something along the lines of:
```suggestion
    enum LimitEnforcement {
        DEFAULT_ENABLED,
        DANGEROUS_BYPASS_LIMITS
    }
```

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
```diff
@@ -62,6 +63,7 @@
  */
 final class QueuedChannel implements Channel {
     private static final Logger log = LoggerFactory.getLogger(QueuedChannel.class);
+    private static final SkipLimits DO_NOT_SKIP_LIMITS = SkipLimits.No;
```
carterkozak: Why not reference the enum directly?

jkozlowski: I was just trying to encode that really both codepaths here should use the same value, cause otherwise it's weird.

dialogue-core/src/main/java/com/palantir/dialogue/core/LimitedChannel.java
jkozlowski: +1 will do manually cause need to do in an IDE.

dialogue-core/src/main/java/com/palantir/dialogue/core/ConcurrencyLimitedChannel.java
```diff
@@ -46,23 +44,11 @@
...
-            case DANGEROUS_DISABLE_SYMPATHETIC_CLIENT_QOS:
-                return new ChannelToLimitedChannelAdapter(channel);
-        }
-        throw new SafeIllegalStateException(
-                "Encountered unknown client QoS configuration", SafeArg.of("ClientQoS", clientQoS));
+        TaggedMetricRegistry metrics = cf.clientConf().taggedMetricRegistry();
+        CautiousIncreaseAggressiveDecreaseConcurrencyLimiter limiter = createLimiter(Behavior.HOST_LEVEL);
+        ConcurrencyLimitedChannelInstrumentation instrumentation =
+                new HostConcurrencyLimitedChannelInstrumentation(cf.channelName(), uriIndex, limiter, metrics);
+        return new ConcurrencyLimitedChannel(channel, limiter, instrumentation);
```
carterkozak, line 51: Nice

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
```diff
@@ -62,6 +63,7 @@
  */
 final class QueuedChannel implements Channel {
     private static final Logger log = LoggerFactory.getLogger(QueuedChannel.class);
+    private static final LimitEnforcement DO_NOT_SKIP_LIMITS = LimitEnforcement.DEFAULT_ENABLED;
```
carterkozak, line 66: nit: not sure the constant makes things clearer

carterkozak (approved):

PR #1323, opened 2021-07-14
[WIP] DialogueChannel can route to a specific host.

jkozlowski:
## Before this PR

Host-specific routing is a concern external to DialogueChannel.

## After this PR
==COMMIT_MSG==
DialogueChannel can route to a specific host.
==COMMIT_MSG==

### DIFF
--- a/dialogue-clients/src/main/java/com/palantir/dialogue/clients/ReloadingClientFactory.java
+++ b/dialogue-clients/src/main/java/com/palantir/dialogue/clients/ReloadingClientFactory.java
@@ -50,6 +50,7 @@ import com.palantir.refreshable.Refreshable;
 import com.palantir.tritium.metrics.registry.TaggedMetricRegistry;
 import java.security.Provider;
 import java.time.Duration;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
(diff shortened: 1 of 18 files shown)


dialogue-clients/src/main/java/com/palantir/dialogue/clients/ReloadingClientFactory.java
```diff
@@ -333,14 +311,42 @@ public String toString() {
...
+        }
+    }
+
+    interface InternalDialogueChannel extends Channel {
+        List<Channel> perHostChannels();
+    }
+
+    private static final class InternalDialogueChannelFromDialogueChannel implements InternalDialogueChannel {
+
+        private final DialogueChannel dialogueChannel;
```
carterkozak, line 328: Not sure why we don't just use the DialogueChannel directly. Don't we want one that also implements EndpointChannelFactory?

jkozlowski: There might be a better way to do it, but if you have a look at ReloadingClientFactory#getInternalDialogueChannel, we have things like AlwaysThrowingChannel, and the return type must have the same super-type.

carterkozak: Got it. Might be helpful to add that detail to InternalDialogueChannel javadoc.

dialogue-core/src/test/java/com/palantir/dialogue/core/DialogueChannelTest.java
```diff
@@ -301,6 +301,9 @@ void nice_tostring() {
                 .isNotEqualTo(channel2.toString());
     }
 
+    @Test
+    public void can_request_executed_on_attachment() {}
```
carterkozak, line 305:
```suggestion
    public void can_request_executed_on_attachment() {
        throw new AssertionError("TODO");
    }
```

dialogue-clients/src/main/java/com/palantir/dialogue/clients/ReloadingClientFactory.java
```diff
@@ -316,7 +260,41 @@ public String toString() {
...
+            ServiceConfiguration serviceConf =
+                    ServiceConfigurationFactory.of(block).get(serviceName);
+
+            DialogueChannel dialogueChannel =
+                    cache.getNonReloadingChannel(params, serviceConf, channelName, OptionalInt.empty());
+            return new InternalDialogueChannelFromDialogueChannel(dialogueChannel);
+        });
+    }
+
+    private static final class AlwaysThrowingChannel implements InternalDialogueChannel {
```
carterkozak:
```suggestion
    private static final class NoUrisDialogueChannel implements InternalDialogueChannel {
```

dialogue-core/src/main/java/com/palantir/dialogue/core/HostLimitedChannel.java
```diff
@@ -0,0 +1,21 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.dialogue.core;
+
+interface HostLimitedChannel extends LimitedChannel {
+    HostIdx getHostIdx();
```
carterkozak: Perhaps compose over inherit: `LimitedChannel channel()` and rename to HostAndLimitedChannel or similar?

PR #1327, opened 2021-07-14
Extract HostAndLimitedChannel.

jkozlowski:
## Before this PR

Various components try to make sure they can refer to host by their original stable ids. We will also need the ability to lookup the channels by their original ids, even after scrambling, so it made sense to extract a shared representation.

## After this PR
==COMMIT_MSG==
Extract HostAndLimitedChannel and a container for them.
==COMMIT_MSG==


dialogue-core/src/main/java/com/palantir/dialogue/core/HostAndLimitedChannels.java
```diff
@@ -0,0 +1,77 @@
...
+
+final class HostAndLimitedChannels {
+
+    private final ImmutableList<HostAndLimitedChannel> channels;
+    private final BiMap<HostIdx, HostAndLimitedChannel> lookups;
+
+    private HostAndLimitedChannels(ImmutableList<HostAndLimitedChannel> channels) {
+        this.channels = channels;
+        this.lookups = HashBiMap.create(channels.size());
+        channels.forEach(hostLimitedChannel -> lookups.put(hostLimitedChannel.getHostIdx(), hostLimitedChannel));
```
carterkozak: nit: Can we use an `ImmutableBiMap` for symmetry with the `ImmutableList`?

dialogue-core/src/main/java/com/palantir/dialogue/core/HostAndLimitedChannels.java
```diff
@@ -0,0 +1,77 @@
...
+        return channels.get(index);
+    }
+
+    HostAndLimitedChannel getByHostIdx(HostIdx hostIdx) {
+        return Preconditions.checkNotNull(lookups.get(hostIdx), "Unknown hostIdx");
+    }
+
+    int unorderedIndexOf(HostAndLimitedChannel hostAndLimitedChannel) {
+        return channels.indexOf(hostAndLimitedChannel);
+    }
```
carterkozak: Could use some javadoc describing the difference between unordered index and hostIndex index. It's hard to reason through why we have both.

dialogue-core/src/main/java/com/palantir/dialogue/core/BalancedNodeSelectionStrategyChannel.java
```diff
@@ -56,18 +55,22 @@
     private static final int UNHEALTHY_SCORE_MULTIPLIER = 2;
 
     private final BalancedScoreTracker tracker;
-    private final ImmutableList<LimitedChannel> channels;
+    private final HostAndLimitedChannels channels;
 
     BalancedNodeSelectionStrategyChannel(
-            ImmutableList<LimitedChannel> channels,
+            HostAndLimitedChannels channels,
```
carterkozak, line 61: Not a suggestion for this change, just musing: Part of me wants to apply BalancedScoreTracker into HostAndLimitedChannels to consolidate our metrics. The data is useful for pin-until-error and single-host configurations, and it would open the door for reloads/strategy changes to retain existing knowledge of dead hosts.

carterkozak:
I need to take a closer look at the stableId/HostIdx pieces in the morning, but things generally look good.

dialogue-core/src/main/java/com/palantir/dialogue/core/HostAndLimitedChannels.java
jkozlowski: yeah, I wasn't quite sure how to best handle this, but in my original attempt the pin until error channel still kept it's own PinChannel thing, which was exactly the same as this thing, but allowed you to shuffle and refer to channels with just an List index. I'm open to suggestions on how to reword/restructure this, but basically the contract is that HostAndLimitedChannels maintains the original HostIdx -> LimitedChannel, even though you can scramble the ordering if you need it for the usecase (pin until error).

dialogue-core/src/main/java/com/palantir/dialogue/core/BalancedNodeSelectionStrategyChannel.java
jkozlowski: How would you use balancedscore tracker in pin until error? surely scores will be empty for the unpinned hosts?

dialogue-core/src/main/java/com/palantir/dialogue/core/HostAndLimitedChannels.java
jkozlowski: Ok, I added a bit of javadoc, and removed the specilized weird methods. Instead just have getChannels and all the usual ImmutableList methods can be used directly.

dialogue-core/src/main/java/com/palantir/dialogue/core/HostAndLimitedChannel.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+package com.palantir.dialogue.core;
+
+import com.palantir.dialogue.DialogueImmutablesStyle;
+import org.immutables.value.Value;
+
+@Value.Immutable
+@DialogueImmutablesStyle
+interface HostAndLimitedChannel {
```
jkozlowski, line 24: Just realized this as an immutable marker makes it hard to implement the executed-on marking channel. Just going to have a look at what I can do

jkozlowski: Done

dialogue-core/src/main/java/com/palantir/dialogue/core/BalancedNodeSelectionStrategyChannel.java
carterkozak: It depends -- if nodes fail frequently, we may rotate across nodes quickly. The score just isn't used by the pin-until-error strategy.

dialogue-core/src/main/java/com/palantir/dialogue/core/HostAndLimitedChannel.java
carterkozak: I'm not sure I follow why this is -- can you elaborate? I found the interface a bit harder to follow.

dialogue-core/src/main/java/com/palantir/dialogue/core/HostAndLimitedChannels.java
jkozlowski: Done

dialogue-core/src/main/java/com/palantir/dialogue/core/BalancedNodeSelectionStrategyChannel.java
jkozlowski: > The score just isn't used by the pin-until-error strategy.

I know, but how would you propose it to be used?

dialogue-core/src/main/java/com/palantir/dialogue/core/HostAndLimitedChannel.java
jkozlowski: Reverted, host marking will live in NodeSelectionChannel

dialogue-core/src/main/java/com/palantir/dialogue/core/BalancedNodeSelectionStrategyChannel.java
carterkozak: Dashboards for investigation. Unclear how useful that would be.

carterkozak:
The index pieces of this PR scare me because indexes can change when the configuration is reloaded, and cause requests to hit a new node even though the original node is still present.

I tried to implement the simplest possible attachment-based sticky API without refactoring much existing code here: https://github.com/palantir/dialogue/pull/1331

I think there are some foot-guns around potential changes to node selection causing sticky-session bugs, but it should also make it much harder to break existing code, or hit the wrong node. What do you think?

PR #1330, opened 2021-07-15
NodeSelectionStrategyChannel can route to a particular host.

jkozlowski:
## Before this PR

Needs to merge after https://github.com/palantir/dialogue/pull/1327.

DialogueChannels cannot route to a particular host.

## After this PR
==COMMIT_MSG==
NodeSelectionStrategyChannel can route to a particular host.
==COMMIT_MSG==


dialogue-core/src/main/java/com/palantir/dialogue/core/PinUntilErrorNodeSelectionStrategyChannel.java
```diff
@@ -154,7 +172,7 @@ public void onFailure(Throwable throwable) {
                 instrumentation.receivedThrowable(pin, channel, throwable, next);
             }
         });
-        return maybeResponse;
+        return ExecutedOnResponseMarker.maybeMarkExecutedOn(request, maybeResponse, channel);
```
jkozlowski: This is obviously quite bad, if we ever impl another strategy, we're bound to forget about this piece.

1. Go back to HostAndLimitedChannel being an interface.
2. Change the API for setting the host to take a LimitedChannel. <- I think this is probably the best way forward

jkozlowski: Latest commit has 2. above.

PR #1331, opened 2021-07-15
Quick prototype of sticky handling mechanisms with minimal code-change

carterkozak:
==COMMIT_MSG==
Quick prototype of sticky handling mechanisms with minimal code-change
==COMMIT_MSG==

The exposed/public API is:

https://github.com/palantir/dialogue/blob/558edccae45b7326780330ba6f9dcf2bf2d55cf2/dialogue-core/src/main/java/com/palantir/dialogue/core/StickyAttachments.java#L32-L47

## Possible downsides?
has no test coverage, I have not verified any of this.

This doesn't give us the tools to expose a `list<channel>` to
improve `PerHostClientFactory`, however that should be feasible
with a little bit of plumbing if we choose to do so.

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/NodeSelectionStrategyChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/NodeSelectionStrategyChannel.java
@@ -24,6 +24,7 @@ import com.google.common.util.concurrent.ListenableFuture;
 import com.palantir.dialogue.Endpoint;
 import com.palantir.dialogue.Request;
 import com.palantir.dialogue.Response;
+import com.palantir.dialogue.core.StickyAttachments.StickyTarget;
 import com.palantir.dialogue.futures.DialogueFutures;
 import com.palantir.logsafe.SafeArg;
 import com.palantir.logsafe.exceptions.SafeRuntimeException;
@@ -74,11 +75,11 @@ final class NodeSelectionStrategyChannel implements LimitedChannel {
 
     static LimitedChannel create(Config cf, ImmutableList<LimitedChannel> channels) {
         if (channels.isEmpty()) {
-            return new ZeroUriNodeSelectionChannel(cf.channelName());
+            return new StickyChannelHandler(new ZeroUriNodeSelectionChannel(cf.channelName()));
         }
 
         if (channels.size() == 1) {
-            return channels.get(0);
+            return new StickyChannelHandler(new StuckRequestHandler(channels.get(0)));
         }
 
         return new NodeSelectionStrategyChannel(
@@ -94,7 +95,8 @@ final class NodeSelectionStrategyChannel implements LimitedChannel {
     @Override
     public Optional<ListenableFuture<Response>> maybeExecute(
             Endpoint endpoint, Request request, LimitEnforcement limitEnforcement) {
-        Optional<ListenableFuture<Response>> maybe = delegate.maybeExecute(endpoint, request, limitEnforcement);
+        Optional<ListenableFuture<Response>> maybe =
+                StickyChannelHandler.maybeExecute(delegate, endpoint, request, limitEnforcement);
         if (!maybe.isPresent()) {
             return Optional.empty();
         }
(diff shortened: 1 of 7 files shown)


dialogue-core/src/main/java/com/palantir/dialogue/core/NodeSelectionStrategyChannel.java
```diff
@@ -74,11 +75,11 @@
 
     static LimitedChannel create(Config cf, ImmutableList<LimitedChannel> channels) {
         if (channels.isEmpty()) {
-            return new ZeroUriNodeSelectionChannel(cf.channelName());
+            return new StickyChannelHandler(new ZeroUriNodeSelectionChannel(cf.channelName()));
```
carterkozak, line 78:
```suggestion
            // If we reloaded away from a channel, we should still attempt to allow remaining requests
            // to execute against the old reference.
            return new StickyChannelHandler(new ZeroUriNodeSelectionChannel(cf.channelName()));
```

dialogue-core/src/main/java/com/palantir/dialogue/core/StickyAttachments.java
```diff
@@ -0,0 +1,71 @@
...
+            ResponseAttachmentKey.create(StickyTarget.class);
+
+    /**
+     * Used to execute requests against the same host.
+     */
+    public static final RequestAttachmentKey<StickyTarget> STICKY = RequestAttachmentKey.create(StickyTarget.class);
+
+    private StickyAttachments() {}
+
+    // package private, used exclusively
```
carterkozak, line 51:
```suggestion
    // package private, used exclusively around the borders of the node selection channel
```

dialogue-core/src/main/java/com/palantir/dialogue/core/PinUntilErrorNodeSelectionStrategyChannel.java
```diff
@@ -133,7 +133,10 @@ LimitedChannel getCurrentChannel() {
         int pin = currentPin.get();
         PinChannel channel = nodeList.get(pin);
 
-        Optional<ListenableFuture<Response>> maybeResponse = channel.maybeExecute(endpoint, request, limitEnforcement);
+        // n.b. StickyAttachments.maybeExecute uses the delegate PinChannel, which bypasses the FutureCallback
```
jkozlowski, line 136: Why not include this instrumentation? Because you'd have to allocate to associate the pin with PinChannel?

jkozlowski: Granted we don't use this strategy with sticky sessions, it's still a bit sucky not to handle it correctly.

jkozlowski: I think if PinChannel knows it's current index, as well as the stable index, you won't have to allocate; then just need to change the recovery logic

```
initialShuffle.indexOf(limitedChannel)
```

To actually only compare the underlying LimitedChannels?

jkozlowski: Discussed: this is fine because the initial request will still be tracked, so if that happens to fail, we'll move hosts anyway. Not worth being complete. @carterkozak just to add a comment why this is fine

dialogue-core/src/main/java/com/palantir/dialogue/core/NodeSelectionStrategyChannel.java
jkozlowski: +1

dialogue-core/src/main/java/com/palantir/dialogue/core/StickyAttachments.java
jkozlowski: +1

dialogue-core/src/main/java/com/palantir/dialogue/core/StickyAttachments.java
```diff
@@ -0,0 +1,71 @@
...
+import com.palantir.dialogue.Request;
+import com.palantir.dialogue.RequestAttachmentKey;
+import com.palantir.dialogue.Response;
+import com.palantir.dialogue.ResponseAttachmentKey;
+import com.palantir.dialogue.core.LimitedChannel.LimitEnforcement;
+import com.palantir.dialogue.futures.DialogueFutures;
+import java.util.Optional;
+import javax.annotation.CheckReturnValue;
+
+public final class StickyAttachments {
```
jkozlowski, line 30: nit: Not public.

dialogue-core/src/main/java/com/palantir/dialogue/core/StickyAttachments.java
```diff
@@ -0,0 +1,71 @@
...
+
+    /**
+     * Added to {@link com.palantir.dialogue.RequestAttachments} to opt into a {@link #STICKY_TOKEN} on the response.
+     */
+    public static final RequestAttachmentKey<Boolean> REQUEST_STICKY_TOKEN = RequestAttachmentKey.create(Boolean.class);
```
jkozlowski, line 35: How about keeping these private, and providing API for setting/getting?

dialogue-core/src/main/java/com/palantir/dialogue/core/NodeSelectionStrategyChannel.java
jkozlowski: I don't think this is a concern: reload creates a new DialogueChannel; existing sticky channels (as defined here)https://github.com/palantir/dialogue/pull/1334/files will continue referring to the old DialogueChannel instance; by construction (as long as we make StickyAttachments package private) there is should be no way for requests from one DialogueChannel (and therefore one set of LimitedChannels) to bleed into another DialogueChannel.

jkozlowski (commented):
Some nits

dialogue-core/src/main/java/com/palantir/dialogue/core/StickyAttachments.java
carterkozak: Yep, I wasn't sure if we'd implement the finalized API pieces within dialogue-core or the client module. Given #1334 these should be package-private.

jkozlowski: Yeah, I strongly feel this should live here because that's how we can hide this complexity and never run into the problem you tried to handle here: https://github.com/palantir/dialogue/pull/1331#discussion_r670848625.

jkozlowski:
Getting finished up here: https://github.com/palantir/dialogue/pull/1336/files

PR #1333, opened 2021-07-16
Per host channels

jkozlowski:
## After this PR
==COMMIT_MSG==

==COMMIT_MSG==

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueChannel.java
@@ -33,16 +33,23 @@ import com.palantir.logsafe.Safe;
 import java.util.OptionalInt;
 import java.util.Random;
 import java.util.concurrent.ScheduledExecutorService;
+import java.util.stream.IntStream;
 
 public final class DialogueChannel implements Channel, EndpointChannelFactory {
     private final EndpointChannelFactory delegate;
     private final Config cf;
     private final StickySessionFactory stickySessionFactory;
+    private final ImmutableList<Channel> perHostChannels;
 
-    private DialogueChannel(Config cf, EndpointChannelFactory delegate, StickySessionFactory stickySessionFactory) {
+    private DialogueChannel(
+            Config cf,
+            EndpointChannelFactory delegate,
+            StickySessionFactory stickySessionFactory,
+            ImmutableList<Channel> perHostChannels) {
         this.cf = cf;
         this.delegate = delegate;
         this.stickySessionFactory = stickySessionFactory;
+        this.perHostChannels = perHostChannels;
     }
 
     @Override
@@ -57,6 +64,10 @@ public final class DialogueChannel implements Channel, EndpointChannelFactory {
 
     public StickySessionFactory stickySessionFactory() {
         return stickySessionFactory;
+    }
+
+    public ImmutableList<Channel> perHost() {
+        return perHostChannels;
     }
 
     public static Builder builder() {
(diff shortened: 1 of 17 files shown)


dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueChannel.java
```diff
@@ -207,7 +218,33 @@ public DialogueChannel build() {
...
+                                return endpointChannel.execute(request);
+                            };
+                        }
+
+                        @Override
+                        public ListenableFuture<Response> execute(Endpoint endpoint, Request request) {
+                            return endpoint(endpoint).execute(request);
+                        }
+                    })
+                    .collect(ImmutableList.toImmutableList());
```
carterkozak, line 247: I think this piece is more or less the alternative described [here](https://github.com/palantir/dialogue/pull/1334#discussion_r671334170), except that we don't leverage it from the StickySessionFactory.

dialogue-core/src/main/java/com/palantir/dialogue/core/NodeSelectingChannel.java
```diff
@@ -0,0 +1,23 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.dialogue.core;
+
+import com.palantir.dialogue.Request;
+
+interface NodeSelectingChannel extends LimitedChannel {
+    void routeToHost(int index, Request request);
```
carterkozak: What if we used an API along the lines of:
```suggestion
    /** Returns a stable list of all known delegate nodes. */
    ImmutableList<LimitedChannel> nodeChannels();
```

carterkozak: Sorry, I think the API would actually look like this:

```suggestion
    Refreshable<? extends List<? extends LimitedChannel> nodeChannels();
```

The refreshable allows us to update the channels with the current node-selection-strategy info as the strategy changes, however within a single DialogueChannel the order would be guaranteed to be stable.
The refreshable would also update on config reload which may or may not impact the list order.

jkozlowski: Done, looks good.

dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueChannel.java
jkozlowski: Yeah: I think it's valuable to provide wholesale fairness implementation. If we're not going to ship the whole thing (i.e. you can support internal usecase for per-transaction fairness, with ability to create a new sticky channel from an existing sticky channel with new queue), I'd be happy to switch to per-host queues even for sticky channels. The reason I added it in https://github.com/palantir/dialogue/pull/1334 is because I wanted to ship the whole thing.

dialogue-core/src/main/java/com/palantir/dialogue/core/NodeSelectingChannel.java
jkozlowski: Hmmm I'm not sure what the refreshable buys us?

1. The order and set of uris is always fixed within a DialogueChannel; it gets Config not Refreshable<Config>. Refreshables are handled in ReloadingClientFactory.
2. just returning a List is fine: we know the original list of URLs does not change, so just using the integer to index into that, and creating a Channel per host as I've done, does the right thing? We can just put put the javadoc as you described, to explain that it should be in the same order as received and call it done?

jkozlowski: Unless you want to push the refreshable down to this layer? I'd be down for that, but definitely not in this piece of work.

carterkozak: I think the channel instances need to change if the server requests a different node selection strategy, that way we can update balanced channel scores based on host-targeted requests.

PR #1334, opened 2021-07-16
Sticky channels and queue override

jkozlowski:
## Before this PR

1. Sticky channels are separate from DialogueChannel and therefore do not share failure/concurrency limit tracking along with balanced requests.
2. Sticky channel implementation pins the host before the first request, therefore forcing all retries to land on the same host.
3. Sticky requests can overwhelm balanced requests.

## After this PR

1. Sticky channels are created directly in DialogueChannel: therefore underlying channels are shared.
2. Sticky channels send first requests as an unpinned request, queuing all subsequent requests until first request is done. First request can be routed as a normal request, crucially for retries it is not pinned to a single host. Once first request completes, subsequent requests are pinned to the same host.
3. We provide a degree of fairness between balanced and sticky requests: sticky channels are generally not allowed to overwhelm non-sticky requests, and sticky requests do not overwhelm each other (see simulation results).
4. If many sticky channels are created simultaneously, they can overtake the queue: as an implementation detail, we have to allow first request on a sticky channel to enter per-endpoint queues, without taking host limits into account. This is a tradeoff against a more expensive implementation, which we'd like to delay until we have more data that this is a problem in practice.

==COMMIT_MSG==
New sticky channel implementation.
==COMMIT_MSG==

## Possible downsides?

Listed above.


dialogue-core/src/main/java/com/palantir/dialogue/core/StickyEnpointChannels2.java
```diff
@@ -0,0 +1,194 @@
...
+                if (stickyTarget != null) {
+                    return executeWithStickyTarget(stickyTarget, request, endpointChannel);
+                }
+
+                ListenableFuture<Response> callInFlightSnapshot = callInFlight;
+                if (callInFlightSnapshot == null) {
+                    // Cannot use DialogueFutures#addDirectCallback because we want ordering of listeners:
+                    // we want the success/failure callbacks to run BEFORE the queued requests inspect the result of
+                    // the first call.
+                    ListenableFuture<Response> result = DialogueFutures.transform(
```
jkozlowski: @carterkozak I originally had this as a callback, but I couldn't convince myself that's safe: ListenableFutures do not guarantee order of callbacks. I guess it's safe because queued callbacks that run before the in-flight callback will just queue themselves again? Would that just OOM, because those callbacks would execute immediately? I think I'm not familiar enough with ListenableFutures impl, so this feels safer.

carterkozak: The ListenableFuture reference implementation notes that they cannot change the order because it would break a bunch of code internally and externally, but it's still better to transform when we require ordering because it's easier to reason about.

dialogue-core/src/main/java/com/palantir/dialogue/core/CautiousIncreaseAggressiveDecreaseConcurrencyLimiter.java
```diff
@@ -130,6 +130,17 @@ void onSuccess(Response result, PermitControl control) {
                 }
             }
 
+            @Override
+            void onFailure(Throwable _throwable, PermitControl control) {
+                control.ignore();
+            }
+        },
+        STICKY() {
```
carterkozak, line 138: As written I think the queue limit can only ever increase, never drop. I think we should use the HOST_LEVEL behavior here because this is still a per-host queue, just a step more isolated.

This way, at the limit, when a host becomes unhealthy (low limit based on the HOST_LEVEL limiter) the sticky session is also reduced to a similar number of permits and cannot dominate the queue.

dialogue-core/src/main/java/com/palantir/dialogue/core/QueueAttachments.java
```diff
@@ -0,0 +1,24 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.dialogue.core;
+
+import com.palantir.dialogue.Channel;
+import com.palantir.dialogue.RequestAttachmentKey;
+
+interface QueueAttachments {
```
carterkozak: nit: I tend to prefer concrete classes to make it clear this isn't meant to be subclassed

dialogue-core/src/main/java/com/palantir/dialogue/core/CautiousIncreaseAggressiveDecreaseConcurrencyLimiter.java
jkozlowski: Have a look at StickyConcurrencyLimitedChannel: it calls Permit#dropped when it fails to schedule on the LimitedChannel it delegates to. That's when it's concurrency limit will drop. It didn't feel like failures of the sticky requests is a significant signal to lowering it's concurrency permit: to me it's signal that the host is overloaded and therefore signal for the host concurrency limiter to limit it's permits. Once that's updated, it will naturally limit the sticky queue by virtue of not allowing more requests through.

dialogue-core/src/main/java/com/palantir/dialogue/core/QueueAttachments.java
jkozlowski: Done

dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueChannel.java
```diff
@@ -179,14 +188,27 @@ public DialogueChannel build() {
                 return new NeverThrowEndpointChannel(channel); // this must come last as a defensive backstop
             };
 
+            Supplier<Channel> stickyChannelSupplier = StickyEnpointChannels2.create(() -> {
+                LimitedChannel stickyLimitedChannel =
+                        StickyConcurrencyLimitedChannel.create(nodeSelectionChannel, cf.channelName());
+                Channel queueOverride = QueuedChannel.createForSticky(cf, stickyLimitedChannel);
```
carterkozak: <just my own notes, not requesting any change>

This provides one queue per sticky session:

This means we get fairness between individual sticky sessions as well as requests which don't use sticky sessions, however in the worst case there may be many sessions in flight which makes it possible for one 'forced' request for each transaction to bypass the queue and limits. I think this is a good trade-off to avoid complexity.

The alternative is to associate queues with hosts so they're shared by all sticky sessions that have stuck to a given host. Each use is likely a bit different, so I like the idea of separate concurrency limits to punish 'bad' transactions without penalizing healthy transactions so much.

dialogue-core/src/main/java/com/palantir/dialogue/core/QueueOverrideChannel.java
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    @Override
+    public ListenableFuture<Response> execute(Endpoint endpoint, Request request) {
+        Channel override = request.attachments().getOrDefault(QueueAttachments.QUEUE_OVERRIDE, null);
+        if (override != null) {
+            return override.execute(endpoint, request);
+        } else {
+            return defaultDelegate.execute(endpoint, request);
+        }
```
carterkozak: No change requested, defer to your judgement:

The `@Nullable` return value of `getOrDefault` is a little odd here, isn't it. I'd want to write this using the default, but that angers `NullAway`:
```suggestion
        Channel channel = request.attachments().getOrDefault(QueueAttachments.QUEUE_OVERRIDE, defaultDelegate);
        return channel.execute(endpoint, request);
```

dialogue-core/src/main/java/com/palantir/dialogue/core/CautiousIncreaseAggressiveDecreaseConcurrencyLimiter.java
carterkozak: Got it, that makes sense!

carterkozak (commented):
I like it :-)

dialogue-core/src/main/java/com/palantir/dialogue/core/QueueOverrideChannel.java
jkozlowski: Yeah, sadly there's no way to tell the null check that the result is null IFF the argument is null? I copied the design for Attachments#put from Iterables#getFirst in terms of the annotations (i.e. both arg and return type are nullable)

carterkozak:
I wonder what it would take to wire up the sticky fairness benchmark to the new mechanism?

dialogue-core/src/main/java/com/palantir/dialogue/core/QueueOverrideChannel.java
carterkozak: A Preconditions.checkNotNull around the call would do it, but it's pretty silly....

jkozlowski:
> I wonder what it would take to wire up the sticky fairness benchmark to the new mechanism?

I already had it done in the original PR, the benchmarks are already setup to switch on sticky mechanism.

dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueChannel.java
jkozlowski: > however in the worst case there may be many sessions in flight which makes it possible for one 'forced' request for each transaction to bypass the queue and limits.

So if you actually go back to the dominant resource fairness, and fairness in general, this is actually a perfectly valid strategy: by the way I implemented this, you only punch through the limits IFF you have zero other requests in flight. That means that in order to achieve min-max fairness you should absolutely let them through, because by definition they have the lowest utilisation of the available resources :)

My point is that this isn't a degenerate case, but rather correct implementation.

carterkozak:
Could you add a picture/diagram showing the intended flow of requests?

If I'm reading this correctly, we're bypassing the original queue any time we have a sticky session, allowing sticky requests to potentially jump ahead of non-sticky requests. I think that can interact poorly with the limit-bypass functionality to block all progress from non-sticky requests behind VIP-pass sticky requests per session.

jkozlowski:
> Could you add a picture/diagram showing the intended flow of requests?

Hmmm, your description is correct, but here's diagram for reference:

Sticky: StickyQueue (StickyConcurrencyLimitedChannel which backs out as soon as it detects contention in NodeSelectingChannel) -> NodeSelectingChannel -> Host level ConcurrencyLimitedChannel -> Endpoint level queue -> Endpoint level ConcurrencyLimitedChannel -> network.

> If I'm reading this correctly, we're bypassing the original queue any time we have a sticky session, allowing sticky requests to potentially jump ahead of non-sticky requests. I think that can interact poorly with the limit-bypass functionality to block all progress from non-sticky requests behind VIP-pass sticky requests per session.

Yes, I have bypassed it. Otherwise, I couldn't figure out a good way to provide signal to per-sticky queue concurrency limit to back off, and we'd have to either a fixed limit there (or come up with some other mechanism). I have a suspicion that because the normal queue gets poked (and therefore runs scheduling) whenever any request completes/arrives there, it should naturally make progress, whilst the sticky queues will back off.

~~I'll see what it looks like with queues in front of each other and some sort of special exception thing.~~ Discussed offline, and we just want docs on how this is constructed.

dialogue-clients/src/main/java/com/palantir/dialogue/clients/DialogueClients.java
```diff
@@ -101,6 +102,18 @@ public static ReloadingFactory create(Refreshable<ServicesConfigBlock> scb) {
...
+    /** A stateful object - should only need one of these. Live reloads under the hood. */
+    @Beta
+    public interface StickyChannelFactory2 {
+        /**
+         * Returns a channel which will route all requests to a single host, even if that host returns some 429s.
+         * Each successive call to this method may get a different channel (or it may return the same one).
+         */
+        Channel getStickyChannel();
+
+        <T> T getCurrentBest(Class<T> clientInterface);
```
jkozlowski: Thoughts on better name?

carterkozak: Maybe just `sticky`?

jkozlowski: Done

simulation/src/test/resources/report.md
```diff
@@ -644,6 +692,36 @@ client=9 endpoint	client_mean=PT0.000001S
 <tr><td><image width=400 src="https://media.githubusercontent.com/media/palantir/dialogue/develop/simulation/src/test/resources/server_side_rate_limits[UNLIMITED_ROUND_ROBIN].png" /></td><td><image width=400 src="server_side_rate_limits[UNLIMITED_ROUND_ROBIN].png" /></td></tr></table>
 
 
+## `server_side_rate_limits_with_sticky2_clients_fairness_across_multiple_clients[CONCURRENCY_LIMITER_PIN_UNTIL_ERROR]`
```
carterkozak: I think I'd prefer if we updated the existing test rather than adding a new one, that will be a better comparison, and we don't intend to invest in the original implementation moving forward.

dialogue-core/src/test/java/com/palantir/dialogue/core/StickyEndpointChannels2Test.java
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    private SettableFuture<Response> expectStickyRequest(ListenableFuture<Response> response, Request request) {
+        SettableFuture<Response> responseSettableFuture = SettableFuture.create();
+        when(endpointChannel.execute(request)).thenAnswer((Answer<ListenableFuture<Response>>) invocation -> {
+            Request actualRequest = invocation.getArgument(0);
+            assertThat(actualRequest).isEqualTo(request);
+            assertThat(response).isDone();
+            assertThat(Futures.getDone(response).attachments().getOrDefault(StickyAttachments.STICKY_TOKEN, null))
+                    .isEqualTo(request.attachments().getOrDefault(StickyAttachments.STICKY, null));
```
carterkozak: We could add a sticky-integrity-validating-channel to the default chain -- when no sticky attachment is on the request, we forward the request as usual, but if there's a sticky token, we assert the sticky response attachment appears on the response. Same for the initial sticky request -> sticky_token. That way if a configuration we haven't thought to test occurs, we can fail fast rather than producing incorrect behavior.

dialogue-core/src/test/java/com/palantir/dialogue/core/StickyEndpointChannels2Test.java
```diff
@@ -0,0 +1,160 @@
...
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+import org.mockito.stubbing.Answer;
+
+@ExtendWith(MockitoExtension.class)
+public final class StickyEndpointChannels2Test {
+
+    @Mock
+    private Endpoint endpoint;
```
carterkozak: I think we have a TestEndpoint we could use to avoid unnecessary mocking

jkozlowski: Done

simulation/src/test/resources/report.md
jkozlowski: Ack

jkozlowski: Should be done.

dialogue-core/src/test/java/com/palantir/dialogue/core/StickyEndpointChannels2Test.java
jkozlowski: Ack

dialogue-core/src/main/java/com/palantir/dialogue/core/StickyConcurrencyLimitedChannel.java
```diff
@@ -0,0 +1,110 @@
...
+            // concurrency limited: we would then have to rely on another request coming in to try scheduling the
+            // first request again. If that request never came, we would lock up.
+            // To combat that we have a way to instruct downstream channel to ignore its capacity limits,
+            // and let the request through.
+            Optional<ListenableFuture<Response>> result = delegate.maybeExecute(
+                    endpoint,
+                    request,
+                    permit.isOnlyInFlight() || limitEnforcement == LimitEnforcement.DANGEROUS_BYPASS_LIMITS
+                            ? LimitEnforcement.DANGEROUS_BYPASS_LIMITS
+                            : LimitEnforcement.DEFAULT_ENABLED);
```
carterkozak: I think this is easier to understand with a slight refactor:
```suggestion
                    permit.isOnlyInFlight()
                            ? LimitEnforcement.DANGEROUS_BYPASS_LIMITS
                            : limitEnforcement);
```

dialogue-core/src/main/java/com/palantir/dialogue/core/StickyEndpointChannels2.java
```diff
@@ -0,0 +1,236 @@
...
+            this.channelFactory = channelFactory;
+        }
+
+        @Override
+        public EndpointChannel endpoint(Endpoint endpoint) {
+            return new StickyEndpointChannel(router, channelFactory.endpoint(endpoint));
+        }
+
+        /**
+         * .
```
carterkozak: javadoc

simulation/src/test/resources/txt/server_side_rate_limits_with_sticky_clients_steady_vs_bursty_client[CONCURRENCY_LIMITER_ROUND_ROBIN].txt
```diff
@@ -1,3 +1,3 @@
-success=97.0%	client_mean=PT24.931845968S	server_cpu=PT1M37.525S    	client_received=10060/10060	server_resps=19505	codes={200=9755, 429=305}
-client=slowAndSteady endpoint	client_mean=PT14.795458286S
-client=oneShotBurst endpoint	client_mean=PT24.992664294S
+success=96.9%	client_mean=PT24.927819953S	server_cpu=PT1M37.475S    	client_received=10060/10060	server_resps=19495	codes={200=9749, 429=311}
+client=slowAndSteady endpoint	client_mean=PT0.450916666S 
+client=oneShotBurst endpoint	client_mean=PT25.074681373S
```
Comment (carterkozak) [inline simulation/src/test/resources/txt/server_side_rate_limits_with_sticky_clients_steady_vs_bursty_client[CONCURRENCY_LIMITER_ROUND_ROBIN].txt:3]: `slowAndSteady` shows significant improvement while `onsShotBurst` is only slightly slower, this is what we're looking for.

carterkozak (commented):
generally lgtm, needs a description in the readme

dialogue-core/src/test/java/com/palantir/dialogue/core/StickyEndpointChannels2Test.java
jkozlowski: Done

dialogue-core/src/main/java/com/palantir/dialogue/core/StickyConcurrencyLimitedChannel.java
jkozlowski: Much better!

dialogue-core/src/main/java/com/palantir/dialogue/core/StickyEndpointChannels2.java
jkozlowski: Nuked

jkozlowski:
👍

carterkozak (approved):

jkozlowski:
dozer

jkozlowski:
hello

carterkozak:
Oh, this targets the wrong branch. Needs to be updated against `develop`.

jkozlowski:
Duh

PR #1336, opened 2021-07-19
Sticky host routing.

jkozlowski:
## Before this PR

DialogueChannels cannot route to a particular host.

## After this PR

*StickyAttachments* can be used to request that special Response attachment be added that, when copied to subsequent requests, pins them to the same node as initial request.

==COMMIT_MSG==
DialogueChannels can route to a particular host.
==COMMIT_MSG==

## Possible downsides?

As discussed in the prototype PR, pin until error pinned requests to not update the pin on failures. This is absolutely possible to do, but quite a bit of extra work so we decided to forego. Initial requests still switch the pins which should be good enough.

### DIFF
--- a/dialogue-core/src/test/java/com/palantir/dialogue/core/DialogueChannelTest.java
+++ b/dialogue-core/src/test/java/com/palantir/dialogue/core/DialogueChannelTest.java
@@ -67,6 +67,7 @@ import org.junit.jupiter.api.extension.ExtendWith;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.EnumSource;
 import org.mockito.Mock;
+import org.mockito.Mockito;
 import org.mockito.junit.jupiter.MockitoExtension;
 import org.mockito.stubbing.Answer;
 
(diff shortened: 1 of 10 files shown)


dialogue-core/src/test/java/com/palantir/dialogue/core/DialogueChannelTest.java
```diff
@@ -301,6 +306,73 @@ void nice_tostring() {
                 .isNotEqualTo(channel2.toString());
     }
 
+    @Test
+    public void test_can_route_to_specific_host() throws ExecutionException {
```
jkozlowski: Likely unnecessary test, will probably drop.

jkozlowski: Or actually make it run for all strategies, as a catch all in case new ones are added and they don't have the handling.

jkozlowski: Done, simple test that checks the attachments are being added.

dialogue-core/src/test/java/com/palantir/dialogue/core/DialogueChannelTest.java
```diff
@@ -301,6 +305,39 @@ void nice_tostring() {
                 .isNotEqualTo(channel2.toString());
     }
 
+    @ParameterizedTest
+    @EnumSource(NodeSelectionStrategy.class)
+    public void test_can_use_sticky_attachments(NodeSelectionStrategy nodeSelectionStrategy) throws ExecutionException {
```
jkozlowski: Ha if only, forgot the NodeSelectionChannel just sidesteps the strategies for one url.

carterkozak (approved):

jkozlowski:
🍾

jkozlowski:
:+1:

PR #1383, opened 2021-08-24
Dialogue Annotation Processor Encoders may reference supertypes

carterkozak:
For example, the new implementation allows an encoder for type
Object to be used for a String parameter, for instance to implement
a toString encoder.

==COMMIT_MSG==
Dialogue Annotation Processor Encoders may reference supertypes, for example an `Object` encoder may be applied to a `BigInteger` parameter.
==COMMIT_MSG==


jkozlowski (approved):
Haha that is indeed simpler.

PR #1389, opened 2021-08-30
Remove several tracing spans per request/connection

carterkozak:
These are lower-value spans, most of which were added to rule
out components of dialogue as the root cause of past network issues.

These aren't necessarily the noisiest, but they represent the
intersection of the noisiest and easiest to remove.

## After this PR
==COMMIT_MSG==
Remove several tracing spans per request/connection
==COMMIT_MSG==

## Possible downsides?
It may be more difficult to discover which pieces of the system are the cause of slowness in the future, however holistically this change should allow us to better utilize our tracing infrastructure.


smoorpal:
This should be a good start. Is it intentional to keep the lone trace `Dialogue ConnectionManager.connect`?

smoorpal (approved):
Approved from my point of view with the previous comment. I'm OK with the `ConnectionMangaer.connect` to remain as it appears like it could be a place where real latency does build up. As I pointed out there are some more traces in QueuedChannel. Given that these are easier I think its perfectly reasonable to get the savings here first.

carterkozak:
I tried to slim down some of the connection initialization traces as well, but I think it's helpful to retain some of that data to differentiate reuses from new connection creation, thus the `ConnectionMangaer.connect` span. We generally expect <1% of requests to require new connections as most are able to reuse persistent http/1.1 (or h2) connections.

smoorpal:
Makes sense and that is consistent with what I am seeing.

> I tried to slim down some of the connection initialization traces as well, but I think it's helpful to retain some of that data to differentiate reuses from new connection creation, thus the `ConnectionMangaer.connect` span. We generally expect <1% of requests to require new connections as most are able to reuse persistent http/1.1 (or h2) connections.

tpetracca (approved):

PR #1394, opened 2021-09-07
Remove unhelpful tracing spans

carterkozak:
## Before this PR
more spans

## After this PR
==COMMIT_MSG==
Remove unhelpful and redundant tracing spans
==COMMIT_MSG==

This removes the dialogue-request-attempt span which largely duplicates the CLIENT_OUTGOING `Dialogue-http-request` span.

Note that I haven't entirely removed the serialize/deserialize spans, but instead limited them to when logging is configured for DEBUG on the individual classes. This effectively disables them, but allows us to begin emitting the spans at runtime.


dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/TracedEncoding.java
```diff
@@ -94,11 +98,15 @@ static String toString(TypeMarker<?> typeMarker) {
 
         @Override
         public void serialize(T value, OutputStream output) throws IOException {
-            Tracer.fastStartSpan(operation);
-            try {
+            if (log.isDebugEnabled()) {
```
tpetracca, line 101: what's the thinking here and below?

carterkozak: We can quickly turn this on at runtime if we need to, but it’s not useful in most scenarios, nor does it make sense to build our own tracing logger configuration point.

tpetracca (approved):

PR #1396, opened 2021-09-08
InstrumentedDnsResolver produces DEBUG logging for DNS resolution

carterkozak:
==COMMIT_MSG==
InstrumentedDnsResolver produces DEBUG logging for DNS resolution
==COMMIT_MSG==

### DIFF
--- a/dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/InstrumentedDnsResolver.java
+++ b/dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/InstrumentedDnsResolver.java
@@ -16,6 +16,10 @@
 
 package com.palantir.dialogue.hc5;
 
+import com.palantir.logsafe.SafeArg;
+import com.palantir.logsafe.UnsafeArg;
+import com.palantir.logsafe.logger.SafeLogger;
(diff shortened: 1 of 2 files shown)


dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/InstrumentedDnsResolver.java
```diff
@@ -32,15 +37,58 @@
 
     @Override
     public InetAddress[] resolve(String host) throws UnknownHostException {
+        boolean debugLoggingEnabled = log.isDebugEnabled();
+        long startNanos = debugLoggingEnabled ? System.nanoTime() : -1L;
```
schlosna, line 43: Is it worth noting that we're explicitly not using timers here as the overhead is undesirable for the common case?

schlosna (approved):
Thanks, overall LGTM, had one question if we should comment why we're not using timers

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/InstrumentedDnsResolver.java
carterkozak: Good idea

PR #1399, opened 2021-09-09
fix #1308: Reduce Dialogue client creation overhead

### LINKED ISSUE #1308: StickyClientFactory creates a new dialogue client for each request to getCurrentBest for a client interface
Creating a new dialogue client involves a lot of potentially expensive allocations, like creating metric registries etc. We can avoid creating clients for the same client interface passed to the sticky client factory by simply caching the client we return.

Relevant code is here: https://github.com/palantir/dialogue/blob/develop/dialogue-clients/src/main/java/com/palantir/dialogue/clients/ReloadingClientFactory.java#L238

carterkozak:
This bounds work to the number of endpoints that are actually
used. In most cases only a small subset of endpoints are
exercised, this is particularly true in many sticky-session
use-cases in which instances are created much more frequently.

Note that our serializers and deserializers can be relatively
expensive, however the expensive components are already
memoized. Per-endpoint channel binding attempts to do additional
work up-front, however it means each client instance we create
was doing _all_ of this work immediately. Now it occurs lazily
the first time an endpoint is used for a more predictable
performance profile.

==COMMIT_MSG==
Reduce Dialogue client creation overhead by deferring per-endpoint initialization work until the endpoint is first used.
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-1399.v2.yml
+++ b/changelog/@unreleased/pr-1399.v2.yml
@@ -1,0 +1,6 @@
+type: improvement
+improvement:
+  description: Reduce Dialogue client creation overhead by deferring per-endpoint
+    initialization work until the endpoint is first used.
+  links:
+  - https://github.com/palantir/dialogue/pull/1399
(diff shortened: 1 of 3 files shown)


jkozlowski (approved):
Exciting, looks like a step in the right direction and I hope it improves things enough. 

Now that we have "one true channel" ™️ for most things, and routing decisions should be handled through attachments, I don't see a point of instantiating all those tiny channels for each client. I was hoping we could reuse all that, and each client would instead just have a tiny bit of transformation logic to apply after Request creation, to apply any necessary attachments. It's a bit more involved, but would get rid of most allocations, and have it to just 1 or 2.

PR #1400, opened 2021-09-09
Provide a static factory method replacement for JaxrsClient.create

carterkozak:
This method takes fewer args, requiring the HostEventSink and
UserAgent to be included in the ClientConfiguration itself, however
we can handle this transformation in our automation.

==COMMIT_MSG==
Provide a static factory method replacement for JaxrsClient.create
==COMMIT_MSG==

### DIFF
--- a/dialogue-clients/src/main/java/com/palantir/dialogue/clients/ChannelCache.java
+++ b/dialogue-clients/src/main/java/com/palantir/dialogue/clients/ChannelCache.java
@@ -75,7 +75,7 @@ final class ChannelCache {
         LIVE_INSTANCES.add(newCache);
 
         int numLiveInstances = LIVE_INSTANCES.size();
-        if ((numLiveInstances > 1 && log.isInfoEnabled()) || log.isDebugEnabled()) {
+        if ((numLiveInstances > 5 && log.isInfoEnabled()) || log.isDebugEnabled()) {
             if (numLiveInstances >= 10) {
                 log.info(
                         "Created ChannelCache instance #{} ({} alive): {}",
--- a/dialogue-clients/src/main/java/com/palantir/dialogue/clients/DialogueClients.java
+++ b/dialogue-clients/src/main/java/com/palantir/dialogue/clients/DialogueClients.java
@@ -50,6 +50,22 @@ public final class DialogueClients {
                 ImmutableReloadingParams.builder().scb(scb).build(), ChannelCache.createEmptyCache());
     }
 
+    /**
+     * Prefer {@link #create(Refreshable)}. Using a configured {@link ReloadingFactory} should always be preferred
+     * over this method, it only exists to ease migration.
+     *
+     * Construct an instance of the given {@code clientInterface} which can be used to make network calls to the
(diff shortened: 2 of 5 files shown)


dialogue-clients/src/main/java/com/palantir/dialogue/clients/ChannelCache.java
```diff
@@ -75,7 +75,7 @@ static ChannelCache createEmptyCache() {
         LIVE_INSTANCES.add(newCache);
 
         int numLiveInstances = LIVE_INSTANCES.size();
-        if ((numLiveInstances > 1 && log.isInfoEnabled()) || log.isDebugEnabled()) {
+        if ((numLiveInstances > 5 && log.isInfoEnabled()) || log.isDebugEnabled()) {
```
jkozlowski, line 78: What's this?

dialogue-clients/src/main/java/com/palantir/dialogue/clients/DialogueClients.java
```diff
@@ -50,6 +50,22 @@ public static ReloadingFactory create(Refreshable<ServicesConfigBlock> scb) {
...
+     *
+     * Construct an instance of the given {@code clientInterface} which can be used to make network calls to the
+     * single conceptual upstream identified by {@code clientConfiguration}.
+     *
+     * Behaviour is undefined if {@code clientConfiguration} contains no URIs.
+     *
+     * @param clientInterface Dialogue client interface annotated with {@link com.palantir.dialogue.DialogueService}
+     * @param clientConfiguration Configuration which <b>must</b> contain a user-agent
+     */
+    public static <T> T create(Class<T> clientInterface, ClientConfiguration clientConfiguration) {
```
jkozlowski, line 65: createLegacy? some other way to indicate people shouldn't really be doing this? I guess you have docs

dialogue-clients/src/main/java/com/palantir/dialogue/clients/ChannelCache.java
carterkozak: LegacyClients has its own client factory, backed by a ChannelCache -- I loosened this constraint because it's common for apps to create a couple instances already (e.g. atlas->lock + wc factory) and the additional instance would incur this logging.
It's probably safe to push up toward 10-100 since this is primarily to detect when a new instance is created on a per-operation basis, but I didn't want to change it too much in this change.

dialogue-clients/src/main/java/com/palantir/dialogue/clients/DialogueClients.java
carterkozak: Ya, I went back and forth on naming a few times here, but landed on `create` matching `jaxrsclient` becuase it's a direct replacement, and we don't provide alternatives for some use-cases (when trust material exists only in memory, for example).

carterkozak: To clarify, I'm not confident I've picked the right name, just recording my thought process :-)

jkozlowski (approved):

PR #1401, opened 2021-09-09
Dialogue path parameters support multi-segment values

carterkozak:
This is useful reading files based on paths from a remote
non-conjure server. Previously we opted not to support this sort
of thing because Dialogue was conjure-first, however adoption
of the annotation processor for non-conjure services has exposed
the need.

## After this PR
==COMMIT_MSG==
Dialogue path parameters support multi-segment values
==COMMIT_MSG==

## Possible downsides?
Potential infinite recursion if `Endpoint` is manually implemented without overriding one render method.

### DIFF
--- a/dialogue-annotations-example/src/test/java/com/palantir/myservice/example/MyServiceIntegrationTest.java
+++ b/dialogue-annotations-example/src/test/java/com/palantir/myservice/example/MyServiceIntegrationTest.java
@@ -19,6 +19,7 @@ package com.palantir.myservice.example;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableMultimap;
 import com.google.common.collect.Iterables;
@@ -314,6 +315,17 @@ public final class MyServiceIntegrationTest {
                         .putAll("q1", "var1", "var2")
                         .build(),
                 new MyCustomParamType("var3"));
+    }
+
+    @Test
+    void testMultiplePathParams() {
+        UUID first = UUID.randomUUID();
+        UUID second = UUID.randomUUID();
+        undertowHandler = exchange -> {
+            exchange.assertMethod(HttpMethod.GET);
+            exchange.assertPath("/multipath/" + first + '/' + second);
+        };
+        myServiceDialogue.multiplePathSegments(ImmutableList.of(first, second));
     }
 
     private void testCustomResponse(int code) {
--- a/dialogue-annotations-example/src/test/java/com/palantir/myservice/example/MyServiceIntegrationTest.java
+++ b/dialogue-annotations-example/src/test/java/com/palantir/myservice/example/MyServiceIntegrationTest.java
@@ -328,6 +328,29 @@ public final class MyServiceIntegrationTest {
         myServiceDialogue.multiplePathSegments(ImmutableList.of(first, second));
     }
 
+    @Test
+    void testMultiplePathParams_empty() {
+        undertowHandler = exchange -> {
+            exchange.assertMethod(HttpMethod.GET);
+            exchange.assertPath("/multipath");
+        };
+        myServiceDialogue.multiplePathSegments(ImmutableList.of());
+    }
+
+    @Test
+    void testMultiplePathParams_escaped() {
+        String first = "a/b";
+        String second = "c/d";
+        undertowHandler = exchange -> {
+            exchange.assertMethod(HttpMethod.GET);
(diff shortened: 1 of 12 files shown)


dialogue-annotations-example/src/test/java/com/palantir/myservice/example/MyServiceIntegrationTest.java
```diff
@@ -316,6 +317,17 @@ void testMultiparams() {
...
 
+    @Test
+    void testMultiplePathParams() {
+        UUID first = UUID.randomUUID();
+        UUID second = UUID.randomUUID();
+        undertowHandler = exchange -> {
+            exchange.assertMethod(HttpMethod.GET);
+            exchange.assertPath("/multipath/" + first + '/' + second);
+        };
+        myServiceDialogue.multiplePathSegments(ImmutableList.of(first, second));
```
raiju, line 328: Do we want a test that includes a `/` char to guard against future refactors messing up assumptions?

raiju: (As in, to ensure it's properly escaped) Or do we assume this is already covered by existing tests?

carterkozak: No harm in adding one, however that's a separate concern from the code impacted by this change. This works nicely because we take in a collection (boundaries defined in code) and simply apply them as segments to the uri using the same method we've always used. The function which applies individual segments is responsible for encoding characters in the value.

dialogue-target/src/main/java/com/palantir/dialogue/Endpoint.java
```diff
@@ -25,7 +28,23 @@
  * as well as the types of the request and response body.
  */
 public interface Endpoint {
-    void renderPath(Map<String, String> params, UrlBuilder url);
+    /**
+     * {@link #renderPath(ListMultimap, UrlBuilder)} is preferred, however at least one {@code renderPath} must be
+     * implemented to avoid infinite recursion.
+     */
+    default void renderPath(Map<String, String> params, UrlBuilder url) {
```
jkozlowski, line 35: Just to reason through this: existing conjure clients will only have the ```renderPath(Map)``` implementation. So when BaseUrl#render comes around, it will call ```renderPath(ListMultimap)``` which will add a tiny bit of overhead with the MultimapAsMap wrapper, and call into ```renderPath(Map)``` implementation. I'm assuming you'll FLUP with a conjure change to implement the right method instead?

dialogue-target/src/main/java/com/palantir/dialogue/Endpoint.java
```diff
@@ -25,7 +28,23 @@
...
+        renderPath(
+                params.isEmpty()
+                        ? ImmutableListMultimap.of()
+                        : ImmutableListMultimap.<String, String>builder()
+                                .putAll(params.entrySet())
+                                .build(),
+                url);
+    }
+
+    default void renderPath(ListMultimap<String, String> params, UrlBuilder url) {
```
jkozlowski, line 45: Tell me if I'm wrong: but maybe instead of implementing this method, we should instead have:

```
PathTemplate pathTemplate();
```

If we had it this way, it would not have been a break at all?

jkozlowski: As an alternative, we could have Endpoint2?

jkozlowski (approved):
lgtm, just a couple of thoughts

dialogue-target/src/main/java/com/palantir/dialogue/Endpoint.java
carterkozak: Exactly!

dialogue-target/src/main/java/com/palantir/dialogue/Endpoint.java
carterkozak: I think one would require the other, I’m not sure we could generate the template for backcompat on existing endpoints.

I don’t recall why we provide a method rather than an accessor for the template though, exposing the template would be cleaner. Rolling out Endpoint2 would cause a lot of churn given how widely used Endpoint is.

jkozlowski: It's generate code though, does it matter?

carterkozak: Endpoint2 would require overloads everywhere in this project for backcompat, right?

I just rolled out of bed and might not be groking something.

jkozlowski: Ah yes, I see. I thought it would only be relevant to the url generation and maybe some metrics, but it's used all over. I guess we could major rev, but probably not worth the pain.

PR #1407, opened 2021-09-13
Add custom list encoder support for path parameters

raiju:
## Before this PR
Multi-segment paths had to be supplied as lists.

## After this PR
==COMMIT_MSG==
Add custom list encoder support for path parameters
==COMMIT_MSG==

By supplying the `listEncoder` property to a header param, we can control the behavior of the serialization into path segments, e.g. automatically encoding maven coordinates into the matching path.


dialogue-annotations/src/main/java/com/palantir/dialogue/annotations/Request.java
```diff
@@ -86,6 +86,8 @@
     @Target(ElementType.PARAMETER)
     @interface PathParam {
         Class<? extends ParamEncoder<?>> encoder() default DefaultParamEncoder.class;
+
+        Class<? extends ListParamEncoder<?>> listEncoder() default UnsetListParamEncoder.class;
```
carterkozak: nit: no reason to create a new type, we could use the `DefaultListParamEncoder` marker interrface -- annotation processors can differentiate between default values and values that are explicitly set.
```suggestion
        Class<? extends ListParamEncoder<?>> listEncoder() default DefaultListParamEncoder.class;
```

dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/data/ParamTypesResolver.java
```diff
@@ -158,23 +162,53 @@ private ParameterEncoderType multimapDefaultEncoder(EndpointName endpointName, V
...
+            EncoderTypeAndMethod listEncoderTypeAndMethod) {
+        Optional<TypeName> encoderTypeName =
+                annotationReflector.getFieldMaybe("encoder", TypeMirror.class).map(TypeName::get);
+
+        Optional<TypeName> listEncoderTypeName = annotationReflector
+                .getFieldMaybe("listEncoder", TypeMirror.class)
+                .map(TypeName::get);
+
+        if (encoderTypeName.isPresent() && listEncoderTypeName.isPresent()) {
+            throw new SafeRuntimeException("Only one of encoder and listEncoder can be set for " + endpointName.get());
```
carterkozak: Perhaps `context.reportError` instead of throwing, we can return empty optional since we're failing the build.

carterkozak (approved):
lgtm, thanks!

PR #1408, opened 2021-09-13
Fail when multiple interfaces use same factory in a single compilation unit.

raiju:
## Before this PR

It's easy to copy-pasta one's way into a mismatched factory annotation.

## After this PR
==COMMIT_MSG==
Fail when multiple interfaces use same factory in a single compilation unit. This is only a sanity check and won't always detect badness (e.g. when using incremental compilation).
==COMMIT_MSG==

Setting to "do not merge" as I'm not convinced this is a good idea (given the unreliability), but wanted to discuss nonetheless.

## Possible downsides?

See the changelog; this check is fundamentally limited in when it can detect these problems. Up for discussion whether it is useful to even have this at all.


raiju:
Ah, there's more fun around the resolution of classes that don't exist yet. Closing...

PR #1428, opened 2021-10-04
Implement configuration-based SOCKS proxy support

carterkozak:
Depends on:
https://github.com/palantir/conjure-java-runtime-api/pull/699
https://github.com/palantir/conjure-java-runtime/pull/2126

## After this PR
==COMMIT_MSG==
Implement configuration-based SOCKS proxy support
==COMMIT_MSG==

## Possible downsides?
More proxies, socks is particularly difficult to test


dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientChannels.java
```diff
@@ -592,6 +593,34 @@ private static InetSocketAddress getSocksProxyAddress() {
...
+                    .map(InetSocketAddress.class::cast)
+                    .findFirst()
+                    .orElseGet(ApacheHttpClientChannels::getSocksProxyAddress);
+            if (address != null) {
+                log.debug("Found SOCKS proxy address", UnsafeArg.of("address", address));
+            } else {
+                log.debug("No SOCKS proxy found");
+            }
+            return address;
+        } catch (RuntimeException e) {
```
CRogers, line 618: This seems a little scary (as in might hide errors). Is this just to avoid breaking existing stuff in prod in case there is a problem with the socks detection?

CRogers:
:+1:

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientChannels.java
carterkozak: Ya, this should log at ERROR rather than DEBUG :+1:

carterkozak: You are correct, I added the catch to avoid adding risk to the production case.

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientChannels.java
```diff
@@ -592,6 +593,34 @@ private static InetSocketAddress getSocksProxyAddress() {
...
+            log.debug("Failed to find a SOCKS proxy", e);
```
carterkozak:
```suggestion
            log.error("Failed to find a SOCKS proxy", e);
```

PR #1431, opened 2021-10-06
Dialogue schedulers allow threads to time out when no work is scheduled

carterkozak:
This should allow dialogue clients to be used in dynamically
loaded/unloaded plugins without leaking scheduler threads.

## After this PR
==COMMIT_MSG==
Dialogue schedulers allow threads to time out when no work is scheduled
==COMMIT_MSG==

### DIFF
--- a/dialogue-core/src/test/java/com/palantir/dialogue/core/DialogueExecutorsTest.java
+++ b/dialogue-core/src/test/java/com/palantir/dialogue/core/DialogueExecutorsTest.java
@@ -1,0 +1,69 @@
+/*
+ * (c) Copyright 2021 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.dialogue.core;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import com.google.common.util.concurrent.Runnables;
+import com.google.common.util.concurrent.ThreadFactoryBuilder;
+import java.time.Duration;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+import org.awaitility.Awaitility;
+import org.junit.jupiter.api.Test;
+
+final class DialogueExecutorsTest {
+    private static final String THREAD_PREFIX = "dialogue-scheduler-";
+
+    @Test
+    void testThreadTimeout() throws Exception {
+        ScheduledExecutorService exec = DialogueExecutors.newSharedSingleThreadScheduler(
+                new ThreadFactoryBuilder()
+                        .setNameFormat(THREAD_PREFIX + "%d")
+                        .setDaemon(true)
+                        .build(),
+                Duration.ofMillis(100));
+        try {
+            assertThat(countExecutorThreads()).isEqualTo(0);
+            exec.schedule(Runnables.doNothing(), 1, TimeUnit.SECONDS).cancel(true);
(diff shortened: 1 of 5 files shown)


henryptung:
FYSA, from ScheduledThreadPoolExecutor javadoc:

```
 * <p>While this class inherits from {@link ThreadPoolExecutor}, a few
 * of the inherited tuning methods are not useful for it. In
 * particular, because it acts as a fixed-sized pool using
 * {@code corePoolSize} threads and an unbounded queue, adjustments
 * to {@code maximumPoolSize} have no useful effect. Additionally, it
 * is almost never a good idea to set {@code corePoolSize} to zero or
 * use {@code allowCoreThreadTimeOut} because this may leave the pool
 * without threads to handle tasks once they become eligible to run.
```

But I'm also very not-convinced that this Javadoc actually applies anymore, since https://github.com/openjdk/jdk/commit/2d19ea519b17529a083a62eb219da532693bbef3 was merged but did not update the Javadoc.

Worth some proper testing just to make sure the pool works as expected (e.g. scheduling tasks further in the future than the specified pool timeout).

henryptung (approved):

dialogue-core/src/test/java/com/palantir/dialogue/core/DialogueExecutorsTest.java
```diff
@@ -0,0 +1,73 @@
...
+    void testThreadTimeout() throws Exception {
+        ScheduledExecutorService exec = DialogueExecutors.newSharedSingleThreadScheduler(
+                new ThreadFactoryBuilder()
+                        .setNameFormat(THREAD_PREFIX + "%d")
+                        .setDaemon(true)
+                        .build(),
+                Duration.ofMillis(100));
+        try {
+            assertThat(countExecutorThreads()).isEqualTo(0);
+            exec.schedule(Runnables.doNothing(), 1, TimeUnit.SECONDS).cancel(true);
```
jkozlowski, line 44: Isn't this a bit racey?

dialogue-core/src/test/java/com/palantir/dialogue/core/DialogueExecutorsTest.java
```diff
@@ -0,0 +1,73 @@
...
+            assertThat(countExecutorThreads()).isEqualTo(1);
+
+            Awaitility.waitAtMost(Duration.ofSeconds(1)).untilAsserted(() -> assertThat(countExecutorThreads())
+                    .as("Created threads should time out")
+                    .isEqualTo(0));
+
+            // Test scheduling beyond the timeout
+            AtomicInteger counter = new AtomicInteger();
+            ScheduledFuture<Integer> scheduledFuture = exec.schedule(counter::incrementAndGet, 1, TimeUnit.SECONDS);
+            assertThat(countExecutorThreads()).isEqualTo(1);
```
jkozlowski, line 54: Tis too?

dialogue-core/src/test/java/com/palantir/dialogue/core/DialogueExecutorsTest.java
```diff
@@ -0,0 +1,73 @@
...
+                    .isEqualTo(0));
+        } finally {
+            exec.shutdownNow();
+            assertThat(exec.awaitTermination(1, TimeUnit.SECONDS))
+                    .as("Executor failed to stop")
+                    .isTrue();
+        }
+    }
+
+    private long countExecutorThreads() {
```
jkozlowski, line 68: Do you think we could add the "plugin" test? How difficult would that be to setup? I'm just worried we'll add some more shared threadpools and this regresses for people again?

jkozlowski (approved):
Some comments, but change looks good

dialogue-core/src/test/java/com/palantir/dialogue/core/DialogueExecutorsTest.java
carterkozak: After this, the timeout duration must elapse before the created thread disappears again. Threads are created when tasks are scheduled, and stay alive until the queue is empty and execution has completed

dialogue-core/src/test/java/com/palantir/dialogue/core/DialogueExecutorsTest.java
carterkozak: Same as above, the executor thread is created by the submitting thread (this one) in order to handle the `counter::incrementAndGet` task, and will exist until keepalive-time after the future has completed

dialogue-core/src/test/java/com/palantir/dialogue/core/DialogueExecutorsTest.java
carterkozak: It's possible, but painful to set up and maintain :-/. Easy to make classpath changes which result in the test becoming invalid over time.

PR #1524, opened 2022-01-07
Query parameters retain input order rather than hash-based order

carterkozak:
In retrospect, the map wasn't a great design choice. A list of pairs
would be simpler, and align more closely with what we're trying
to accomplish. No matter, a map implementation which retains
insertion order is sufficient.

## Before this PR
seemingly random query parameter order, which can cause some systems to fail signed request validation.

## After this PR
==COMMIT_MSG==
Query parameters retain input order rather than hash-based order
==COMMIT_MSG==

## Possible downsides?
These maps may be slightly more expensive to create?


xRuiAlves (approved):
Looks good! 👍

PR #1534, opened 2022-01-13
Fix Sticky2 early queued response handling

carterkozak:
Previously the initial response future was mishandled in a way
that caused cancellation to 'leak' between queued futures, this
meant that pending responses could be dropped on the floor rather
than being deserialized as expected.

Secondly, this fixes an issue in which certain sticky2 requests
could be retried in error if they fail with an exception after the
initial request completed with a valid response.

## After this PR
==COMMIT_MSG==
Fix Sticky2 early queued response handling
==COMMIT_MSG==

## Possible downsides?
:ghost: Futures are scary :ghost:

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/StickyEndpointChannels2.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/StickyEndpointChannels2.java
@@ -17,8 +17,10 @@
 package com.palantir.dialogue.core;
 
 import com.google.common.util.concurrent.AsyncFunction;
+import com.google.common.util.concurrent.FutureCallback;
 import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
+import com.google.common.util.concurrent.SettableFuture;
 import com.palantir.dialogue.Channel;
 import com.palantir.dialogue.Endpoint;
 import com.palantir.dialogue.EndpointChannel;
@@ -27,7 +29,6 @@ import com.palantir.dialogue.Request;
 import com.palantir.dialogue.Response;
 import com.palantir.dialogue.futures.DialogueFutures;
 import com.palantir.logsafe.exceptions.SafeIllegalStateException;
-import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.function.Supplier;
@@ -134,6 +135,10 @@ final class StickyEndpointChannels2 implements Supplier<Channel> {
 
         private final InFlightCallSuccessTransformer successTransformer = new InFlightCallSuccessTransformer();
         private final InFlightCallFailureTransformer failureTransformer = new InFlightCallFailureTransformer();
+        private final AsyncFunction<Throwable, Response> cancellationTransformer = _input -> {
+            failed();
+            return Futures.immediateCancelledFuture();
+        };
 
         @Nullable
         private volatile Consumer<Request> stickyTarget;
@@ -160,30 +165,35 @@ final class StickyEndpointChannels2 implements Supplier<Channel> {
                     // the first call.
                     result = DialogueFutures.transform(
                             executeWithStickyToken(request, endpointChannel), successTransformer);
-                    result = DialogueFutures.catchingAllAsync(result, failureTransformer);
+                    result = DialogueFutures.catchingAllAsync(result, failureTransformer, cancellationTransformer);
                     callInFlight = result;
                 } else {
                     // Each subsequent (parallel) call may be independently cancelled, that cancellation
                     // must not leak to other pending calls.
-                    result = Futures.nonCancellationPropagating(callInFlightSnapshot);
-                    // State must be tracked to prevent duplicate requests when the original in-flight request
-                    // completes, triggering waiting requests, which may fail via an exception and get
-                    // unintentionally retried by 'catchingAllAsync'.
-                    AtomicBoolean handled = new AtomicBoolean();
-                    result = DialogueFutures.transformAsync(result, response -> {
-                        if (!handled.getAndSet(true)) {
-                            return execute(request, endpointChannel);
-                        } else {
-                            return Futures.immediateFuture(response);
-                        }
+                    SettableFuture<Response> response = SettableFuture.create();
+                    DialogueFutures.addDirectListener(callInFlightSnapshot, () -> {
+                        ListenableFuture<Response> queuedRequestResponse = execute(request, endpointChannel);
+                        DialogueFutures.addDirectCallback(queuedRequestResponse, new FutureCallback<>() {
+                            @Override
+                            public void onSuccess(Response result) {
+                                if (!response.set(result)) {
+                                    result.close();
+                                }
+                            }
+
+                            @Override
+                            public void onFailure(Throwable throwable) {
+                                response.setException(throwable);
+                            }
+                        });
+                        // If the returned future is cancelled, this request should be as well.
+                        DialogueFutures.addDirectListener(response, () -> {
+                            if (queuedRequestResponse.isCancelled()) {
+                                queuedRequestResponse.cancel(false);
+                            }
+                        });
                     });
-                    result = DialogueFutures.catchingAllAsync(result, throwable -> {
-                        if (!handled.getAndSet(true)) {
-                            return execute(request, endpointChannel);
-                        } else {
-                            return Futures.immediateFailedFuture(throwable);
-                        }
-                    });
+                    return response;
                 }
                 return result;
             }
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/StickyEndpointChannels2.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/StickyEndpointChannels2.java
@@ -16,9 +16,7 @@
 
 package com.palantir.dialogue.core;
 
-import com.google.common.util.concurrent.AsyncFunction;
 import com.google.common.util.concurrent.FutureCallback;
-import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
 import com.google.common.util.concurrent.SettableFuture;
 import com.palantir.dialogue.Channel;
@@ -28,9 +26,7 @@ import com.palantir.dialogue.EndpointChannelFactory;
 import com.palantir.dialogue.Request;
 import com.palantir.dialogue.Response;
 import com.palantir.dialogue.futures.DialogueFutures;
-import com.palantir.logsafe.exceptions.SafeIllegalStateException;
 import java.util.function.Consumer;
-import java.util.function.Function;
 import java.util.function.Supplier;
 import javax.annotation.Nullable;
 import javax.annotation.concurrent.GuardedBy;
@@ -133,11 +129,16 @@ final class StickyEndpointChannels2 implements Supplier<Channel> {
     @ThreadSafe
     private static final class StickyRouter {
 
-        private final InFlightCallSuccessTransformer successTransformer = new InFlightCallSuccessTransformer();
-        private final InFlightCallFailureTransformer failureTransformer = new InFlightCallFailureTransformer();
-        private final AsyncFunction<Throwable, Response> cancellationTransformer = _input -> {
-            failed();
-            return Futures.immediateCancelledFuture();
+        private final FutureCallback<Response> initialRequestCallback = new FutureCallback<>() {
+            @Override
+            public void onSuccess(Response response) {
+                successfulCall(response);
+            }
+
+            @Override
+            public void onFailure(Throwable _throwable) {
+                failed();
+            }
         };
 
         @Nullable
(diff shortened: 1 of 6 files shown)


dialogue-core/src/main/java/com/palantir/dialogue/core/StickyEndpointChannels2.java
```diff
@@ -159,11 +161,29 @@ public String toString() {
                     result = DialogueFutures.transform(
                             executeWithStickyToken(request, endpointChannel), successTransformer);
                     result = DialogueFutures.catchingAllAsync(result, failureTransformer);
-                    callInFlight = Futures.nonCancellationPropagating(result);
+                    callInFlight = result;
```
jkozlowski, line 166: This was totally a bug, I had the wrong cancellation semantics in mind.

dialogue-core/src/main/java/com/palantir/dialogue/core/StickyEndpointChannels2.java
```diff
@@ -159,11 +161,29 @@ public String toString() {
...
-                    result = DialogueFutures.transformAsync(result, _input -> execute(request, endpointChannel));
-                    result = DialogueFutures.catchingAllAsync(result, _throwable -> execute(request, endpointChannel));
+                    // Each subsequent (parallel) call may be independently cancelled, that cancellation
+                    // must not leak to other pending calls.
+                    result = Futures.nonCancellationPropagating(callInFlightSnapshot);
+                    // State must be tracked to prevent duplicate requests when the original in-flight request
+                    // completes, triggering waiting requests, which may fail via an exception and get
+                    // unintentionally retried by 'catchingAllAsync'.
+                    AtomicBoolean handled = new AtomicBoolean();
+                    result = DialogueFutures.transformAsync(result, response -> {
```
jkozlowski: What would be wrong with this impl:

```
                    result = Futures.nonCancellationPropagating(callInFlightSnapshot);
                    SettableFuture<Response> response = SettableFuture.create();
                    DialogueFutures.addDirectCallback(result, new FutureCallback<>() {
                        @Override
                        public void onSuccess(Response result) {
                            response.setFuture(execute(request, endpointChannel));
                        }

                        @Override
                        public void onFailure(Throwable _ignored) {
                            response.setFuture(execute(request, endpointChannel));
                        }
                    });
```

carterkozak: `setFuture` can leak responses silently, so we have to wire it up ourselves. I've pushed something similar which I think handles that for us, with a great many additional test cases.

I caught another bug in which `catchingAllAsync` didn't catch cancellation (by design) which left this object in a bad state...

dialogue-core/src/main/java/com/palantir/dialogue/core/StickyEndpointChannels2.java
```diff
@@ -158,12 +165,35 @@ public String toString() {
...
+                    callInFlight = result;
                 } else {
-                    result = callInFlightSnapshot;
-                    result = DialogueFutures.transformAsync(result, _input -> execute(request, endpointChannel));
-                    result = DialogueFutures.catchingAllAsync(result, _throwable -> execute(request, endpointChannel));
+                    // Each subsequent (parallel) call may be independently cancelled, that cancellation
+                    // must not leak to other pending calls.
+                    SettableFuture<Response> response = SettableFuture.create();
+                    DialogueFutures.addDirectListener(callInFlightSnapshot, () -> {
+                        ListenableFuture<Response> queuedRequestResponse = execute(request, endpointChannel);
```
carterkozak: optimization: This block could be wrapped in a `if (!response.isDone()) {`

dialogue-core/src/main/java/com/palantir/dialogue/core/StickyEndpointChannels2.java
jkozlowski: > setFuture can leak responses silently, so we have to wire it up ourselves. I've pushed something similar which I think handles that for us, with a great many additional test cases.

wait what? Is that a bug or by design?

dialogue-futures/src/main/java/com/palantir/dialogue/futures/DialogueFutures.java
```diff
@@ -51,7 +51,14 @@
 
     public static <T> ListenableFuture<T> catchingAllAsync(
             ListenableFuture<T> input, AsyncFunction<Throwable, T> function) {
-        return new DialogueDirectAsyncCatchingFuture<T>(input, function);
+        return new DialogueDirectAsyncCatchingFuture<>(input, function);
+    }
+
+    public static <T> ListenableFuture<T> catchingAllAsync(
+            ListenableFuture<T> input,
+            AsyncFunction<Throwable, T> function,
```
jkozlowski: nit: "onException" here and above.

dialogue-core/src/main/java/com/palantir/dialogue/core/StickyEndpointChannels2.java
```diff
@@ -158,12 +165,35 @@ public String toString() {
...
+                    result = DialogueFutures.catchingAllAsync(result, failureTransformer, cancellationTransformer);
+                    callInFlight = result;
                 } else {
-                    result = callInFlightSnapshot;
-                    result = DialogueFutures.transformAsync(result, _input -> execute(request, endpointChannel));
-                    result = DialogueFutures.catchingAllAsync(result, _throwable -> execute(request, endpointChannel));
+                    // Each subsequent (parallel) call may be independently cancelled, that cancellation
+                    // must not leak to other pending calls.
+                    SettableFuture<Response> response = SettableFuture.create();
+                    DialogueFutures.addDirectListener(callInFlightSnapshot, () -> {
```
jkozlowski, line 173: So, RetryingChannel has a simlar sort of snippet (and this is actually quite prevalent across the codebase, which is where I originally got the transformAsync+catchingAll in my original version of this code.

```
            result = DialogueFutures.transformAsync(result, this::handleHttpResponse);
            result = DialogueFutures.catchingAllAsync(result, this::handleThrowable);
```

Shall we add a transformOrCatchingAll thing to DialogueFutures (so basically addDirectListener, but one that returns futures) and migrate everything to this?

dialogue-core/src/main/java/com/palantir/dialogue/core/StickyEndpointChannels2.java
```diff
@@ -132,6 +135,10 @@ public String toString() {
 
         private final InFlightCallSuccessTransformer successTransformer = new InFlightCallSuccessTransformer();
         private final InFlightCallFailureTransformer failureTransformer = new InFlightCallFailureTransformer();
+        private final AsyncFunction<Throwable, Response> cancellationTransformer = _input -> {
```
jkozlowski: This is the same impl as failureTransformer, can we use the same thing? When would you want to handle cancellation differently than error?

dialogue-core/src/main/java/com/palantir/dialogue/core/StickyEndpointChannels2.java
```diff
@@ -158,12 +165,35 @@ public String toString() {
...
+                    // Each subsequent (parallel) call may be independently cancelled, that cancellation
+                    // must not leak to other pending calls.
+                    SettableFuture<Response> response = SettableFuture.create();
+                    DialogueFutures.addDirectListener(callInFlightSnapshot, () -> {
+                        ListenableFuture<Response> queuedRequestResponse = execute(request, endpointChannel);
+                        DialogueFutures.addDirectCallback(queuedRequestResponse, new FutureCallback<>() {
+                            @Override
+                            public void onSuccess(Response result) {
+                                if (!response.set(result)) {
+                                    result.close();
```
jkozlowski: Whoa this is sooo subtle.

dialogue-core/src/main/java/com/palantir/dialogue/core/StickyEndpointChannels2.java
jkozlowski, line 173: This whole block is kinda scary, one wrong move and you're broken. Maybe trying to implement this queue as futures is not the way to go? I feel like at this point this needs to be a more specific implementation, like the QueuedChannel; you have a specific QueuedCall etc., with the SettableFuture as a field etc.

Whilst thinking about this I also realized that with this implementation, each of the requests are basically building a mini queue: if you're queued, you keep building a queue of futures, if the inFlightCall fails, you'll get notified, and attach yourself to the new inFlightCall; but the original future obviously needs to stay around, because that's what was returned to the caller. So you end up building a queue of linked futures, for each request and each failure.

dialogue-core/src/main/java/com/palantir/dialogue/core/StickyEndpointChannels2.java
jkozlowski: Also, seems like a repeated snippet, e.g. in QueuedChannel#ForwardAndSchedule.

dialogue-core/src/main/java/com/palantir/dialogue/core/StickyEndpointChannels2.java
jkozlowski: I think this queue of queues thing is maybe a problem, but I tried to pass down the SettableFuture at each step and it just gets messy, so probably not a worthwhile optimisation.

What I feel like we're missing is futures that have an explicit cleanup/resource attached to them, and then semantics on how the ownership of the resource is passed around. Kinda like Stream#close exists and also #onClose: feels like any sufficiently advanced flow data-structure should have a way of attaching resources that need closing/cleaning up.

dialogue-core/src/main/java/com/palantir/dialogue/core/StickyEndpointChannels2.java
```diff
@@ -172,7 +202,12 @@ public String toString() {
         private final class InFlightCallSuccessTransformer implements Function<Response, Response> {
             @Override
             public Response apply(Response response) {
-                successfulCall(response);
+                try {
+                    successfulCall(response);
+                } catch (Throwable t) {
```
jkozlowski: Is this overly paranoid?

dialogue-core/src/main/java/com/palantir/dialogue/core/StickyEndpointChannels2.java
```diff
@@ -158,12 +165,35 @@ public String toString() {
...
+                                }
+                            }
+
+                            @Override
+                            public void onFailure(Throwable throwable) {
+                                response.setException(throwable);
+                            }
+                        });
+                        // If the returned future is cancelled, this request should be as well.
+                        DialogueFutures.addDirectListener(response, () -> {
```
jkozlowski: So this replaces the Futures.nonCancellationPropagating

dialogue-core/src/main/java/com/palantir/dialogue/core/StickyEndpointChannels2.java
```diff
@@ -158,12 +165,35 @@ public String toString() {
                     // the first call.
```
jkozlowski: With this comment was I was getting at is: ListenableFutures does not guarantee ordering of listeners. I want to make sure that after this call is complete, I first update my internal state (stickTarget etc.) BEFORE anything that's queued up on this future can inspect this internal state.

I thought that #addDirectCallback would basically not guarantee this at all, hence why I went for transform + catchingAllAsync. Maybe there's a better way to achieve it or I am misunderstanding something?

jkozlowski (commented):
I trust that you're probably correct here, but I'm also scared of all this. So there's a bunch of questions of whether we should introduce some new primitives/abstractions.

dialogue-core/src/main/java/com/palantir/dialogue/core/StickyEndpointChannels2.java
jkozlowski: Ok, I've read the impl a bit, and I think this should totally be a DialogueFutures.addDirectCallback, instead of this malarky: the future internally builds a linked list, in-order of the calls (so the order is deterministic if done from single thread, which we do in here.

jkozlowski:
```
        private final FutureCallback<Response> callInFlightResponseCallback = new FutureCallback<>() {
            @Override
            public void onSuccess(Response result) {
                try {
                    successfulCall(result);
                } catch (Throwable t) {
                    result.close();
                    throw new SafeIllegalStateException("Failed to update state with successful call", t);
                }
            }

            @Override
            public void onFailure(Throwable _ignored) {
                failed();
            }
        };

// and then
result = executeWithStickyToken(request, endpointChannel);
                    callInFlight = result;
                    DialogueFutures.addDirectCallback(result, callInFlightResponseCallback);
```

the #onFailure should be executed on any exception, including cancellation.

dialogue-core/src/main/java/com/palantir/dialogue/core/StickyEndpointChannels2.java
carterkozak: By design, guava future.cancel always moves a future to the cancelled state, however it can be invoked while a transformation is in progress:

consumerFuture.cancel -> hmm <- producerFuture.complete

DialogueFutures methods return false from cancel if there's a transformation in progress, so we can add a listener to consume the resulting future.

dialogue-futures/src/main/java/com/palantir/dialogue/futures/DialogueFutures.java
carterkozak: 👍

dialogue-core/src/main/java/com/palantir/dialogue/core/StickyEndpointChannels2.java
carterkozak: In the RetryingChannel case, we rely on cancel not being handled/retried -- we want it to cancel the request asap. In the RetryingChannel case, the retryer tracks the retry count, so there's no risk of additional retries beyond what we expect.

dialogue-core/src/main/java/com/palantir/dialogue/core/StickyEndpointChannels2.java
carterkozak: It's important that we return `Futures.immediateCancelledFuture()` instead of `immediateFailedFuture(throwable)` in this case, otherwise the final future will be failed(CancellationException) (`get()` throws `ExecutionException` caused by `CancellationException` instead of a `CancellationException`).

jkozlowski: Ah missed that

dialogue-core/src/main/java/com/palantir/dialogue/core/StickyEndpointChannels2.java
carterkozak: There's no great way for us to know, unfortunately

dialogue-core/src/main/java/com/palantir/dialogue/core/StickyEndpointChannels2.java
carterkozak: Good call, it's much simpler that way. We don't need to worry about catch/close in that case either since it's not a transformation.

jkozlowski (approved):
As per previous comment, we should merge to fix the leaks, but this way of handling this is not maintainable: if each transformation chain needs to worry about resource leak handling (in a way that's non-obvious, as opposed to e.g. try-with-resources), any code we write is likely to start leaky, and then get fixed up.

dialogue-core/src/main/java/com/palantir/dialogue/core/StickyEndpointChannels2.java
jkozlowski: Of course I was wrong here: see internal issue PDS-241899; The future does internally build a linked list, except that when a future completes is sets the value immediately and then executes the listeners. And that's all great, but #addListener skips the queue and executes the listener immediately if this happens :(

So I'll introduce an intermediate SettableFuture here, to make sure that the callInFlightCallback executes first and updates the internal state BEFORE executing all the futures waiting on it!

PR #1542, opened 2022-01-18
Fix cancellation race resulting in a leaked response on retry 

carterkozak:
==COMMIT_MSG==
Fix cancellation race resulting in a leaked response on retry
==COMMIT_MSG==

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
@@ -319,7 +319,7 @@ final class RetryingChannel implements EndpointChannel {
                     },
                     backoffNanoseconds,
                     TimeUnit.NANOSECONDS);
-            return wrap(Futures.transformAsync(scheduled, input -> input, DialogueFutures.safeDirectExecutor()));
+            return wrap(DialogueFutures.transformAsync(scheduled, input -> input));
         }
 
         private long getBackoffNanoseconds() {
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
@@ -329,7 +329,7 @@ final class RetryingChannel implements EndpointChannel {
                             return;
                         }
                         ListenableFuture<Response> delegateResult = delegate.execute(request);
-                        DialogueFutures.addDirectCallback(delegateResult, new FutureCallback<Response>() {
+                        DialogueFutures.addDirectCallback(delegateResult, new FutureCallback<>() {
                             @Override
                             public void onSuccess(Response result) {
                                 if (!responseFuture.set(result)) {
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
@@ -19,11 +19,12 @@ package com.palantir.dialogue.core;
 import com.codahale.metrics.Meter;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Suppliers;
+import com.google.common.util.concurrent.FutureCallback;
 import com.google.common.util.concurrent.Futures;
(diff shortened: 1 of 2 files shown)


dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -305,21 +306,55 @@ private RetryingCallback(
...
+            // Code beyond this line implements the following without risking leaking responses:
+            // ListenableScheduledFuture<ListenableFuture<Response>> scheduled = scheduler.schedule(
+            //     () -> {
+            //         backoffSpan.complete(RetryingCallbackTranslator.INSTANCE, this);
+            //         return delegate.execute(request);
+            //     }, backoffNanoseconds, TimeUnit.NANOSECONDS);
+            // return wrap(DialogueFutures.transformAsync(scheduled, input -> input));
+            SettableFuture<Response> responseFuture = SettableFuture.create();
+            // Scheduler future must not be cancelled, otherwise responses will leak.
+            scheduler.schedule(
```
carterkozak, line 325: Not thrilled with this, ideally we'd be able to cancel the scheduledfuture, but it's not a big deal.

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -305,21 +306,55 @@ private RetryingCallback(
...
                     () -> {
                         backoffSpan.complete(RetryingCallbackTranslator.INSTANCE, this);
-                        return delegate.execute(request);
+                        if (responseFuture.isDone()) {
```
jkozlowski, line 328: Is there a chance to add a test?

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -305,21 +306,55 @@ private RetryingCallback(
...
+                            return;
+                        }
+                        ListenableFuture<Response> delegateResult = delegate.execute(request);
+                        DialogueFutures.addDirectCallback(delegateResult, new FutureCallback<Response>() {
```
jkozlowski: nit: FutureCallback<>

jkozlowski (approved):
This is very much in the same spirit as the other PR: it's scary and we now have the same resource handling code scattered all over. I think we should merge to stem the bleeding but seriously reconsider how to make this more bulletproof.

jkozlowski:
@carterkozak there's a direct usage of Futures methods (outside of Futures.immediate* family of methods) in DefaultClients:

```
if (!future.cancel(true)) {
                Futures.addCallback(future, CancelListener.INSTANCE, DialogueFutures.safeDirectExecutor());
            }
```

Shall we add a quick errorprone/checkstyle thing to ban these?

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
carterkozak: Unfortunately not that I'm aware of because the failure was the result of a race. I was able to put together an external reproducer which caused the leak every 30,000 requests or so, but not the sort of thing that would make a reasonable unit test.

carterkozak:
That addCallback is fine, that's almost exactly the same as the DialogueFutures version, except ours doesn't allow `? super T` so the generics wouldn't work -- the transformation methods risk leaks, the callback/listener methods are for convenience

carterkozak:
> Shall we add a quick errorprone/checkstyle thing to ban these?

That's a tough one, perhaps worth the cost at this point? error-prone checks can cause problems when they're defined in the same project where they're used, so it would require a fair bit of boilerplate work. It's not entirely clear where we would require the checks: All closeables, or just futures of Response?

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
carterkozak:
```suggestion
                        DialogueFutures.addDirectCallback(delegateResult, new FutureCallback<>() {
```

PR #1545, opened 2022-01-19
[WIP] DialogueListenableFuture

jkozlowski:
## Before this PR
This PR is complete garbage, implementation wise, but is a conversation starter.

What if our leaf futures started a context (that has a single useful method #failureCallback) which allows us to add callbacks that should run, should the future fail in any way. The leaf future that returns the Response from the network layer, simply add the Response#close as a callback to the future it returns.

Then all transformation futures (transform/transformAsync/catchingAll) simply need to share this context and propagate it up the chain. If any future fails, we close the context. Closing the context means all the callbacks get run, and forgotten (you can't run #close multiple times).

Subtleties:
1. What happens when context is already closed: I currently think the callback should run immediately to not leak, but where does the close exception go if it happens?
2. This is obviously a simplistic solution which assumes there are no forks in the chain of futures: so there's isn't a transformAsync applied twice where the returned futures do completely different things and append new failureCallbacks. In this implementation failure anywhere along either fork would close all resources. I feel like that's probably an assumption we can make for our codebase, implementing forking of contexts sounds not fun.

To make this more specific, we could instead have DialogueResponseFuture, so this is only tied to responses, and not e.g. closing timers or any other stuff.

### DIFF
--- a/dialogue-futures/src/main/java/com/palantir/dialogue/futures/DialogueDirectAsyncCatchingFuture.java
+++ b/dialogue-futures/src/main/java/com/palantir/dialogue/futures/DialogueDirectAsyncCatchingFuture.java
@@ -19,6 +19,7 @@ package com.palantir.dialogue.futures;
 import com.google.common.util.concurrent.AsyncFunction;
 import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
+import java.io.Closeable;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executor;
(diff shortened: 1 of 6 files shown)


carterkozak:
Have you seen [ClosingFuture](https://github.com/google/guava/blob/master/guava/src/com/google/common/util/concurrent/ClosingFuture.java)? We requested the feature here: https://github.com/google/guava/issues/3975 however it's not a clean migration, and would almost certainly result in friction (both bugs and api changes) to roll out.

jkozlowski:
Boom! great, haven't seen that

PR #1578, opened 2022-02-09
Check before allocating args for debug logging

schlosna:
## Before this PR

Some hot paths were allocating a lot of `SafeArg` only used for debug logging, even when debug logging is not enabled.

Example that popped in JFR:
```
SafeArg com.palantir.logsafe.SafeArg.of(String, Object)
Optional com.palantir.conjure.java.dialogue.serde.JacksonEmptyContainerLoader.constructEmptyInstance(Type, TypeMarker, int)
```

## After this PR
==COMMIT_MSG==
Check before allocating args for debug logging
==COMMIT_MSG==

### DIFF
--- a/dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/JacksonEmptyContainerLoader.java
+++ b/dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/JacksonEmptyContainerLoader.java
@@ -76,17 +76,21 @@ final class JacksonEmptyContainerLoader implements EmptyContainerDeserializer {
             if (parameter.isPresent()) {
                 return invokeStaticFactoryMethod(method, parameter.get());
             } else {
-                log.debug(
-                        "Found a @JsonCreator, but couldn't construct the parameter",
-                        SafeArg.of("type", type),
-                        SafeArg.of("parameter", parameter));
+                if (log.isDebugEnabled()) {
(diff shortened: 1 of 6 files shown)


dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/JacksonEmptyContainerLoader.java
```diff
@@ -150,7 +154,9 @@ private static int decrement(int maxRecursion, TypeMarker<?> originalType) {
         try {
             return Optional.ofNullable(method.invoke(null, parameter));
         } catch (IllegalAccessException | InvocationTargetException e) {
-            log.debug("Reflection instantiation failed", e);
+            if (log.isDebugEnabled()) {
+                log.debug("Reflection instantiation failed", e);
+            }
```
carterkozak, line 159: No allocation in this one, but it's future-proof :+1:

carterkozak (approved):
Great stuff, thank you!

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/JacksonEmptyContainerLoader.java
schlosna: yeah, I may have over-optimized in applying some human automation

PR #1582, opened 2022-02-11
Track host resolution failure metrics

raiju:
## After this PR
==COMMIT_MSG==
Track host resolution failure metrics
==COMMIT_MSG==


carterkozak (commented):
#1583 shows what is likely a better place for such logic to live

dialogue-apache-hc5-client/src/main/metrics/dialogue-apache-client-metrics.yml
```diff
@@ -39,3 +39,8 @@ namespaces:
         type: timer
         tags: [client-name, client-type]
         docs: Reports the time spent creating a new connection. This includes both connecting the socket and the full TLS handshake.
+
+      connection.resolution.error:
+        type: counter
+        tags: [client-name]
+        docs: Number of connections that failed due to host resolution failures.
```
carterkozak:
```suggestion
        type: meter
        tags: [client-name]
        docs: Rate that connections have failed due to host resolution.
```

carterkozak: The counter type can both increment and decrement, using a meter is a touch easier on the indexer :-)

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/UnknownHostCounter.java
```diff
@@ -0,0 +1,43 @@
...
+    private final Counter counter;
+
+    static UnknownHostCounter of(String clientName, TaggedMetricRegistry metrics) {
+        return new UnknownHostCounter(clientName, DialogueClientMetrics.of(metrics));
+    }
+
+    UnknownHostCounter(String clientName, DialogueClientMetrics metrics) {
+        this.clientName = clientName;
+        this.counter = metrics.connectionResolutionError(clientName);
+    }
```
carterkozak: I'd drop this class in favor of holding the connection-resolution-meter in InstrumentedDnsResolver. We can pass the TaggedMetricRegistry (or DialogueClientMetrics if we already have one) to InstrumentedDnsResolver where it can extract the `Meter` it cares about in the ctor.

dialogue-apache-hc5-client/src/test/java/com/palantir/dialogue/hc5/ApacheHttpClientChannelsTest.java
```diff
@@ -227,4 +258,20 @@ private long poolGaugeValue(TaggedMetricRegistry metrics, String clientName, Str
...
+                .map(Map.Entry::getValue)
+                .collect(Collectors.toSet());
+        if (matchingCounters.isEmpty()) {
+            return 0L;
+        }
+
+        Metric counter = Iterables.getOnlyElement(matchingCounters);
+        assertThat(counter).isInstanceOf(Counter.class);
+        return ((Counter) counter).getCount();
+    }
```
carterkozak: We can simplify + inline
```suggestion
    private long unknownHostCount(TaggedMetricRegistry metrics, String clientName) {
        return  DialogueClientMetrics.of(metrics).connectionResolutionError(clientName);
    }
```

raiju: Are gauges different? If no, should I also fix the other method (I stole this structure from)?

carterkozak: Gauges are different, they're runnable tasks which return a value, so the equivalent method would attempt to register a new one.
We added some methods for gauges (likely after that test was written) which allow us to build the MetricName object. That may clean things up, but I think it's out of scope for this PR.

raiju: Gauges are different! 😄

raiju:
@carterkozak Can you take another look?

Do you have an opinion on separating the host -> IP address & canonical host name resolution? For my use-case they should be similar. Happy to add it as an arg to the meter/isolate it to IP resolution if you'd prefer.

dialogue-apache-hc5-client/src/test/java/com/palantir/dialogue/hc5/ApacheHttpClientChannelsTest.java
```diff
@@ -124,6 +125,38 @@ public void metrics() throws Exception {
...
+                            .getCount())
+                    .isZero();
+
+            Channel channel =
+                    ApacheHttpClientChannels.createSingleUri("http://unknown-host-for-testing.unused", client);
+            ListenableFuture<Response> future =
+                    channel.execute(TestEndpoint.GET, Request.builder().build());
+
+            try (Response response = Futures.getChecked(future, UnknownHostException.class)) {
+                fail("This request should have failed with an unknown host exception!", response.code());
```
carterkozak: I think this uses `String.format` syntax (`response.code()` is ignored).

dialogue-apache-hc5-client/src/test/java/com/palantir/dialogue/hc5/ApacheHttpClientChannelsTest.java
```diff
@@ -124,6 +125,38 @@ public void metrics() throws Exception {
...
+
+        try (ApacheHttpClientChannels.CloseableClient client =
+                ApacheHttpClientChannels.createCloseableHttpClient(conf, "testClient")) {
+
+            TaggedMetricRegistry metrics = conf.taggedMetricRegistry();
+
+            assertThat(DialogueClientMetrics.of(metrics)
+                            .connectionResolutionError("testClient")
+                            .getCount())
+                    .isZero();
```
carterkozak: pedantic nit: We can hold the `Meter` instance and check it again on line 153 rather than looking it up a second time.

carterkozak:
lgtm, let's fix up the assertion message before merging :+1:

carterkozak:
> Do you have an opinion on separating the host -> IP address & canonical host name resolution? For my use-case they should be similar. Happy to add it as an arg to the meter/isolate it to IP resolution if you'd prefer.

In practice I think we're only resolving host -> IP, not canonical host names. Looks good as written :-)

raiju:
@carterkozak Should be fixed now

PR #1586, opened 2022-02-11
Dialogue request and response serializers are cached

carterkozak:
This allows instances to be reused when multiple endpoints
consume or return the same type. More importantly, this reduces
the cost of sticky channels where client instances are recreated.

## After this PR
==COMMIT_MSG==
Dialogue request and response serializers are cached, reducing the cost of sticky sessions which frequently re-create client interfaces
==COMMIT_MSG==

### DIFF
--- a/dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ConjureBodySerDe.java
+++ b/dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ConjureBodySerDe.java
@@ -17,6 +17,7 @@
 package com.palantir.conjure.java.dialogue.serde;
 
 import com.github.benmanes.caffeine.cache.Caffeine;
+import com.github.benmanes.caffeine.cache.CaffeineSpec;
 import com.github.benmanes.caffeine.cache.LoadingCache;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Lists;
@@ -64,7 +65,8 @@ final class ConjureBodySerDe implements BodySerDe {
     ConjureBodySerDe(
             List<WeightedEncoding> rawEncodings,
             ErrorDecoder errorDecoder,
-            EmptyContainerDeserializer emptyContainerDeserializer) {
+            EmptyContainerDeserializer emptyContainerDeserializer,
+            CaffeineSpec cacheSpec) {
         List<WeightedEncoding> encodings = decorateEncodings(rawEncodings);
         this.encodingsSortedByWeight = sortByWeight(encodings);
         Preconditions.checkArgument(encodings.size() > 0, "At least one Encoding is required");
@@ -83,8 +85,8 @@ final class ConjureBodySerDe implements BodySerDe {
         // Class unloading: Not supported, Jackson keeps strong references to the types
         // it sees: https://github.com/FasterXML/jackson-databind/issues/489
         this.serializers =
-                Caffeine.newBuilder().build(token -> new EncodingSerializerRegistry<>(defaultEncoding, token));
-        this.deserializers = Caffeine.newBuilder()
+                Caffeine.from(cacheSpec).build(token -> new EncodingSerializerRegistry<>(defaultEncoding, token));
+        this.deserializers = Caffeine.from(cacheSpec)
                 .build(token -> new EncodingDeserializerRegistry<>(
                         encodingsSortedByWeight, errorDecoder, emptyContainerDeserializer, token));
     }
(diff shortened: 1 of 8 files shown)


dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ConjureBodySerDe.java
```diff
@@ -80,6 +80,13 @@
                 emptyContainerDeserializer,
                 BinaryEncoding.OPTIONAL_MARKER);
         this.emptyBodyDeserializer = new EmptyBodyDeserializer(errorDecoder);
+        // Class unloading: Not supported, Jackson keeps strong references to the types
+        // it sees: https://github.com/FasterXML/jackson-databind/issues/489
+        this.serializers =
+                Caffeine.newBuilder().build(token -> new EncodingSerializerRegistry<>(defaultEncoding, token));
+        this.deserializers = Caffeine.newBuilder()
+                .build(token -> new EncodingDeserializerRegistry<>(
+                        encodingsSortedByWeight, errorDecoder, emptyContainerDeserializer, token));
```
schlosna: What's the rough heap retention footprint per serializer/deserializer? Should we bound these to avoid growing indefinitely? Wiring config through to here seems tricky, so wondering if we just set an arbitrary max size limit to keep top N with time expiration, or heuristically max weight it by max heap?

```suggestion
        this.serializers = Caffeine.newBuilder()
                .maximumSize(1_000) // reasonable bounds
                .expireAfterAccess(Duration.ofMinutes(1))
                .build(token -> new EncodingSerializerRegistry<>(defaultEncoding, token));
        this.deserializers = Caffeine.newBuilder()
                .maximumSize(1_000) // reasonable bounds
                .expireAfterAccess(Duration.ofMinutes(1))
                .build(token -> new EncodingDeserializerRegistry<>(
                        encodingsSortedByWeight, errorDecoder, emptyContainerDeserializer, token));
```

carterkozak: I need to do some validation on the delta between what's retained by the objectmapper and what's retained by this class.

I had considered using time-based retention as well, which would help with the frequently-rebuild-client case, an upper bound on cache size is always good idea as well.

schlosna: created https://github.com/palantir/dialogue/pull/1587 for thoughts

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/DefaultConjureRuntime.java
```diff
@@ -67,8 +73,19 @@ public Clients clients() {
...
         private Builder() {}
 
+        /**
+         * Specify the serializer and deserializer cache configuration.
+         */
+        @CanIgnoreReturnValue
+        public Builder serializerCacheSpecification(String specification) {
+            cacheSpec = CaffeineSpec.parse(specification);
+            return this;
+        }
```
carterkozak: Thinking about this a bit more, I'm not sure we want to expose this as API. It controls an internal detail that should be tuned to help our consumers, but not something they should tune themselves or depend upon because it could change.

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/DefaultConjureRuntime.java
```diff
@@ -29,6 +31,9 @@
  * {@link DefaultConjureRuntime} provides functionality required by generated handlers.
  */
 public final class DefaultConjureRuntime implements ConjureRuntime {
+    @VisibleForTesting
+    static final CaffeineSpec DEFAULT_SERDE_CACHE_SPEC =
+            CaffeineSpec.parse("maximumSize=1000,expireAfterAccess=1m,softValues");
```
carterkozak: I've been burned a few too many times by soft references, my preference is either hard or weak references. Soft references tend to cause gc pressure issues for some collectors, and behave identically to weak references in others (Shenandoah iirc, although it's possible this has been updated).

In this case the value will be held as a strong reference in any existing client that's still used, so I'd opt toward weak values.

schlosna: switched to weak values

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/DefaultConjureRuntime.java
schlosna: removed

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/DefaultConjureRuntime.java
carterkozak: Thanks!

carterkozak:
I'm happy with this, @fawind what do you think?

fawind (approved):
Makes sense to me!

PR #1595, opened 2022-02-19
Respect Retry-After header in 429 and 503 responses

pkoenig10:
Not sure whether this omission was intentional or whether we just never got around to implementing it.

## Before this PR
Dialogue does not respect the `Retry-After` header on 429 and 503 responses. Retried requests always use the next exponential backoff.

## After this PR
Dialogue respect the `Retry-After` header on 429 and 503 responses. If a `Retry-After` header is present, that backoff values will be used instead of the exponential backoff value.

### DIFF
--- a/changelog/@unreleased/pr-1595.v2.yml
+++ b/changelog/@unreleased/pr-1595.v2.yml
@@ -1,0 +1,6 @@
+type: improvement
+improvement:
+  description: Dialogue now clients now respect the Retry-After header in 429 and
+    503 responses.
+  links:
+  - https://github.com/palantir/dialogue/pull/1595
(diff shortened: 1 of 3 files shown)


carterkozak (commented):
This is intentionally not supported, I’d like to understand your use case. In general it’s easy to force incorrect behavior from a server when we really want to fail over to another host which may not be constrained on the same resource.

pkoenig10:
Yeah that makes sense, I figured that was the case.

I just came across a couple places where we're using `QosException.throttle` with a duration and noticed that Dialogue clients just ignored it.

Maybe we should consider deprecating those `QosException.throttle` methods to indicate they won't result in the expected behavior?

ash211:
It's definitely surprising that 429's `Retry-After` isn't respected.  It's mentioned in https://datatracker.ietf.org/doc/html/rfc6585#section-4 as an example, and in https://www.rfc-editor.org/rfc/rfc9110#field.retry-after

After a long internal discussion about getting Dialogue to respect this (essentially to merge this PR), I ended up throwing my own Conjure error that's essentially a RetryAfter error.  This is required to be a well-behaved client to Azure OpenAI's endpoints ([docs](https://learn.microsoft.com/en-us/azure/ai-services/openai/how-to/provisioned-get-started#what-should--i-do-when-i-receive-a-429-response)) with a Dialogue client.

PR #1606, opened 2022-02-25
Detect usage of deprecated cipher suites

carterkozak:
## After this PR
==COMMIT_MSG==
Detect usage of deprecated cipher suites
==COMMIT_MSG==

## Risks

Unlike servers, when a deprecated cipher is used, we don't have confidence that its removal would cause requests to fail. Servers ultimately choose which cipher to use, it's relatively common for java 8 servers to prefer less secure CBC ciphers over GCM due to the faster native implementation (GCM ciphers are much faster on modern jvms), however we can gauge the potential scale of impact prior to removing ciphers this way.


carterkozak:
Just waiting on cjr 7.29.0 to publish with https://github.com/palantir/conjure-java-runtime/pull/2272 for this to compile.

dialogue-apache-hc5-client/src/main/metrics/dialogue-apache-client-metrics.yml
```diff
@@ -16,36 +16,70 @@ namespaces:
...
+          - name: client-type
+            values: [ apache-hc5 ]
         docs: Rate that connections have failed due to host resolution.
+
+      connection.insecure.cipher:
+        type: meter
+        tags:
+          - name: client-name
+          - name: client-type
+            values: [ apache-hc5 ]
```
jkozlowski, line 82: Does this provide value now? Do we run with any other client? I guess might as well leave lest we switch to something else?

jkozlowski (approved):

dialogue-apache-hc5-client/src/main/metrics/dialogue-apache-client-metrics.yml
carterkozak: Ya, I figured it was best to add to all these metrics and include the static value to simplify our code.

PR #1619, opened 2022-03-02
Sticky sessions reuse metric references

carterkozak:
Previously these would be re-loaded from the metric registry
for each session (which are created very frequently). By reusing
metric references, we avoid MetricName creation, hashing,
and registry (concurrenthashmap) traversal.

## After this PR
==COMMIT_MSG==
Sticky sessions reuse metric references for better performance
==COMMIT_MSG==


jkozlowski (approved):
👍  🥇

PR #1629, opened 2022-03-10
Java 11 clients use TLSv1.3 in addition to TLSv1.2

carterkozak:
Previously TLSv1.3 was gated on JDK-15+ due to issues in early
releases of java 11. These have been resolved in later releases,
at this point we want to allow all clients to use modern protocols
to give us the option to remove TLSv1.2 support as early as possible
in case vulnerabilities are discovered.

Background:
https://webtide.com/openjdk-11-and-tls-1-3-issues/ describes the original issues we ran into
Tracked by https://bugs.openjdk.java.net/browse/JDK-8213202
Fixed in jdk12, and backported to jdk11 (11.0.3) here:
https://bugs.openjdk.java.net/browse/JDK-8218094
https://www.oracle.com/java/technologies/javase/11-0-3-relnotes.html

## After this PR
==COMMIT_MSG==
Java 11 clients use TLSv1.3 in addition to TLSv1.2
==COMMIT_MSG==

### DIFF
--- a/dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/TlsProtocols.java
+++ b/dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/TlsProtocols.java
@@ -16,26 +16,14 @@
 
 package com.palantir.dialogue.hc5;
 
-import com.google.common.primitives.Ints;
-
 /** Internal utility functionality to slowly roll out new TLS protocol support. */
 final class TlsProtocols {
 
-    private static final boolean JAVA_15_OR_LATER = isJava15OrLater();
     private static final String TLS_V1_2 = "TLSv1.2";
     private static final String TLS_V1_3 = "TLSv1.3";
 
     static String[] get() {
-        if (JAVA_15_OR_LATER) {
-            return new String[] {TLS_V1_3, TLS_V1_2};
-        } else {
-            return new String[] {TLS_V1_2};
-        }
-    }
-
-    private static boolean isJava15OrLater() {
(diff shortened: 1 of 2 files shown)


dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/TlsProtocols.java
```diff
@@ -16,26 +16,14 @@
 
 package com.palantir.dialogue.hc5;
 
-import com.google.common.primitives.Ints;
-
 /** Internal utility functionality to slowly roll out new TLS protocol support. */
 final class TlsProtocols {
 
-    private static final boolean JAVA_15_OR_LATER = isJava15OrLater();
```
fawind, line 24: Should we keep a check here that the client JDK is 11.0.3 or later? Probably not strictly necessary if we can provide some guarantees on the server side?

fawind (approved):
Makes sense, feel free to merge!

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/TlsProtocols.java
carterkozak: I've gone back and forth on the client-side a few times, heh. On the one hand, I'd like to ensure we can safely deprecate tlsv1.2 asap (no plans to do so, but we often don't get a long runway). However, if folks do run on old jdks in odd environments, it's difficult to tell, and we don't actually expect to see any 11.0.2 or older in our managed environments, so the two _should_ be equivalent.

I might as well add the check here since we don't expect to see it, and if our expectations are incorrect, we won't see odd tls failures.

PR #1708, opened 2022-05-13
Support request compression using the `compress-request` endpoint tag

carterkozak:
Using the `compress-request` tag signifies prior knowledge that the
remote server supports GZIP content-encoding on request bodies.
Unlike response compression, there is no form of negotiation.

This is associated with individual endpoints, regardless of the content type (json/smile/octet-stream/etc). I considered a custom implementation such that small requests are not compressed while large requests are, however there are two drawback which make it less desirable:
1. More complex to write (the request must be partially buffered until a threshold is reached, then either compressed, or sent)
2. Due to the lack of negotiation, this may fail against some servers. Tests _generally_ use small inputs which are not likely to reach a compression threshold, thus may not expose potential failures if such a threshold exists.

==COMMIT_MSG==
Support request compression using the `compress-request` endpoint tag
==COMMIT_MSG==

## Possible downsides?
Requires prior knowledge. Our standard webserver provides this guarantee, however conjure is much broader than our reference implementation, thus the feature may be used on an opt-in basis.

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/ContentEncodingChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/ContentEncodingChannel.java
@@ -1,5 +1,5 @@
 /*
- * (c) Copyright 2020 Palantir Technologies Inc. All rights reserved.
+ * (c) Copyright 2022 Palantir Technologies Inc. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/ContentEncodingChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/ContentEncodingChannel.java
@@ -126,6 +126,7 @@ final class ContentEncodingChannel implements EndpointChannel {
      * Specialized implementation of {@link GZIPOutputStream} which uses {@link Deflater#BEST_SPEED} to reduce
      * CPU utilization with a slight cost to compression ratio.
      */
+    @SuppressWarnings("FilterOutputStreamSlowMultibyteWrite") // false positive
     private static final class BestSpeedGzipOutputStream extends GZIPOutputStream {
 
         BestSpeedGzipOutputStream(OutputStream out) throws IOException {
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/ContentEncodingChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/ContentEncodingChannel.java
@@ -1,0 +1,146 @@
+/*
+ * (c) Copyright 2020 Palantir Technologies Inc. All rights reserved.
+ *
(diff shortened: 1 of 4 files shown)


dialogue-core/src/main/java/com/palantir/dialogue/core/ContentEncodingChannel.java
```diff
@@ -0,0 +1,146 @@
+/*
+ * (c) Copyright 2020 Palantir Technologies Inc. All rights reserved.
```
carterkozak:
```suggestion
 * (c) Copyright 2022 Palantir Technologies Inc. All rights reserved.
```

dialogue-core/src/main/java/com/palantir/dialogue/core/ContentEncodingChannel.java
```diff
@@ -0,0 +1,147 @@
...
+/**
+ * Adds support for transparently encoding sending <code>Content-Encoding: gzip</code> requests
+ * in a client agnostic way based on a Conjure endpoint tag. This requires prior knowledge that
+ * the remote server is capable of handling compressed data.
+ */
+final class ContentEncodingChannel implements EndpointChannel {
+
+    private static final String ENABLEMENT_TAG = "compress-request";
+    private static final String GZIP = "gzip";
+    private static final int BUFFER_SIZE = 8 * 1024;
```
fawind, line 44: For my own learning, how did you pick this buffer size? Just eyeballing something that looks reasonable or are there usual limits we have to account for?

fawind (approved):
Looks pretty clean! Binary opt-in/out makes sense to help the few endpoints first that benefit most from this!

dialogue-core/src/main/java/com/palantir/dialogue/core/ContentEncodingChannel.java
carterkozak: 8k is fairly standard, it matches guava ByteStreams.copy and (iirc) apache commons IoUtils.copy[Large]. It tends to perform well where larger values don't have much impact on performance. In our webservers we use 16k reusable buffers, however there's limited cost for the large size since they're pooled, yet doesn't impact performance in many cases.
I've done benchmarking, but I can't find my writeup/results at the moment.

carterkozak: 8k is a common i/o buffer size, used by [guava](https://github.com/google/guava/blob/acb004c10193c1c476a9a9ba8683eeff910e4a82/guava/src/com/google/common/io/ByteStreams.java#L60) and [apache commons-io](https://github.com/apache/commons-io/blob/495abe13df04c96eed0d690202d1c0e053bef29a/src/main/java/org/apache/commons/io/IOUtils.java#L128-L131).
I've done some benchmarking to ensure this is accurate elsewhere, but I can't seem to find my old results.
In this instance, the important piece is that we configure the gzip stream and buffering stream with similar sizes to avoid unnecessary churn, and buffering inputs to the gzip stream such that the cost isn't dominated by interactions with native code (`Deflater` implementation is native, non-java).

PR #1709, opened 2022-05-13
Handle raw Response out of the box.

jkozlowski:
## Before this PR

This has come up a few times, people usually copy the MyResponseDeserializer type into their repos. I feel like "*/*" is a good default to handle this and actually from reading the comments is likely what we intended from the start just never handled it.

## After this PR
==COMMIT_MSG==
Handle raw Response out of the box.
==COMMIT_MSG==

## Possible downsides?

It is a new default, but should only really apply to new code. If someone has already defined their own deserializer, it will be honoured.


dialogue-annotations-processor/src/test/resources/com/palantir/myservice/service/MyServiceDialogueServiceFactory.java.generated
```diff
@@ -66,7 +67,7 @@ public final class MyServiceDialogueServiceFactory implements DialogueServiceFac
                     endpointChannelFactory.endpoint(Endpoints.customResponse);
 
             private final Deserializer<Response> customResponseDeserializer = new ErrorHandlingDeserializerFactory(
-                            new MyResponseDeserializer(), new ConjureErrorDecoder())
+                            new ResponseDeserializer(), new ConjureErrorDecoder())
```
carterkozak, line 70: This is a little bit interesting: It's not entirely clear what behavior we expect in this case -- should a 500 response status still throw an exception, or do we want to get back a `Response` object with code=500?

carterkozak: I suppose we'd want to allow custom error decoders, which would make it pretty weird to reason about if the default error decoder differed for `Response` vs other types

carterkozak (approved):
Thanks!

dialogue-annotations-example/src/main/java/com/palantir/myservice/example/MyService.java
```diff
@@ -52,18 +53,15 @@ String greet(
...
     // Should we support custom static request headers via
     // method level annotations? e.g.
     // @Request.Header(name="Accept", value="text/plain")
     // This is the dialogue Response object
-    @Request(
-            method = HttpMethod.PUT,
-            path = "/custom/request1",
-            accept = MyResponseDeserializer.class,
-            errorDecoder = ErrorDecoder.None.class)
+    @MustBeClosed
```
carterkozak, line 63: This is huge, thanks!

dialogue-annotations-processor/src/test/resources/com/palantir/myservice/service/MyServiceDialogueServiceFactory.java.generated
jkozlowski: Good question: I feel like 500s should still be the remit of the error decoder.

PR #1711, opened 2022-05-13
Disable socks system proxy auto-detect if not specified

rtxv15:
## Before this PR
super.createSocket(context) calls new Socket(). This defaults to a SocksSocketImpl [here](https://github.com/AdoptOpenJDK/openjdk-jdk11/blob/19fb8f93c59dfd791f62d41f332db9e306bc1422/src/java.base/share/classes/java/net/Socket.java#L505). In #connect, this grabs the default system proxy selector if a proxy was not set ([here](https://github.com/AdoptOpenJDK/openjdk-jdk11/blob/19fb8f93c59dfd791f62d41f332db9e306bc1422/src/java.base/share/classes/java/net/SocksSocketImpl.java#L367)).
We've started out with a proxy selector from the client configuration that returned no proxy, and ended up using the system default one.
I don't think this behaviour matches that of an HTTP proxy either which won't default to the system one.

## After this PR
==COMMIT_MSG==
When no proxy is set, respect this instead of defaulting to the system socks proxy.
==COMMIT_MSG==

## Possible downsides?
None I believe. It is still possible to use the system socks proxy if that is the desired behaviour, by passing in the system ProxySelector in the configuration. This is the default behaviour in ClientConfiguration when no proxy is provided, so the default behaviour will stay the same after this PR.

### DIFF
--- a/dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientChannels.java
+++ b/dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientChannels.java
@@ -480,9 +480,9 @@ public final class ApacheHttpClientChannels {
                             new InstrumentedHostnameVerifier(
                                     new DefaultHostnameVerifier(), name, conf.taggedMetricRegistry())) {
                         @Override
-                        public Socket createSocket(final HttpContext context) throws IOException {
+                        public Socket createSocket(final HttpContext _context) {
                             return socksProxyAddress == null
-                                    ? super.createSocket(context)
+                                    ? new Socket(Proxy.NO_PROXY)
                                     : new Socket(new Proxy(Proxy.Type.SOCKS, socksProxyAddress));
                         }
                     };
@@ -491,9 +491,9 @@ public final class ApacheHttpClientChannels {
                     RegistryBuilder.<ConnectionSocketFactory>create()
                             .register(URIScheme.HTTP.id, new PlainConnectionSocketFactory() {
                                 @Override
-                                public Socket createSocket(final HttpContext context) throws IOException {
+                                public Socket createSocket(final HttpContext _context) {
                                     return socksProxyAddress == null
-                                            ? super.createSocket(context)
+                                            ? new Socket(Proxy.NO_PROXY)
                                             : new Socket(new Proxy(Proxy.Type.SOCKS, socksProxyAddress));
                                 }
                             })


rtxv15:
The new codepath creates a PlainSocketImpl instead [here](https://github.com/AdoptOpenJDK/openjdk-jdk11/blob/19fb8f93c59dfd791f62d41f332db9e306bc1422/src/java.base/share/classes/java/net/Socket.java#L146)

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientChannels.java
```diff
@@ -480,9 +480,9 @@ public CloseableClient build() {
                             new InstrumentedHostnameVerifier(
                                     new DefaultHostnameVerifier(), name, conf.taggedMetricRegistry())) {
                         @Override
-                        public Socket createSocket(final HttpContext context) throws IOException {
+                        public Socket createSocket(final HttpContext _context) {
                             return socksProxyAddress == null
-                                    ? super.createSocket(context)
+                                    ? new Socket(Proxy.NO_PROXY)
```
schlosna, line 485: shouldn't this use `rawSocketFactory` to create the socket?

carterkozak: It is, in a non-obvious way: The rawSocketFactory (SSLSocketFactory) is used with the vanilla socket created here, via [createSocket(Socket,String,int,boolean)](https://github.com/openjdk/jdk/blob/573eaceca559a8a0832b1e1a7181b2f21d3978c7/src/java.base/share/classes/javax/net/ssl/SSLSocketFactory.java#L281-L282).
Magic happens [here](https://github.com/apache/httpcomponents-client/blob/c395aad5adb4e9ee312559937c08382ea8215e7c/httpclient5/src/main/java/org/apache/hc/client5/http/ssl/SSLConnectionSocketFactory.java#L270-L274)

rtxv15: That sounds more correct. I don't think it matches the previous behaviour which wouldn't have been using that one ([here](https://github.com/apache/httpcomponents-client/blob/47d0946372e757ca6e7a366e340f0dba6686b012/httpclient5/src/main/java/org/apache/hc/client5/http/ssl/SSLConnectionSocketFactory.java#L194) using a default one, unless the config one is set to the default?), but sounds like it should have been. I'll change it

EDIT: Ignore, missed Carter's comment

carterkozak:
I've recreated without a fork to allow CI/changelog/etc to run correctly #1713

rtxv15:
Thanks! Closing in favour of that one

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientChannels.java
schlosna: Ahh, the naming confused me, so this is creating the actual raw socket, that will then be wrapped with TLS

PR #1713, opened 2022-05-13
Disable socks system proxy auto-detect if not specified

carterkozak:
Credit to @rtxv15 for #1711 which I've re-created without a fork to allow CI to run in a standard way.

The SSLSocketFactory creates plain sockets by default using
`new Socket()` rather than specifying `NO_PROXY` in cases that
a proxy is not expected. We explictly avoid using proxy controls
set at the jvm level for non-socks proxies, and never intentionally
used socks proxies from the jvm in dialogue. Socks proxies
are used for testing in some cases, but not in any production
environments, thus the risk introduced by this change is minimal.

==COMMIT_MSG==
Disable socks system proxy auto-detect if not specified
==COMMIT_MSG==

### DIFF
--- a/dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientChannels.java
+++ b/dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientChannels.java
@@ -56,6 +56,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.ScheduledFuture;
+import java.util.function.Supplier;
 import java.util.stream.LongStream;
 import javax.annotation.Nullable;
 import javax.net.ssl.SSLSocketFactory;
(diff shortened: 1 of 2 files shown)


dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientChannels.java
```diff
@@ -472,32 +473,33 @@ public CloseableClient build() {
...
-                        @Override
-                        public Socket createSocket(final HttpContext context) throws IOException {
-                            return socksProxyAddress == null
-                                    ? super.createSocket(context)
-                                    : new Socket(new Proxy(Proxy.Type.SOCKS, socksProxyAddress));
-                        }
-                    };
+            Supplier<Socket> simpleSocketCreator = socksProxyAddress == null
+                    ? () -> new Socket(Proxy.NO_PROXY)
+                    : () -> new Socket(new Proxy(Proxy.Type.SOCKS, socksProxyAddress));
```
carterkozak, line 478: Factored out rather than duplicating because we want identical behavior between the two

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientChannels.java
```diff
@@ -472,32 +473,33 @@ public CloseableClient build() {
...
+                                                    conf.taggedMetricRegistry(), rawSocketFactory, name),
+                                            TlsProtocols.get(),
+                                            supportedCipherSuites(
+                                                    CipherSuites.allCipherSuites(), rawSocketFactory, name),
+                                            new InstrumentedHostnameVerifier(
+                                                    new DefaultHostnameVerifier(), name, conf.taggedMetricRegistry())) {
+                                        @Override
+                                        public Socket createSocket(HttpContext _context) {
+                                            return simpleSocketCreator.get();
+                                        }
```
carterkozak, line 501: Updated to use the supplier, but also moved the implementation inline here matching `PlainConnectionSocketFactory` -- it was odd that one was defined inline, and the other was not.

schlosna (approved):

PR #1714, opened 2022-05-13
Retryer differentiates 308 status QoSException and non-QosException

carterkozak:
Distinction is based on the presence of a `Location` header. Note
that the default error decoder still requires strictly conjure
spec, logging an error in the case of a 308 without location info,
however non-conjure dialogue-annotations clients bring their own
error decoders to handle expected cases.

==COMMIT_MSG==
Retryer differentiates 308 status QoSException and non-QosException
==COMMIT_MSG==

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/Responses.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/Responses.java
@@ -15,13 +15,18 @@
  */
 package com.palantir.dialogue.core;
 
+import com.google.common.net.HttpHeaders;
 import com.palantir.dialogue.Response;
 
 /** Utility functionality for {@link Response} handling. */
 final class Responses {
 
     static boolean isRetryOther(Response response) {
-        return response.code() == 308;
+        // Note that a 308 status may be a non-retryable signal, for instance google sometimes
+        // uses a '308 Resume Incomplete', so we must verify the presence of a Location header
+        // to differentiate the two.
+        return response.code() == 308
+                && response.getFirstHeader(HttpHeaders.LOCATION).isPresent();
     }
 
     static boolean isTooManyRequests(Response response) {
--- a/dialogue-core/src/test/java/com/palantir/dialogue/core/CautiousIncreaseAggressiveDecreaseConcurrencyLimiterTest.java
+++ b/dialogue-core/src/test/java/com/palantir/dialogue/core/CautiousIncreaseAggressiveDecreaseConcurrencyLimiterTest.java
@@ -17,6 +17,7 @@
 package com.palantir.dialogue.core;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.ArgumentMatchers.eq;
 import static org.mockito.Mockito.mock;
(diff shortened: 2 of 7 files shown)


dialogue-core/src/main/java/com/palantir/dialogue/core/Responses.java
```diff
@@ -15,13 +15,18 @@
...
 /** Utility functionality for {@link Response} handling. */
 final class Responses {
 
     static boolean isRetryOther(Response response) {
-        return response.code() == 308;
+        // Note that a 308 status may be a non-retryable signal, for instance google sometimes
+        // uses a '308 Resume Incomplete', so we must verify the presence of a Location header
+        // to differentiate the two.
+        return response.code() == 308
+                && response.getFirstHeader(HttpHeaders.LOCATION).isPresent();
```
carterkozak, line 29: This is the only change to main sources, the rest is fixing up tests which made assumptions that didn't align with conjure QoS retry-other + test coverage for the ways this impacts usage.

dialogue-core/src/test/java/com/palantir/dialogue/core/CautiousIncreaseAggressiveDecreaseConcurrencyLimiterTest.java
```diff
@@ -185,16 +186,39 @@ public void onSuccess_releasesSuccessfully(Behavior behavior) {
...
-            assertThat(limiter.getLimit()).as("For status %d", code).isCloseTo(max * 0.9, Percentage.withPercentage(5));
-        }
+        double max = limiter.getLimit();
+        limiter.acquire(LimitEnforcement.DEFAULT_ENABLED).get().onSuccess(response);
+        assertThat(limiter.getLimit()).as("For status %d", 308).isCloseTo(max * 0.9, Percentage.withPercentage(5));
+    }
+
+    @Test
+    public void onSuccess_successIfResponseIndicatesNonQos308() {
+        // This represents google chunked-upload APIs which respond '308 Resume Incomplete' to indicate
```
jkozlowski, line 202: https://cloud.google.com/storage/docs/performing-resumable-uploads

I must say I had trouble following the problem here and it looks like there's quite [complicated history](https://insanecoding.blogspot.com/2014/02/http-308-incompetence-expected.html?m=1) for 308s.

dialogue-core/src/test/java/com/palantir/dialogue/core/TimingEndpointChannelTest.java
```diff
@@ -151,8 +168,37 @@ private void runRequest() {
             Preconditions.checkState(
                     maybeCode.isPresent() ^ maybeThrowable.isPresent(), "Either code of throwable need to be present");
             maybeCode.ifPresent(code -> {
-                Response response = mock(Response.class);
-                when(response.code()).thenReturn(code);
+                Response response = new Response() {
```
jkozlowski, line 171: Should we have TestResponse or something?

jkozlowski (approved):
Thanks, good find!

PR #1735, opened 2022-06-07
Closing a response drains at most 64k of data (was unlimited)

carterkozak:
This bounds the worst case time spent calling close, which could
be very long when multi-gigabyte files are downloaded over a slow
connection, especially when only a small number of bytes are
required from the beginning dynamically.

==COMMIT_MSG==
Closing a response drains at most 64k of data (was unlimited)
==COMMIT_MSG==


dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientBlockingChannel.java
```diff
@@ -321,6 +354,29 @@ public String toString() {
...
+     * destructive operation which should only occur in order to close the stream.
+     */
+    private static boolean hasSubstantialRemainingData(CloseableHttpResponse response) {
+        try {
+            HttpEntity entity = response.getEntity();
+            if (entity == null || !entity.isStreaming()) {
+                return false;
+            }
+            InputStream stream = entity.getContent();
+            // Fast check: The stream has been fully exhausted in hte expected case,
```
fawind:
```suggestion
            // Fast check: The stream has been fully exhausted in the expected case,
```

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientBlockingChannel.java
```diff
@@ -307,11 +322,29 @@ public ResponseAttachments attachments() {
...
             client = null;
-            try {
-                response.close();
-            } catch (IOException | RuntimeException e) {
-                log.warn("Failed to close response", e);
+            // Avoid attempting to close a response that has already been closed.
+            if (clientSnapshot != null) {
+                try {
+                    // Check if the response has been fully drained. If not, we close the connection rather than
+                    // potentially
```
fawind: ... draining the full response ?

carterkozak:
```suggestion
                    // potentially reading massive data unnecessarily.
```

fawind (approved):

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientBlockingChannel.java
```diff
@@ -321,6 +354,29 @@ public String toString() {
...
+                return false;
+            }
+            InputStream stream = entity.getContent();
+            // Fast check: The stream has been fully exhausted in the expected case,
+            // no need to create buffers for drainage unless we know there's data to drain.
+            if (stream.read() == -1) {
+                return false;
+            }
+            return REMAINING_CONTENT_CONNECTION_DISCARD_THRESHOLD
+                    == ByteStreams.exhaust(ByteStreams.limit(stream, REMAINING_CONTENT_CONNECTION_DISCARD_THRESHOLD));
```
jkozlowski, line 374: So this naturally allocates a buffer just for this check: is there a way we can do this without allocating? Also we allocate 2 implementations of an InputStream

carterkozak: That's correct, however in the expected case the `stream.read() == -1` check on line 370 avoids this path entirely because the stream has been fully exhausted. For what it's worth, the call to `close()` in both `ChunkedInputStream` and `ContentLengthInputStream` (from hc) allocate new buffers in order to drain remaining contents, so this shouldn't be dramatically worse.

PR #1740, opened 2022-06-14
Dialogue response body stream safety

carterkozak:
## Before this PR
Warnings logged when the response body was closed before fully consuming content due to a `runtime.discardEndpoint()` followed by `response.close()` which internally attempts to consume remaining data from a stream that has been disconnected.

## After this PR
==COMMIT_MSG==
Dialogue response body stream safety improvements to prevent accidental stream reuse and confusing warnings.
==COMMIT_MSG==

### DIFF
--- a/dialogue-apache-hc5-client/build.gradle
+++ b/dialogue-apache-hc5-client/build.gradle
@@ -24,4 +24,6 @@ dependencies {
     testImplementation project(':dialogue-serde')
     testImplementation 'org.awaitility:awaitility'
     testImplementation 'org.junit.jupiter:junit-jupiter'
+    testRuntimeOnly 'org.apache.logging.log4j:log4j-slf4j-impl'
+    testRuntimeOnly 'org.apache.logging.log4j:log4j-core'
 }
--- a/dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientBlockingChannel.java
+++ b/dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientBlockingChannel.java
@@ -15,6 +15,7 @@
  */
 package com.palantir.dialogue.hc5;
 
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ListMultimap;
 import com.google.common.collect.MultimapBuilder;
 import com.google.common.collect.Multimaps;
@@ -264,6 +265,9 @@ final class ApacheHttpClientBlockingChannel implements BlockingChannel {
         @Nullable
         private ListMultimap<String, String> headers;
 
+        @Nullable
+        private InputStream responseBody;
+
         HttpClientResponse(
                 ApacheHttpClientChannels.CloseableClient client,
                 CloseableHttpResponse response,
@@ -275,10 +279,19 @@ final class ApacheHttpClientBlockingChannel implements BlockingChannel {
 
         @Override
         public InputStream body() {
+            InputStream snapshot = this.responseBody;
+            if (snapshot == null) {
+                snapshot = createResponseBody();
+                this.responseBody = snapshot;
+            }
+            return snapshot;
+        }
+
+        private InputStream createResponseBody() {
             HttpEntity entity = response.getEntity();
             if (entity != null) {
                 try {
-                    return new ResponseInputStream(client, entity.getContent(), this);
+                    return new ResponseInputStream(entity.getContent(), this);
                 } catch (IOException e) {
                     throw new SafeRuntimeException("Failed to get response stream", e);
                 }
@@ -339,6 +352,9 @@ final class ApacheHttpClientBlockingChannel implements BlockingChannel {
                                             clientSnapshot.clientConfiguration().taggedMetricRegistry())
                                     .connectionClosedPartiallyConsumedResponse(clientSnapshot.name())
                                     .mark();
+                            // Important not to call response.close which still has access to the underlying
+                            // socket channel, which may be in use by another exchange.
+                            return;
                         }
                     }
                     response.close();
@@ -346,6 +362,10 @@ final class ApacheHttpClientBlockingChannel implements BlockingChannel {
                     log.warn("Failed to close response", e);
                 }
             }
+        }
+
+        boolean isOpen() {
+            return client != null;
         }
 
         @Override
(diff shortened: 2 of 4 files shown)


dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientBlockingChannel.java
```diff
@@ -455,37 +475,73 @@ public void close() {}
 
     private static final class ResponseInputStream extends FilterInputStream {
 
-        // Client reference is used to prevent premature termination
-        @Nullable
-        private ApacheHttpClientChannels.CloseableClient client;
```
carterkozak, line 460: This `client` ref was unnecessary due to the strong reference to `HttpClientResponse response`.

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientBlockingChannel.java
```diff
@@ -455,37 +475,73 @@ public void close() {}
...
             this.response = response;
         }
 
         @Override
-        public void close() throws IOException {
-            try {
-                try {
-                    response.close();
-                } finally {
-                    super.close();
```
carterkozak, line 479: This `super.close()` call is unnecessary as it may attempt to drain remaining bytes, which is already handled by response.close(). super.close may throw if response.close chose to terminate the connection rather than fully draining.

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientBlockingChannel.java
```diff
@@ -275,10 +279,19 @@ private static final class HttpClientResponse implements Response {
 
         @Override
         public InputStream body() {
+            InputStream snapshot = this.responseBody;
+            if (snapshot == null) {
+                snapshot = createResponseBody();
```
carterkozak, line 284: entity.getContent creates a new state tracking wrapper around the delegate InputStream for each invocation, by holding a reference to the single delegate, we can be more confident that state is consistent.

dialogue-apache-hc5-client/build.gradle
```diff
@@ -24,4 +24,6 @@ dependencies {
     testImplementation project(':dialogue-serde')
     testImplementation 'org.awaitility:awaitility'
     testImplementation 'org.junit.jupiter:junit-jupiter'
+    testRuntimeOnly 'org.apache.logging.log4j:log4j-slf4j-impl'
+    testRuntimeOnly 'org.apache.logging.log4j:log4j-core'
```
carterkozak, line 28: With the logging implementation in place, the reported warnings were very obvious in tests.

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientBlockingChannel.java
```diff
@@ -455,37 +475,73 @@ public void close() {}
...
-        public void close() throws IOException {
-            try {
-                try {
-                    response.close();
-                } finally {
-                    super.close();
-                }
-            } finally {
-                client = null;
+        public int read() throws IOException {
```
jkozlowski, line 486: Is there some extra testing we can write to check the behavior you're fixing?

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientBlockingChannel.java
```diff
@@ -455,37 +475,73 @@ public void close() {}
...
+            checkOpen();
+            return super.read();
+        }
+
+        @Override
+        public int read(byte[] buffer) throws IOException {
+            checkOpen();
```
jkozlowski, line 493: How expensive are these checkOpen calls?

jkozlowski: Would the delegate not already do these checks for you?

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientBlockingChannel.java
```diff
@@ -455,37 +475,73 @@ public void close() {}
...
+        public long skip(long num) throws IOException {
+            checkOpen();
+            return super.skip(num);
+        }
+
+        @Override
+        public void close() {
+            if (response.isOpen()) {
+                response.close();
+                // no need to close the delegate stream itself, closing the response is sufficient
```
jkozlowski, line 513: Is that always the case? what if they have an implementation that like holds onto a pooled bytebuffer? does calling close on the response guarantee to close any opened input streams?

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientBlockingChannel.java
```diff
@@ -339,6 +352,9 @@ public void close() {
                                             clientSnapshot.clientConfiguration().taggedMetricRegistry())
                                     .connectionClosedPartiallyConsumedResponse(clientSnapshot.name())
                                     .mark();
+                            // Do not call response.close which internally attempts to drain the response
```
jkozlowski, line 355: I didn't review the original PR, but runtime has #markConnectionNonReusable which maybe handles this whole business of closing things a bit more gracefully?

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientBlockingChannel.java
carterkozak: Yep, the client doesn’t pool anything beyond connections, and buffers are created with connections (and collected when the connection is closed)

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientBlockingChannel.java
carterkozak: Unfortunately it doesn’t help us solve the problem at hand because close still fully drains the response, regardless of the length.

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientBlockingChannel.java
carterkozak: It’s just a null check, idea is to co solo safe state between response.close and stream.close which wouldn’t be guaranteed to agree otherwise

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientBlockingChannel.java
jkozlowski: I found this: https://hc.apache.org/httpcomponents-client-4.5.x/current/tutorial/pdf/httpclient-tutorial.pdf

```
The difference between closing the content stream and closing the response is that the former will
attempt to keep the underlying connection alive by consuming the entity content while the latter
immediately shuts down and discards the connection.
```

Further reading they give the exact example of what we're trying to achieve:

```
There can be situations, however, when only a small portion of the entire response content needs to be
retrieved and the performance penalty for consuming the remaining content and making the connection
reusable is too high, in which case one can terminate the content stream by closing the response.
CloseableHttpClient httpclient = HttpClients.createDefault();
HttpGet httpget = new HttpGet("http://localhost/");
CloseableHttpResponse response = httpclient.execute(httpget);
try {
 HttpEntity entity = response.getEntity();
 if (entity != null) {
 InputStream instream = entity.getContent();
 int byteOne = instream.read();
 int byteTwo = instream.read();
 // Do not need the rest
 }
} finally {
 response.close();
}
```

jkozlowski: These docs are confusing me, maybe this is for a different version?

carterkozak: Those are the 4.x docs

jkozlowski (approved):
I think this makes sense, but feels awfully complicated to achieve this (Basically discard a connection. Upstream support would be nice

PR #1760, opened 2022-07-15
Dialogue doesn't allow response compression on range requests

carterkozak:
==COMMIT_MSG==
Dialouge doesn't allow response compression on range requests
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-1760.v2.yml
+++ b/changelog/@unreleased/pr-1760.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Dialouge doesn't allow response compression on range requests
+  links:
+  - https://github.com/palantir/dialogue/pull/1760
--- a/changelog/@unreleased/pr-1760.v2.yml
+++ b/changelog/@unreleased/pr-1760.v2.yml
@@ -1,5 +1,5 @@
 type: improvement
 improvement:
-  description: Dialouge doesn't allow response compression on range requests
+  description: Dialogue doesn't allow response compression on range requests
   links:
   - https://github.com/palantir/dialogue/pull/1760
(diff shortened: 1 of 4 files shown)


changelog/@unreleased/pr-1760.v2.yml
```diff
@@ -0,0 +1,5 @@
+type: improvement
+improvement:
+  description: Dialouge doesn't allow response compression on range requests
```
Dgleish: nit: Dialouge -> Dialogue

carterkozak:
```suggestion
  description: Dialogue doesn't allow response compression on range requests
```

carterkozak: :facepalm:

dialogue-core/src/main/java/com/palantir/dialogue/core/RangeAcceptsIdentityEncodingChannel.java
```diff
@@ -0,0 +1,57 @@
...
+import com.google.common.net.HttpHeaders;
+import com.google.common.util.concurrent.ListenableFuture;
+import com.palantir.dialogue.EndpointChannel;
+import com.palantir.dialogue.Request;
+import com.palantir.dialogue.Response;
+
+/**
+ * Range requests generally don't support content-compression in responses due to ambiguity around whether the
+ * byte-range refers to compressed content, or the original content, which are completely different.
+ */
```
sfackler, line 29: Do servers actually reject range requests with `Accept-Encoding: gzip` or do they just not compress the response?

carterkozak: Depends on the server. Some don't compress, though that's not required. Some configurations compress, then take a byte-range of the result, others take a byte-range, and compress that result. In theory, as I understand it, `Content-Encoding` should be applied prior to the byte-range, however that completely invalidates the negotiation piece of response compression as the range must be aware of this. Folks recommend `Transport-Encoding: gzip`, which is negotiated by sending the `TE` header, but that's not supported by much infrastructure, and cannot be relied upon at this point. Also may not survive through reverse proxies as it's considered a transport mechanism, incurring additional cost/bytes over the wire.

jkozlowski (approved):

PR #1776, opened 2022-08-04
Add diamonds when initializing ErrorHandlingDeserializingFactory

a-k-g:
## Before this PR
Compiler emits an unchecked call warning in the generated code.

## After this PR
==COMMIT_MSG==
Add diamonds when initializing ErrorHandlingDeserializingFactory in generated code
==COMMIT_MSG==

## Possible downsides?
Don't think there's any.

### DIFF
--- a/changelog/@unreleased/pr-1776.v2.yml
+++ b/changelog/@unreleased/pr-1776.v2.yml
@@ -1,0 +1,6 @@
+type: fix
+fix:
+  description: Add diamonds when initializing ErrorHandlingDeserializingFactory in
+    generated code
+  links:
+  - https://github.com/palantir/dialogue/pull/1776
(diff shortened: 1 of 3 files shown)


carterkozak:
> Compiler emits an unchecked call warning in the generated code.

Ideally we could fail compilation if more of these are introduced. Perhaps adding -Werror or -Xlint options to integrationinput?

carterkozak (approved):
thanks!

a-k-g:
>Ideally we could fail compilation if more of these are introduced. Perhaps adding -Werror or -Xlint options to integrationinput?

That's a great flag. I've made https://github.com/palantir/dialogue/pull/1777. Is this what you had in mind?

PR #1777, opened 2022-08-04
Stricter compilation in annotation processor test

a-k-g:
## Before this PR
FLUP for https://github.com/palantir/dialogue/pull/1776
Our annotation processor can generate code which may have compiler warnings. 

## After this PR
We can enforce that doesn't happen by setting the appropriate compiler flags in the test.

==COMMIT_MSG==
Enable stricter compilation flags in annotation processor test. No user facing change.
==COMMIT_MSG==

## Possible downsides?
Don't think so.

### DIFF
--- a/changelog/@unreleased/pr-1777.v2.yml
+++ b/changelog/@unreleased/pr-1777.v2.yml
@@ -1,0 +1,6 @@
+type: improvement
+improvement:
+  description: Enable stricter compilation flags in annotation processor test. No
+    user facing change.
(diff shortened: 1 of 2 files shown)


carterkozak:
Love it, thanks! Have you verified that it fails without your previous PR?

a-k-g:
Yes, I've verified that.

carterkozak (approved):
Thanks * 2 :-]

PR #1781, opened 2022-08-11
bump undertow to 2.2.19

jamesthomp:
I am running into https://issues.redhat.com/browse/UNDERTOW-2116 in an internal product, which is fixed in 2.2.19.

## After this PR
==COMMIT_MSG==
Bump undertow to 2.2.19.Final
==COMMIT_MSG==


carterkozak:
robots will handle this, undertow is only used in tests here.

PR #1782, opened 2022-08-18
Support alias parameter types

pkoenig10:
This is particularly useful if you have Conjure alias types you'd like to use in a Dialogue annotations interface. But it works for any types that have a `get` method which returns a Conjure primitive type.

This can be viewed as the counterpart to the Conjure Undertow annotations feature to [find a static factory decoder method](https://github.com/palantir/conjure-java/blob/b68a179b34e52a05495b9b9b795f6113b9a8ece7/conjure-undertow-processor/src/main/java/com/palantir/conjure/java/undertow/processor/data/ParamTypesResolver.java#L372-L393).

The consequential changes are in `ArgumentTypesResolver` - most of rest is just cleaning up some of the tests.

### DIFF
--- a/dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/data/ArgumentTypesResolver.java
+++ b/dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/data/ArgumentTypesResolver.java
@@ -18,12 +18,10 @@ package com.palantir.dialogue.annotations.processor.data;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableMap;
-import com.google.common.collect.Multimap;
+import com.google.common.collect.MoreCollectors;
 import com.palantir.common.streams.KeyedStream;
 import com.palantir.conjure.java.lib.SafeLong;
 import com.palantir.dialogue.RequestBody;
-import com.palantir.dialogue.annotations.processor.data.ArgumentType.OptionalType;
-import com.palantir.logsafe.Preconditions;
 import com.palantir.ri.ResourceIdentifier;
 import com.palantir.tokens.auth.AuthHeader;
 import com.palantir.tokens.auth.BearerToken;
@@ -31,12 +29,15 @@ import com.squareup.javapoet.ClassName;
 import com.squareup.javapoet.TypeName;
 import java.time.OffsetDateTime;
 import java.util.List;
-import java.util.Map;
 import java.util.Optional;
+import java.util.OptionalDouble;
 import java.util.OptionalInt;
+import java.util.OptionalLong;
 import java.util.UUID;
 import java.util.function.Function;
-import javax.lang.model.element.Element;
+import javax.lang.model.element.ElementKind;
+import javax.lang.model.element.ExecutableElement;
+import javax.lang.model.element.Modifier;
 import javax.lang.model.element.VariableElement;
 import javax.lang.model.type.TypeMirror;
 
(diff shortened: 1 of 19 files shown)


dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/data/ArgumentTypesResolver.java
```diff
@@ -66,87 +67,117 @@ public final class ArgumentTypesResolver {
...
-            return Optional.of(ArgumentTypes.primitive(typeName, planSerDeMethodName(typeName), Optional.empty()));
-        } else if (listType.map(innerType -> isPrimitive(TypeName.get(innerType)))
-                .orElse(false)) {
-            TypeName innerTypeName = TypeName.get(listType.get());
-            return Optional.of(
-                    ArgumentTypes.primitive(typeName, planSerDeMethodName(innerTypeName), Optional.of(innerTypeName)));
-        } else if (isRawRequestBody(actualTypeMirror)) {
-            return Optional.of(ArgumentTypes.rawRequestBody(TypeName.get(actualTypeMirror)));
-        } else if (optionalType.isPresent()) {
-            // TODO(12345): We only want to go one level down: don't allow Optional<Optional<Type>>.
```
pkoenig10, line 98: This PR fixes this TODO. See the comment below for details.

dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/data/ArgumentTypesResolver.java
```diff
@@ -66,87 +67,117 @@ public final class ArgumentTypesResolver {
...
+                            .innerType(getPrimitiveType(context.getTypeMirror(Double.class))
+                                    .get())
+                            .build()));
+        }
+
+        return context.getGenericInnerType(Optional.class, typeMirror)
+                .map(innerType -> getPrimitiveType(innerType)
+                        .or(() -> getAliasType(innerType))
+                        .orElseGet(() -> getCustomType(typeMirror)))
+                .map(innerType -> ArgumentTypes.optional(
```
pkoenig10, line 146: After this PR, we no longer support the following types:

- `Optional<Optional<T>>`
- `Optional<List<T>>`

We almost certainly never intended to support these types because servers cannot distinguish between `empty()` and `of(empty())`. It is trivial for clients to flatten these values into `Optional<T>` and `List<T>`, respectively.

dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/data/ArgumentTypesResolver.java
```diff
@@ -66,87 +67,117 @@ public final class ArgumentTypesResolver {
...
 
-    private String planSerDeMethodName(TypeName in) {
-        String typeName = PARAMETER_SERIALIZER_TYPES.get(in.box());
-        return Preconditions.checkNotNull(typeName, "Unknown type");
+    private Optional<ArgumentType> getListType(TypeMirror typeMirror) {
+        TypeName typeName = TypeName.get(typeMirror);
+
+        return context.getGenericInnerType(List.class, typeMirror)
+                .flatMap(this::getPrimitiveSerializerMethodName)
+                .map(methodName -> ArgumentTypes.list(typeName, methodName));
```
pkoenig10, line 103: It would be nice to also support lists of aliases, but this requires a more involved refactor to support arbitrary inner types. So I've left this unsupported for now.

dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/data/ArgumentTypesResolver.java
```diff
@@ -66,87 +67,117 @@ public final class ArgumentTypesResolver {
...
+                    typeName,
+                    ImmutableOptionalType.builder()
+                            .isPresentMethodName("isPresent")
+                            .valueGetMethodName("getAsInt")
+                            .innerType(getPrimitiveType(context.getTypeMirror(Integer.class))
+                                    .get())
+                            .build()));
+        }
+
+        if (context.isSameTypes(typeMirror, OptionalLong.class)) {
```
pkoenig10, line 120: We were missing some optional types, so I've added them in this PR

jkozlowski:
As I understand it, you want to just fallback to #get method on a type, if there's no custom converter specified? Memory might be bad here, but I think this is initially how I implemented this and @carterkozak suggested we make this more explicit with the factories and conversions, specifically to break off from the history of using magic methods in this space.

If we've done this for conjure undertow I guess that's precedent, but I'd just like to confirm our thinking and be intentional.

pkoenig10:
> you want to just fallback to #get method on a type, if there's no custom converter specified?

Yes, that is what I'm proposing in this PR. While it's certainly possible to create custom encoders, it's tedious and punishes users for using strongly typed wrappers for things like IDs, etc.

This PR intentionally limits the alias depth to 1 (the `get` method must return a Conjure primitive - it cannot return another alias), which should alleviate some of the concerns.

dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/data/ArgumentTypesResolver.java
```diff
@@ -66,87 +67,117 @@ public final class ArgumentTypesResolver {
...
+
+    private Optional<ArgumentType> getRawRequestBodyType(TypeMirror typeMirror) {
+        if (!context.isAssignable(typeMirror, RequestBody.class)) {
+            return Optional.empty();
+        }
+
+        return Optional.of(ArgumentTypes.rawRequestBody(TypeName.get(typeMirror)));
+    }
+
+    private Optional<ArgumentType> getAliasType(TypeMirror typeMirror) {
```
carterkozak, line 163: Would this detect `java.util.Optional` as an alias?

pkoenig10: No, the return type of `Optional.get` is a `TypeVar` which is not a key in our map of primitive types.

Screen Shot 2022-08-29 at 8 33 59 AM

pkoenig10: Also, we always call `getOptionalType` before `getAliasType`. Checking for an alias type should always be the last thing we do before falling back to a custom type.

carterkozak:
The design looks reasonable to me, but I haven't thoroughly reviewed. Perhaps @jkozlowski or @fawind can review while I'm away?

dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/data/ArgumentTypesResolver.java
```diff
@@ -66,87 +67,117 @@ public final class ArgumentTypesResolver {
...
-            return mapType;
-        } else {
-            return Optional.of(ArgumentTypes.customType(typeName));
-        }
+        return getPrimitiveType(typeMirror)
+                .or(() -> getListType(typeMirror))
+                .or(() -> getOptionalType(typeMirror))
+                .or(() -> getRawRequestBodyType(typeMirror))
+                .or(() -> getAliasType(typeMirror))
+                .orElseGet(() -> ArgumentTypes.customType(TypeName.get(typeMirror)));
```
fawind, line 82: Nice, this is much easier to follow!

dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/data/ArgumentTypesResolver.java
fawind: Makes sense. We should also fail loudly on compile when we encounter these cases (where users do not provide an explicit encoder).

fawind (approved):
Sorry for the delay, lgtm!

PR #1790, opened 2022-09-22
Modulate response output writes to 16KiB blocks

schlosna:
## Before this PR
When Dialogue would execute `RequestBody#writeTo`, a `RequestBody` implementation may attempt to write a large buffer to the resulting output stream causing suboptimal cipher throughput due to the large size.

See https://github.com/palantir/hadoop-crypto/pull/586 & https://github.com/palantir/hadoop-crypto/pull/587

## After this PR
==COMMIT_MSG==
Modulate response output writes to 16KiB blocks

Block size of 16 KB is small enough to allow cipher implementations to become hot and optimize properly when given large inputs. Otherwise large array writes into a javax.crypto.CipherOutputStream fail to use intrinsified implementations. If 16 KB blocks aren't enough to produce hot methods, the I/O is small and infrequent enough that performance isn't relevant.  For more information, see the details around com.sun.crypto.provider.GHASH::processBlocks in
https://github.com/palantir/hadoop-crypto/pull/586#issuecomment-964394587
==COMMIT_MSG==

### DIFF
--- a/dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientBlockingChannel.java
+++ b/dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientBlockingChannel.java
@@ -40,6 +40,7 @@ import com.palantir.logsafe.logger.SafeLoggerFactory;
 import com.palantir.tracing.api.TraceHttpHeaders;
 import java.io.ByteArrayInputStream;
 import java.io.FilterInputStream;
+import java.io.FilterOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
@@ -48,6 +49,7 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.OptionalInt;
 import java.util.OptionalLong;
@@ -449,7 +451,7 @@ final class ApacheHttpClientBlockingChannel implements BlockingChannel {
 
         @Override
         public void writeTo(OutputStream outStream) throws IOException {
-            requestBody.writeTo(outStream);
+            requestBody.writeTo(new ModulatingOutputStream(outStream));
         }
 
         @Override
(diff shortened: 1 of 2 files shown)


dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientBlockingChannel.java
```diff
@@ -473,6 +475,42 @@ public String toString() {
         public void close() {}
     }
 
+    /**
+     * {@link ModulatingOutputStream} limits the size of individual writes to the wrapped {@link OutputStream}
+     * in order to prevent degraded performance on large buffers as described in
+     * <a href="https://github.com/palantir/hadoop-crypto/pull/586">hadoop-crypto#586</a>.
+     */
+    static final class ModulatingOutputStream extends FilterOutputStream {
```
schlosna, line 483: Adjusted naming from `ChunkingOutputStream` to `ModulatingOutputStream` to avoid confusion with multipart chunks since this is just wrapping & writing to a single OutputStream in 16KB blocks

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientBlockingChannel.java
```diff
@@ -473,6 +475,42 @@ public String toString() {
...
+            int currentOffset = off;
+            int remaining = len;
+            while (remaining > 0) {
+                int toWrite = Math.min(remaining, BLOCK_SIZE);
+                out.write(buffer, currentOffset, toWrite);
+                currentOffset += toWrite;
+                remaining -= toWrite;
+            }
+        }
+    }
```
carterkozak, line 517: Might be helpful to override `write(int)` as a simple passthrough to `out` and avoid array allocation

schlosna: done, though I don't think it is actually necessary as `FilterOutputStream#write(int)` just does `out.write(b);`

carterkozak: Ah right, sorry for the runaround!

carterkozak (approved):
Thanks!

PR #1797, opened 2022-09-30
Add JDK version to user agent

schlosna:
## Before this PR
Diagnosing the JDK version a specific Dialogue client is using to make requests is difficult. 

## After this PR
==COMMIT_MSG==
Adds JDK version to the dialogue user agent string, e.g. jdk/17.0.4.1
==COMMIT_MSG==


dialogue-core/src/main/java/com/palantir/dialogue/core/UserAgentEndpointChannel.java
```diff
@@ -60,7 +61,7 @@ public ListenableFuture<Response> execute(Request request) {
     }
 
     private static UserAgent augmentUserAgent(UserAgent baseAgent, Endpoint endpoint) {
-        return tryAddEndpointAgent(baseAgent, endpoint).addAgent(DIALOGUE_AGENT);
+        return tryAddEndpointAgent(baseAgent, endpoint).addAgent(DIALOGUE_AGENT).addAgent(JDK_AGENT);
```
schlosna: I'm considering exposing something like `UserAgent addAgents(Iterable<Agent>)` to `conjure-java-runtime-api` to avoid multiple immutable copies, but that can be a follow on PR

schlosna: https://github.com/palantir/conjure-java-runtime-api/pull/902

schlosna:
Closing in favor of https://github.com/palantir/dialogue/pull/1802

PR #1808, opened 2022-10-18
Allow for a pluggable ErrorDecoder in a DefaultConjureRuntime

ferozco:
## Before this PR
`DefaultConjureRuntime` exposes a way for you to provide custom deserialization logic with `Encoding`s but it does not expose a way to override the way that non-200 responses are deserialized and converted to exceptions limiting the usage when interacting with servers that do not respond with Conjure compatible errors.

There is currently a way for you to inject a custom errorDecoder using the annotations but it only works if you are defining your API in Java.

## After this PR
==COMMIT_MSG==
Allow for a pluggable ErrorDecoder in a DefaultConjureRuntime
==COMMIT_MSG==

This change makes it possible to use Conjure generated APIs with non-Conjure error messages

## Possible downsides?
We renamed the public `ErrorDecoder` enum to `ConjureErrorDecoder` so that we could expose a well named `ErrorDecoder` interface. Unfortunately that does come at the cost of an API break. We can avoid the break by changing around the names a bit but I think it would leave the code base with a bit of a smell. Looking through the commit history it seems like the class was made public in this [PR](https://github.com/palantir/dialogue/pull/1267) 

Defer to the maintainers on what they think is best.


palantirtech:
Thanks for your interest in [palantir/dialogue](https://github.com/palantir/dialogue), @ferozco! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

ferozco:
Hey @carterkozak would you mind taking a look?

ferozco:
Friendly ping on this @carterkozak @tpetracca

PR #1843, opened 2023-01-10
DialogueNodeSelectionStrategy specifies a locale for upper-casing

carterkozak:
==COMMIT_MSG==
DialogueNodeSelectionStrategy specifies a locale for upper-casing
==COMMIT_MSG==

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueNodeSelectionStrategy.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueNodeSelectionStrategy.java
@@ -25,6 +25,7 @@ import com.palantir.logsafe.exceptions.SafeIllegalStateException;
 import com.palantir.logsafe.logger.SafeLogger;
 import com.palantir.logsafe.logger.SafeLoggerFactory;
 import java.util.List;
+import java.util.Locale;
 
 /**
  * Supported node selection strategies which can either be user provided or received over the wire from servers.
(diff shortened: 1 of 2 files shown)


dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueNodeSelectionStrategy.java
```diff
@@ -50,7 +51,7 @@ static List<DialogueNodeSelectionStrategy> fromHeader(String header) {
      * which users can't normally configure.
      */
     private static DialogueNodeSelectionStrategy safeValueOf(String string) {
-        String uppercaseString = string.toUpperCase();
+        String uppercaseString = string.toUpperCase(Locale.ROOT);
```
schlosna: should we bump `DefaultLocale` to `error` in `build.gradle` line 89?
```
            options.errorprone.error('DefaultLocale')
```

An alternative to avoid the potential allocation which would fail fast based on length check as opposed to the switch `hashCode`, then do case insensitive `regionMatches`:

```
        if (string.equalsIgnoreCase("PIN_UNTIL_ERROR")) {
            return PIN_UNTIL_ERROR;
        } else if (string.equalsIgnoreCase("PIN_UNTIL_ERROR_WITHOUT_RESHUFFLE")) {
            return PIN_UNTIL_ERROR_WITHOUT_RESHUFFLE;
        } else if (string.equalsIgnoreCase("BALANCED")) {
            return BALANCED;
        }
```

schlosna (approved):
overall LGTM, left a couple questions/comments

dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueNodeSelectionStrategy.java
carterkozak: :+1: Updated with equalsIgnoreCase, and enabled DefaultLocale at warn (which will fail builds due to `-Werror`)

Issue #1814, opened 2022-11-07
Closing a streaming response body hangs if the server doesn't write


sfackler:
## What happened?

When closing the response body for a request to an endpoint that the server never finishes writing to, the close call hangs until the socket-level read timeout is reached inside of [`hasSubstantialRemainingData`](https://github.com/palantir/dialogue/blob/develop/dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientBlockingChannel.java#L379-L400). Currently the only way to avoid the hang is to intentionally not close the response body, which can then trigger the leak detection warnings in Dialogue.

## What did you want to happen?

Dialogue should be able to handle hung responses. Either the hasSubstantialRemainingData call should happen under a fairly aggressive timeout, or there should be some way to close the response without trying to consume any more of it.

PR #1853, opened 2023-02-01
Set a small socket timeout before closing partially consumed responses

carterkozak:
This works around a degenerate case in which closure isn't negotiated in a timely manner, and the client consumes far more bytes than desired in `SSLSocketInputRecord.deplete`.

## After this PR
==COMMIT_MSG==
Set a small socket timeout before closing partially consumed responses
==COMMIT_MSG==

### DIFF
--- a/dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientBlockingChannel.java
+++ b/dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientBlockingChannel.java
@@ -59,6 +59,7 @@ import javax.annotation.Nullable;
 import org.apache.hc.client5.http.ConnectTimeoutException;
 import org.apache.hc.client5.http.classic.ExecRuntime;
 import org.apache.hc.client5.http.impl.classic.CloseableHttpResponse;
+import org.apache.hc.client5.http.io.ConnectionEndpoint;
 import org.apache.hc.client5.http.protocol.HttpClientContext;
 import org.apache.hc.core5.function.Supplier;
 import org.apache.hc.core5.http.Header;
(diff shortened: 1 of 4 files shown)


dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientBlockingChannel.java
```diff
@@ -347,6 +353,16 @@ public void close() {
...
                         if (runtime != null) {
+                            // Attempt to set the smallest possible socket timeout before closing the connection.
+                            // In some degenerate cases, remote servers may not close connections, causing the client
+                            // to consume network resources and time in SSLSocketInputRecord.deplete.
+                            ConnectionEndpoint maybeEndpoint = ConnectionEndpointAccess.getConnectionEndpoint(runtime);
+                            assert maybeEndpoint != null || !runtime.isEndpointConnected()
+                                    : "Expected ConnectionEndpointAccess.getConnectionEndpoint "
+                                            + "to extract a ConnectionEndpoint";
+                            if (maybeEndpoint != null) {
+                                maybeEndpoint.setSocketTimeout(Timeout.ONE_MILLISECOND);
```
schlosna, line 364: Should we set this to 0 so that the deplete `tryToRead` arg is false
* https://github.com/openjdk/jdk17u/blob/2fe42855c48c49b515b97312ce64a5a8ef3af407/src/java.base/share/classes/sun/security/ssl/SSLSocketImpl.java#L1221
* https://github.com/openjdk/jdk17u/blob/0a6252780a862e6d5646a28c2b22a1c03eff37bf/src/java.base/share/classes/sun/security/ssl/SSLSocketInputRecord.java#L494

carterkozak: I don't think so -- that would only skip the single-byte read call, not the `skip` where most of the time is spent in a loop (with potentially no timeout)

schlosna: Yeah, makes sense so we should expect the skip to timeout after 1ms with no bytes

sfackler:
With some adjustments this could also handle #1814.

carterkozak:
Yep, I'd thought the same thing -- planning to take a look at it after some issues I'm currently digging into.

PR #1886, opened 2023-03-16
Validate path params

pkoenig10:
Similar to https://github.com/palantir/conjure-java/pull/1876, but for Dialogue.

## Before this PR

`dialogue-annotations-processor` does not validate that the path template parameters match the `@PathParam` method parameters.

## After this PR

`dialogue-annotations-processor` validates that the path template parameters match the `@PathParam` method parameters.


carterkozak (approved):
Looks great, thank you!

PR #1898, opened 2023-04-03
Don't use Jersey to parse path templates

pkoenig10:
## Before this PR
`dialogue-annotations` will parse path template segments like `{foo:something}` as a variable path segment with the name `foo`.

This is because we use Jersey's `UriTemplateParser`, which supports patterns in path template segments, to parse path templates.

This is confusing for `dialogue-annotations` users because it appears that Dialogue respects these patterns when in reality it does not.

## After this PR

Don't use Jersey's `UriTemplateParser` to parse path templates.

The validation added in https://github.com/palantir/dialogue/pull/1886 will ensure a path template segment with a pattern causes the build to fail.


dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/data/HttpPathParser.java
```diff
@@ -17,14 +17,16 @@
...
 import java.util.List;
 import java.util.Optional;
+import java.util.stream.Collectors;
 import javax.lang.model.element.Element;
-import org.glassfish.jersey.uri.internal.UriTemplateParser;
 
 public final class HttpPathParser {
 
+    private static final Splitter SPLITTER =
+            Splitter.on('/').omitEmptyStrings(); // omit empty segments; typically the first segment is empty
```
carterkozak, line 28: This may change behavior when a pair of slashes is expected -- can we normalize out a leading slash instead?

pkoenig10: Is this not equivalent to the existing `segment.isEmpty()` check?

Happy to change this - I agree that only omitting empty segments on the first segment is better behavior. Just want to make sure I correctly understand what the code is currently doing.

pkoenig10: I'm a little worried about changing the behavior here if an existing path template has repeated slashes.

dialogue-annotations-processor/src/test/java/com/palantir/dialogue/annotations/processor/DialogueRequestAnnotationsProcessorTest.java
```diff
@@ -82,18 +81,6 @@ public void testWrongFactoryAnnotation() {
...
-    @Test
-    public void testHttpPathIsParsed() {
-        Compilation compilation = compileTestClass(TEST_CLASSES_BASE_DIR, UnparseableHttpPath.class);
-        assertThat(compilation)
-                .hadErrorContaining("Failed to parse http path: threw an exception java.lang.IllegalArgumentException: "
-                        + "Invalid syntax in the template \"/greet/{oops\". "
-                        + "Check if a path parameter is terminated with a \"}\".")
-                .inFile(compilation.sourceFiles().get(0))
-                .onLineContaining("String greet(@Request.PathParam String greeting)");
-        assertThat(compilation).hadErrorCount(2);
```
carterkozak, line 94: 👍 Makes sense that `{oops` is considered a valid path segment

carterkozak:
Great stuff, thanks @pkoenig10! I'm happy with this once the edge case is addressed

dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/data/HttpPathParser.java
carterkozak: oh hmm, I'd missed that. I agree that we shouldn't change behavior here, I can revisit this later.

carterkozak (approved):

PR #1912, opened 2023-04-18
Dialogue attempts not to retain context classloaders

carterkozak:
Dialogue didn't behave well in systems using classloaders which must be unloaded later.

## After this PR
==COMMIT_MSG==
Dialogue attempts not to retain context classloaders
==COMMIT_MSG==

## Possible downsides?
The cleaner thread is no longer named something that points to dialogue. We could use the same ThreadFactory utility that I adopted elsewhere, however the cleaner is the most important thread to clean up (the others will expire) and the default constructor uses an internal `InnocuousThread` implementation which does an even better job of avoiding holding external references.

### DIFF
--- a/dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/CleanerSupport.java
+++ b/dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/CleanerSupport.java
@@ -16,17 +16,15 @@
 
 package com.palantir.dialogue.hc5;
 
-import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import java.lang.ref.Cleaner;
 import java.lang.ref.Cleaner.Cleanable;
 
 /** Reflective shim to allow consumers on new runtime versions to take advantage of java.lang.ref.Cleaner. */
 final class CleanerSupport {
 
-    private static final Cleaner cleaner = Cleaner.create(new ThreadFactoryBuilder()
-            .setDaemon(true)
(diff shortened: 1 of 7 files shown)


dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/CleanerSupport.java
```diff
@@ -16,17 +16,15 @@
...
 /** Reflective shim to allow consumers on new runtime versions to take advantage of java.lang.ref.Cleaner. */
 final class CleanerSupport {
 
-    private static final Cleaner cleaner = Cleaner.create(new ThreadFactoryBuilder()
-            .setDaemon(true)
-            .setNameFormat("dialogue-cleaner-%d")
-            .build());
+    // Ideally we would use a name pattern like 'dialogue-cleaner-%d', however it's more important
+    // that this cleaner does not retain context classloader references.
+    private static final Cleaner cleaner = Cleaner.create();
```
jkozlowski, line 27: Could not use ThreadFactoryBuilder and impl from scratch?

carterkozak: Not quite, cleaner has some special case logic to clear threadlocal data after cleaning objects when the thread implementation is the internal default (InnocuousThread) here. Unfortunately it becomes complex to duplicate that logic (requires Unsafe apis)

PR #1920, opened 2023-04-24
Eagerly throw on null request parameters

westinrm:
## Before this PR
Null header names, query parameter names/values, and path parameter names/values will throw exceptions during request execution or URL rendering respectively without any useful stack trace, making debugging the root cause of these issues difficult.

For example, this is the error when a null query parameter value is used:
```
java.lang.NullPointerException: Cannot invoke "String.getBytes(java.nio.charset.Charset)" because "source" is null
    at com.palantir.dialogue.core.BaseUrl$UrlEncoder.encode(BaseUrl.java:279)
    at com.palantir.dialogue.core.BaseUrl$UrlEncoder.encodeQueryNameOrValue(BaseUrl.java:271)
    at com.palantir.dialogue.core.BaseUrl$DefaultUrlBuilder.queryParam(BaseUrl.java:164)
    ...
    at com.palantir.dialogue.core.BaseUrl.render(BaseUrl.java:58)
    at com.palantir.dialogue.hc5.ApacheHttpClientBlockingChannel.execute(ApacheHttpClientBlockingChannel.java:99)
    ...
```

## After this PR
==COMMIT_MSG==
Eagerly throw on null request parameters

By throwing when the null values are inserted into the Request builder it's possible to get a helpful stack trace instead of one that just traces back into one of Dialogue's async threads.
==COMMIT_MSG==

## Possible downsides?
- Slightly worse performance due to having to check all entries inserted against null

### DIFF
--- a/dialogue-target/src/main/java/com/palantir/dialogue/Request.java
+++ b/dialogue-target/src/main/java/com/palantir/dialogue/Request.java
@@ -22,7 +22,6 @@ import com.google.common.collect.ListMultimap;
 import com.google.common.collect.Multimap;
 import com.google.common.collect.Multimaps;
 import com.palantir.logsafe.Preconditions;
-import com.palantir.logsafe.exceptions.SafeNullPointerException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.LinkedHashMap;
@@ -205,18 +204,12 @@ public final class Request {
         }
 
         public Request.Builder putAllHeaderParams(Multimap<String, ? extends String> entries) {
-            if (entries.containsKey(null)) {
-                throw new SafeNullPointerException("Header name must not be null");
-            }
             mutableHeaderParams().putAll(entries);
             return this;
         }
 
         public Request.Builder putQueryParams(String key, String... values) {
             Preconditions.checkNotNull(key, "Query parameter name must not be null");
-            for (String value : values) {
-                Preconditions.checkNotNull(value, "Query parameter value must not be null");
-            }
             mutableQueryParams().putAll(key, Arrays.asList(values));
             return this;
         }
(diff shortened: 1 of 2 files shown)


dialogue-target/src/main/java/com/palantir/dialogue/Request.java
```diff
@@ -187,6 +188,7 @@ public Request.Builder putHeaderParams(String key, String... values) {
         }
 
         public Request.Builder putHeaderParams(String key, String value) {
+            Preconditions.checkArgumentNotNull(key, "Header name must not be null");
```
pkoenig10: These should probably be `checkNotNull`:
- It's what we use in our other builders and constructors
- If one of these values is null it's very likely an application bug and not due to user input

dialogue-target/src/main/java/com/palantir/dialogue/Request.java
```diff
@@ -197,26 +199,38 @@ public Request.Builder headerParams(Multimap<String, ? extends String> entries)
         }
 
         public Request.Builder putAllHeaderParams(String key, Iterable<String> values) {
+            Preconditions.checkArgumentNotNull(key, "Header name must not be null");
```
carterkozak: Let's match conjure builder semantics where we `Preconditions.checkNotNull` individual elements (like this one), but we don't iterate through collections (like `putAllHeaderParams(Multimap<String, ? extends String> entries)` below)

carterkozak: I am curious how your bug came about, I would have expected the recent nullaway rollout to catch that sort of thing before hitting prod. Perhaps there's a case we're missing?

westinrm: I still have no idea how my bug came about, which is part of why I want to make this change.
I'm okay trying just checking non-collections for now, though I doubt there are ever enough query/path/header parameters that this will actually be a notable performance hit.

dialogue-target/src/main/java/com/palantir/dialogue/Request.java
```diff
@@ -207,16 +209,21 @@ public Request.Builder putAllHeaderParams(Multimap<String, ? extends String> ent
...
 
         public Request.Builder putQueryParams(String key, String... values) {
+            Preconditions.checkNotNull(key, "Query parameter name must not be null");
             mutableQueryParams().putAll(key, Arrays.asList(values));
             return this;
         }
 
         public Request.Builder putQueryParams(String key, String value) {
+            Preconditions.checkNotNull(key, "Query parameter name must not be null");
+            Preconditions.checkNotNull(value, "Query parameter value must not be null");
```
carterkozak, line 219: Why do we check key+value here, but only key in `public Request.Builder putHeaderParams(String key, String value) {`?

westinrm: headers actually permit null values from what I could determine, while query and path parameters don't

carterkozak: if null headers don't result in failures, I'd consider that a bug, but reasonable not to make it throw in this change 👍

carterkozak (approved):

PR #1923, opened 2023-04-25
Remove redundant SSL work from client config creation

jpcorreia99:
## Before this PR
Currently, we do redundant work when[ building the client configuration](https://github.com/palantir/dialogue/blob/069b94db3f53fceda0fb3e4bbc729e1e4e39a71d/dialogue-clients/src/main/java/com/palantir/dialogue/clients/AugmentClientConfig.java#L64-L95). 

The call to `ClientConfigurations.of` already initializes a full `SSLContext`, however an SSLContext  is initialized again [here](https://github.com/palantir/dialogue/blob/069b94db3f53fceda0fb3e4bbc729e1e4e39a71d/dialogue-clients/src/main/java/com/palantir/dialogue/clients/AugmentClientConfig.java#L68-L70) since we want to optionally use a security provider and also control cache size.

This PR uses the new overloaded `ClientConfigurations.of()` method that accepts a `TrustContextFactory`. This interface allows us to fully control how SSL related classes are provided and removes the redundant SSL/IO work that was being done in the initial `of()` call.

## After this PR
==COMMIT_MSG==
Use TrustContextFactory when building ClientConfiguration instead of overwriting already performed SSL work.
==COMMIT_MSG==


dialogue-clients/src/main/java/com/palantir/dialogue/clients/AugmentClientConfig.java
```diff
@@ -93,4 +90,26 @@ static ClientConfiguration getClientConf(ServiceConfiguration serviceConfig, Aug
 
         return builder.build();
     }
+
+    private static TrustContextFactory buildTrustContextFactory(AugmentClientConfig augment) {
```
carterkozak: perhaps `buildTrustContextFactory(Optional<Provider> augment)` to make it clear what data we use from the AugmentClientConfig?

jpcorreia99: Handled in https://github.com/palantir/dialogue/pull/1923/commits/29d64ac261593516cfbdb84d4c30cf528f984ddb

dialogue-clients/build.gradle
```diff
@@ -16,6 +16,7 @@ dependencies {
     implementation 'com.google.errorprone:error_prone_annotations'
     implementation 'com.google.guava:guava'
     implementation 'com.palantir.conjure.java.runtime:keystores'
+    implementation 'com.palantir.conjure.java.api:ssl-config'
```
jpcorreia99, line 19: Had to add this line as CI was failing because of implicit dependencies

carterkozak (approved):

PR #1947, opened 2023-06-07
Provide more info on field type mismatch

tengfliao:
## Before this PR

The code compilation might otherwise look okay until the annotation processor runs and it's not clear what the type mismatch is since the fields and types aren't reported.

## After this PR
==COMMIT_MSG==

Provide more information for debugging when the @DialogueService annotation processor detects type mismatches.
==COMMIT_MSG==


carterkozak (approved):
Thanks!

PR #1948, opened 2023-06-07
make the meshmode enum public in order to allow use of stripMeshPrefix

bavardage:
## Before this PR
There are some places where we need to strip the mesh mode prefix from urls before handing those urls to other non-dialogue clients (e.g. conda, gradle package managers, etc).

I shared some other context in an internal slack channel.

## After this PR
==COMMIT_MSG==
Make MeshMode enum public to allow use of stripMeshPrefix
==COMMIT_MSG==

It would be good if, when we do this modification, we do this using a common library function (rather than just inlining the url manipulation). This will make it easier to track where this manipulation needs to be done, and will make it easier to unwind should we ever change the way we produce mesh urls.

### DIFF
--- a/changelog/@unreleased/pr-1948.v2.yml
+++ b/changelog/@unreleased/pr-1948.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Make MeshMode enum public to allow use of stripMeshPrefix
+  links:
+  - https://github.com/palantir/dialogue/pull/1948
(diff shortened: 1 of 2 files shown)


carterkozak (changes requested):
MeshMode is an implementation detail. In practice, it could be implemented by registering `mesh-http` and `mesh-https` schemes with the apache httpclient as well, but making this class public would lock us into a specific pattern. Given the function is a fairly simple one-liner, I'd prefer to duplicate it rather than create a public library utility in dialogue.

The complexity isn't parsing, but rather ensuring that anywhere the mesh uri is used takes the mesh into account and doesn't multiple retries.

bavardage:
>  I'd prefer to duplicate it 

not duplicating means we can unwind this

>  it could be implemented by registering mesh-http and mesh-https schemes with the apache httpclient as well

there are clients we do not control

PR #1951, opened 2023-06-13
Dialogue guards against HTTPCLIENT-1924 client self-closure

carterkozak:
## Before this PR
The underlying http client closes itself when an Error is thrown.

## After this PR
==COMMIT_MSG==
Dialogue guards against HTTPCLIENT-1924 client self-closure
==COMMIT_MSG==

## Possible downsides?
In cases that would previously self-close, we're guaranteed that a connection will leak. Our connection pool allows two billion entries, so it's not the worst thing in the world so long as it doesn't happen in a hot loop, which is unlikely, and in the unlikely worst case causes similar issues to the self-closure problem.

### DIFF
--- a/dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/InstrumentedPoolingHttpClientConnectionManager.java
+++ b/dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/InstrumentedPoolingHttpClientConnectionManager.java
@@ -19,6 +19,7 @@ package com.palantir.dialogue.hc5;
 import com.codahale.metrics.Timer;
 import com.codahale.metrics.Timer.Context;
 import com.palantir.logsafe.SafeArg;
+import com.palantir.logsafe.exceptions.SafeRuntimeException;
 import com.palantir.logsafe.logger.SafeLogger;
 import com.palantir.logsafe.logger.SafeLoggerFactory;
 import com.palantir.tracing.CloseableTracer;
@@ -47,6 +48,7 @@ final class InstrumentedPoolingHttpClientConnectionManager
     private final TaggedMetricRegistry registry;
     private final String clientName;
     private final Timer connectTimer;
+    private volatile boolean closed;
 
     InstrumentedPoolingHttpClientConnectionManager(
             PoolingHttpClientConnectionManager manager, TaggedMetricRegistry registry, String clientName) {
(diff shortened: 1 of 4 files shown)


dialogue-apache-hc5-client/src/test/java/com/palantir/dialogue/hc5/ErrorRecoveryTest.java
```diff
@@ -0,0 +1,122 @@
...
+            return Optional.empty();
+        }
+    }
+
+    private enum ErrorThrowingRequestBody implements RequestBody {
+        INSTANCE;
+
+        @Override
+        public void writeTo(OutputStream _output) {
+            throw new StackOverflowError();
```
carterkozak, line 106: This makes for a nice test, and I suspect is similar to the issue which prompted this change, but in practice I don't think httpclient should actually close the connection pool when an error is thrown from this portion of the request. I will follow up separately on that.

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/InstrumentedPoolingHttpClientConnectionManager.java
```diff
@@ -58,12 +60,41 @@
...
     @Override
     public void close() {
-        manager.close();
+        if (!closed) {
+            log.warn(
+                    "Dialogue ConnectionManager close invoked unexpectedly and ignored",
+                    SafeArg.of("clientName", clientName),
+                    new SafeRuntimeException("stacktrace"));
+            // Note: manager.close is not invoked here, see closeUnderlyingConnectionManager.
+        }
```
carterkozak, line 69: Relevant upstream link is: https://github.com/apache/httpcomponents-client/blob/5b61e132c3871ddfa967ab21b3af5d6d738bc6e8/httpclient5/src/main/java/org/apache/hc/client5/http/impl/classic/MainClientExec.java#L161-L164

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/InstrumentedPoolingHttpClientConnectionManager.java
```diff
@@ -58,12 +60,41 @@
...
+    }
+
+    /**
+     * This method is used to close the underlying connection manager, while the {@link #close()} methods are
+     * overridden specifically not to do so in order to avoid unexpected closure in MainClientExec when
+     * an Error is encountered due to HTTPCLIENT-1924.
+     * https://github.com/apache/httpcomponents-client/blob/5b61e132c3871ddfa967ab21b3af5d6d738bc6e8/
+     * httpclient5/src/main/java/org/apache/hc/client5/http/impl/classic/MainClientExec.java#L161-L164
```
carterkozak, line 89: hyperlink: [MainClientExec.java#L161-L164](https://github.com/apache/httpcomponents-client/blob/5b61e132c3871ddfa967ab21b3af5d6d738bc6e8/httpclient5/src/main/java/org/apache/hc/client5/http/impl/classic/MainClientExec.java#L161-L164)

mpritham (approved):

PR #1971, opened 2023-07-10
Failing test for retry after connection failure

carterkozak:
We should retry failures which don't consume requests.

==COMMIT_MSG==
Failing test for retry after connection failure
==COMMIT_MSG==


carterkozak:
Integrated into #1972

PR #1972, opened 2023-07-11
RetryingChannel: Retry request if message is not consumed.

mpritham:
## Before this PR
Currently we only retry requests that have a repeatable body. We should also retry non-repeatable requests if the body is not consumed.

## After this PR
==COMMIT_MSG==
Update RetryingChannel to retry requests if the request body is not consumed, or the request is repeatable.
==COMMIT_MSG==

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
@@ -52,6 +52,7 @@ import java.net.SocketTimeoutException;
 import java.time.Duration;
 import java.util.Optional;
 import java.util.OptionalInt;
+import java.util.OptionalLong;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ThreadLocalRandom;
 import java.util.concurrent.TimeUnit;
@@ -231,10 +232,7 @@ final class RetryingChannel implements EndpointChannel {
         private RetryingCallback(
                 Endpoint endpoint, Request request, Optional<SafeRuntimeException> callsiteStacktrace) {
             this.endpoint = endpoint;
-            this.request = Request.builder()
-                    .from(request)
-                    .body(request.body().map(ConsumptionTrackingRequestBody::new))
-                    .build();
+            this.request = trackNonRepeatableBodyConsumption(request);
             this.callsiteStacktrace = callsiteStacktrace;
         }
 
@@ -251,18 +249,16 @@ final class RetryingChannel implements EndpointChannel {
         }
 
         private boolean requestCanBeRetried() {
-            return request.body()
-                    .map(body -> {
-                        if (body instanceof ConsumptionTrackingRequestBody) {
-                            return ((ConsumptionTrackingRequestBody) body).requestBodyCanBeRetried();
-                        } else {
-                            throw new SafeIllegalStateException(
-                                    "Expected a ConsumptionTrackingRequestBody but got an unexpected request body type",
-                                    SafeArg.of("bodyClassName", body.getClass()));
-                        }
-                    })
-                    // If the request body is empty, we can retry the request.
-                    .orElse(true);
+            // If the request body is empty, we can retry the request.
+            if (request.body().isEmpty()) {
+                return true;
+            }
+
+            RequestBody body = request.body().get();
+            if (body instanceof ConsumptionTrackingRequestBody) {
+                return ((ConsumptionTrackingRequestBody) body).requestBodyCanBeRetried();
+            }
+            return body.repeatable();
         }
 
         private ListenableFuture<Response> wrap(ListenableFuture<Response> input) {
@@ -451,7 +447,6 @@ final class RetryingChannel implements EndpointChannel {
 
         ConsumptionTrackingRequestBody(RequestBody delegate) {
             this.delegate = delegate;
-            this.consumed = false;
         }
 
         @Override
@@ -466,6 +461,11 @@ final class RetryingChannel implements EndpointChannel {
         }
 
         @Override
+        public OptionalLong contentLength() {
+            return delegate.contentLength();
+        }
+
+        @Override
         public boolean repeatable() {
             return delegate.repeatable();
         }
@@ -477,7 +477,7 @@ final class RetryingChannel implements EndpointChannel {
             delegate.close();
         }
 
-        public boolean requestBodyCanBeRetried() {
+        boolean requestBodyCanBeRetried() {
             return repeatable() || !consumed;
         }
 
@@ -485,6 +485,16 @@ final class RetryingChannel implements EndpointChannel {
         public String toString() {
             return "ConsumptionTrackingRequestBody{" + delegate + '}';
         }
+    }
+
+    private static Request trackNonRepeatableBodyConsumption(Request request) {
+        if (request.body().isEmpty() || request.body().get().repeatable()) {
+            return request;
+        }
+        return Request.builder()
+                .from(request)
+                .body(request.body().map(ConsumptionTrackingRequestBody::new))
+                .build();
     }
 
     /**
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
@@ -228,13 +228,13 @@ final class RetryingChannel implements EndpointChannel {
         private final DetachedSpan span = DetachedSpan.start("Dialogue-RetryingChannel");
         private int failures = 0;
 
-        private Optional<RequestBodyWrapper> maybeWrappedRequestBody;
-
         private RetryingCallback(
                 Endpoint endpoint, Request request, Optional<SafeRuntimeException> callsiteStacktrace) {
             this.endpoint = endpoint;
-            this.request = request;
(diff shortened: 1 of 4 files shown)


dialogue-core/src/test/java/com/palantir/dialogue/core/RetryingChannelTest.java
```diff
@@ -569,6 +572,46 @@ public void close() {}
         verify(channel, times(1)).execute(any());
     }
```
mpritham, line 574: This test is no longer needed.

mpritham:
Hm, I don't think this will actually work because the wrapped request body isn't sent to the client, so `writeTo` won't be called on it. But some tests should fail... checking.

carterkozak:
It's concerning that we don't have any failing tests in this case. The change to RetryingChannelTest sort of stands to reason because it wasn't implemented expecting the need to consume a request body, but we should be able to get mockito to consume the request object instead of rewriting all the logic.

carterkozak:
> Hm, I don't think this will actually work because the wrapped request body isn't sent to the client, so writeTo won't be called on it.

Perhaps [ContentEncodingChannel.java](https://github.com/palantir/dialogue/blob/4cb67dbc427c22745b5d2e063db67cb3eec4477c/dialogue-core/src/main/java/com/palantir/dialogue/core/ContentEncodingChannel.java) can provide some inspiration.

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -235,10 +228,13 @@ private final class RetryingCallback {
         private final DetachedSpan span = DetachedSpan.start("Dialogue-RetryingChannel");
         private int failures = 0;
 
+        private Optional<RequestBodyWrapper> maybeWrappedRequestBody;
+
         private RetryingCallback(
                 Endpoint endpoint, Request request, Optional<SafeRuntimeException> callsiteStacktrace) {
             this.endpoint = endpoint;
             this.request = request;
+            this.maybeWrappedRequestBody = request.body().map(RequestBodyWrapper::new);
```
jkozlowski: I think you need to actually change request to RequestBodyWrapper so that's the body you're actually using for the writes? and then consumed needs to be volatile at least?

mpritham: > I think you need to actually change request to RequestBodyWrapper so that's the body you're actually using for the writes?

Agree. I think this is what Carter was also getting at with the pointer to `ContentEncodingChannel`.

> and then consumed needs to be volatile at least?

Missed this, thanks! I think this makes sense as we'd like the writes to the `consumed` field complete before any `read` from this retrying logic.

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -433,6 +445,48 @@ private String channelName() {
         }
     }
 
+    private static final class ConsumptionTrackingRequestBody implements RequestBody {
```
carterkozak, line 444: We're missing a delegating override on `contentLength()`

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -433,6 +445,48 @@ private String channelName() {
...
+        }
+
+        @Override
+        public void close() {
+            // Closing the body returns all resources, and the body can be considered consumed.
+            this.consumed = true;
+            delegate.close();
+        }
+
+        public boolean requestBodyCanBeRetried() {
```
carterkozak: No need for a public modifier on a private class -- it's required on the `@Override` methods because interface method overrides must always be public.
```suggestion
       boolean requestBodyCanBeRetried() {
```

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -238,7 +231,10 @@ private final class RetryingCallback {
         private RetryingCallback(
                 Endpoint endpoint, Request request, Optional<SafeRuntimeException> callsiteStacktrace) {
             this.endpoint = endpoint;
-            this.request = request;
+            this.request = Request.builder()
+                    .from(request)
+                    .body(request.body().map(ConsumptionTrackingRequestBody::new))
+                    .build();
```
carterkozak: Let's factor this out into a `private static Request trackNonRepeatableBodyConsumption(Request input) {` method (no strong opinion on the name, this is just an example).
If the request body is repeatable or not present, we should pass along the request as is without wrapping/copying.

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -254,6 +250,21 @@ ListenableFuture<Response> execute() {
             return result;
         }
 
+        private boolean requestCanBeRetried() {
+            return request.body()
+                    .map(body -> {
```
carterkozak: I'd prefer to implement this without lambdas, avoiding a slight bit of overhead (the `map` function is a java object, which must be allocated and freed -- modern GCs are incredible, but not as efficient as avoiding allocations where it's relatively simple to do so).

In most places I'd prefer the way you've written this as is, but the constraints change a bit deeper in our infra.

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -254,6 +250,21 @@ ListenableFuture<Response> execute() {
...
+                        if (body instanceof ConsumptionTrackingRequestBody) {
+                            return ((ConsumptionTrackingRequestBody) body).requestBodyCanBeRetried();
+                        } else {
+                            throw new SafeIllegalStateException(
+                                    "Expected a ConsumptionTrackingRequestBody but got an unexpected request body type",
+                                    SafeArg.of("bodyClassName", body.getClass()));
```
carterkozak: If we implement the suggestion about only wrapping request bodies as necessary, this can become `return body.repeatable();`

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -433,6 +445,48 @@ private String channelName() {
         }
     }
 
+    private static final class ConsumptionTrackingRequestBody implements RequestBody {
+        private final RequestBody delegate;
+        private volatile boolean consumed;
+
+        ConsumptionTrackingRequestBody(RequestBody delegate) {
+            this.delegate = delegate;
+            this.consumed = false;
```
carterkozak: booleans default to `false`, explicitly setting a volatile variable to its default variable is more expensive than the implicit assignment because it forces volatile assignment semantics unnecessarily (edge cases can apply if a constructor passes the partially-initialized `this` reference elsewhere, but that's not considered good practice anyhow).

We could write an error-prone rule for this, however there's a small legibility cost, and it doesn't matter much outside of hot performance-sensitive codepaths.

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
mpritham: 🤦‍♂️ Thank you

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
mpritham: > If the request body is repeatable or not present, we should pass along the request as is without wrapping/copying.

Nice. This is better: only using the tracking logic if needed.

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
mpritham: I see. Thanks.

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
mpritham: Interesting consideration.

> edge cases can apply if a constructor passes the partially-initialized this reference elsewhere

I suppose this case is when, in the constructor, initialization is done in multiple threads which read the volatile variable. Seems scary to debug.

dialogue-core/src/test/java/com/palantir/dialogue/core/RetryingChannelTest.java
```diff
@@ -531,15 +536,62 @@ public void retriesSocketTimeout_connectionTimeout() throws ExecutionException,
...
 
     @Test
-    public void nonRetryableRequestBodyIsNotRetried() throws ExecutionException, InterruptedException {
-        when(channel.execute(any())).thenReturn(FAILED).thenReturn(SUCCESS);
+    public void requestWithNonRepeatableBodyRetriedWhenConnectionFails()
```
carterkozak, line 536: Can we rearrange the change to the tests for a cleaner diff? I'd prefer if we could make the minimal possible change to the existing test such that it still asserts the same thing, and add new test cases to cover new considerations. It makes reviewing much easier :-)

carterkozak (approved):
great stuff, thanks!

PR #1977, opened 2023-07-21
Use blocking interface in annotation processor example

pkoenig10:
We've seen consumers use async interfaces when it's not necessary simply because that's what appears to be suggested by the README. We'd prefer consumers to use sync interfaces where possible because they are easier to reason about.


carterkozak (approved):
Thanks, @pkoenig10!

PR #1983, opened 2023-07-28
BinaryRequestBody and ContentBody use InputStream.transferTo(OutputStream)

schlosna:
## Before this PR
Dialogue used Guava `ByteStreams.copy` for 8KiB byte buffer chunk copies.

## After this PR
==COMMIT_MSG==
`BinaryRequestBody` and `ContentBody` use `InputStream.transferTo(OutputStream)`

Allow for optimization when underlying input stream (such as `ByteArrayInputStream`, `ChannelInputStream`) overrides `transferTo(OutputStream)` to avoid extra array allocations and copy larger chunks at a time (e.g. allowing 16KiB chunks via `ApacheHttpClientBlockingChannel.ModulatingOutputStream` from #1790).

When running on JDK 21+, this also enables 16KiB byte chunk copies via `InputStream.transferTo(OutputStream)` per JDK-8299336, where as on JDK < 21 and when using Guava `ByteStreams.copy` 8KiB byte chunk copies are used.

References:
* https://github.com/palantir/hadoop-crypto/pull/586
* https://bugs.openjdk.org/browse/JDK-8299336
* https://bugs.openjdk.org/browse/JDK-8067661
* https://bugs.openjdk.org/browse/JDK-8265891
* https://bugs.openjdk.org/browse/JDK-8273038
* https://bugs.openjdk.org/browse/JDK-8279283
* https://bugs.openjdk.org/browse/JDK-8296431

==COMMIT_MSG==

### DIFF
--- a/dialogue-annotations/src/main/java/com/palantir/dialogue/annotations/InputStreamContentBody.java
+++ b/dialogue-annotations/src/main/java/com/palantir/dialogue/annotations/InputStreamContentBody.java
@@ -16,7 +16,6 @@
 
 package com.palantir.dialogue.annotations;
 
-import com.google.common.io.ByteStreams;
 import com.palantir.logsafe.Preconditions;
 import com.palantir.logsafe.UnsafeArg;
 import com.palantir.logsafe.logger.SafeLogger;
@@ -39,7 +38,7 @@ final class InputStreamContentBody implements ContentBody {
 
     @Override
     public void writeTo(OutputStream output) throws IOException {
-        ByteStreams.copy(inputStream, output);
+        inputStream.transferTo(output);
     }
 
     @Override
--- a/changelog/@unreleased/pr-1983.v2.yml
+++ b/changelog/@unreleased/pr-1983.v2.yml
@@ -1,0 +1,19 @@
+type: improvement
+improvement:
+  description: |-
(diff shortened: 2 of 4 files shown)


carterkozak (approved):

dialogue-annotations/src/main/java/com/palantir/dialogue/annotations/InputStreamContentBody.java
```diff
@@ -39,7 +38,7 @@ final class InputStreamContentBody implements ContentBody {
 
     @Override
     public void writeTo(OutputStream output) throws IOException {
-        ByteStreams.copy(inputStream, output);
+        inputStream.transferTo(output);
```
jkozlowski, line 41: Should we get this into error-prone if this is now the preferred approach? I know I'm guilty of many uses of this method

schlosna: good call, I'd like to write up an error-prone rule to migrate folks soon.

schlosna: https://github.com/palantir/gradle-baseline/pull/2616

PR #1986, opened 2023-08-01
Implement a sticky session interface for multiple client interfaces

carterkozak:
==COMMIT_MSG==
Implement a sticky session interface for multiple client interfaces
==COMMIT_MSG==

### DIFF
--- a/dialogue-clients/src/main/java/com/palantir/dialogue/clients/DialogueClients.java
+++ b/dialogue-clients/src/main/java/com/palantir/dialogue/clients/DialogueClients.java
@@ -128,6 +128,26 @@ public final class DialogueClients {
         Channel getStickyChannel();
 
         <T> T sticky(Class<T> clientInterface);
+
+        /**
+         * Creates a new {@link StickyChannelSession} which can be used to create multiple
+         * clients bound to the same session.
+         */
+        StickyChannelSession session();
+    }
+
+    /**
+     * Represents a single session of {@link StickyChannelFactory2} where the underlying channel will not change.
+     * This is useful when multiple client interfaces must be bound to use the same host.
+     */
+    public interface StickyChannelSession {
+        /**
+         * Returns a channel which will route all requests to a single host, even if that host returns some 429s.
+         * Each successive call to this method may get a different channel (or it may return the same one).
+         */
+        Channel getStickyChannel();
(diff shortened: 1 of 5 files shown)


dialogue-clients/src/main/java/com/palantir/dialogue/clients/DialogueClients.java
```diff
@@ -128,6 +128,26 @@ public interface StickyChannelFactory2 {
...
+         * clients bound to the same session.
+         */
+        StickyChannelSession session();
+    }
+
+    /**
+     * Represents a single session of {@link StickyChannelFactory2} where the underlying channel will not change.
+     * This is useful when multiple client interfaces must be bound to use the same host.
+     */
+    public interface StickyChannelSession {
```
carterkozak, line 143: Naming: should this interface have a 2 somewhere? Unlike the sticky channels factory, it is meant to be stored for reuse, so there's upside to having a cleaner name. I don't have a strong opinion either way.

dialogue-clients/src/main/java/com/palantir/dialogue/clients/DialogueClients.java
```diff
@@ -128,6 +128,26 @@ public interface StickyChannelFactory2 {
...
+        /**
+         * Returns a channel which will route all requests to a single host, even if that host returns some 429s.
+         * Each successive call to this method may get a different channel (or it may return the same one).
+         */
+        Channel getStickyChannel();
+
+        <T> T sticky(Class<T> clientInterface);
```
carterkozak, line 150: I considered naming these `getChannel` and `create(Class<T> clientInterface)` since the interface name implies sticky, but figured matching the sticky factory would be a bit more obvious for readers

jkozlowski (approved):
I forgot how this code works, I thought it would be harder to knock out. THanks!

dialogue-clients/src/main/java/com/palantir/dialogue/clients/DialogueClients.java
jkozlowski: I think at this point it might be time to remove the previous sticky channel implementation, that was the plan all along. 2+ seems enough to prove out it's correctness? :P

PR #1996, opened 2023-08-10
Attempt to avoid gzip response compression between backend services

carterkozak:
## Before this PR
Gzip compression is expensive.

## After this PR
==COMMIT_MSG==
Avoid gzip response compression between backend services
==COMMIT_MSG==

## Possible downsides?
It's possible that we may stop using gzip response compression somewhere it's helpful, most such environments don't have multiple addressable nodes, though. We do not run single nodes of production services, so I expect this heuristic to be quite accurate.
In ideal world, I might prefer to add metadata to the client request which the server can use to intelligently opt out of compression -- that would be fairly similar to setting the accept-encoding header on the client, and increase complexity of the rollout.

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/ContentDecodingChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/ContentDecodingChannel.java
@@ -55,19 +55,34 @@ final class ContentDecodingChannel implements EndpointChannel {
     private static final String GZIP = "gzip";
 
     private final EndpointChannel delegate;
+    private final boolean sendAcceptGzip;
 
-    ContentDecodingChannel(EndpointChannel delegate) {
+    private ContentDecodingChannel(EndpointChannel delegate, boolean sendAcceptGzip) {
         this.delegate = Preconditions.checkNotNull(delegate, "Channel is required");
+        this.sendAcceptGzip = sendAcceptGzip;
+    }
+
+    static EndpointChannel create(Config cf, EndpointChannel delegate) {
+        boolean sendAcceptGzip = shouldSendAcceptGzip(cf);
+        return new ContentDecodingChannel(delegate, sendAcceptGzip);
+    }
+
+    private static boolean shouldSendAcceptGzip(Config cf) {
+        // In mesh mode or environments which appear to be within an environment,
+        // prefer not to request compressed responses. This heuristic assumes response
+        // compression should not be used in a service mesh, nor when load balancing
+        // is handled by the client.
+        return cf.mesh() == MeshMode.DEFAULT_NO_MESH && cf.clientConf().uris().size() == 1;
     }
 
     @Override
     public ListenableFuture<Response> execute(Request request) {
-        Request augmentedRequest = acceptEncoding(request);
+        Request augmentedRequest = acceptEncoding(request, sendAcceptGzip);
         return DialogueFutures.transform(delegate.execute(augmentedRequest), ContentDecodingChannel::decompress);
     }
 
-    private static Request acceptEncoding(Request request) {
-        if (request.headerParams().containsKey(ACCEPT_ENCODING)) {
+    private static Request acceptEncoding(Request request, boolean sendAcceptGzip) {
+        if (!sendAcceptGzip || request.headerParams().containsKey(ACCEPT_ENCODING)) {
             // Do not replace existing accept-encoding values
             return request;
         }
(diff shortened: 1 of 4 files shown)


dialogue-core/src/main/java/com/palantir/dialogue/core/ContentDecodingChannel.java
```diff
@@ -55,19 +55,34 @@ final class ContentDecodingChannel implements EndpointChannel {
...
+        // prefer not to request compressed responses. This heuristic assumes response
+        // compression should not be used in a service mesh, nor when load balancing
+        // is handled by the client.
+        return cf.mesh() == MeshMode.DEFAULT_NO_MESH && cf.clientConf().uris().size() == 1;
     }
 
     @Override
     public ListenableFuture<Response> execute(Request request) {
-        Request augmentedRequest = acceptEncoding(request);
+        Request augmentedRequest = acceptEncoding(request, sendAcceptGzip);
```
carterkozak, line 86: needs a comment:

In cases where gzip is not expected, we continue to handle gzipped responses to avoid abrupt failures against servers which hard-code `Content-Encoding: gzip` responses without checking request headers.

dialogue-core/src/main/java/com/palantir/dialogue/core/ContentDecodingChannel.java
```diff
@@ -53,21 +54,43 @@ final class ContentDecodingChannel implements EndpointChannel {
     private static final String CONTENT_ENCODING = "content-encoding";
     private static final String CONTENT_LENGTH = "content-length";
     private static final String GZIP = "gzip";
+    private static final String PREFER_COMPRESSED_RESPONSE_TAG = "prefer-compressed-response";
```
carterkozak, line 57: Note that I have opted _not_ to add this to the readme quite yet, so we can understand when and why folks opt into it.

mpritham: Thoughts on making this `@VisibleForTesting` so we can use it in the tests when defining `PreferCompressedResponseEndpoint`?

dialogue-core/src/main/java/com/palantir/dialogue/core/ContentDecodingChannel.java
```diff
@@ -53,21 +54,43 @@ final class ContentDecodingChannel implements EndpointChannel {
...
+        Request augmentedRequest = acceptEncoding(request, sendAcceptGzip);
+        // In cases where gzip is not expected, we continue to handle gzipped responses to avoid abrupt failures
+        // against servers which hard-code 'Content-Encoding: gzip' responses without checking request headers.
         return DialogueFutures.transform(delegate.execute(augmentedRequest), ContentDecodingChannel::decompress);
     }
 
-    private static Request acceptEncoding(Request request) {
-        if (request.headerParams().containsKey(ACCEPT_ENCODING)) {
+    private static Request acceptEncoding(Request request, boolean sendAcceptGzip) {
+        if (!sendAcceptGzip || request.headerParams().containsKey(ACCEPT_ENCODING)) {
```
mpritham, line 93: Seems like `sendAcceptGzip` is now some extra state that `ContentDecodingChannel` holds. Curious what you think of adding it as a param in `toString` like `"ContentDecodingChannel{delegate=" + delegate + " sendAcceptGzip=" + sendAcceptGzip + '}'`?

mpritham (approved):
lgtm! some clarifications.

dialogue-core/src/main/java/com/palantir/dialogue/core/ContentDecodingChannel.java
carterkozak: Good idea, updated!

dialogue-core/src/main/java/com/palantir/dialogue/core/ContentDecodingChannel.java
carterkozak: I'd actually written that initially, but decided I would prefer not to share the reference because it adds friction if the value changes (the test will fail) and make folks consider the implications for consumers before moving forward. In practice I don't think it makes a great deal of difference either way, but that was my line of reasoning :-)

mpritham: > it adds friction if the value changes (the test will fail) and make folks consider the implications for consumers before moving forward

I like this reasoning! Sounds good.

PR #2016, opened 2023-09-13
UrlBuilder supports adding multiple path segments at once

schlosna:
## Before this PR
For high volume services, adding the URL path segments can trigger reallocations as the list resizes.

Calls to non-deprecated endpoints incurred the cost of creating the deprecation metric name.

## After this PR
==COMMIT_MSG==
UrlBuilder supports adding multiple path segments at once.
Deprecation metric is only created when a deprecated endpoint is invoked.
==COMMIT_MSG==

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/DeprecationWarningChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/DeprecationWarningChannel.java
@@ -19,6 +19,7 @@ package com.palantir.dialogue.core;
 import com.codahale.metrics.Meter;
 import com.github.benmanes.caffeine.cache.Cache;
 import com.github.benmanes.caffeine.cache.Caffeine;
+import com.google.common.base.Suppliers;
 import com.google.common.util.concurrent.FutureCallback;
 import com.google.common.util.concurrent.ListenableFuture;
 import com.palantir.dialogue.Endpoint;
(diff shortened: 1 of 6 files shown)


dialogue-core/src/main/java/com/palantir/dialogue/core/DeprecationWarningChannel.java
```diff
@@ -73,19 +75,19 @@ public ListenableFuture<Response> execute(Request request) {
     }
 
     private FutureCallback<Response> createCallback(String channelName, Endpoint endpoint) {
-        Meter meter = metrics.deprecations(endpoint.serviceName());
-
+        // lazily create meter metric name only if deprecated endpoint is accessed
+        Supplier<Meter> meterSupplier = Suppliers.memoize(() -> metrics.deprecations(endpoint.serviceName()));
```
carterkozak, line 79: I imagine this came up with something that used sticky sessions? iirc each sticky session client re-creates endpointchannel instances, which is why we memoize some other metrics.

dialogue-target/src/main/java/com/palantir/dialogue/PathTemplate.java
```diff
@@ -80,7 +80,7 @@ public void fill(ListMultimap<String, String> parameters, UrlBuilder url) {
             if (segment.fixed != null) {
                 url.pathSegment(segment.fixed);
             } else {
-                parameters.get(segment.variable).forEach(url::pathSegment);
+                url.pathSegments(parameters.get(segment.variable));
```
carterkozak, line 83: 👍

carterkozak (approved):
thanks!

dialogue-core/src/main/java/com/palantir/dialogue/core/DeprecationWarningChannel.java
schlosna: yep, I saw this on some of the major clients that use sticky sessions

PR #2020, opened 2023-09-18
dialogue.client.connection.create differentiates success and failure

carterkozak:
This metric is helpful for safely reducing connect timeouts, unfortunately some failures will always result in a timeout, so we need to rule out failed attempts.
==COMMIT_MSG==
dialogue.client.connection.create differentiates success and failure
==COMMIT_MSG==

### DIFF
--- a/dialogue-apache-hc5-client/src/main/metrics/dialogue-apache-client-metrics.yml
+++ b/dialogue-apache-hc5-client/src/main/metrics/dialogue-apache-client-metrics.yml
@@ -55,6 +55,9 @@ namespaces:
           - name: client-name
           - name: client-type
             values: [ apache-hc5 ]
+          - name: result
+            values: [ success, failure ]
+            docs: Describes whether or not a connection was successfully established.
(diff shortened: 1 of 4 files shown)


dialogue-apache-hc5-client/src/main/metrics/dialogue-apache-client-metrics.yml
```diff
@@ -55,6 +55,9 @@ namespaces:
           - name: client-name
           - name: client-type
             values: [ apache-hc5 ]
+          - name: result
+            values: [ success, failure ]
```
amerguy, line 59: do we care about timing failures ?  should we instead just time successes ? failures timing is tied to timeouts

carterkozak: We care about understanding the worst case timing. It depends on the type of problem we're trying to solve at the time, though.

mpritham (approved):

PR #2023, opened 2023-09-28
Add client request size metric channel

andybradshaw:
## Before this PR

Currently, the best way to understand the size of requests made by a particular client is to analyze server request logs. While this does provide some high-level insight into the behavior of a client, it doesn't provide any information about the clients understanding of if a request is "repeatable" or not. This is important as not all retry approaches will have the same opinion on the repeatability of a request (e.g. envoy has a payload size constraint), so we need to know which large requests dialogue currently believes would get retried. 

## After this PR
==COMMIT_MSG==
Add a request size metric channel, which records the size of payloads written by the client. 
==COMMIT_MSG==

## Possible downsides?

This is currently as histogram, but I don't believe we need all of the metrics that would come out of it, and it's likely quite expensive to add this metric to all clients. I think we probably only care about the `p99` or `max` values here, as average request size is largely irrelevant.

### DIFF
--- a/dialogue-clients/metrics.md
+++ b/dialogue-clients/metrics.md
@@ -71,7 +71,11 @@ Dialogue-specific metrics that are not necessarily applicable to other client im
 - `dialogue.client.request.queued.time` tagged `channel-name` (timer): Time spent waiting in the queue before execution.
 - `dialogue.client.request.endpoint.queued.time` tagged `channel-name`, `service-name`, `endpoint` (timer): Time spent waiting in the queue before execution on a specific endpoint due to server QoS.
 - `dialogue.client.request.sticky.queued.time` tagged `channel-name` (timer): Time spent waiting in the sticky queue before execution attempt.
-- `dialogue.client.requests.size` tagged `channel-name`, `service-name`, `endpoint`, `retryable` (histogram): Size of requests
+- `dialogue.client.requests.size` (histogram): Size of requests
+  - `repeatable` values (`true`,`false`)
+  - `channel-name`
+  - `service-name`
+  - `endpoint`
 - `dialogue.client.create` tagged `client-name`, `client-type` (meter): Marked every time a new client is created.
 
 ### dialogue.concurrencylimiter
--- a/dialogue-clients/metrics.md
+++ b/dialogue-clients/metrics.md
@@ -71,6 +71,7 @@ Dialogue-specific metrics that are not necessarily applicable to other client im
 - `dialogue.client.request.queued.time` tagged `channel-name` (timer): Time spent waiting in the queue before execution.
 - `dialogue.client.request.endpoint.queued.time` tagged `channel-name`, `service-name`, `endpoint` (timer): Time spent waiting in the queue before execution on a specific endpoint due to server QoS.
 - `dialogue.client.request.sticky.queued.time` tagged `channel-name` (timer): Time spent waiting in the sticky queue before execution attempt.
+- `dialogue.client.requests.size` tagged `service-name`, `endpoint` (histogram): Size of requests
 - `dialogue.client.create` tagged `client-name`, `client-type` (meter): Marked every time a new client is created.
 
 ### dialogue.concurrencylimiter
--- a/dialogue-clients/metrics.md
+++ b/dialogue-clients/metrics.md
@@ -71,7 +71,7 @@ Dialogue-specific metrics that are not necessarily applicable to other client im
 - `dialogue.client.request.queued.time` tagged `channel-name` (timer): Time spent waiting in the queue before execution.
 - `dialogue.client.request.endpoint.queued.time` tagged `channel-name`, `service-name`, `endpoint` (timer): Time spent waiting in the queue before execution on a specific endpoint due to server QoS.
 - `dialogue.client.request.sticky.queued.time` tagged `channel-name` (timer): Time spent waiting in the sticky queue before execution attempt.
-- `dialogue.client.requests.size` (histogram): Size of requests
+- `dialogue.client.requests.size` (histogram): Histogram of the sizes of requests larger than a threshold (1 MiB).
   - `repeatable` values (`true`,`false`)
   - `channel-name`
   - `service-name`
--- a/dialogue-clients/metrics.md
+++ b/dialogue-clients/metrics.md
@@ -71,7 +71,7 @@ Dialogue-specific metrics that are not necessarily applicable to other client im
 - `dialogue.client.request.queued.time` tagged `channel-name` (timer): Time spent waiting in the queue before execution.
 - `dialogue.client.request.endpoint.queued.time` tagged `channel-name`, `service-name`, `endpoint` (timer): Time spent waiting in the queue before execution on a specific endpoint due to server QoS.
 - `dialogue.client.request.sticky.queued.time` tagged `channel-name` (timer): Time spent waiting in the sticky queue before execution attempt.
-- `dialogue.client.requests.size` tagged `service-name`, `endpoint` (histogram): Size of requests
+- `dialogue.client.requests.size` tagged `channel-name`, `service-name`, `endpoint`, `retryable` (histogram): Size of requests
 - `dialogue.client.create` tagged `client-name`, `client-type` (meter): Marked every time a new client is created.
 
 ### dialogue.concurrencylimiter
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/RequestSizeMetricsChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/RequestSizeMetricsChannel.java
@@ -24,6 +24,8 @@ import com.palantir.dialogue.EndpointChannel;
 import com.palantir.dialogue.Request;
 import com.palantir.dialogue.RequestBody;
 import com.palantir.dialogue.Response;
+import com.palantir.logsafe.logger.SafeLogger;
+import com.palantir.logsafe.logger.SafeLoggerFactory;
 import com.palantir.tritium.metrics.registry.TaggedMetricRegistry;
 import java.io.FilterOutputStream;
 import java.io.IOException;
@@ -31,6 +33,7 @@ import java.io.OutputStream;
 import java.util.Optional;
 
 final class RequestSizeMetricsChannel implements EndpointChannel {
+    private static final SafeLogger log = SafeLoggerFactory.get(RequestSizeMetricsChannel.class);
     private final EndpointChannel delegate;
     private final Histogram requestSize;
 
@@ -102,7 +105,7 @@ final class RequestSizeMetricsChannel implements EndpointChannel {
             try {
                 out.close();
             } catch (IOException e) {
-                throw new RuntimeException(e);
+                log.warn("Failed to close tracking output stream", e);
             }
             delegate.close();
         }
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/RequestSizeMetricsChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/RequestSizeMetricsChannel.java
@@ -1,0 +1,141 @@
+/*
+ * (c) Copyright 2023 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.dialogue.core;
+
+import com.codahale.metrics.Histogram;
+import com.google.common.util.concurrent.ListenableFuture;
+import com.palantir.conjure.java.client.config.ClientConfiguration;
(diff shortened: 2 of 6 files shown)


dialogue-core/src/main/java/com/palantir/dialogue/core/RequestSizeMetricsChannel.java
```diff
@@ -0,0 +1,144 @@
...
+import java.io.IOException;
+import java.io.OutputStream;
+import java.util.Optional;
+
+final class RequestSizeMetricsChannel implements EndpointChannel {
+    private static final SafeLogger log = SafeLoggerFactory.get(RequestSizeMetricsChannel.class);
+    private final EndpointChannel delegate;
+    private final Histogram requestSize;
+
+    static EndpointChannel create(Config cf, EndpointChannel channel, Endpoint endpoint) {
```
carterkozak, line 47: We'll need to wire this up from `DialogueChannel` so that it's part of the chain of channels used to send requests

dialogue-core/src/main/metrics/dialogue-core-metrics.yml
```diff
@@ -56,6 +56,10 @@ namespaces:
         type: timer
         tags: [ channel-name ]
         docs: Time spent waiting in the sticky queue before execution attempt.
+      requests.size:
+        type: histogram
+        tags: [service-name, endpoint]
```
carterkozak: Let's include `channel-name` which is roughly equivalent to the service name in a service discovery block (this shouldn't increase cardinality)

dialogue-core/src/main/java/com/palantir/dialogue/core/RequestSizeMetricsChannel.java
```diff
@@ -0,0 +1,144 @@
...
+        Request augmentedRequest = wrap(request);
+        return delegate.execute(augmentedRequest);
+    }
+
+    private Request wrap(Request request) {
+        Optional<RequestBody> body = request.body();
+        if (body.isEmpty()) {
+            // No need to record empty bodies
+            return request;
+        }
```
carterkozak, line 83: Currently this will record metrics for both requests which can and cannot be retried, without distinction. Should we tag metrics based on whether or not retires are allowed? (we could hold two histograms and pass the correct one to `RequestSizeRecordingRequestBody` based on `repeatable()`).

dialogue-core/src/main/java/com/palantir/dialogue/core/RequestSizeMetricsChannel.java
```diff
@@ -0,0 +1,144 @@
...
+
+        return Request.builder()
+                .from(request)
+                .body(new RequestSizeRecordingRequestBody(body.get(), this.requestSize))
+                .build();
+    }
+
+    private class RequestSizeRecordingRequestBody implements RequestBody {
```
carterkozak: I think this can be static
```suggestion
    private static final class RequestSizeRecordingRequestBody implements RequestBody {
```

dialogue-core/src/main/java/com/palantir/dialogue/core/RequestSizeMetricsChannel.java
```diff
@@ -0,0 +1,144 @@
...
+            // we'll never actually write to this output stream, but is safe to perform all operations on in case a
+            // client closes without calling write.
+            this.out = new SizeTrackingOutputStream(OutputStream.nullOutputStream(), size);
+        }
+
+        @Override
+        public void writeTo(OutputStream output) throws IOException {
+            out = new SizeTrackingOutputStream(output, size);
+            delegate.writeTo(out);
+        }
```
carterkozak: Instead of maintaining a field, what do you think about something along these lines?

```suggestion
        @Override
        public void writeTo(OutputStream output) throws IOException {
            CountingOutputStream countingOut = new CountingOutputStream(out)
            delegate.writeTo(countingOut);
            size.update(countingOut.getCount());
        }
```

This way, if the delegate calls close on `countingOut`, it is passed through to `out`, however we don't require close to be called on the wrapper stream here. Keeps everything nicely contained in this method.

The guava [CountingOutputStream](https://github.com/google/guava/blob/master/guava/src/com/google/common/io/CountingOutputStream.java) handles counting bytes for us :-]

dialogue-core/src/main/java/com/palantir/dialogue/core/RequestSizeMetricsChannel.java
andybradshaw: Yep! wasn't sure if we'd want to do that separately or not, and also wasn't sure where in the chain to add this... maybe alongside the timing endpoint channel?

carterkozak: Hmm, the timing endpoint channel is in a conditional to protect against an edge case where clients may arbitrarily build new `Endpoint` objects (if they're not using standard dialogue generated bindings). I don't want to miss out on that data, so perhaps we can avoid fanout by setting a relatively high threshold for reporting data, something like 1mb+.

We'll need to restructure a bit of the code to avoid creating a histogram until we've seen a large request for the first time (which is impossible-ish on GET endpoints, and many PUTs/POSTs will never reach it) by creating a memoized supplier for the histogram rather than passing the histogram directly.

Setting a threshold improves the probability of capturing maximums as well, since we use a sampling reservoir, we only hold ~1024 samples at a given time (with heavy recency bias), so reducing small and uninteresting samples will increase the probability we capture outliers.

dialogue-core/src/main/java/com/palantir/dialogue/core/RequestSizeMetricsChannel.java
```diff
@@ -0,0 +1,124 @@
...
+        @Override
+        public void writeTo(OutputStream output) throws IOException {
+            CountingOutputStream countingOut = new CountingOutputStream(output);
+            delegate.writeTo(countingOut);
+            if (countingOut.getCount() > MIN_REPORTED_REQUEST_SIZE) {
+                size.get().update(countingOut.getCount());
+            }
+        }
+
+        @Override
```
carterkozak, line 111: We need to override and pass through the `default` interface methods from `RequestBody` (some tests are failing due to the missing `Content-Length` header)

dialogue-core/src/main/metrics/dialogue-core-metrics.yml
```diff
@@ -56,6 +56,10 @@ namespaces:
         type: timer
         tags: [ channel-name ]
         docs: Time spent waiting in the sticky queue before execution attempt.
+      requests.size:
+        type: histogram
+        tags: [channel-name, service-name, endpoint, retryable]
```
carterkozak: Thoughts on updating from `retryable` to `repeatable` matching the RequestBody API?

Defer to your judgement, just mentioning it in case you aren't familiar with teh feature: Metric-schema allows us to enumerate tag values which may improve the generated documentation slightly (and avoids unnecessary string values in code):
```yaml
tags:
  - name: repeatable
     values: ['true', 'false']
```

dialogue-core/src/test/java/com/palantir/dialogue/core/RequestSizeMetricsChannelTest.java
```diff
@@ -0,0 +1,185 @@
...
+                        .channelFactory(factory)
+                        .rawConfig(ClientConfiguration.builder()
+                                .from(TestConfigurations.create("https://foo:10001"))
+                                .taggedMetricRegistry(registry)
+                                .build())
+                        .build(),
+                r -> {
+                    try {
+                        RequestBody body = r.body().get();
+                        body.writeTo(baos);
```
carterkozak: `ByteArrayOutputStream baos` is otherwise unused, we can use the guava `ByteStreams.nullOutputStream()` here instead, to throw away bytes immediately:
```suggestion
                        body.writeTo(ByteStreams.nullOutputStream());
```

dialogue-core/src/test/java/com/palantir/dialogue/core/RequestSizeMetricsChannelTest.java
```diff
@@ -0,0 +1,185 @@
...
+                        body.close();
+                    } catch (IOException e) {
```
carterkozak: We could try/with here:

```java
try (RequestBody body = r.body().get()) {
    body.writeTo(ByteStreams.nullOutputStream());
} catch (/*..etc*/
```

dialogue-core/src/test/java/com/palantir/dialogue/core/RequestSizeMetricsChannelTest.java
```diff
@@ -0,0 +1,185 @@
...
+                        throw new RuntimeException(e);
+                    }
```
carterkozak, line 162: here as well

dialogue-core/src/test/java/com/palantir/dialogue/core/RequestSizeMetricsChannelTest.java
```diff
@@ -0,0 +1,185 @@
...
+                .retryable("true")
+                .build()
+                .getSnapshot();
+        assertThat(snapshot.size()).isEqualTo(1);
+        assertThat(snapshot.get99thPercentile()).isEqualTo(expected.length);
+    }
+
+    @Test
+    public void small_request_not_recorded() throws ExecutionException, InterruptedException {
+        TaggedMetricRegistry registry = DefaultTaggedMetricRegistry.getDefault();
```
carterkozak: Let's avoid the singleton instance, this way our test is less likely to be impacted by any other tests in the system:
```suggestion
        TaggedMetricRegistry registry = new DefaultTaggedMetricRegistry();
```

dialogue-core/src/test/java/com/palantir/dialogue/core/RequestSizeMetricsChannelTest.java
```diff
@@ -0,0 +1,185 @@
...
+import org.mockito.junit.jupiter.MockitoExtension;
+
+@ExtendWith(MockitoExtension.class)
+public class RequestSizeMetricsChannelTest {
+    @Mock
+    DialogueChannelFactory factory;
+
+    @Test
+    public void records_request_size_metrics() throws ExecutionException, InterruptedException {
+        TaggedMetricRegistry registry = DefaultTaggedMetricRegistry.getDefault();
```
carterkozak: `new DefaultTaggedMetricRegistry();` here as well

dialogue-core/src/test/java/com/palantir/dialogue/core/RequestSizeMetricsChannelTest.java
```diff
@@ -0,0 +1,185 @@
...
+import java.util.concurrent.ExecutionException;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+
+@ExtendWith(MockitoExtension.class)
+public class RequestSizeMetricsChannelTest {
+    @Mock
+    DialogueChannelFactory factory;
```
carterkozak: I'd avoid using mockito here since I don't think we're directly interacting with the factory -- this will simplify test execution a bit.

```diff
-@ExtendWith(MockitoExtension.class)
public class RequestSizeMetricsChannelTest {
-    @Mock
-    DialogueChannelFactory factory;
+    private static final DialogueChannelFactory STUB_FACTORY = _ignored -> {
+                throw new AssertionError("DialogueChannelFactory should not be used");
+            };
```

dialogue-core/src/test/java/com/palantir/dialogue/core/RequestSizeMetricsChannelTest.java
```diff
@@ -0,0 +1,181 @@
...
+import java.io.IOException;
+import java.io.OutputStream;
+import java.nio.charset.StandardCharsets;
+import java.util.OptionalLong;
+import java.util.concurrent.ExecutionException;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.junit.jupiter.MockitoExtension;
+
+@ExtendWith(MockitoExtension.class)
```
carterkozak: I believe the extension is no longer necessary because we have no fields annotated with `@Mock`
```suggestion

```

dialogue-clients/metrics.md
```diff
@@ -71,6 +71,11 @@ Dialogue-specific metrics that are not necessarily applicable to other client im
 - `dialogue.client.request.queued.time` tagged `channel-name` (timer): Time spent waiting in the queue before execution.
 - `dialogue.client.request.endpoint.queued.time` tagged `channel-name`, `service-name`, `endpoint` (timer): Time spent waiting in the queue before execution on a specific endpoint due to server QoS.
 - `dialogue.client.request.sticky.queued.time` tagged `channel-name` (timer): Time spent waiting in the sticky queue before execution attempt.
+- `dialogue.client.requests.size` (histogram): Size of requests
```
carterkozak: We may want to describe the reporting size threshold in this metric description (though I suppose it doesn't claim to report the size of every request ;-)

Can update the the metric-schema/docs later if needs be.

carterkozak (approved):
Great stuff, thanks for the contribution! :-)

dialogue-clients/metrics.md
andybradshaw: Updated! Yeah, good to call this out... might even be worth renaming the metric to `large_requests` or something to explicitly call out that there's some distinction/threshold that's used for it.

carterkozak: I think the current name gives us flexibility if we want to change thresholds later on, I don't mind it as is if you're happy with it. Thanks for updating!

PR #2026, opened 2023-10-03
Memoize some sticky channel metrics

schlosna:
## Before this PR

Allocations in constructors for some metrics that may not be used, so lazily create and memoize them as needed:

image

## After this PR
==COMMIT_MSG==
Memoize some sticky channel metrics
==COMMIT_MSG==

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/TimingEndpointChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/TimingEndpointChannel.java
@@ -18,6 +18,7 @@ package com.palantir.dialogue.core;
 
 import com.codahale.metrics.Timer;
 import com.github.benmanes.caffeine.cache.Ticker;
+import com.google.common.base.Suppliers;
 import com.google.common.util.concurrent.FutureCallback;
 import com.google.common.util.concurrent.ListenableFuture;
 import com.google.common.util.concurrent.RateLimiter;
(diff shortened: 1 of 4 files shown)


dialogue-core/src/main/java/com/palantir/dialogue/core/TimingEndpointChannel.java
```diff
@@ -52,18 +54,18 @@ final class TimingEndpointChannel implements EndpointChannel {
...
-                .build();
-        this.failureTimer = metrics.response()
+                .build());
+        this.failureTimer = Suppliers.memoize(() -> metrics.response()
                 .channelName(channelName)
                 .serviceName(endpoint.serviceName())
                 .endpoint(endpoint.endpointName())
                 .status("failure")
-                .build();
+                .build());
```
carterkozak, line 68: I wonder if we should provide a lazy TaggedMetricRegistry implementation which registers non-gauge metrics the first time they're updated? Probably not, since the metric implementations aren't interfaces but rather concrete implementations.

carterkozak:
This looks reasonable to me. I didn't review the code reorganization in DialogueChannel very closely because this PR is still in draft.

dialogue-core/src/main/java/com/palantir/dialogue/core/TimingEndpointChannel.java
schlosna: I've debated adding a cache in generated metric-schema `*Metrics` instances to reduce the frequency of `MetricName` creation, though most places can store a reference to the metric itself, but it would help these sticky channels.

carterkozak (approved):

PR #2053, opened 2023-10-31
Introduce dialogue metrics for individual Connect attempts

carterkozak:
These are aimed at safely reducing our connect timeout defaults.

## Before this PR
Only full connection time metrics which include connect+handshake across all IP addresses associated with the requested hostname.

## After this PR
==COMMIT_MSG==
Introduce dialogue metrics for individual Connect attempts
==COMMIT_MSG==

## Possible downsides?
There's an unfortunate amount of code duplication required to instrument this. I'll chat with the project httpcomponents folks about possibly splitting `socket.connect` into a separate `protected` method which can be overridden.

### DIFF
--- a/dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/InstrumentedSslConnectionSocketFactory.java
+++ b/dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/InstrumentedSslConnectionSocketFactory.java
@@ -1,0 +1,206 @@
+/*
+ * (c) Copyright 2023 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.dialogue.hc5;
+
+import com.codahale.metrics.Timer;
+import com.palantir.dialogue.hc5.DialogueClientMetrics.ConnectionConnect_Result;
+import com.palantir.logsafe.Preconditions;
+import com.palantir.logsafe.SafeArg;
+import com.palantir.logsafe.UnsafeArg;
+import com.palantir.logsafe.exceptions.SafeIoException;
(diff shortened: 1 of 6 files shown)


dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/InstrumentedSslConnectionSocketFactory.java
```diff
@@ -0,0 +1,206 @@
...
+                log.debug(
+                        "Connecting socket to {} ({}) with timeout {}",
+                        SafeArg.of("clientName", clientName),
+                        UnsafeArg.of("remoteAddress", remoteAddress),
+                        SafeArg.of("connectTimeout", connectTimeout));
+            }
+            // Run this under a doPrivileged to support lib users that run under a SecurityManager this allows granting
+            // connect permissions only to this library
+            try {
+                AccessController.doPrivileged((PrivilegedExceptionAction<Object>) () -> {
```
mpritham: nit: since we don't use the returned value, what are your thoughts on

```suggestion
                AccessController.doPrivileged((PrivilegedExceptionAction<Void>) () -> {
```

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/InstrumentedSslConnectionSocketFactory.java
```diff
@@ -0,0 +1,206 @@
...
+                    doConnect(
+                            sock,
+                            remoteAddress,
+                            Timeout.defaultsToDisabled(connectTimeout).toMillisecondsIntBound());
+                    return null;
+                });
+            } catch (PrivilegedActionException e) {
+                Throwable cause = e.getCause();
+                if (cause instanceof IOException) {
```
mpritham, line 139: Clarification: would we always want to throw our `SafeIoException` with the additional clientName info?

carterkozak: nope -- the original exception type (as a subtype of IOException, like socketexception, etc) is meaningful

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/InstrumentedSslConnectionSocketFactory.java
carterkozak: :+1:

mpritham (approved):

carterkozak:
We should be able to simplify this instrumentation later on assuming the following makes it into httpcomponents-client 5.3: https://issues.apache.org/jira/browse/HTTPCLIENT-2305

carterkozak:
Proposed change has merged, so we can simplify things a fair bit once httpcomponents-client 5.3 is released :-)

PR #2067, opened 2023-11-17
Add native support for InputStream responses to dialogue annotations processor

aldexis:
## Before this PR
The dialogue annotations processor does not support InputStream responses natively and instead requires using a custom deserializer, which is a shame since the [BodySerDe already provides a default inputStreamDeserializer](https://github.com/palantir/dialogue/blob/85bb924ae5773397bb8248198d7a0e3c943b3946/dialogue-target/src/main/java/com/palantir/dialogue/BodySerDe.java#L36-L42).

## After this PR
==COMMIT_MSG==
Add native support for InputStream responses to dialogue annotations processor
==COMMIT_MSG==

Note that this also keeps using the provided custom deserializer, if any, to avoid unintentionally breaking existing implementations that may rely on possibly different setups.

## Possible downsides?
This introduces a slight change in that the Void response now also favors the custom deserializer when it would previously ignore it. I consider this is desired as we shouldn't ignore a custom deserializer that is explicitly provided. I would also expect that nobody ever bothered providing a deserializer for Void (and am not sure there would be much of a point)


dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/generate/ServiceImplementationGenerator.java
```diff
@@ -166,21 +167,34 @@ private Optional<FieldSpec> deserializer(ReturnType type) {
...
+
+        // If no custom deserializer is used, use the runtime-provided deserializers for specific types
+        if (!type.isUsingCustomDeserializer()) {
+            if (type.isVoid()) {
+                deserializer = CodeBlock.of(
+                        "new $T($L.bodySerDe().emptyBodyDeserializer(), new $T())",
+                        ErrorHandlingVoidDeserializer.class,
+                        serviceDefinition.conjureRuntimeArgName(),
+                        errorDecoderType);
+            } else if (type.isInputStream()) {
```
carterkozak: should we handle `Optional<InputStream>` as well? https://github.com/palantir/dialogue/blob/85bb924ae5773397bb8248198d7a0e3c943b3946/dialogue-target/src/main/java/com/palantir/dialogue/BodySerDe.java#L44-L45

dialogue-annotations-processor/src/test/java/com/palantir/myservice/service/MyService.java
```diff
@@ -47,6 +48,12 @@ String greet(
     @Request(method = HttpMethod.GET, path = "/greeting", accept = CustomStringDeserializer.class)
     ListenableFuture<String> getGreetingAsync();
 
+    @Request(method = HttpMethod.GET, path = "/input-stream")
+    InputStream inputStream();
+
+    @Request(method = HttpMethod.GET, path = "/input-stream-custom", accept = CustomInputStreamDeserializer.class)
+    InputStream customInputStream();
```
carterkozak: I think we're missing some test coverage for these, I don't believe the ErrorHandlingDeserializer will do what we want because unlike `ErrorHandlingDeserializerFactory`, the `ErrorHandlingDeserializer` closes responses in all cases before returning the result from the delegate deserializer.

dialogue-annotations/src/main/java/com/palantir/dialogue/annotations/ErrorHandlingDeserializer.java
```diff
@@ -0,0 +1,49 @@
...
+ */
+
+package com.palantir.dialogue.annotations;
+
+import com.palantir.dialogue.Deserializer;
+import com.palantir.dialogue.Response;
+import com.palantir.logsafe.Preconditions;
+import java.util.Optional;
+
+public class ErrorHandlingDeserializer<T> implements Deserializer<T> {
```
carterkozak: This should be final and based on the `ErrorHandlingDeserializerFactory` `Deserizlizer` implementation instead

dialogue-annotations-processor/src/test/java/com/palantir/myservice/service/MyService.java
aldexis: Indeed - I had pattern-matched from ErrorHandlingVoidSerializer but realize that this should behave differently. I'm not quite sure how I'd actually test the behavior here though. Would you have recommendations?

Also, I'd be inclined to remove `ErrorHandlingVoidDeserializer` as well in this case and make them all based on `ErrorHandlingDeserializerFactory`. Any objection to my doing so?

dialogue-annotations/src/main/java/com/palantir/dialogue/annotations/ErrorHandlingDeserializer.java
aldexis: Since I'm trying to pass the BodySerDe's deserializers, would you recommend making a custom `DeserializerFactory` class that just returns the BodySerDe ones? Should I maybe make DeserializerFactory into a functional interface and pass BodySerDe#deserializer? Or would that not work for InputStream and Void? (which would seem a bit counter-intuitive to me but seems plausible based on the ConjureBodySerde code)

dialogue-annotations-processor/src/test/java/com/palantir/myservice/service/MyService.java
carterkozak: Testing shouldn't be too bad, using the generated client and attempting to read a byte from the returned InputStream would be sufficient. We expect to read a byte, but if the read fails because the stream is closed, we know we have a bug.

> Also, I'd be inclined to remove `ErrorHandlingVoidDeserializer`

I don't think we can do this because `ErrorHandlingVoidDeserializer` is used in existing generated code, removal would cause runtime failures on transitive dependencies!

dialogue-annotations/src/main/java/com/palantir/dialogue/annotations/ErrorHandlingDeserializer.java
carterkozak: I think we should follow the pattern used by `ResponseDeserializer`, define an InputStreamDeserialzier + OptionalInputStreamDeserialzier which extend StdDeserializer, and use those directly from codegen

carterkozak: Let's ignore OptionalInputStreamDeserialzier initially as well, since that will require some more complex reflection on `Optional<InputStream>` to prevent the error handling deserializer from closing the response prematurely. We can handle that in a separate PR

dialogue-annotations-processor/src/test/java/com/palantir/myservice/service/MyService.java
aldexis: Oh good catch on the ErrorHandlingVoidDeserializer! That's a bit sad 😅 

I'll see if I can implement a test 🤞

dialogue-annotations/src/main/java/com/palantir/dialogue/annotations/ErrorHandlingDeserializer.java
aldexis: Do you mean that we should _not_ be using the deserializers from BodySerDe? Or could I instead pass the runtime as a parameter to the deserializer constructor in the generated code?
I personally like the idea of reusing the already defined deserializers since I assume they have been implemented for a reason :)

dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/data/ReturnTypesResolver.java
```diff
@@ -82,15 +83,29 @@ private Optional<TypeName> orDefaultDeserializerFactory(
...
     private boolean isResponseType(TypeMirror type) {
         return context.isSameTypes(type, Response.class);
     }
 
+    private boolean isInputStreamType(TypeMirror type) {
+        return context.isSameTypes(type, InputStream.class);
+    }
+
+    private boolean usesBodySerde(TypeMirror type) {
+        return context.isSameTypes(type, InputStreamDeserializer.class);
```
aldexis: I tried to have something a bit more generic by making a `BodySerDeBasedDeserializer<T>` class that `InputStreamDeserializer` extends and use `context.isAssignable` here, but that didn't seem to work.

I haven't figured out a nicer way to make it possible to rely on the BodySerDe's existing InputStreamDeserializer either

dialogue-annotations/src/main/java/com/palantir/dialogue/annotations/ErrorHandlingDeserializer.java
aldexis: I tried an approach - let me know what you think of it :)

carterkozak: I like the idea of passing through the BodySerDe (or even full runtime reference), but I'd prefer to investigate that change separately from this one -- I think we want this implementation to match `Json.java` which handles the default json serialization structure: https://github.com/palantir/dialogue/blob/60bfc2aca0cd6b2054cfe1887b62e20c3d01c9c6/dialogue-annotations/src/main/java/com/palantir/dialogue/annotations/Json.java#L43-L50

aldexis: Ah got it! Let me look into that and push something through! Would you prefer I extract the `DEFAULT_BODY_SERDE` as a constant or just create two of them? (I'm going to go for a shared constant unless you think it's a bad idea)

aldexis:
@carterkozak Thanks for the review (and sorry for the delay - was busy last week). This should be ready for a re-review!

dialogue-annotations/src/main/java/com/palantir/dialogue/annotations/Constants.java
```diff
@@ -0,0 +1,29 @@
...
+import com.palantir.conjure.java.dialogue.serde.Encodings;
+import com.palantir.dialogue.BodySerDe;
+
+public final class Constants {
+
+    public static final BodySerDe DEFAULT_BODY_SERDE =
+            DefaultConjureRuntime.builder().encodings(Encodings.json()).build().bodySerDe();
+
+    private Constants() {}
+}
```
carterkozak: This shouldn't be public, I'd have probably constructed another `DefaultConjureRuntime.builder().encodings(Encodings.json()).build().bodySerDe()` within `InputStreamDeserializer` since it's not terribly expensive to create, and doing so is a one-time cost at class initialization time. Sharing one instance via package private access also works, but the naming is tough -- this isn't really a constant, more of a singleton.

carterkozak:
aside from the public `Constants` class, this looks good to me -- thanks!

dialogue-annotations/src/main/java/com/palantir/dialogue/annotations/Constants.java
aldexis: Out of curiosity, why do we want to keep such internals package-private? (I realize the rest does it and I should have thought of it, but I'm not sure I completely understand the reason for it)

I initially moved it back to private fields, but remembered why I initially went with a shared class, which is that this ensures we use the same parameters/setup for both of these. This way, if we want to push some change for one of them, we don't need to remember to find and update the others too, bringing us better consistency in the implementation.

aldexis:
Should be updated now @carterkozak!

dialogue-annotations/src/main/java/com/palantir/dialogue/annotations/Constants.java
carterkozak: > Out of curiosity, why do we want to keep such internals package-private?

The less we expose to consumers, the greater freedom we retain to change and improve the implementation without a long, expensive migration campaign. If a custom deserializer implementation wanted to reference this class, we'd prefer they used the standard API that we're using here (`DefaultConjureRuntime.builder().//etc`) because we're already locked into supporting that one, so it doesn't expand the protected api surface.

Also, it's not entirely clear that a change to one of them should necessarily impact the other, there are lots of types of changes which would impact jackson ser/de that wouldn't have any impact on binary streaming, however it is possible we could get clever and break streaming if we aren't careful.

aldexis: Both of these make sense! Would you rather I move back to private fields in this case? I'm happy to do so as I just need to revert the last commit :)

carterkozak: I think either way is fine :-] Happy to merge as is

carterkozak (approved):

dialogue-annotations/src/main/java/com/palantir/dialogue/annotations/Constants.java
aldexis: Sounds great, thank you for the iterations and sharing your knowledge!

PR #2076, opened 2023-12-08
Support enum types including optional & list enums

skedida:
## Before this PR

Conjure-style enums are not supported in particular for Dialogue annotations. This would result in an error like the following

```
Parameter '<name>' with custom type '<EnumClassName>' must declare an encoder
```
## After this PR
==COMMIT_MSG==
Support enum parameter types
==COMMIT_MSG==

Implemented a similar mechanism as https://github.com/palantir/dialogue/pull/1782 to support the following:
- `MyEnum` type
- `Optional<MyEnum>` type
- `List<MyEnum>` type and `List<MyAlias>` type

### DIFF
--- a/dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/data/ArgumentTypesResolver.java
+++ b/dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/data/ArgumentTypesResolver.java
@@ -74,6 +74,10 @@ public final class ArgumentTypesResolver {
     public ArgumentType getArgumentType(VariableElement param) {
         TypeMirror typeMirror = param.asType();
 
+        return getArgumentType(typeMirror);
+    }
+
+    private ArgumentType getArgumentType(TypeMirror typeMirror) {
         return getPrimitiveType(typeMirror)
                 .or(() -> getListType(typeMirror))
                 .or(() -> getOptionalType(typeMirror))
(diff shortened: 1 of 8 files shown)


pkoenig10 (changes requested):
This feels rather dangerous to me. I'm not sure it's reasonable to assume that all enums should be serialized using `toString` by default. It feels very easy for someone to just add an enum parameter without being aware of how it is actually serialized.

I think there's enough potential for footguns here where I'd prefer this to be explicitly opt-in by declaring a encoder. It should be pretty easy to define an encoder that works for any enum type (although I'm not convinced that something like that should be provided by this library).

dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/data/ArgumentTypesResolver.java
```diff
@@ -177,6 +188,30 @@ private Optional<ArgumentType> getAliasType(TypeMirror typeMirror) {
                 .map(methodName -> ArgumentTypes.alias(typeName, methodName));
     }
 
+    private Optional<ArgumentType> getEnumType(TypeMirror typeMirror) {
+        TypeName typeName = TypeName.get(typeMirror);
+
+        return context.maybeAsDeclaredType(typeMirror).stream()
```
skedida, line 194: @pkoenig10 you are right. I think all enums being serialized like that has foot guns. I wanted to bring to the table whether we can mitigate that risk since my hope is to minimize manual encoding as much as possible.

(Side note: for List<Alias>, I created a separate PR to allow support for that: https://github.com/palantir/dialogue/pull/2079)

For enums, one approach we can take is limiting to only allowing enums that adhere to the following structure:

```
class MyEnum {
    
    Value get() { ... }

    enum Value {
        VALUE_1,
        VALUE_2
    }
}
```

This would require: 
1) An enclosing class such as `MyEnum` that has a `get()` field that returns a `Value` enum
2) The `Value` enum has a constructor that doesn't take any args. I.e. if the enum is `VALUE_1("something")`, don't serialize it
3) `toString()` method is not overridden

If these are not satisfied, then we can require a custom deserializer.  This doesn't eliminate the risk but we can minimize it. But it depends on the risk tolerance here. In addition, note that even with aliases, there's some small risk we are already taking since we assume we can serialize `MyAlias.get()` and ignore other fields in `MyAlias` - someone could technically have other fields that affect the serialization.

Thoughts on whether that is acceptable or any ideas?

pkoenig10: Can you explain what you trying to accomplish with this change? It's not clear to me what problem we're trying to solve here.

PR #2079, opened 2023-12-11
Support list<alias> in annotations

skedida:
## Before this PR

Query and header parameters of `list<SomeAliasType>` were not supported.

## After this PR
==COMMIT_MSG==
Support list<Alias> in annotations
==COMMIT_MSG==

- Add support for `List<Alias>`


dialogue-annotations-processor/src/test/resources/com/palantir/myservice/service/MyServiceDialogueServiceFactory.java.generated
```diff
@@ -194,11 +191,9 @@ public final class MyServiceDialogueServiceFactory implements DialogueServiceFac
                 if (header5.isPresent()) {
                     _request.putHeaderParams("h5", _parameterSerializer.serializeDouble(header5.getAsDouble()));
                 }
-                _request.putAllHeaderParams(
-                        "h6",
-                        header6.stream()
-                                .map(_parameterSerializer::serializeString)
-                                .collect(Collectors.toList()));
```
pkoenig10: nit: Can we avoid changing this from `collect` to `forEach` in this PR? It's probably not a meaningful difference in practice, but it's probably marginally better to put parameters in bulk. If we have reason to believe that switching to `forEach` would be an improvement, let's make that change in a separate PR with the requisite justification/benchmarks/etc.

dialogue-annotations-example/src/main/java/com/palantir/myservice/example/MyService.java
```diff
@@ -76,13 +76,15 @@ void params(
             @Request.QueryParam("q3") Optional<String> query3,
             // Alias types are supported for @QueryParam and @Header
             @Request.QueryParam("q4") MyAliasType query4,
+            @Request.QueryParam("q5") List<MyAliasType> query5,
```
pkoenig10, line 79: We should add tests here that produce the generated code for this case.

https://github.com/palantir/dialogue/blob/bfa0a4b5bd70beb12c18004b2f3fb0a89bbf82a8/dialogue-annotations-processor/src/test/java/com/palantir/myservice/service/MyService.java

pkoenig10:
Prior context in https://github.com/palantir/dialogue/pull/1782#discussion_r948566913.

It's reasonable to support this, I just didn't take the time to do this when initially adding support for alias types.

dialogue-annotations-example/src/main/java/com/palantir/myservice/example/MyService.java
skedida: I updated it

dialogue-annotations-processor/src/test/resources/com/palantir/myservice/service/MyServiceDialogueServiceFactory.java.generated
skedida: Good point. I mostly did it because it was not meaningfully different and because conjure-java uses a for loop & a `put` instead of `putAll` https://github.com/palantir/conjure-java/blob/0e8ad1430736965cd4c926ad6bcd1ff25dd06e54/conjure-java-core/src/main/java/com/palantir/conjure/java/services/dialogue/DefaultStaticFactoryMethodGenerator.java#L380-L388

I update the code to be `putAll` instead

dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/data/ArgumentTypesResolver.java
```diff
@@ -97,10 +97,17 @@ private Optional<ArgumentType> getPrimitiveType(TypeMirror typeMirror) {
 
     private Optional<ArgumentType> getListType(TypeMirror typeMirror) {
         TypeName typeName = TypeName.get(typeMirror);
+        Optional<TypeMirror> innerTypeMirror = context.getGenericInnerType(List.class, typeMirror);
 
-        return context.getGenericInnerType(List.class, typeMirror)
-                .flatMap(this::getPrimitiveSerializerMethodName)
-                .map(methodName -> ArgumentTypes.list(typeName, methodName));
+        return innerTypeMirror
```
pkoenig10: nit: Inline the `getGenericInnerType` call - `innerTypeMirror` is only used here

```suggestion
        return context.getGenericInnerType(List.class, typeMirror)
```

dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/generate/ServiceImplementationGenerator.java
```diff
@@ -280,27 +281,22 @@ private CodeBlock generatePlainSerializer(
...
                 return maybeParameterEncoderType.map(this::parameterEncoderType).orElseGet(() -> {
                     return CodeBlock.of(
-                            "$L.$L($S, $L.$L($L));",
-                            REQUEST,
-                            singleValueMethod,
-                            key,
-                            PARAMETER_SERIALIZER,
-                            parameterSerializerMethodName,
-                            argName);
+                            "$L.$L($S, $L);", REQUEST, singleValueMethod, key, generateSerializerCall(argName, type));
```
pkoenig10: This `generateSerializerCall` is weird. We're already matching on `type` above, so it's unusual to pass the full `type` into a helper method where we match on it again.

This feels like too much DRY - let's keep this as it was.

dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/data/ArgumentTypesResolver.java
skedida: Fixed

dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/generate/ServiceImplementationGenerator.java
skedida: Fixed

dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/generate/ServiceImplementationGenerator.java
```diff
@@ -393,4 +395,38 @@ public CodeBlock multimapParam() {
             }
         });
     }
+
+    private CodeBlock generateListElementSerializerCall(CodeBlock argName, ArgumentType type) {
```
pkoenig10:
```suggestion
    private CodeBlock generateListElementSerializerCall(CodeBlock elementName, ArgumentType type) {
```

dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/data/ArgumentTypesResolver.java
```diff
@@ -97,10 +97,15 @@ private Optional<ArgumentType> getPrimitiveType(TypeMirror typeMirror) {
...
         return context.getGenericInnerType(List.class, typeMirror)
-                .flatMap(this::getPrimitiveSerializerMethodName)
-                .map(methodName -> ArgumentTypes.list(typeName, methodName));
+                .flatMap(this::getListInnerType)
+                .map(argumentType -> ArgumentTypes.list(
+                        typeName,
+                        ImmutableListType.builder().innerType(argumentType).build()));
+    }
+
+    private Optional<ArgumentType> getListInnerType(TypeMirror typeMirror) {
```
pkoenig10: nit: Inline this - it's only used in one place and inlining is consistent with the design of `getOptionalType` below

dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/generate/ServiceImplementationGenerator.java
```diff
@@ -294,13 +295,14 @@ public CodeBlock primitive(TypeName _typeName, String parameterSerializerMethodN
             }
 
             @Override
-            public CodeBlock list(TypeName _typeName, String parameterSerializerMethodName) {
+            public CodeBlock list(TypeName _typeName, ListType listType) {
                 return maybeParameterEncoderType.map(this::parameterEncoderType).orElseGet(() -> {
+                    CodeBlock elementName = CodeBlock.of("$L$L", argName, "Element");
                     CodeBlock asList = CodeBlock.of(
-                            "$L.stream().map($L::$L).collect($T.toList())",
+                            "$L.stream().map($L -> $L).collect($T.toList())",
```
pkoenig10: It might be better to make this a series of maps:
- It allows us to use method references which (in theory) allows to compiler to avoid allocation
- It avoids the need to define a variable here, which avoid potential name conflicts

```suggestion
                            "$L.stream()$L.collect($T.toList())",
```

pkoenig10 (dismissed):
Thanks!

skedida:
@pkoenig10 the review got dismissed because i merged (i think), I should have rebased. Could you reapprove?

dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/generate/ServiceImplementationGenerator.java
```diff
@@ -393,4 +394,51 @@ public CodeBlock multimapParam() {
...
     }
+
+    private CodeBlock generateListElementSerializerCall(CodeBlock argName, ArgumentType type) {
+        CodeBlock elementName = CodeBlock.of("$L$L", argName, "Element");
+        return type.match(new ArgumentType.Cases<>() {
+            @Override
+            public CodeBlock primitive(TypeName _typeName, String parameterSerializerMethodName) {
+                return CodeBlock.of(
+                        ".$L($L -> $L.$L($L))",
+                        MAP,
```
carterkozak: A couple minor nits: I think we're better off inlining the `.map(` literal for readability, and removing the constant. Javapoet prefers to use `$N` for names ([docs](https://github.com/square/javapoet?tab=readme-ov-file#n-for-names)). Positional arguments may be helpful to reference the same component twice (certainly not always something we want to do, as it can harm readability) [docs](https://github.com/square/javapoet?tab=readme-ov-file#positional-arguments).

```suggestion
                        ".map($1N -> $2N.$3N($1N))",
```

carterkozak: Isn't this equivalent to `".map($2N::$3N)"`? Using a method reference is a bit easier to read when possible

dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/generate/ServiceImplementationGenerator.java
```diff
@@ -393,4 +394,51 @@ public CodeBlock multimapParam() {
...
+
+            @Override
+            public CodeBlock alias(TypeName _typeName, String parameterSerializerMethodName) {
+                return CodeBlock.of(
+                        ".$L($L -> $L.$L($L.get()))",
+                        MAP,
+                        elementName,
+                        PARAMETER_SERIALIZER,
+                        parameterSerializerMethodName,
+                        elementName);
```
carterkozak: Thoughts on splitting this into two maps on method references instead of a single complex lambda? I thought that's what @pkoenig10 was asking in https://github.com/palantir/dialogue/pull/2079#discussion_r1424348673

`.map(AliasType::get).map(_parameterSerializer::parameterSerializerMethodName)`

dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/generate/ServiceImplementationGenerator.java
skedida: Fixed

dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/generate/ServiceImplementationGenerator.java
skedida: Fixed

dialogue-annotations-processor/src/test/resources/com/palantir/myservice/service/MyServiceDialogueServiceFactory.java.generated
```diff
@@ -208,6 +216,12 @@ public final class MyServiceDialogueServiceFactory implements DialogueServiceFac
                 if (header11.isPresent()) {
                     _request.putAllHeaderParams("h11", paramsHeader11Encoder.toParamValues(header11.get()));
                 }
+                _request.putAllHeaderParams(
+                        "h12",
+                        header12.stream()
+                                .map(MyAliasType::get)
+                                .map(_parameterSerializer::serializeString)
+                                .collect(Collectors.toList()));
```
carterkozak, line 224: 👍 nice, I like the smaller diff here -- thanks!

carterkozak (approved):

PR #2093, opened 2024-01-03
Add reason to propagated QoS errors

pkoenig10:
When QoS exceptions are propagated, it would be helpful for the reason to indicate the source of the `QosException`.

### DIFF
--- a/dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ErrorDecoder.java
+++ b/dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ErrorDecoder.java
@@ -17,11 +17,13 @@
 package com.palantir.conjure.java.dialogue.serde;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.common.annotations.VisibleForTesting;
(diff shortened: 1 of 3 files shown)


dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ErrorDecoder.java
```diff
@@ -57,6 +59,9 @@ public enum ErrorDecoder {
     private static final SafeLogger log = SafeLoggerFactory.get(ErrorDecoder.class);
     private static final ObjectMapper MAPPER = ObjectMappers.newClientObjectMapper();
 
+    @VisibleForTesting
+    static final QosReason QOS_REASON = QosReason.of("client-qos-response");
```
pkoenig10, line 63: Wasn't sure what to use for the reason - I'm not particularly in favor of this name and open to suggestions.

carterkozak: I don't have anything better in mind, we can likely change this later without breaking folks.

carterkozak (approved):

PR #2094, opened 2024-01-03
Simplify InstrumentedSslConnectionSocketFactory

carterkozak:
==COMMIT_MSG==
Simplify InstrumentedSslConnectionSocketFactory
==COMMIT_MSG==

This takes advantage of an upstream change we made to avoid duplicating logic, which risked falling behind upstream bug fixes:
https://github.com/apache/httpcomponents-client/pull/499

### DIFF
--- a/dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/InstrumentedSslConnectionSocketFactory.java
+++ b/dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/InstrumentedSslConnectionSocketFactory.java
@@ -18,54 +18,28 @@ package com.palantir.dialogue.hc5;
 
 import com.codahale.metrics.Timer;
 import com.palantir.dialogue.hc5.DialogueClientMetrics.ConnectionConnect_Result;
-import com.palantir.logsafe.Preconditions;
-import com.palantir.logsafe.SafeArg;
-import com.palantir.logsafe.UnsafeArg;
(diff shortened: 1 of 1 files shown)


dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/InstrumentedSslConnectionSocketFactory.java
```diff
@@ -101,106 +72,21 @@ public Socket createSocket(HttpContext _context) {
...
-            sslSocket.setEnabledCipherSuites(supportedCipherSuites);
-        } else {
-            sslSocket.setEnabledCipherSuites(TlsCiphers.excludeWeak(sslSocket.getEnabledCipherSuites()));
-        }
-        Timeout handshakeTimeout = tlsConfig.getHandshakeTimeout();
-        if (handshakeTimeout != null) {
-            sslSocket.setSoTimeout(handshakeTimeout.toMillisecondsIntBound());
-        }
-
-        prepareSocket(sslSocket);
```
carterkozak, line 187: Example of how this can be dangerous -- this `prepareSocket` method was deprecated in the weeks since we added this instrumented implementation in favor of an overload with an additional `HttpContext context` parameter. While not breaking, that sort of change can cause us to miss out on bugfixes and new feature support when we replace logic as opposed to wrapping it as this PR does.

bjlaub (approved):

PR #2114, opened 2024-01-24
Handle ETIMEDOUT SocketExceptions

carterkozak:
==COMMIT_MSG==
Handle ETIMEDOUT SocketExceptions
==COMMIT_MSG==

## Possible downsides?
Unfortunately we don't have great ways to test this.
It's possible that a similar failure may manifest from connection establishment, rather than the read path, in which case a retry may be warranted, however I suspect that would manifest differently.

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
@@ -445,6 +445,11 @@ final class RetryingChannel implements EndpointChannel {
         }
     }
 
+    /**
+     * Checks if the input is the result of a TCP level {@code ETIMEDOUT} error. This means that SYN segment did
+     * not result in an ACK from the remote server, most likely because connections have been severed in an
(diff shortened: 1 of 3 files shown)


dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -399,6 +400,9 @@ private boolean shouldAttemptToRetry(Throwable throwable) {
                             // String matches CJR RemotingOkHttpCall.shouldRetry
                             && socketTimeout.getMessage().contains("connect timed out");
                 }
+                if (isEtimedoutException(throwable)) {
```
bjlaub, line 403: minor nit: it might be worth a note explaining that the native `read` implementation sometimes throws `SocketException` with this specific text and that that is distinct from `SocketTimeoutException`.

otherwise lgtm

bjlaub (dismissed):

bjlaub (approved):

PR #2118, opened 2024-01-26
[WIP] Dialogue may consider each resolved IP as a separate node

carterkozak:
This is a work in progress. Will need some docs describing the behavior change.
==COMMIT_MSG==
Dialogue may consider each resolved IP as a separate node
==COMMIT_MSG==

## Possible downsides?
restructuring some core pieces of code, any bugs at this layer won'y be great...
We handle reloads in a fairly coarse way, currently DNS updates don't cause the client to fully reload. Once this is implemented, they will.


dialogue-clients/src/test/java/com/palantir/dialogue/clients/DialogueClientsDnsResolutionTest.java
```diff
@@ -0,0 +1,180 @@
...
+        assertThat(activeDnsRefreshTasks.getCount()).isEqualTo(1L);
+        assertThat(clientWeakRef.get())
+                .as("This reference is expected to be present before the strong client ref is unset")
+                .isNotNull();
+        // This reference _must_ be freed to allow the cleaner to do its job.
+        client = null;
+        attemptToGarbageCollect(clientWeakRef);
+        Awaitility.waitAtMost(Duration.ofSeconds(15)).untilAsserted(() -> assertThat(activeDnsRefreshTasks.getCount())
+                .isEqualTo(0L));
+    }
```
carterkozak: This test is expected to be failing at this point, I haven't implemented dns refreshing on this codepath quite yet

dialogue-clients/src/test/java/com/palantir/dialogue/clients/DialogueClientsDnsResolutionTest.java
```diff
@@ -0,0 +1,182 @@
...
+
+    @AfterEach
+    public void after() {
+        if (undertow != null) {
+            undertow.stop();
+        }
+    }
+
+    @Test
+    @Disabled // Not implemented yet
```
carterkozak, line 81: This must be implemented before we can consider merging

dialogue-clients/src/main/java/com/palantir/dialogue/clients/ServicesConfigBlockWithResolvedHosts.java
```diff
@@ -0,0 +1,130 @@
...
+    @DoNotLog
+    public String toString() {
+        return "ServicesConfigBlockWithResolvedHosts{scb=" + scb + ", resolvedAddresses=" + resolvedAddresses + '}';
+    }
+
+    static Refreshable<ServicesConfigBlockWithResolvedHosts> from(Refreshable<ServicesConfigBlock> input) {
+        // TODO(ckozak): refresh DNS info
+        //   e.g. wiring DialogueDnsResolution.refreshableUris(serviceName, registry, config.uris(), proxySelector)
+        //   currently this only handles DNS resolution once when the configuration is updated.
+        return input.map(block -> block == null ? null : new ServicesConfigBlockWithResolvedHosts(block, from(block)));
```
carterkozak, line 90: This is the biggest TODO at the moment: we're doing DNS resolution eagerly when the `Refreshable<ServicesConfigBlock>` changes, however the resulting refreshable is _not_ yet updated based on dns polling in the background.

At the moment we should be able to test this when nodes are already running, but we shouldn't expect it to work when dns changes.

Note that the other entrypoint of `DialogueClients.create(Class<T>,ClientConfiguration)` is implemented with dns polling at this point

dialogue-clients/src/main/java/com/palantir/dialogue/clients/DialogueDnsResolution.java
```diff
@@ -0,0 +1,193 @@
...
+                .channelName(channelName)
+                .result(Resolution_Result.FAILURE)
+                .build();
+        ImmutableList<TargetUri> initial = resolveUris(channelName, uris, selector, successTimer, failureTimer);
+        SettableRefreshable<ImmutableList<TargetUri>> refreshable = Refreshable.create(initial);
+        Counter scheduledTasks = metrics.scheduled(channelName);
+        ScheduledFuture<?> dnsRefreshFuture = SINGLETON_SCHEDULER.scheduleWithFixedDelay(
+                () -> {
+                    try {
+                        refreshable.update(resolveUris(channelName, uris, selector, successTimer, failureTimer));
```
bjlaub, line 79: i'm pretty ignorant about how all of this wires up... how does the list of `uris` within this thread get updated when the client configuration on disk changes? or does that allocate an entirely new channel, causing this thread to eventually die when the old one is gc'd?

carterkozak: The `ScheduledFuture` is encapsulated within the returned `RefreshableUris` object, so that when `RefreshableUris.stop()` is invoked, the scheduled task is removed.

That's handled here in the non-reloading case, but isn't fully implemented for the reloading case yet: https://github.com/palantir/dialogue/pull/2118/files/3be9eebaf03d51853105389ca18e6bc45db0465f#diff-6d1086f7aca5ef649647a82c52683f1d85aa8c3a9568e9b4510669f1e749fc89R98-R101

bjlaub: ah i see, sorry i missed your other comment

PR #2134, opened 2024-02-13
Dialogue client channel creation may target a specific address

carterkozak:
This represents a small piece of dns-based node discovery. Plumbing to handle automatic background dns refresh will come separately.

==COMMIT_MSG==
Dialogue client channel creation may target a specific address
==COMMIT_MSG==


dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueChannelFactory.java
```diff
@@ -30,6 +32,8 @@ public interface DialogueChannelFactory {
     interface ChannelArgs {
         String uri();
 
+        Optional<InetAddress> resolvedAddress();
```
bjlaub, line 35: is it possible for there to be multiple resolved addresses here?

carterkozak: Not the way this is constructed, as it specifies a particular target here: https://github.com/palantir/dialogue/pull/2134/files#diff-1a7529217d0eef0aaa07e3e259b4907f671eab95a67619174a7a6d4707cd0bb2R49

If an empty optional is provided, this falls back to the current behavior (which is why it's safe to merge).

I could see a model where we provide both ipv4 and ipv6 for a given node on a single client, however that opens a few new cans of worms that I'd prefer we avoid. We can always add a separate field to support that sort of thing in the future it it turns out to be important.

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/DialogueRoutePlanner.java
```diff
@@ -0,0 +1,94 @@
...
+    private final HttpRoutePlanner delegate;
+
+    DialogueRoutePlanner(ProxySelector proxySelector) {
+        delegate = new SystemDefaultRoutePlanner(proxySelector);
+    }
+
+    @Override
+    public HttpRoute determineRoute(HttpHost host, HttpContext context) throws HttpException {
+        HttpRoute route = delegate.determineRoute(host, context);
+        if (route.getTargetHost().getAddress() == null) {
```
bjlaub, line 46: when will this actually be true?

carterkozak: This is always true in the cases I've observed, it's a defensive check to make sure we're not clobbering the target address that has already been bound.

bjlaub: oh that's interesting - so name resolution is currently happening further down (or perhaps not at all in hc5) and this is just a convenient way of setting a resolved address ahead of time?

carterkozak: Yep -- this sets the address ahead of time, which tells hc5 that it doesn't need to interact with the dns resolver at all. That happens here: https://github.com/apache/httpcomponents-client/blob/661589e386e204fbdba1fbd7dac4f90ca5222c7f/httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/DefaultHttpClientConnectionOperator.java#L169-L176

bjlaub:
👍

PR #2141, opened 2024-02-14
WIP dns resolution worker v2

bjlaub:
## After this PR
==COMMIT_MSG==

==COMMIT_MSG==


dialogue-clients/src/main/java/com/palantir/dialogue/clients/DialogueDnsResolver.java
```diff
@@ -0,0 +1,31 @@
...
+ * limitations under the License.
+ */
+
+package com.palantir.dialogue.clients;
+
+import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.ImmutableSetMultimap;
+import java.net.InetAddress;
+
+public interface DialogueDnsResolver {
```
carterkozak: probably want to delete this, it should duplicate the `DialogueDnsResolver` declared in dialogue-core in the target branch

dialogue-clients/src/main/java/com/palantir/dialogue/clients/ServicesConfigBlockWithResolvedHosts.java
```diff
@@ -0,0 +1,70 @@
...
+import com.google.common.collect.ImmutableSetMultimap;
+import com.palantir.conjure.java.api.config.service.ServicesConfigBlock;
+import com.palantir.logsafe.DoNotLog;
+import com.palantir.logsafe.logger.SafeLogger;
+import com.palantir.logsafe.logger.SafeLoggerFactory;
+import java.net.InetAddress;
+import java.util.Objects;
+
+@DoNotLog
+final class ServicesConfigBlockWithResolvedHosts {
```
carterkozak: It might be a bit safer to use immutables to generate these sorts of types, since it'll include all fields in equals/hashCode

dialogue-clients/src/main/java/com/palantir/dialogue/clients/ReloadingClientFactory.java
```diff
@@ -512,4 +542,49 @@ public String toString() {
             return "LazilyMappedRefreshable{" + delegate + '}';
         }
     }
+
+    private static final class InternalDialogueChannelConfiguration {
```
carterkozak: probably safest to use immutables here as well, matching other internal types used in this module (e.g. the ChannelCacheKey/etc)

dialogue-clients/src/main/java/com/palantir/dialogue/clients/ServicesConfigBlockWithResolvedHosts.java
```diff
@@ -0,0 +1,34 @@
...
+import com.palantir.conjure.java.api.config.service.ServicesConfigBlock;
+import com.palantir.logsafe.DoNotLog;
+import java.net.InetAddress;
+import org.immutables.value.Value;
+
+@DoNotLog
+@Value.Immutable
+interface ServicesConfigBlockWithResolvedHosts {
+    @Value.Parameter
+    ServicesConfigBlock scb();
```
bjlaub, line 29: @carterkozak thoughts on making this an `Optional`? a thing i'm concerned about is: we ideally need to set an initial value for a `SettableRefreshable` of this type (see here: https://github.com/palantir/dialogue/pull/2141/files#diff-6d1086f7aca5ef649647a82c52683f1d85aa8c3a9568e9b4510669f1e749fc89R75-R76) which will be provided before the full round of initial name resolutions has completed. We could construct an "empty" `ServicesConfigBlock` but i worry that becomes a bit error-prone if the types within change to being required at build time.

carterkozak: If those types changed, we’d break a lot more than this — I don’t think it would be possible to add required types to that type.

dialogue-clients/src/main/java/com/palantir/dialogue/clients/DialogueDnsResolutionWorker.java
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    @Override
+    public void run() {
+        while (!shutdownRequested) {
+            try {
+                // check for updates to scb state first
+                ServicesConfigBlock updatedInputState = updatesQueue.poll(5, TimeUnit.SECONDS);
+                doUpdate(updatedInputState);
+            } catch (InterruptedException e) {
```
carterkozak, line 73: When an InterruptedException is thrown, the thread interrupted state is reset back to false, so the check on line 79 will not end up setting `shutdownRequested` to `true`.

bjlaub: doh, tyvm. fixed

dialogue-clients/src/main/java/com/palantir/dialogue/clients/DialogueClients.java
```diff
@@ -47,6 +47,19 @@
 public final class DialogueClients {
 
     public static ReloadingFactory create(Refreshable<ServicesConfigBlock> scb) {
+        // TODO(blaub): move the logic from RelaodingClientFactory ctor that launches a DialogueDnsResolutionWorker
```
bjlaub, line 50: @carterkozak i suspect we'll want to do something like:
- change `ReloadingParams` to hold a `Refreshable<ServicesConfigBlockWIthResolvedHosts>` instead
- create the dns worker thread here, and have it `subscribe` to the `scb` parameter
- make the thread output updates to the `Refreshable<ServicesConfigBlockWIthResolvedHosts>` on the `ReloadingParams` instance we create

open question is what to do about the `reloading()` method on `ReloadingFactory`, which comes from `ConjureClients.ToReloadingFactory` interface and creates a new factory with params that set a different ServicesConfigBlock refreshable. Perhaps call `dispose()` on the original subscription and re-wire everything to the new one without spawning a new thread?

PR #2149, opened 2024-02-20
[feature/dns] dns resource lifecycle

carterkozak:
==COMMIT_MSG==
dns resource lifecycle
==COMMIT_MSG==

### DIFF
--- a/dialogue-clients/src/main/java/com/palantir/dialogue/clients/DialogueDnsResolutionWorker.java
+++ b/dialogue-clients/src/main/java/com/palantir/dialogue/clients/DialogueDnsResolutionWorker.java
@@ -18,33 +18,34 @@ package com.palantir.dialogue.clients;
 
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.ImmutableSetMultimap;
+import com.google.errorprone.annotations.concurrent.GuardedBy;
 import com.palantir.conjure.java.api.config.service.ServicesConfigBlock;
 import com.palantir.dialogue.core.DialogueDnsResolver;
 import com.palantir.logsafe.logger.SafeLogger;
 import com.palantir.logsafe.logger.SafeLoggerFactory;
+import com.palantir.refreshable.SettableRefreshable;
+import java.lang.ref.WeakReference;
 import java.net.InetAddress;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.util.Objects;
-import java.util.function.Consumer;
 import javax.annotation.Nullable;
 
 final class DialogueDnsResolutionWorker implements Runnable {
     private static final SafeLogger log = SafeLoggerFactory.get(DialogueDnsResolutionWorker.class);
 
     @Nullable
+    @GuardedBy("this")
     private ServicesConfigBlock inputState;
-
-    @Nullable
-    private ServicesConfigBlockWithResolvedHosts resolvedState;
 
     private volatile boolean shutdownRequested;
     private final DialogueDnsResolver resolver;
-    private final Consumer<ServicesConfigBlockWithResolvedHosts> receiver;
+    private final WeakReference<SettableRefreshable<ServicesConfigBlockWithResolvedHosts>> receiver;
 
-    DialogueDnsResolutionWorker(DialogueDnsResolver resolver, Consumer<ServicesConfigBlockWithResolvedHosts> receiver) {
+    DialogueDnsResolutionWorker(
+            DialogueDnsResolver resolver, SettableRefreshable<ServicesConfigBlockWithResolvedHosts> receiver) {
         this.resolver = resolver;
-        this.receiver = receiver;
+        this.receiver = new WeakReference<>(receiver);
         this.shutdownRequested = false;
     }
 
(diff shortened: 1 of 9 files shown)


dialogue-clients/src/main/java/com/palantir/dialogue/clients/ReloadingClientFactory.java
```diff
@@ -134,6 +173,12 @@ default DialogueDnsResolver dnsResolver() {
         }
 
         Optional<ExecutorService> blockingExecutor();
+
+        static Builder builder() {
+            return new Builder();
+        }
+
+        class Builder extends ImmutableReloadingParams.Builder {}
```
carterkozak: This part should be reverted, unnecessary diff

dialogue-clients/src/test/java/com/palantir/dialogue/clients/DialogueDnsResolutionWorkerTest.java
```diff
@@ -144,7 +144,8 @@ public void testResolvedAddressesChangesAfterStartup() throws Exception {
                 .build();
         SettableRefreshable<ServicesConfigBlock> inputRefreshable = Refreshable.create(initialState);
         SettableRefreshable<ServicesConfigBlockWithResolvedHosts> receiverRefreshable = Refreshable.create(null);
-        DialogueDnsResolutionWorker worker = new DialogueDnsResolutionWorker(resolver, receiverRefreshable::update);
+        DialogueDnsResolutionWorker worker =
+                new DialogueDnsResolutionWorker(resolver, Duration.ofSeconds(5), receiverRefreshable);
```
carterkozak, line 148: I think we may want to externally trigger state changes in this tests dns resolver, rather than leveraging interactions to change the state.

I don't want to do that in this PR though, since it's already complex enough

dialogue-clients/src/main/java/com/palantir/dialogue/clients/ReloadingClientFactory.java
```diff
@@ -170,27 +229,37 @@ public PerHostClientFactory perHost(String serviceName) {
         Preconditions.checkNotNull(serviceName, "serviceName");
         String channelName = ChannelNames.reloading(serviceName, params);
 
-        Refreshable<List<DialogueChannel>> perHostDialogueChannels = params.scb()
+        Refreshable<List<DialogueChannel>> perHostDialogueChannels = getServiceRefreshable(serviceName)
```
carterkozak: Updated to handle `perHost` here, the only remaining caller into the legacy `getNonReloadingChannel` over is the non-reloading method which takes a fully realized ServiceConfiguration as input. We can handle that in a separate PR, as well as the `getNonReloading(Class<T> clazz, ClientConfiguration clientConf)` overload

dialogue-clients/src/test/java/com/palantir/dialogue/otherpackage/DialogueClientsTest.java
```diff
@@ -169,9 +165,9 @@ void perHost_attributes_concurrencylimiter_metrics_to_the_right_host() {
                         .putServices(
                                 "my-service",
                                 PartialServiceConfiguration.builder()
-                                        .addUris("https://my-service-0.fake.palantir.com")
-                                        .addUris("https://my-service-1.fake.palantir.com")
-                                        .addUris("https://my-service-2.fake.palantir.com")
+                                        .addUris("https://127.0.0.1/0")
+                                        .addUris("https://127.0.0.1/1")
+                                        .addUris("https://127.0.0.1/2")
```
carterkozak, line 170: Updated because the manifestation is slightly different -- this way we allow the actual client to progress as expected.
Without using localhost here, the dns resolution component reduces the target to no TargetUris as dns resolution fails.

dialogue-clients/src/main/java/com/palantir/dialogue/clients/DialogueDnsResolutionWorker.java
```diff
@@ -64,15 +70,23 @@ public void run() {
         while (!shutdownRequested) {
             try {
                 // check for updates to scb state first
-                Thread.sleep(5000L);
+                Thread.sleep(dnsRefreshInterval.toMillis());
                 try {
+                    if (receiver.get() == null) {
+                        shutdownRequested = true;
+                        log.info("Output refreshable has been garbage collected, no need to continue polling");
```
bjlaub, line 77: should there be an `else` or a `break` here?

carterkozak: Ya, we should probably add an `else` here to avoid unnecessary work. `doUpdate` must always handle the not-present case either way, as the gc may run between doing a pre-check and later on.

dialogue-clients/src/main/java/com/palantir/dialogue/clients/ReloadingClientFactory.java
```diff
@@ -382,105 +456,106 @@ private ClientConfiguration hydrate(ClientConfiguration configuration) {
         return configuration;
     }
 
-    // TODO(ckozak): Each of the 'map(' blocks should be extracted to another method. I'm not doing that
-    // yet be cause it will make the diff much harder to review.
-    @SuppressWarnings("CyclomaticComplexity")
+    private Refreshable<InternalDialogueChannelConfiguration> getServiceRefreshable(String serviceName) {
```
bjlaub: minor nit: i'm not wild about this method name, perhaps something like `configurationForService` as this returns a configuration that can be used to create a channel for this service? not super picky though

bjlaub (dismissed):

dialogue-clients/src/main/java/com/palantir/dialogue/clients/ReloadingClientFactory.java
carterkozak: 👍 One of the two hardest problems in CS

PR #2150, opened 2024-02-20
[feature/dns] add tests for invalid partial configurations

bjlaub:
Verify that when one service's partial configuration in the ServicesConfigBlock is invalid, it does not prevent other services from creating and using channels.


dialogue-clients/src/main/java/com/palantir/dialogue/clients/ReloadingClientFactory.java
```diff
@@ -425,7 +425,7 @@ private Refreshable<InternalDialogueChannel> getInternalDialogueChannel(String s
 
                     if (maybeServiceConf.isEmpty()) {
                         return new EmptyInternalDialogueChannel(() -> new SafeIllegalStateException(
-                                "Service not configured (config block not present)",
+                                "Service not configured (config block invalid or not present)",
```
carterkozak: this change is breaking tests

carterkozak (dismissed):

carterkozak:
lgtm, minor merge conflict blocking the merge though

PR #2156, opened 2024-02-21
[feature/dns] dns resolution uses only ipv4 or ipv6 addresses, but not both

bjlaub:
## Before this PR
It was possible for DefaultDialogueDnsResolver to resolve a hostname to a Set<InetAddress> containing e.g. pairs of (Inet4Address, Inet6Address). This can lead to problems downstream as dialogue won't be able to distinguish them as possibly pointing at the same host, and instead will try to treat them as different nodes.

## After this PR
Filter DNS resolution results to either IPv4 or IPv6 addresses, but not both. We assume that if a resolution result contains a mix of both IPv4 and IPv6 addresses, then they correspond to the same physical node, and we ignore the IPv6 addresses.

==COMMIT_MSG==
dns resolution uses only ipv4 or ipv6 addresses, but not both
==COMMIT_MSG==

## Possible downsides?
It is possible a name resolves to a set of IPv4 and IPv6 addresses that actually point to two different nodes; in that case, we'd be ignoring a valid resolved address.

### DIFF
--- a/dialogue-clients/src/main/java/com/palantir/dialogue/clients/ProtocolVersionFilteringDialogueDnsResolver.java
+++ b/dialogue-clients/src/main/java/com/palantir/dialogue/clients/ProtocolVersionFilteringDialogueDnsResolver.java
@@ -19,14 +19,15 @@ package com.palantir.dialogue.clients;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.ImmutableSetMultimap;
 import com.palantir.dialogue.core.DialogueDnsResolver;
+import com.palantir.logsafe.logger.SafeLogger;
+import com.palantir.logsafe.logger.SafeLoggerFactory;
 import java.net.Inet4Address;
 import java.net.Inet6Address;
 import java.net.InetAddress;
-import java.util.Collection;
-import java.util.Set;
-import java.util.stream.Collectors;
 
 class ProtocolVersionFilteringDialogueDnsResolver implements DialogueDnsResolver {
+    private static final SafeLogger log = SafeLoggerFactory.get(ProtocolVersionFilteringDialogueDnsResolver.class);
+
     private final DialogueDnsResolver delegate;
 
     ProtocolVersionFilteringDialogueDnsResolver(DialogueDnsResolver delegate) {
@@ -42,32 +43,44 @@ class ProtocolVersionFilteringDialogueDnsResolver implements DialogueDnsResolver
     public ImmutableSetMultimap<String, InetAddress> resolve(Iterable<String> hostnames) {
         ImmutableSetMultimap<String, InetAddress> resolved = delegate.resolve(hostnames);
         ImmutableSetMultimap.Builder<String, InetAddress> result = ImmutableSetMultimap.builder();
-        resolved.asMap().forEach((host, addresses) -> result.putAll(host, filter(addresses)));
+        for (String host : resolved.keySet()) {
+            result.putAll(host, filter(resolved.get(host)));
+        }
         return result.build();
     }
 
     // TODO(dns): report metrics here
-    private static ImmutableSet<InetAddress> filter(Collection<InetAddress> addresses) {
+    private static ImmutableSet<InetAddress> filter(ImmutableSet<InetAddress> addresses) {
         // Assume that if resolved addresses contain a mix of ipv4 and ipv6 addresses, then they
         // likely point to the same host, just using a different protocol. In that case, we discared the ipv6
         // addresses and prefer only ipv4.
         //
         // If the set of resolved addresses contains only ipv4 or only ipv6 addresses, then it is left unmodified.
 
-        Set<InetAddress> ipv4Addresses =
-                addresses.stream().filter(addr -> addr instanceof Inet4Address).collect(Collectors.toUnmodifiableSet());
-        Set<InetAddress> ipv6Addresses =
-                addresses.stream().filter(addr -> addr instanceof Inet6Address).collect(Collectors.toUnmodifiableSet());
-
-        if (ipv4Addresses.isEmpty() && ipv6Addresses.isEmpty()) {
-            return ImmutableSet.of();
-        } else if (ipv6Addresses.isEmpty()) {
-            return ImmutableSet.copyOf(ipv4Addresses);
-        } else if (ipv4Addresses.isEmpty()) {
-            return ImmutableSet.copyOf(ipv6Addresses);
-        } else {
-            // only include ipv4 addresses
-            return ImmutableSet.copyOf(ipv4Addresses);
+        int numIpv4 = 0;
+        int numIpv6 = 0;
+        int numUnknown = 0;
+        ImmutableSet.Builder<InetAddress> onlyIpv4Addresses = ImmutableSet.builder();
+        for (InetAddress address : addresses) {
+            if (address instanceof Inet4Address) {
+                ++numIpv4;
+                onlyIpv4Addresses.add(address);
+            } else if (address instanceof Inet6Address) {
+                ++numIpv6;
+            } else {
+                ++numUnknown;
+            }
         }
+
+        if (numUnknown > 0) {
+            log.warn("name resolution result contains an address that is neither IPv4 nor IPv6");
+            return addresses;
+        }
+
+        if (numIpv4 > 0 && numIpv6 > 0) {
+            return onlyIpv4Addresses.build();
+        }
+
+        return addresses;
     }
 }
--- a/dialogue-clients/src/main/java/com/palantir/dialogue/clients/ProtocolVersionFilteringDialogueDnsResolver.java
+++ b/dialogue-clients/src/main/java/com/palantir/dialogue/clients/ProtocolVersionFilteringDialogueDnsResolver.java
@@ -19,6 +19,7 @@ package com.palantir.dialogue.clients;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.ImmutableSetMultimap;
 import com.palantir.dialogue.core.DialogueDnsResolver;
+import com.palantir.logsafe.UnsafeArg;
 import com.palantir.logsafe.logger.SafeLogger;
 import com.palantir.logsafe.logger.SafeLoggerFactory;
 import java.net.Inet4Address;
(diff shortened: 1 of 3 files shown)


dialogue-clients/src/main/java/com/palantir/dialogue/clients/ProtocolVersionFilteringDialogueDnsResolver.java
```diff
@@ -0,0 +1,73 @@
...
+        // Assume that if resolved addresses contain a mix of ipv4 and ipv6 addresses, then they
+        // likely point to the same host, just using a different protocol. In that case, we discared the ipv6
+        // addresses and prefer only ipv4.
+        //
+        // If the set of resolved addresses contains only ipv4 or only ipv6 addresses, then it is left unmodified.
+
+        Set<InetAddress> ipv4Addresses =
+                addresses.stream().filter(addr -> addr instanceof Inet4Address).collect(Collectors.toUnmodifiableSet());
+        Set<InetAddress> ipv6Addresses =
+                addresses.stream().filter(addr -> addr instanceof Inet6Address).collect(Collectors.toUnmodifiableSet());
```
carterkozak: slight preference to avoid streams in this sort of recurring task, since their performance is not great (high allocation, and hotspot has a tough time optimizing them)

dialogue-clients/src/main/java/com/palantir/dialogue/clients/ProtocolVersionFilteringDialogueDnsResolver.java
```diff
@@ -0,0 +1,73 @@
...
+
+        if (ipv4Addresses.isEmpty() && ipv6Addresses.isEmpty()) {
```
carterkozak: If we have any InetAddresses provided as input which don't appear to be `Inet4Address` or `Inet6Address`, we should probably log a warning and return the input.

dialogue-clients/src/main/java/com/palantir/dialogue/clients/ProtocolVersionFilteringDialogueDnsResolver.java
carterkozak: We should make sure the input order is preserved (we specify `ImmutableSet` in the api contract because it enforces uniqueness while retaining original iteration order)

dialogue-clients/src/main/java/com/palantir/dialogue/clients/ProtocolVersionFilteringDialogueDnsResolver.java
```diff
@@ -0,0 +1,73 @@
...
+            return ImmutableSet.of();
+        } else if (ipv6Addresses.isEmpty()) {
+            return ImmutableSet.copyOf(ipv4Addresses);
+        } else if (ipv4Addresses.isEmpty()) {
+            return ImmutableSet.copyOf(ipv6Addresses);
+        } else {
+            // only include ipv4 addresses
+            return ImmutableSet.copyOf(ipv4Addresses);
+        }
```
carterkozak: I think this can be simplified to something closer to:

if (includesIpV4 && includesIpV6) {
    return allIpV4Addresses;
} else {
    return inputParameter;
}

dialogue-clients/src/main/java/com/palantir/dialogue/clients/ProtocolVersionFilteringDialogueDnsResolver.java
bjlaub: do we want to handle cases where there is a mix of e.g. `(Inet4Address, RandomSubclassOfInetAddress)` or `(Inet6Address, RandomSubclassOfInetAddress)`? or just return the input if any single item is not one of the subclasses we were expecting?

dialogue-clients/src/main/java/com/palantir/dialogue/clients/ProtocolVersionFilteringDialogueDnsResolver.java
```diff
@@ -0,0 +1,86 @@
...
+                onlyIpv4Addresses.add(address);
+            } else if (address instanceof Inet6Address) {
+                ++numIpv6;
+            } else {
+                ++numUnknown;
+            }
+        }
+
+        if (numUnknown > 0) {
+            log.warn("name resolution result contains an address that is neither IPv4 nor IPv6");
```
bjlaub: is it worth logging the unknown address types here?

dialogue-clients/src/main/java/com/palantir/dialogue/clients/ProtocolVersionFilteringDialogueDnsResolver.java
bjlaub: we can iterate over the input in the preserved order (see the most recent commit) but there's no guarantee the resulting ImmutableSet here will have the same iteration order as the input since the contents have possibly changed.

dialogue-clients/src/main/java/com/palantir/dialogue/clients/ProtocolVersionFilteringDialogueDnsResolver.java
carterkozak: Log some kind of warning/error and return the unmodified input imo -- if we're outside the bounds of what we understand, we can do more harm than good

dialogue-clients/src/main/java/com/palantir/dialogue/clients/ProtocolVersionFilteringDialogueDnsResolver.java
carterkozak: Yep, that would be great!

dialogue-clients/src/main/java/com/palantir/dialogue/clients/ProtocolVersionFilteringDialogueDnsResolver.java
```diff
@@ -0,0 +1,86 @@
...
+        // Assume that if resolved addresses contain a mix of ipv4 and ipv6 addresses, then they
+        // likely point to the same host, just using a different protocol. In that case, we discared the ipv6
+        // addresses and prefer only ipv4.
+        //
+        // If the set of resolved addresses contains only ipv4 or only ipv6 addresses, then it is left unmodified.
+
+        int numIpv4 = 0;
+        int numIpv6 = 0;
+        int numUnknown = 0;
+        ImmutableSet.Builder<InetAddress> onlyIpv4Addresses = ImmutableSet.builder();
```
carterkozak: dns resolution isn't going to be a super hot codepath, so this is probably fine. I'd probably split the detection and filtering into separate loops so that we avoid creating new collections unless we need to do so.

dialogue-clients/src/main/java/com/palantir/dialogue/clients/ProtocolVersionFilteringDialogueDnsResolver.java
carterkozak: we could log within iteration when we encounter the unknown types (may log a few times per address, but that won't be much more frequent than per-hostname)

dialogue-clients/src/main/java/com/palantir/dialogue/clients/ProtocolVersionFilteringDialogueDnsResolver.java
bjlaub: yeah that makes sense, will do

dialogue-clients/src/main/java/com/palantir/dialogue/clients/ProtocolVersionFilteringDialogueDnsResolver.java
```diff
@@ -0,0 +1,93 @@
...
+        for (InetAddress address : addresses) {
+            if (address instanceof Inet4Address) {
+                ++numIpv4;
+            } else if (address instanceof Inet6Address) {
+                ++numIpv6;
+            } else {
+                ++numUnknown;
+                log.warn(
+                        "name resolution result contains an address that is neither IPv4 nor IPv6",
+                        UnsafeArg.of("address", address));
```
bjlaub, line 74: i never know whether to log something like this as safe or unsafe 🤷

carterkozak: hostnames and IP addresses are usually unsafe. Some specific instances can be safe, but dialogue operates at a lower level where it can be fed both, so we must assume the worst.

dialogue-clients/src/main/java/com/palantir/dialogue/clients/ProtocolVersionFilteringDialogueDnsResolver.java
```diff
@@ -0,0 +1,93 @@
...
+        }
+
+        if (numIpv4 > 0 && numIpv6 > 0) {
+            ImmutableSet.Builder<InetAddress> onlyIpv4Addresses = ImmutableSet.builder();
+            for (InetAddress address : addresses) {
+                if (address instanceof Inet4Address) {
+                    onlyIpv4Addresses.add(address);
+                }
+            }
+            return onlyIpv4Addresses.build();
```
carterkozak, line 94: Let's log at info level in this case, describing the number of ipv4 and ipv6 addresses, and that we're only going to use the ipv4 values to avoid double-counting hosts.

dialogue-clients/src/test/java/com/palantir/dialogue/clients/ProtocolVersionFilteringDialogueDnsResolverTest.java
```diff
@@ -0,0 +1,78 @@
...
+import org.junit.jupiter.api.Test;
+
+class ProtocolVersionFilteringDialogueDnsResolverTest {
+
+    @Test
+    void returns_only_ipv4_when_input_contains_only_ipv4() throws UnknownHostException {
+        ImmutableSet<InetAddress> resolvedAddresses = ImmutableSet.of(
+                InetAddress.getByAddress(new byte[] {1, 1, 1, 1}), InetAddress.getByAddress(new byte[] {2, 2, 2, 2}));
+
+        DialogueDnsResolver delegate = mock(DialogueDnsResolver.class);
```
carterkozak: It might be easier if we split the `MapBasedDnsResolver` out of `DialogueClientsDnsIntegrationTest` to reuse in this test. I find mockito a giant pain to work with in favor of simpler stubs. This way we can mutate the underlying multimap, and assert the results are what we expect, with the compile-time guarantee that all methods are implemented.

dialogue-clients/src/test/java/com/palantir/dialogue/clients/ProtocolVersionFilteringDialogueDnsResolverTest.java
```diff
@@ -0,0 +1,78 @@
...
+        when(delegate.resolve(anyString())).thenReturn(resolvedAddresses);
+
+        DialogueDnsResolver resolver = new ProtocolVersionFilteringDialogueDnsResolver(delegate);
+
+        assertThat(resolver.resolve("foo")).isEqualTo(resolvedAddresses);
```
carterkozak: I don't think equality on ImmutableSet takes order into account, so we may want to prefer `containsExactly` or `containsExactlyElementsOf` to ensure the order is what we expect.

dialogue-clients/src/test/java/com/palantir/dialogue/clients/ProtocolVersionFilteringDialogueDnsResolverTest.java
bjlaub: let's do that in a separate change, then i'll rebase: https://github.com/palantir/dialogue/pull/2158

dialogue-clients/src/test/java/com/palantir/dialogue/clients/ProtocolVersionFilteringDialogueDnsResolverTest.java
bjlaub: 👍 fixed

carterkozak:
👍 thanks!

PR #2157, opened 2024-02-21
fix flaky DialogueDnsResolutionWorkerTest

bjlaub:
## Before this PR
DialogueDnsResolutionWorkerTest had some truly terrible logic that tried to "rotate" resolved IPs every 5 seconds, leading to some fragile awaitility logic.

## After this PR
RotatingStaticDnsResolver adds an explicit `rotate()` method that we can control from the test itself to simulate a new set of resolved addresses.

==COMMIT_MSG==
fix flaky DialogueDnsResolutionWorkerTest
==COMMIT_MSG==


carterkozak (approved):
👍

PR #2163, opened 2024-02-22
[feature/dns] Support non-reloading factory methods

carterkozak:
==COMMIT_MSG==
Support non-reloading factory methods
==COMMIT_MSG==


dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/DefaultConjureRuntime.java
```diff
@@ -33,7 +33,7 @@
 public final class DefaultConjureRuntime implements ConjureRuntime {
     @VisibleForTesting
     static final CaffeineSpec DEFAULT_SERDE_CACHE_SPEC =
-            CaffeineSpec.parse("maximumSize=1000,expireAfterAccess=1m,weakValues");
+            CaffeineSpec.parse("maximumSize=1000,expireAfterAccess=1m,weakKeys,weakValues");
```
carterkozak, line 36: `weakKeys` is required to allow clients to be garbage collected. Serializer instances are keyed on anonymous types declared to capture type info, however these types can also capture more than they intend to.

bjlaub (dismissed):

PR #2164, opened 2024-02-22
[feature/dns][stacked] Implement a dialogue dns node discovery feature flag

carterkozak:
==COMMIT_MSG==
Implement a dialogue dns node discovery feature flag
==COMMIT_MSG==

### DIFF
--- a/dialogue-clients/src/main/java/com/palantir/dialogue/clients/ReloadingClientFactory.java
+++ b/dialogue-clients/src/main/java/com/palantir/dialogue/clients/ReloadingClientFactory.java
@@ -105,6 +105,7 @@ final class ReloadingClientFactory implements DialogueClients.ReloadingFactory {
         ApacheHttpClientChannels.CloseableClient apacheClient = clientBuilder.build();
         return new LiveReloadingChannel(
                 DnsSupport.pollForChanges(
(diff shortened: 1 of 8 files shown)


dialogue-clients/src/main/java/com/palantir/dialogue/clients/ReloadingClientFactory.java
```diff
@@ -156,6 +162,11 @@ default Duration dnsRefreshInterval() {
             return Duration.ofSeconds(5);
         }
 
+        @Value.Default
+        default boolean dnsNodeDiscovery() {
+            return true;
```
carterkozak: Testing with the wrong default initially, to make sure I haven't broken anything I don't expect. Once I flip this, I suspect a few tests may need slight modification.

bjlaub (commented):
is it worth a test to verify that we don't schedule any background tasks when dnsNodeDiscovery is false?

carterkozak:
Good idea, I've updated the test which sets the feature flag explicitly off.

bjlaub (approved):
lgtm, thanks!

carterkozak:
Once #2163 is approved I'll merge the stack at once.

PR #2166, opened 2024-02-23
[feature/dns] enable dns node discovery via environment variable

bjlaub:
## Before this PR
Flag can only be enabled via ReloadingParams.

## After this PR
Allow enabling dns node discovery via an environment variable, which will possibly ease testing and rollout without having to make sweeping code changes.

==COMMIT_MSG==
enable dns node discovery via environment variable
==COMMIT_MSG==

### DIFF
--- a/dialogue-clients/src/main/java/com/palantir/dialogue/clients/ReloadingClientFactory.java
+++ b/dialogue-clients/src/main/java/com/palantir/dialogue/clients/ReloadingClientFactory.java
@@ -127,6 +127,12 @@ final class ReloadingClientFactory implements DialogueClients.ReloadingFactory {
                 params.runtime().clients());
     }
 
+    // allow enabling dns node discovery by setting an environment variable
+    // note that this can still be disabled explicitly via the factory, but having a global flag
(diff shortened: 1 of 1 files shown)


dialogue-clients/src/main/java/com/palantir/dialogue/clients/ReloadingClientFactory.java
```diff
@@ -164,7 +170,7 @@ default Duration dnsRefreshInterval() {
 
         @Value.Default
         default boolean dnsNodeDiscovery() {
-            return false;
+            return "true".equals(System.getenv(ENABLE_DNS_NODE_DISCOVERY_ENV_VAR));
```
carterkozak: might as well use equalsIgnoreCase

carterkozak:
no need to write a test for the envvar imo

dialogue-clients/src/main/java/com/palantir/dialogue/clients/ReloadingClientFactory.java
bjlaub: done

carterkozak (approved):

PR #2167, opened 2024-02-23
[feature/dns] metric for timing dns polling sweep

bjlaub:
## After this PR
==COMMIT_MSG==
metric for timing dns polling sweep
==COMMIT_MSG==

### DIFF
--- a/dialogue-clients/metrics.md
+++ b/dialogue-clients/metrics.md
@@ -55,7 +55,7 @@ Connection pool metrics from the dialogue Apache client.
 Dialogue DNS metrics.
 - `client.dns.tasks` (counter): Number of active Dialogue DNS update background tasks currently scheduled.
   - `kind`: Describes the type of component polling for DNS updates.
-- `client.dns.resolveTime` (timer): Measures the time taken to complete a full pass polling for DNS updates.
+- `client.dns.refresh` (timer): Measures the time taken to complete a full pass polling for DNS updates.
   - `kind`: Describes the type of component polling for DNS updates.
 
 ## Dialogue Core
--- a/dialogue-clients/metrics.md
+++ b/dialogue-clients/metrics.md
@@ -55,6 +55,7 @@ Connection pool metrics from the dialogue Apache client.
 Dialogue DNS metrics.
 - `client.dns.tasks` (counter): Number of active Dialogue DNS update background tasks currently scheduled.
   - `kind`: Describes the type of component polling for DNS updates.
+- `client.dns.resolveTime` (timer): Measures the time taken to complete a full pass polling for DNS updates.
 
 ## Dialogue Core
 
--- a/dialogue-clients/metrics.md
+++ b/dialogue-clients/metrics.md
@@ -56,6 +56,7 @@ Dialogue DNS metrics.
 - `client.dns.tasks` (counter): Number of active Dialogue DNS update background tasks currently scheduled.
   - `kind`: Describes the type of component polling for DNS updates.
 - `client.dns.resolveTime` (timer): Measures the time taken to complete a full pass polling for DNS updates.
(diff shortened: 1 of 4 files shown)


dialogue-clients/src/main/java/com/palantir/dialogue/clients/DialogueDnsResolutionWorker.java
```diff
@@ -64,7 +69,7 @@ public void run() {
                 // this logging may be helpful in informing us of problems in the system.
                 log.info("Output refreshable has been garbage collected, no need to continue polling");
             } else {
-                doUpdate(null);
+                updateTimer.time(() -> doUpdate(null));
```
bjlaub: whoops, this is the wrong place to put this, i think we need to measure within doUpdate.

dialogue-clients/src/main/java/com/palantir/dialogue/clients/DialogueDnsResolutionWorker.java
```diff
@@ -104,5 +111,7 @@ private synchronized void doUpdate(@Nullable INPUT updatedInputState) {
                 log.info("Attempted to update DNS output refreshable which has already been garbage collected");
             }
         }
+        long end = System.nanoTime();
+        updateTimer.update(Duration.ofNanos(end - start));
```
carterkozak: We may want to put this inside the `if (inputState != null) {` conditional to make sure we're actually measuring work

dialogue-clients/src/main/metrics/dialogue-core-metrics.yml
```diff
@@ -11,3 +11,6 @@ namespaces:
           - name: kind
             docs: Describes the type of component polling for DNS updates.
         docs: Number of active Dialogue DNS update background tasks currently scheduled.
+      resolveTime:
+        type: timer
```
carterkozak, line 15: Can we tag this with the `kind` tag used above, from the `DnsPollingSpec`?

bjlaub: yep will do

dialogue-clients/src/main/java/com/palantir/dialogue/clients/DialogueDnsResolutionWorker.java
```diff
@@ -103,6 +109,8 @@ private synchronized void doUpdate(@Nullable INPUT updatedInputState) {
             } else {
                 log.info("Attempted to update DNS output refreshable which has already been garbage collected");
             }
+            long end = System.nanoTime();
+            updateTimer.update(Duration.ofNanos(end - start));
```
carterkozak: nit: let's avoid Duration allocation and use `updateTimer.update(end - start, TimeUnit.NANOSECONDS)`

dialogue-clients/metrics.md
```diff
@@ -55,6 +55,8 @@ Connection pool metrics from the dialogue Apache client.
 Dialogue DNS metrics.
 - `client.dns.tasks` (counter): Number of active Dialogue DNS update background tasks currently scheduled.
   - `kind`: Describes the type of component polling for DNS updates.
+- `client.dns.resolveTime` (timer): Measures the time taken to complete a full pass polling for DNS updates.
```
carterkozak: perhaps something like `client.dns.refresh` to make the name reflect that this is a full sweep

bjlaub: done

dialogue-clients/src/main/java/com/palantir/dialogue/clients/DialogueDnsResolutionWorker.java
bjlaub: fixed

carterkozak (approved):

PR #2168, opened 2024-02-23
[feature/dns] Connect timers are tagged with the dns strategy

carterkozak:
==COMMIT_MSG==
[feature/dns] Connect timers are tagged with the dns strategy
==COMMIT_MSG==

### DIFF
--- a/dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/InstrumentedPlainConnectionSocketFactory.java
+++ b/dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/InstrumentedPlainConnectionSocketFactory.java
@@ -1,0 +1,67 @@
+/*
+ * (c) Copyright 2024 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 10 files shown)


dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/InstrumentedPlainConnectionSocketFactory.java
```diff
@@ -0,0 +1,67 @@
...
+        try {
+            Socket result = super.connectSocket(connectTimeout, socket, host, remoteAddress, localAddress, context);
+            success = true;
+            return result;
+        } finally {
+            long durationNanos = System.nanoTime() - startNanos;
+            Timer timer = connectInstrumentation.timer(success, context);
+            timer.update(durationNanos, TimeUnit.NANOSECONDS);
+        }
+    }
```
carterkozak, line 61: This fixes a longstanding metrics omission that I found in `testPlainHttp` where only the `connectionCreate` timer was updated, not `connectionConnect` because only the tls connection socket factory was wrapped

bjlaub: perhaps we should also update the docs:? https://github.com/palantir/dialogue/pull/2168/files#diff-2c10ff3d3757eb3774598e1eee65be06017260f98a612b18067e9850ededf136R86

bjlaub: oh wait i misread the docs, disregard!

bjlaub (approved):

PR #2171, opened 2024-02-26
handle invalid URIs

carterkozak:
## Before this PR
The DNS feature changed the way invalid URIs were handled, where -1 was supported.

## After this PR
==COMMIT_MSG==
handle invalid URIs
==COMMIT_MSG==

### DIFF
--- a/dialogue-clients/src/main/java/com/palantir/dialogue/clients/ReloadingClientFactory.java
+++ b/dialogue-clients/src/main/java/com/palantir/dialogue/clients/ReloadingClientFactory.java
@@ -498,6 +498,7 @@ final class ReloadingClientFactory implements DialogueClients.ReloadingFactory {
         });
     }
 
+    @SuppressWarnings("checkstyle:CyclomaticComplexity")
     private ImmutableList<TargetUri> getTargetUris(
             @Safe String serviceNameForLogging,
             Collection<String> uris,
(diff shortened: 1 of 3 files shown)


dialogue-clients/src/main/java/com/palantir/dialogue/clients/ReloadingClientFactory.java
```diff
@@ -531,6 +534,13 @@ private ImmutableList<TargetUri> getTargetUris(
                 }
             }
         }
+        if (targetUris.isEmpty() && failedToParse) {
+            // Handle cases like "host:-1", but only when _all_ uris are invalid
+            log.warn(
+                    "Failed to parse all URIs, falling back to legacy DNS approach for service '{}'",
+                    SafeArg.of("service", serviceNameForLogging));
+            return uris.stream().map(TargetUri::of).collect(ImmutableList.toImmutableList());
```
carterkozak, line 542: Later on we parse uri strings into the `URL` type. I wrote an implementation of this fix which uses URL throughout, however that has other issues (requires known schemes, so it would throw if `mesh-http://` uris are read in places we don't test). I think this is the safest path forward.

dialogue-clients/src/main/java/com/palantir/dialogue/clients/ReloadingClientFactory.java
```diff
@@ -531,6 +534,13 @@ private ImmutableList<TargetUri> getTargetUris(
                 }
             }
         }
+        if (targetUris.isEmpty() && failedToParse) {
```
bjlaub, line 537: do we need to handle a case where e.g. `uris` contains a mix of both valid and invalid URIs? With this change, we will drop the invalid ones completely.

carterkozak: Yep, the comment calls this out by design

bjlaub (approved):

dialogue-clients/src/main/java/com/palantir/dialogue/clients/ReloadingClientFactory.java
carterkozak: In _most_ cases, an invalid uri won't be able to handle requests, so we want to avoid the possibility of sending data there. When it does work, conversion to URI is likely to fail to list proxies anyhow, so behavior may not be correct :-/

PR #2173, opened 2024-02-26
Increment a metric when URIs fail to parse

carterkozak:
This includes additional logging in the port=-1 case as well, since URI parsing doesn't in fact fail, but produces a uri with no host component.

We'll add a default monitor on this, letting folks know things are going sideways

==COMMIT_MSG==
Increment a metric when URIs fail to parse
==COMMIT_MSG==


dialogue-clients/src/main/java/com/palantir/dialogue/clients/ReloadingClientFactory.java
```diff
@@ -597,22 +597,34 @@ private static boolean usesProxy(ProxySelector proxySelector, URI uri) {
     }
 
     @Nullable
-    private static URI tryParseUri(@Safe String serviceName, @Unsafe String uri) {
+    private static URI tryParseUri(TaggedMetricRegistry metrics, @Safe String serviceName, @Unsafe String uri) {
         try {
-            return new URI(uri);
+            URI result = new URI(uri);
+            if (result.getHost() == null) {
```
bjlaub, line 603: +1

bjlaub (approved):

PR #2176, opened 2024-02-27
Implement a DNS fallback cache to avoid failures in DNS outages

carterkozak:
==COMMIT_MSG==
Implement a DNS fallback cache to avoid failures in DNS outages
==COMMIT_MSG==

### DIFF
--- a/dialogue-clients/src/main/java/com/palantir/dialogue/clients/CachingFallbackDnsResolver.java
+++ b/dialogue-clients/src/main/java/com/palantir/dialogue/clients/CachingFallbackDnsResolver.java
@@ -1,0 +1,74 @@
+/*
+ * (c) Copyright 2024 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


dialogue-clients/src/main/java/com/palantir/dialogue/clients/CachingFallbackDnsResolver.java
```diff
@@ -0,0 +1,74 @@
...
+    public ImmutableSet<InetAddress> resolve(String hostname) {
+        ImmutableSet<InetAddress> result = delegate.resolve(hostname);
+        if (result.isEmpty()) {
+            ImmutableSet<InetAddress> maybeFallback = fallbackCache.getIfPresent(hostname);
+            if (maybeFallback != null) {
+                lookupFallback.mark();
+                log.info(
+                        "DNS resolution failed for host '{}', however fallback addresses are present in the cache",
+                        UnsafeArg.of("hostname", hostname));
+                return maybeFallback;
```
carterkozak, line 64: Ideally we'd pass through service name info to this point, to understand where we're seeing these failures, however that would take some substantial refactoring

bjlaub:
👍

PR #2179, opened 2024-02-27
add more specific metrics and logging for DNS lookup failures

bjlaub:
## Before this PR
DNS lookups that result in an UnknownHostException are logged, but the exception message is logged unsafely which can make it hard to determine what the actual failure was. Under the hood, the error returned by `getaddrinfo()` is converted to a string via `gai_strerror` and used as the message for the generated `UnknownHostException`, which can be useful in determining what type of failure actually triggered the exception.

## After this PR
Try to match strings generated by `gai_strerror` against the exception message, and log only the `EAI_*` error types and corresponding messages from the operating system safely. Additionally we report a meter metric with the error type (e.g. `EAI_NONAME` for when a DNS lookup fails because the name is not actually known to the nameserver).

==COMMIT_MSG==
add more specific metrics and logging for DNS lookup failures
==COMMIT_MSG==

### DIFF
--- a/dialogue-clients/metrics.md
+++ b/dialogue-clients/metrics.md
@@ -68,6 +68,8 @@ Dialogue DNS metrics.
     - `success`: DNS resolution succeeded using `InetAddress.getAllByName`.
     - `fallback`: DNS resolution using the primary mechanism failed, however addresses were available in the fallback cache.
     - `failure`: No addresses could be resolved for the given hostname.
+- `client.dns.lookupError` (meter): DNS resolver query failures.
+  - `error-type`: Describes the error type returned by getaddrinfo() when lookup fails.
 
 ### client.uri
 Dialogue URI parsing metrics.
--- a/dialogue-clients/metrics.md
+++ b/dialogue-clients/metrics.md
@@ -68,7 +68,7 @@ Dialogue DNS metrics.
     - `success`: DNS resolution succeeded using `InetAddress.getAllByName`.
     - `fallback`: DNS resolution using the primary mechanism failed, however addresses were available in the fallback cache.
     - `failure`: No addresses could be resolved for the given hostname.
-- `client.dns.lookupError` (meter): DNS resolver query failures.
+- `client.dns.failure` (meter): DNS resolver query failures.
   - `error-type`: Describes the error type returned by getaddrinfo() when lookup fails.
 
 ### client.uri
--- a/dialogue-clients/src/main/java/com/palantir/dialogue/clients/DefaultDialogueDnsResolver.java
+++ b/dialogue-clients/src/main/java/com/palantir/dialogue/clients/DefaultDialogueDnsResolver.java
@@ -16,14 +16,18 @@
 
 package com.palantir.dialogue.clients;
 
+import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import com.palantir.dialogue.core.DialogueDnsResolver;
 import com.palantir.logsafe.Preconditions;
+import com.palantir.logsafe.SafeArg;
 import com.palantir.logsafe.UnsafeArg;
 import com.palantir.logsafe.logger.SafeLogger;
 import com.palantir.logsafe.logger.SafeLoggerFactory;
 import java.net.InetAddress;
 import java.net.UnknownHostException;
+import java.util.Map;
+import org.immutables.value.Value;
 
 enum DefaultDialogueDnsResolver implements DialogueDnsResolver {
     INSTANCE;
@@ -41,8 +45,60 @@ enum DefaultDialogueDnsResolver implements DialogueDnsResolver {
             }
             return ImmutableSet.copyOf(results);
         } catch (UnknownHostException e) {
-            log.warn("Unknown host '{}'", UnsafeArg.of("hostname", hostname), e);
+            ExtractedGaiError gaiError = extractGaiErrorString(e);
+            log.warn(
+                    "Unknown host '{}'",
+                    SafeArg.of("gaiErrorMessage", gaiError.getErrorMessage()),
+                    SafeArg.of("gaiErrorType", gaiError.getErrorType()),
+                    UnsafeArg.of("hostname", hostname),
+                    e);
             return ImmutableSet.of();
         }
+    }
+
+    // these strings were taken from glibc-2.39, but likely have not changed in quite a while
+    // strings may be different on BSD systems like macos
+    // TODO(dns): update this list to try to match against known strings on other platforms
+    private static final Map<String, String> EXPECTED_GAI_ERROR_STRINGS = ImmutableMap.<String, String>builder()
+            .put("Address family for hostname not supported", "EAI_ADDRFAMILY")
+            .put("Temporary failure in name resolution", "EAI_AGAIN")
+            .put("Bad value for ai_flags", "EAI_BADFLAGS")
+            .put("Non-recoverable failure in name resolution", "EAI_FAIL")
+            .put("ai_family not supported", "EAI_FAMILY")
+            .put("Memory allocation failure", "EAI_MEMORY")
+            .put("No address associated with hostname", "EAI_NODATA")
+            .put("Name or service not known", "EAI_NONAME")
+            .put("Servname not supported for ai_socktype", "EAI_SERVICE")
+            .put("ai_socktype not supported", "EAI_SOCKTYPE")
+            .put("System error", "EAI_SYSTEM")
+            .put("Processing request in progress", "EAI_INPROGRESS")
+            .put("Request canceled", "EAI_CANCELED")
+            .put("Request not canceled", "EAI_NOTCANCELED")
+            .put("All requests done", "EAI_ALLDONE")
+            .put("Interrupted by a signal", "EAI_INTR")
+            .put("Parameter string not correctly encoded", "EAI_IDN_ENCODE")
+            .put("Result too large for supplied buffer", "EAI_OVERFLOW")
+            .buildOrThrow();
+
+    @Value.Immutable
+    interface ExtractedGaiError {
+        @Value.Parameter
+        String getErrorMessage();
+
+        @Value.Parameter
+        String getErrorType();
+    }
+
+    private static ExtractedGaiError extractGaiErrorString(UnknownHostException exception) {
+        try {
+            for (Map.Entry<String, String> entry : EXPECTED_GAI_ERROR_STRINGS.entrySet()) {
+                if (exception.getMessage() != null && exception.getMessage().contains(entry.getKey())) {
+                    return ImmutableExtractedGaiError.of(entry.getKey(), entry.getValue());
+                }
+            }
+        } catch (Exception e) {
+            return ImmutableExtractedGaiError.of("unknown", "unknown");
+        }
+        return ImmutableExtractedGaiError.of("unknown", "unknown");
     }
 }
--- a/dialogue-clients/src/main/java/com/palantir/dialogue/clients/DefaultDialogueDnsResolver.java
+++ b/dialogue-clients/src/main/java/com/palantir/dialogue/clients/DefaultDialogueDnsResolver.java
@@ -16,7 +16,6 @@
 
 package com.palantir.dialogue.clients;
 
-import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import com.palantir.dialogue.core.DialogueDnsResolver;
 import com.palantir.logsafe.Preconditions;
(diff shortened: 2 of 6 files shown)


dialogue-clients/src/main/java/com/palantir/dialogue/clients/DefaultDialogueDnsResolver.java
```diff
@@ -41,8 +45,60 @@ public ImmutableSet<InetAddress> resolve(String hostname) {
             }
             return ImmutableSet.copyOf(results);
         } catch (UnknownHostException e) {
-            log.warn("Unknown host '{}'", UnsafeArg.of("hostname", hostname), e);
+            ExtractedGaiError gaiError = extractGaiErrorString(e);
+            log.warn(
+                    "Unknown host '{}'",
+                    SafeArg.of("gaiErrorMessage", gaiError.getErrorMessage()),
+                    SafeArg.of("gaiErrorType", gaiError.getErrorType()),
```
carterkozak: Thoughts on adding a `meter` metric tagged with the `errorType` so we can keep an eye on things more broadly?

bjlaub: yeah i think that makes sense, will add

dialogue-clients/src/main/java/com/palantir/dialogue/clients/DefaultDialogueDnsResolver.java
```diff
@@ -41,8 +45,60 @@ public ImmutableSet<InetAddress> resolve(String hostname) {
...
+        @Value.Parameter
+        String getErrorMessage();
+
+        @Value.Parameter
+        String getErrorType();
+    }
+
+    private static ExtractedGaiError extractGaiErrorString(UnknownHostException exception) {
+        try {
+            for (Map.Entry<String, String> entry : EXPECTED_GAI_ERROR_STRINGS.entrySet()) {
```
carterkozak: Should we account for cached failures with an initial check that the message doesn't exactly match the input hostname value? (Perhaps `return ImmutableExtractedGaiError.of("cached", "cached");` in that case)

dialogue-clients/src/main/java/com/palantir/dialogue/clients/DefaultDialogueDnsResolver.java
```diff
@@ -41,8 +45,72 @@ public ImmutableSet<InetAddress> resolve(String hostname) {
...
+            return ImmutableExtractedGaiError.of("null", "null");
+        }
+
+        try {
+            StackTraceElement[] trace = exception.getStackTrace();
+            if (trace.length > 0) {
+                StackTraceElement top = trace[0];
+                if ("java.net.InetAddress$CachedLookup".equals(top.getClassName())) {
+                    return ImmutableExtractedGaiError.of("cached", "cached");
+                }
```
carterkozak: I think the cached case failures can be checked most easily using `Objects.equals(requestedHostname, exception.getMessage())` -- either way we should add a test for this.

dialogue-clients/src/main/java/com/palantir/dialogue/clients/DefaultDialogueDnsResolver.java
```diff
@@ -41,8 +45,72 @@ public ImmutableSet<InetAddress> resolve(String hostname) {
...
+
+                for (Map.Entry<String, String> entry : EXPECTED_GAI_ERROR_STRINGS.entrySet()) {
```
carterkozak: I don't think this loop needs to be predicated on `if (trace.length > 0) {`

dialogue-clients/src/main/java/com/palantir/dialogue/clients/DefaultDialogueDnsResolver.java
```diff
@@ -41,8 +45,72 @@ public ImmutableSet<InetAddress> resolve(String hostname) {
...
+        String getErrorMessage();
+
+        @Value.Parameter
+        String getErrorType();
+    }
+
+    private static ExtractedGaiError extractGaiErrorString(UnknownHostException exception) {
+        if (exception == null) {
+            return ImmutableExtractedGaiError.of("null", "null");
+        }
```
carterkozak: We probably don't need to handle null here since we're calling from our own code on `catch (UnknownHostException`

dialogue-clients/src/main/java/com/palantir/dialogue/clients/DefaultDialogueDnsResolver.java
```diff
@@ -41,8 +45,72 @@ public ImmutableSet<InetAddress> resolve(String hostname) {
...
+            .put("Request canceled", "EAI_CANCELED")
+            .put("Request not canceled", "EAI_NOTCANCELED")
+            .put("All requests done", "EAI_ALLDONE")
+            .put("Interrupted by a signal", "EAI_INTR")
+            .put("Parameter string not correctly encoded", "EAI_IDN_ENCODE")
+            .put("Result too large for supplied buffer", "EAI_OVERFLOW")
+            .buildOrThrow();
+
+    @Value.Immutable
+    interface ExtractedGaiError {
```
carterkozak: I think we can combine the map and type here into a single `enum` e.g.
```java
enum DNS_ERROR {
    EAI_ADDRFAMILY("Address family for hostname not supported"),
    EAI_AGAIN("Temporary failure in name resolution"),
    /*etc*/
    CACHED(), // explicitly avoid setting a substring matcher for 'cached' and 'unknown' special cases
    UNKNOWN()
}

File: dialogue-clients/metrics.md [hunk 7]
```diff
@@ -68,6 +68,8 @@ Dialogue DNS metrics.
     - `success`: DNS resolution succeeded using `InetAddress.getAllByName`.
     - `fallback`: DNS resolution using the primary mechanism failed, however addresses were available in the fallback cache.
     - `failure`: No addresses could be resolved for the given hostname.
+- `client.dns.lookupError` (meter): DNS resolver query failures.
```
carterkozak: let's match the `failure` wording above, perhaps naming this `client.dns.failure`

dialogue-clients/src/main/java/com/palantir/dialogue/clients/DefaultDialogueDnsResolver.java
```diff
@@ -41,8 +49,81 @@ public ImmutableSet<InetAddress> resolve(String hostname) {
...
+            this.errorMessage = Optional.of(errorMessage);
+        }
+
+        @Safe
+        String errorMessage() {
+            return errorMessage.orElseGet(this::name);
+        }
+    }
+
+    private static GaiError extractGaiErrorString(UnknownHostException exception) {
```
carterkozak: perhaps `extractGaiError` since this returns `GaiError`, not `String`

dialogue-clients/src/test/java/com/palantir/dialogue/clients/DefaultDialogueDnsResolverTest.java
```diff
@@ -66,7 +70,42 @@ void malformedIpv6_noBrackets() {
...
+    @Test
+    void unknown_host() {
+        assumeThat(System.getProperty("os.name").toLowerCase().startsWith("linux"))
+                .describedAs("GAI Error Strings are only defined for Linux environments")
+                .isTrue();
+
+        TaggedMetricRegistry registry = new DefaultTaggedMetricRegistry();
+        DialogueDnsResolver resolver = new DefaultDialogueDnsResolver(registry);
+
+        String badHost = "alksdjflajsdlkfjalksjflkadjsf.com";
```
carterkozak: Anyone could [pay ~$10](https://porkbun.com/checkout/search?prb=9261189adc&q=alksdjflajsdlkfjalksjflkadjsf.com&tlds=&idnLanguage=&search=search&csrf_pb=3e031623f4516ff491faa3eac63fdf5a) to break this test.

Let's use something like:
```java
String badHost = UUID.randomUUID() + ".palantir.com";
```

dialogue-clients/src/test/java/com/palantir/dialogue/clients/DefaultDialogueDnsResolverTest.java
```diff
@@ -66,7 +70,42 @@ void malformedIpv6_noBrackets() {
...
+        ImmutableSet<InetAddress> result = resolver.resolve(badHost);
+
+        assertThat(result).isEmpty();
+
+        // should resolve from cache
+        ImmutableSet<InetAddress> result2 = resolver.resolve(badHost);
+        assertThat(result2).isEmpty();
+        ClientDnsMetrics metrics = ClientDnsMetrics.of(registry);
+        assertThat(metrics.lookupError("CACHED").getCount()).isGreaterThan(0);
```
carterkozak: isEqualTo?

dialogue-clients/src/test/java/com/palantir/dialogue/clients/DefaultDialogueDnsResolverTest.java
```diff
@@ -66,7 +70,42 @@ void malformedIpv6_noBrackets() {
...
+
+    @Test
+    void unknown_host_from_cache() {
+        TaggedMetricRegistry registry = new DefaultTaggedMetricRegistry();
+        DialogueDnsResolver resolver = new DefaultDialogueDnsResolver(registry);
+
+        String badHost = "alksdjflajsdlkfjalksjflkadjsf.com";
+        ImmutableSet<InetAddress> result = resolver.resolve(badHost);
+
+        assertThat(result).isEmpty();
```
carterkozak, line 100: let's add `assertThat(metrics.lookupError("EAI_NONAME").getCount()).isEqualTo(1);` in addition to this assert

bjlaub: added

dialogue-clients/src/test/java/com/palantir/dialogue/clients/DefaultDialogueDnsResolverTest.java
bjlaub: done

dialogue-clients/src/test/java/com/palantir/dialogue/clients/DefaultDialogueDnsResolverTest.java
bjlaub: fixed

dialogue-clients/src/main/java/com/palantir/dialogue/clients/DefaultDialogueDnsResolver.java
bjlaub: done

dialogue-clients/metrics.md
bjlaub: fixed

dialogue-clients/src/test/java/com/palantir/dialogue/clients/DefaultDialogueDnsResolverTest.java
carterkozak: I think that only went into one of the two tests using this pattern, we need to update both

dialogue-clients/src/test/java/com/palantir/dialogue/clients/DefaultDialogueDnsResolverTest.java
```diff
@@ -66,7 +71,43 @@ void malformedIpv6_noBrackets() {
...
+        String badHost = UUID.randomUUID() + ".palantir.com";
+        ImmutableSet<InetAddress> result = resolver.resolve(badHost);
+
+        assertThat(result).isEmpty();
+        assertThat(metrics.failure("EAI_NONAME").getCount()).isEqualTo(1);
+
+        // should resolve from cache
+        ImmutableSet<InetAddress> result2 = resolver.resolve(badHost);
+        assertThat(result2).isEmpty();
+        assertThat(metrics.failure("CACHED").getCount()).isGreaterThan(0);
```
carterkozak: `.isEqualTo(1);`

dialogue-clients/src/main/java/com/palantir/dialogue/clients/DefaultDialogueDnsResolver.java
```diff
@@ -41,8 +50,77 @@ public ImmutableSet<InetAddress> resolve(String hostname) {
             }
             return ImmutableSet.copyOf(results);
         } catch (UnknownHostException e) {
-            log.warn("Unknown host '{}'", UnsafeArg.of("hostname", hostname), e);
+            GaiError gaiError = extractGaiError(e, hostname);
+            log.warn(
+                    "Unknown host '{}'",
```
carterkozak:
```suggestion
                    "Unknown host '{}'. {}: {}",
```

carterkozak (approved):

PR #2180, opened 2024-03-01
improvement: Avoid logging at INFO on each DNS request

okushchenko:
I noticed that this log line is getting logged often in my service, most likely every time a DNS resolution is happening, which results in a lot of noise in the logs, making it more difficult to debug production issues.

It doesn't look like this log line belongs at the INFO level, so downgrading it to DEBUG.


carterkozak:
Superseded by #2181

PR #2182, opened 2024-03-03
Enable DNS node discovery by default

carterkozak:
==COMMIT_MSG==
Enable DNS node discovery by default
==COMMIT_MSG==

In previous releases, we kept this off by default, behind a feature flag. We've run significant testing, and are ready to enable by default at this point.


mpritham:
👍

PR #2183, opened 2024-03-04
Dialogue 'no addresses via DNS' exception is more standard

carterkozak:
Exception behavior matches legacy dns resolution mode

==COMMIT_MSG==
Dialogue 'no addresses via DNS' exception is more standard
==COMMIT_MSG==

### DIFF
--- a/dialogue-clients/src/main/java/com/palantir/dialogue/clients/SafeUnknownHostException.java
+++ b/dialogue-clients/src/main/java/com/palantir/dialogue/clients/SafeUnknownHostException.java
@@ -1,0 +1,50 @@
+/*
+ * (c) Copyright 2024 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


dialogue-clients/src/main/java/com/palantir/dialogue/clients/SafeUnknownHostException.java
```diff
@@ -0,0 +1,50 @@
...
+import com.palantir.logsafe.Arg;
+import com.palantir.logsafe.Safe;
+import com.palantir.logsafe.SafeLoggable;
+import com.palantir.logsafe.exceptions.SafeExceptions;
+import java.net.UnknownHostException;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+
+final class SafeUnknownHostException extends UnknownHostException implements SafeLoggable {
```
bjlaub, line 29: is there a reason we can't extract https://github.com/palantir/dialogue/blob/develop/dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/InstrumentedDnsResolver.java#L154 and make it shared?

carterkozak: Ya, I'd meant to comment about the duplication -- unfortunately we don't have a great put that exception where it's not public, but can be shared.

carterkozak: That one lives in the dialogue-apache-hc5-client module, and this one lives in dialogue-clients

bjlaub (approved):

PR #2184, opened 2024-03-04
Use markdown sections in README for linkability

ash211:
## Before this PR
There is no anchor icon next to the production usage and under the hood sections in the README, because they are bold text instead of a nested markdown section.

## After this PR
==COMMIT_MSG==
Use markdown sections in README for linkability
==COMMIT_MSG==

By using the markdown syntax, now github will render a link icon next to the title, and users can create links directly to these sections.

## Possible downsides?
None known


carterkozak (approved):
tyvm

PR #2191, opened 2024-03-08
Dialogue DNS logging provides higher signal with less noise

carterkozak:
==COMMIT_MSG==
Dialogue DNS logging provides higher signal with less noise
==COMMIT_MSG==

### DIFF
--- a/dialogue-clients/src/main/java/com/palantir/dialogue/clients/DnsSupport.java
+++ b/dialogue-clients/src/main/java/com/palantir/dialogue/clients/DnsSupport.java
@@ -22,6 +22,7 @@ import com.google.common.base.Suppliers;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import com.palantir.dialogue.core.DialogueDnsResolver;
 import com.palantir.dialogue.core.DialogueExecutors;
+import com.palantir.logsafe.Unsafe;
 import com.palantir.logsafe.logger.SafeLogger;
 import com.palantir.logsafe.logger.SafeLoggerFactory;
 import com.palantir.refreshable.Disposable;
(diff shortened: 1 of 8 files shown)


dialogue-clients/src/main/java/com/palantir/dialogue/clients/DnsSupport.java
```diff
@@ -106,6 +111,18 @@ static <I> Refreshable<DnsResolutionResults<I>> pollForChanges(
...
+    @Nullable
+    static String tryGetHost(@Unsafe String uriString) {
+        try {
+            URI uri = new URI(uriString);
+            return uri.getHost();
+        } catch (URISyntaxException | RuntimeException e) {
+            log.debug("Failed to parse URI", e);
+            return null;
+        }
+    }
```
carterkozak, line 124: Extracted from a lambda in `DialogueDnsResolutionWorker` to reuse within `describeHostname`. There's a similar method in `ReloadingClientFactory` which doesn't share this implementation because it's where we provide higher priority logging and metrics for invalid URIs -- we don't want to duplicate that N times for a single failure.

dialogue-clients/src/main/java/com/palantir/dialogue/clients/ReloadingClientFactory.java
```diff
@@ -186,7 +186,7 @@ public <T> T getNonReloading(Class<T> clazz, ServiceConfiguration serviceConf) {
         Channel channel = new LiveReloadingChannel(
                 DnsSupport.pollForChanges(
                                 params.dnsNodeDiscovery(),
-                                DnsPollingSpec.SERVICE_CONFIG,
+                                DnsPollingSpec.serviceConfig(channelName),
```
carterkozak, line 189: I updated these so that we'll get more specific metric tags describing ongoing work, and used the info within the background DNS polling task failure logging so we don't have to guess about the unsafe URI.

mpritham (approved):

PR #2192, opened 2024-03-09
Add methods to obtain named per-host clients

pkoenig10:
## Before this PR
There are some useful capabilities not currently supported by the per-host channels:
- The ability to order the channels, in a way that is consistent across node and Dialogue versions
- The ability to compare channel targets before and after a configuration reload
- The ability to determine if a channel target is ourselves

## After this PR
The `PerHostClientFactory` now provides two additional methods, `getNamedPerHostChannels` and `getNamedPerHost`, that return channels keyed by `PerHostTarget`. `PerHostTarget` allows callers to introspect the target and determine whether that target refers to themselves.

### DIFF
--- a/dialogue-clients/src/main/java/com/palantir/dialogue/clients/PerHostTarget.java
+++ b/dialogue-clients/src/main/java/com/palantir/dialogue/clients/PerHostTarget.java
@@ -1,0 +1,62 @@
+/*
+ * (c) Copyright 2024 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.dialogue.clients;
+
+import com.palantir.dialogue.core.TargetUri;
+
+public final class PerHostTarget {
+
+    private final TargetUri targetUri;
+    private final boolean isSelf;
+
+    PerHostTarget(TargetUri targetUri, boolean isSelf) {
+        this.targetUri = targetUri;
+        this.isSelf = isSelf;
+    }
+
+    public TargetUri targetUri() {
+        return targetUri;
+    }
+
+    public boolean isSelf() {
+        return isSelf;
+    }
+
+    @Override
+    public String toString() {
+        return "Target{targetUri='" + targetUri + "', isSelf=" + isSelf + '}';
+    }
+
+    @Override
+    public boolean equals(Object other) {
+        if (this == other) {
+            return true;
+        }
+        if (other == null || getClass() != other.getClass()) {
+            return false;
+        }
+        PerHostTarget perHostTarget = (PerHostTarget) other;
+        return targetUri.equals(perHostTarget.targetUri) && isSelf == perHostTarget.isSelf;
+    }
+
+    @Override
+    public int hashCode() {
+        int result = targetUri.hashCode();
+        result = 31 * result + Boolean.hashCode(isSelf);
+        return result;
+    }
+}
--- a/dialogue-clients/src/main/java/com/palantir/dialogue/clients/PerHostTarget.java
+++ b/dialogue-clients/src/main/java/com/palantir/dialogue/clients/PerHostTarget.java
@@ -21,24 +21,18 @@ import com.palantir.dialogue.core.TargetUri;
 public final class PerHostTarget {
 
     private final TargetUri targetUri;
-    private final boolean isSelf;
 
-    PerHostTarget(TargetUri targetUri, boolean isSelf) {
+    PerHostTarget(TargetUri targetUri) {
         this.targetUri = targetUri;
-        this.isSelf = isSelf;
     }
 
     public TargetUri targetUri() {
         return targetUri;
     }
 
-    public boolean isSelf() {
-        return isSelf;
-    }
-
     @Override
     public String toString() {
-        return "Target{targetUri='" + targetUri + "', isSelf=" + isSelf + '}';
+        return "Target{targetUri='" + targetUri + '}';
     }
 
     @Override
@@ -50,13 +44,11 @@ public final class PerHostTarget {
             return false;
         }
         PerHostTarget perHostTarget = (PerHostTarget) other;
-        return targetUri.equals(perHostTarget.targetUri) && isSelf == perHostTarget.isSelf;
+        return targetUri.equals(perHostTarget.targetUri);
     }
 
     @Override
     public int hashCode() {
-        int result = targetUri.hashCode();
-        result = 31 * result + Boolean.hashCode(isSelf);
-        return result;
+        return targetUri.hashCode();
     }
 }
--- a/dialogue-clients/src/main/java/com/palantir/dialogue/clients/ReloadingClientFactory.java
+++ b/dialogue-clients/src/main/java/com/palantir/dialogue/clients/ReloadingClientFactory.java
@@ -66,10 +66,8 @@ import com.palantir.logsafe.logger.SafeLoggerFactory;
 import com.palantir.refreshable.Refreshable;
 import com.palantir.tritium.metrics.registry.TaggedMetricRegistry;
 import java.net.InetAddress;
-import java.net.NetworkInterface;
 import java.net.Proxy;
 import java.net.ProxySelector;
-import java.net.SocketException;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.security.Provider;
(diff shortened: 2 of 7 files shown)


dialogue-clients/src/main/java/com/palantir/dialogue/clients/PerHostTarget.java
```diff
@@ -0,0 +1,62 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.dialogue.clients;
+
+import com.palantir.dialogue.core.TargetUri;
+
+public final class PerHostTarget {
```
pkoenig10, line 21: We had originally discussed designing `PerHostTarget` to look like:

```java
class PerHostTarget {
    // An opaque hash of the uri and resolvedAddress
    String name;
    boolean isSelf;
}
```

The problem with this approach is that the same logical target may produce different names across Dialogue versions. For example, if we add a new field to `TargetUri` that gets incorporated into the hash. This is bad for a couple reasons:
- Consumers using different Dialogue version will disagree on the sorted order of these channels.
- Consumers may attempt to persist these names and later compare them against names produced by a different Dialogue version.

The only capabilities we require from `PerHostTarget` are that it has a consistent ordering for the same logical target across Dialogue version.

This PR instead accomplishes this by making `TargetUri` implement `Comparable<TargetUri>`. If we add new fields to `TargetUri`, we can implement `compareTo` such that the order remains consistent.

carterkozak: I wanted to avoid leaking uri-info to callers, as dialogue aims to provide an abstraction that makes such info unnecessary. However, I don't think it will be possible to provide enough precision for our users without making this data available in some form. `TargetUri` is used to direct requests to individual upstreams (with greater specificity than a URI), so any future additions to abstract away something like nodes behind a reverse proxy (no plans to do this, just speculating about potential future changes) would have to be represented in `TargetUri`, which would be similarly helpful here -- so there shouldn't be any need to expose similar fields from a new type here. I think retaining the `PerHostTarget` wrapper around `TargetUri` buys us some freedom to add new metadata for per-host-channels in the future if needs be later on.

The `isSelf` accessor was an attempt to avoid avoid leaking these details, but I think if we do expose the `TargetUri`, we needn't expose `isSelf` (that sort of check could become library functionality elsewhere, but it isn't required for the dialogue feature imo). The TargetUri will allow consumers to be more accurate in their own `isSelf` check, for instance when N services are running on the same host, the caller is likely to know which port it's listening on, in order to provide further differentiation from other services on the same host.

dialogue-clients/src/main/java/com/palantir/dialogue/clients/ReloadingClientFactory.java
```diff
@@ -239,46 +243,75 @@ public PerHostClientFactory perHost(String serviceName) {
...
+                                new PerHostTarget(targetUri, isSelf),
+                                cache.getNonReloadingChannel(
+                                        params,
+                                        singleUriServiceConf,
+                                        ImmutableList.of(targetUri),
+                                        channelName,
+                                        OptionalInt.of(i)));
                     }
-                    return list.build();
+                    return map.buildKeepingLast();
```
carterkozak, line 266: I think we want this to fail loudly/catastrophically if something goes wrong

dialogue-clients/src/main/java/com/palantir/dialogue/clients/ReloadingClientFactory.java
```diff
@@ -223,12 +227,12 @@ public PerHostClientFactory perHost(String serviceName) {
         Preconditions.checkNotNull(serviceName, "serviceName");
         String channelName = ChannelNames.reloading(serviceName, params);
 
-        Refreshable<List<DialogueChannel>> perHostDialogueChannels = configurationForService(serviceName)
+        Refreshable<Map<PerHostTarget, Channel>> perHostChannels = configurationForService(serviceName)
```
carterkozak, line 228: Slight preference to internally represent things like this using the immutables type for clarity
```suggestion
        Refreshable<ImmutableMap<PerHostTarget, Channel>> perHostChannels = configurationForService(serviceName)
```

This also has the minor upside of making code like 288 a bit simpler:
```diff
-ImmutableList.copyOf(channels.values())
+channels.values().asList()
```

dialogue-clients/src/main/java/com/palantir/dialogue/clients/ReloadingClientFactory.java
```diff
@@ -239,46 +243,75 @@ public PerHostClientFactory perHost(String serviceName) {
...
                         .collect(ImmutableList.toImmutableList()));
             }
 
+            @Override
+            public <T> Refreshable<Map<PerHostTarget, T>> getNamedPerHost(Class<T> clientInterface) {
+                return perHostChannels.map(channels -> channels.entrySet().stream()
+                        .collect(ImmutableMap.toImmutableMap(
+                                Map.Entry::getKey,
+                                entry -> Reflection.callStaticFactoryMethod(
+                                        clientInterface, entry.getValue(), params.runtime()))));
```
carterkozak, line 281: slight preference to avoid `stream()` where possible
```java
ImmutableMap.copyOf(Maps.transformValues(
    channels,
    channel -> Reflection.callStaticFactoryMethod(clientInterface, entry.getValue(), params.runtime())));
```

dialogue-clients/src/main/java/com/palantir/dialogue/clients/ReloadingClientFactory.java
```diff
@@ -239,46 +243,75 @@ public PerHostClientFactory perHost(String serviceName) {
...
+            public Refreshable<Map<PerHostTarget, Channel>> getNamedPerHostChannels() {
+                return perHostChannels;
             }
 
             @Override
             public <T> Refreshable<List<T>> getPerHost(Class<T> clientInterface) {
-                return getPerHostChannels().map(channels -> channels.stream()
-                        .map(chan -> Reflection.callStaticFactoryMethod(clientInterface, chan, params.runtime()))
+                return perHostChannels.map(channels -> channels.values().stream()
+                        .map(channel -> Reflection.callStaticFactoryMethod(clientInterface, channel, params.runtime()))
```
carterkozak: Thoughts on delegation here?
```java
return getNamedPerHost(clientInterface).map(result -> result.values().asList());
```

dialogue-clients/src/main/java/com/palantir/dialogue/clients/ReloadingClientFactory.java
pkoenig10: Why is this code better? We have to allocate some temporary variables in either case.

We're already using streams in a number of places in this class.

carterkozak: Streams are difficult for hotspot to optimize, and I find to-map stream collectors a bit more difficult to read at a glance (this makes it more clear to devs that the keys aren't being modified, and the transformValues call makes it clear to the internals that the resulting ImmutableMap can be pre-sized).

Almost certainly won't make a measurable difference in practice, my goal in the comment is to share how I'm thinking about this, it's entirely up to you whether or not you'd like to change the snippet :-)

dialogue-clients/src/main/java/com/palantir/dialogue/clients/ReloadingClientFactory.java
pkoenig10: Is this safe? There's no guarantee that the `TargetUri` values are unique.

dialogue-clients/src/main/java/com/palantir/dialogue/clients/ReloadingClientFactory.java
pkoenig10: Should we make the existing methods default methods on `PerHostClientFactory`? Or would you prefer to keep the implementation here?

pkoenig10: Delegation would require me to change the return type declared in `PerHostClientFactory` to be `ImmutableMap`, which probably isn't what we want to do here.

It's also nice to have one less refreshable subscriber here.

dialogue-clients/src/main/java/com/palantir/dialogue/clients/ReloadingClientFactory.java
carterkozak: Hmm, I suppose an example might look something like this:
```yaml
uris:
  - https://target/api
  - https://target/api
```

In such a case, I think I'd prefer if we pruned out duplicates earlier on, otherwise we risk effectively doubling the concurrency limits on repeated values. I think we can improve things across the board by updating `ReloadingClientFactory.getTargetUris` with something like this to retain order, but filter duplicates:
```diff
-return ImmutableList.copyOf(targetUris);
+return ImmutableSet.copyOf(targetUris).toList();
```

dialogue-clients/src/main/java/com/palantir/dialogue/clients/ReloadingClientFactory.java
carterkozak: > Delegation would require me to change the return type declared in PerHostClientFactory to be ImmutableMap, which probably isn't what we want to do here.

Ah, we didn't handle PECS (Producer Extends Consumer Super) very well on the `PerHostClientFactory` existing interface methods -- ideally we may have declared something like `Refreshable<? extends List<? extends Channel>> getPerHostChannels();` for greater flexibility in the implementation, however that does force callers who don't use the `var` keyword to declare a fair bit more cruft.

Happy to leave this as is

dialogue-clients/src/main/java/com/palantir/dialogue/clients/ReloadingClientFactory.java
pkoenig10: This runs into the problem mentioned below in https://github.com/palantir/dialogue/pull/2192#discussion_r1520152894.

You can't return a `Refreshable<ImmutableMap>` from a method that returns `Refreshable<Map>`.

carterkozak (dismissed):
lgtm, thanks!

carterkozak (approved):

PR #2195, opened 2024-03-11
Add PerHostTarget builder

pkoenig10:
Small follow-up from https://github.com/palantir/dialogue/pull/2192

Without providing a builder it's impossible to create an external implementation of `PerHostClientFactory`.


pkoenig10:
Closing until we have a legitimate use case for external implementations.

PR #2221, opened 2024-03-22
Expose a TypeMarker factory to wrap an arbitrary Type

carterkozak:
This mirrors a change we've just introduced to conjure-undertow here: https://github.com/palantir/conjure-java/pull/2264

Dialogue takes advantage of this new factory method in ConjureBodySerDe to avoid holding class references in serializer caches, which introduced some friction in our dns-polling lifecycle management. I haven't reverted the change to specify `weakKeys` in the caches, however the keys should never be freed unless classes are unloaded (e.g. plugins).

==COMMIT_MSG==
Expose a TypeMarker factory to wrap an arbitrary Type
==COMMIT_MSG==

### DIFF
--- a/dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ConjureBodySerDe.java
+++ b/dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ConjureBodySerDe.java
@@ -40,6 +40,7 @@ import com.palantir.logsafe.logger.SafeLoggerFactory;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.lang.reflect.Type;
 import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ConjureBodySerDe.java
```diff
@@ -112,13 +113,13 @@ private ImmutableList<Encoding> sortByWeight(List<WeightedEncoding> encodings) {
...
     public <T> Serializer<T> serializer(TypeMarker<T> token) {
-        return (Serializer<T>) serializers.get(token);
+        return (Serializer<T>) serializers.get(token.getType());
     }
 
     @Override
     @SuppressWarnings("unchecked")
     public <T> Deserializer<T> deserializer(TypeMarker<T> token) {
-        return (Deserializer<T>) deserializers.get(token);
+        return (Deserializer<T>) deserializers.get(token.getType());
```
carterkozak, line 122: Changes to this file would also make sense in a separate PR

pkoenig10 (approved):

PR #2252, opened 2024-04-22
Implement a tag to opt out of per-endpoint limits

carterkozak:
## Before this PR
No way to opt out of the per-endpoint concurrency limiters

## After this PR
==COMMIT_MSG==
Implement a tag to opt out of per-endpoint limits
==COMMIT_MSG==

## Possible downsides?
This could be misused, I suppose? I'd prefer to make the option available for the edge cases where it may be useful, rather than harm functionality due to one or two degenerate cases.

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueChannel.java
@@ -165,6 +165,9 @@ public final class DialogueChannel implements Channel, EndpointChannelFactory {
                         new TraceEnrichingChannel(channel, DialogueTracing.tracingTags(cf, uriIndexForInstrumentation));
                 channel = cf.isConcurrencyLimitingEnabled()
                         ? new ChannelToEndpointChannel(endpoint -> {
+                            if (endpoint.tags().contains("dialogue-no-endpoint-limit")) {
+                                return tracingChannel;
+                            }
                             LimitedChannel limited = ConcurrencyLimitedChannel.createForEndpoint(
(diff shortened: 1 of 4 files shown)


dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueChannel.java
```diff
@@ -165,6 +165,9 @@ public DialogueChannel build() {
                         new TraceEnrichingChannel(channel, DialogueTracing.tracingTags(cf, uriIndexForInstrumentation));
                 channel = cf.isConcurrencyLimitingEnabled()
                         ? new ChannelToEndpointChannel(endpoint -> {
+                            if (endpoint.tags().contains("dialogue-no-endpoint-limit")) {
```
carterkozak: Open to suggestions on the name. I wanted to clearly namespace it to dialogue, rather than all clients, because this is very much an implementation detail of dialogue.

I don't expect this feature to be used very often, so the name may not be terribly important.

pkoenig10: I don't have a strong opinion. I might prefer something to make it clear that this is for concurrency limiting as opposed to request size, etc.

```suggestion
                            if (endpoint.tags().contains("dialogue-disable-endpoint-concurrency-limiting")) {
```

pkoenig10 (dismissed):

dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueChannel.java
carterkozak: I like it, thanks! Updated

pkoenig10:
👍

PR #2254, opened 2024-04-23
Provide a consistent exception when no dns records are resolved

carterkozak:
==COMMIT_MSG==
Provide a consistent exception when no dns records are resolved
==COMMIT_MSG==


carterkozak:
do-not-merge since we're rewriting a chunk of dialogue target refreshing code

PR #2259, opened 2024-04-29
Dialogue queued request spans describe the queue type

carterkozak:
==COMMIT_MSG==
Dialogue queued request spans describe the queue type
==COMMIT_MSG==

Previously we couldn't distinguish per-endpoint and per-channel queues from trace graphs alone.


dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
```diff
@@ -78,9 +85,15 @@ final class QueuedChannel implements Channel {
     // avoid creating unnecessary data.
     private volatile boolean shouldRecordQueueMetrics;
 
-    QueuedChannel(LimitedChannel delegate, String channelName, QueuedChannelInstrumentation metrics, int maxQueueSize) {
+    QueuedChannel(
+            LimitedChannel delegate,
+            @Safe String channelName,
+            @Safe String queueType,
```
priyanshukm, line 91: Is queueType basically channelType ? are queue and channel being used interchangeably here ?

carterkozak: nope -- the diagram in the readme describes two types of queues (it doesn't go into much detail about sticky, but you can ignore that for most use-cases). There are two queues each request progresses through before it's sent to a remote server.

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
```diff
@@ -437,4 +452,14 @@ public Timer requestQueuedTime() {
             return requestQueuedTimeSupplier.get();
         }
     }
+
+    private enum QueuedChannelTagTranslator implements TagTranslator<QueuedChannel> {
```
mpritham, line 456: I think an alternative is to pass in a `Map<String, String>` into `span().complete(...)`. I'm not actually sure what would be more performant. Using this enum is cleaner though.

mpritham (approved):
lgtm!

PR #2263, opened 2024-05-01
Dialogue fails fast when existing channels are revoked from DNS

carterkozak:
We don't want to spend ages retrying requests to dead nodes, where we expect every attempt to result in a connect timeout. It's much better to fail quickly rather than retry+backoff for minutes.

==COMMIT_MSG==
Dialogue fails fast when existing channels are revoked from DNS
==COMMIT_MSG==

## Possible downsides?
if dns entries disappear and reappear, we may have failed requests for nothing. It's possible dns entires may be revoked while requests are still routable, but I don't think we should prioritize this case as it's generally not something we expect to support.


bjlaub (approved):
lgtm

PR #2274, opened 2024-05-06
Avoid expensive refreshable creation in a potentially hot path

carterkozak:
==COMMIT_MSG==
Avoid expensive refreshable creation in a potentially hot path
==COMMIT_MSG==

### DIFF
--- a/dialogue-clients/src/main/java/com/palantir/dialogue/clients/ReloadingClientFactory.java
+++ b/dialogue-clients/src/main/java/com/palantir/dialogue/clients/ReloadingClientFactory.java
@@ -482,10 +482,15 @@ final class ReloadingClientFactory implements DialogueClients.ReloadingFactory {
         public Supplier<Channel> stickyChannels() {
             return () -> this;
         }
+
+        @Override
+        public EndpointChannel endpoint(Endpoint _endpoint) {
+            return _request -> Futures.immediateFailedFuture(exceptionSupplier.get());
+        }
     }
 
     /* Abstracts away DialogueChannel so that we can handle no-service/no-uri case in #getInternalDialogueChannel. */
-    private interface InternalDialogueChannel extends Channel {
+    private interface InternalDialogueChannel extends Channel, EndpointChannelFactory {
         Supplier<Channel> stickyChannels();
     }
 
@@ -500,6 +505,11 @@ final class ReloadingClientFactory implements DialogueClients.ReloadingFactory {
         @Override
         public ListenableFuture<Response> execute(Endpoint endpoint, Request request) {
             return dialogueChannel.execute(endpoint, request);
+        }
+
+        @Override
+        public EndpointChannel endpoint(Endpoint endpoint) {
+            return dialogueChannel.endpoint(endpoint);
         }
 
         @Override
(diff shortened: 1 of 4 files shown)


dialogue-core/src/main/java/com/palantir/dialogue/core/ContentDecodingChannel.java
```diff
@@ -81,7 +81,9 @@ private static BooleanSupplier shouldSendAcceptGzip(Config cf, Endpoint endpoint
         // is handled by the client. Note that this will also opt out of response
         // compression when the target host resolves to multiple IP addresses.
         if (cf.mesh() == MeshMode.DEFAULT_NO_MESH) {
-            return cf.uris().map(targets -> targets.size() == 1)::get;
+            // Avoid using refreshable here, as this can be called very frequently, and
+            // refreshable.map can be expensive.
+            return () -> cf.uris().get().size() == 1;
```
carterkozak, line 86: We don't want to create a refreshable.map target for each endpoint. We could extract a refreshable.map to a higher layer, but this is probably simpler (list.size is just a field read)

dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueChannel.java
```diff
@@ -178,16 +172,20 @@ public DialogueChannel build() {
...
                         SafeArg.of("channel", cf.channelName()),
                         UnsafeArg.of("uris", cf.clientConf().uris()),
                         SafeArg.of("numUris", cf.clientConf().uris().size()),
-                        UnsafeArg.of("targets", current),
-                        SafeArg.of("numTargets", current.size()));
-                return NodeSelectionStrategyChannel.create(cf, current);
+                        UnsafeArg.of("targets", targetUris),
+                        SafeArg.of("numTargets", targetUris.size()));
+                ImmutableList<LimitedChannel> targetChannels = createHostChannels(cf, targetUris);
+                return NodeSelectionStrategyChannel.create(cf, targetChannels);
```
carterkozak, line 188: combined two refreshable.map calls since we're really doing one thing here, should be easier to follow

dialogue-clients/src/main/java/com/palantir/dialogue/clients/ReloadingClientFactory.java
```diff
@@ -482,10 +482,15 @@ public String toString() {
...
+
+        @Override
+        public EndpointChannel endpoint(Endpoint _endpoint) {
+            return _request -> Futures.immediateFailedFuture(exceptionSupplier.get());
+        }
     }
 
     /* Abstracts away DialogueChannel so that we can handle no-service/no-uri case in #getInternalDialogueChannel. */
-    private interface InternalDialogueChannel extends Channel {
+    private interface InternalDialogueChannel extends Channel, EndpointChannelFactory {
```
carterkozak, line 493: This is the most important change, we have a lot of code which attempts to cast Channel to EndpointChannelFactory and use `endpoint(endpoint)` on the result, failing that it will call the `Channel.execute(Endpoint,Request)` method, which internally delegates to the preferred `endpoint` method. Unfortunately that ends up re-computing a great deal of info on each invocation.

This cruft exists to avoid breaking dialogue API :-/

bjlaub (dismissed):

bjlaub:
👍

PR #2276, opened 2024-05-08
RetryOtherValidatingChannel handles mesh URIs

carterkozak:
In a previous change, we removed special handling of mesh uris because the configuration URIs weren't used in practice, in favor of the TargetUris list, which removes mesh-prefixes automatically. One edge case we failed to deal with was mesh-uris impact on RetryOtherValidatingChannel. We've updated the implementation to handle mesh-uris, and out of an abundance of caution opted to strip mesh prefixes from the configuration used elsewhere to avoid potential impact.

==COMMIT_MSG==
RetryOtherValidatingChannel handles mesh URIs
==COMMIT_MSG==


dialogue-clients/src/test/java/com/palantir/dialogue/otherpackage/DialogueClientsIntegrationTest.java
```diff
@@ -341,6 +341,21 @@ void test_unlimited_timeouts() {
...
+                .withUserAgent(TestConfigurations.AGENT)
+                .withMaxNumRetries(0)
+                .getNonReloading(
+                        SampleServiceBlocking.class,
+                        ServiceConfiguration.builder()
+                                .addUris("mesh-" + getUri(undertow))
+                                .security(TestConfigurations.SSL_CONFIG)
+                                .build());
+        assertThatCode(client::voidToVoid).as("request should succeed").doesNotThrowAnyException();
+    }
```
carterkozak: Note that this passes on develop, but I figured we should have full test coverage of this case just in case!

dialogue-clients/src/test/java/com/palantir/dialogue/otherpackage/DialogueClientsIntegrationTest.java
```diff
@@ -341,6 +341,21 @@ void test_unlimited_timeouts() {
                 .doesNotThrowAnyException();
     }
 
+    @Test
+    void test_mesh_uri() {
+        undertowHandler = exchange -> exchange.setStatusCode(204);
+        SampleServiceBlocking client = DialogueClients.create(Refreshable.only(null))
+                .withUserAgent(TestConfigurations.AGENT)
+                .withMaxNumRetries(0)
```
carterkozak: This line is unnecessary
```suggestion
```

bjlaub (approved):

PR #2279, opened 2024-05-13
Add support for routing through an HTTPS proxy

### LINKED ISSUE #2278: Add support for routing through HTTPS proxies
Currently, dialogue does not support connecting to HTTPS proxies. These are HTTP proxies where the HTTP CONNECT request to the proxy is also secured with its own TLS handshake.

Eric-Alvarez:
Closes https://github.com/palantir/dialogue/issues/2278
## Before this PR

HTTPS proxies are not supported in dialogue
## After this PR
==COMMIT_MSG==

==COMMIT_MSG==
If an HTTPS proxy is configured then dialogue will be able to successfully route through it


dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/HttpsProxyDefaultRoutePlanner.java
```diff
@@ -0,0 +1,102 @@
...
+import org.apache.hc.client5.http.impl.routing.DefaultRoutePlanner;
+import org.apache.hc.client5.http.impl.routing.SystemDefaultRoutePlanner;
+import org.apache.hc.core5.http.HttpException;
+import org.apache.hc.core5.http.HttpHost;
+import org.apache.hc.core5.http.protocol.HttpContext;
+
+/**
+ * Identical to {@link SystemDefaultRoutePlanner} but adds support for connecting to an HTTPS proxy.
+ */
+public final class HttpsProxyDefaultRoutePlanner extends DefaultRoutePlanner {
```
eric-alvarez: I would have preferred to override SystemDefaultRoutePlanner in case there's any changes there in the future, but the proxySelector is private so I would override `determineProxy` and copy everything anyway

carterkozak: This class should be package private because it's an implementation detail of dialogue -- we don't want folks to access this directly

eric-alvarez: Changed

changelog/@unreleased/pr-2279.v2.yml
```diff
@@ -0,0 +1,7 @@
+type: feature
+feature:
+  description: |-
+    Add support for routing through HTTPS proxies in the apache client when configured
+    <!-- User-facing outcomes this PR delivers -->
```
carterkozak: nit: delete this line

eric-alvarez: Deleted, does changelog bot work differently here 🤔

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/HttpsProxyDefaultRoutePlanner.java
```diff
@@ -0,0 +1,102 @@
...
+import java.util.List;
+import javax.annotation.CheckForNull;
+import org.apache.hc.client5.http.impl.routing.DefaultRoutePlanner;
+import org.apache.hc.client5.http.impl.routing.SystemDefaultRoutePlanner;
+import org.apache.hc.core5.http.HttpException;
+import org.apache.hc.core5.http.HttpHost;
+import org.apache.hc.core5.http.protocol.HttpContext;
+
+/**
+ * Identical to {@link SystemDefaultRoutePlanner} but adds support for connecting to an HTTPS proxy.
```
carterkozak, line 34: Let's link the original source [here](https://github.com/apache/httpcomponents-client/blob/rel/v5.3.1/httpclient5/src/main/java/org/apache/hc/client5/http/impl/routing/SystemDefaultRoutePlanner.java) and describe that the original is licensed under apache 2.0.

eric-alvarez: A bit confused, it is linked, did you want it linked differently? Or you want a link to http://www.apache.org/licenses/LICENSE-2.0

Added a line below "Original version is licensed under Apache License, Version 2.0."

carterkozak: I'd like to have a link to the original, at the version used at the point in time we modified from. This way if there are future changes, we can tell whether or not they may be impactful. The javadoc link will point toward whatever version we happen to depend on at any point in time

eric-alvarez: Makes sense, updated the docs

carterkozak (approved):
thanks!

PR #2348, opened 2024-09-16
Test latest httpclient release

carterkozak:
==COMMIT_MSG==
Test latest httpclient release
==COMMIT_MSG==


dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientChannels.java
```diff
@@ -485,40 +463,56 @@ public CloseableClient build() {
...
+                            instrumentedDnsResolver,
+                            RegistryBuilder.<TlsSocketStrategy>create()
+                                    .register(URIScheme.HTTPS.id, tlsStrategy)
+                                    .build()) {
+                        private static final String CONNECT_BEGAN_ATTRIBUTE = "onBeforeSocketConnectNanoTime";
+
+                        @Override
+                        protected void onBeforeSocketConnect(HttpContext httpContext, HttpHost endpointHost) {
+                            super.onBeforeSocketConnect(httpContext, endpointHost);
+                            httpContext.setAttribute(CONNECT_BEGAN_ATTRIBUTE, System.nanoTime());
```
carterkozak, line 495: I don't love boxing this `long`->`Long`.
The `onBefore`/`onAfter` methods don't allow us to collect timing info when requests fail either, since `onAfterSocketConnect` is never called in that case

PR #2360, opened 2024-09-23
fix #1076: Dialogue allows colons in http paths and query parameters

### LINKED ISSUE #1076: URL validation fails when absolute path contains a colon
## What happened?
Dialogue will fail to initialize a client for a URL of the form `https://example.com:443/foo.bar:1/baz`, due to the colon character, however https://tools.ietf.org/html/rfc3986#section-3.3 indicates this should be allowed as long as the path is not a relative one, which these are not. RFC is a bit tough to parse, but I checked with some other URL validators and this form appears valid. It also works for go clients.

issue is in this fn https://github.com/palantir/dialogue/blob/7d12856c839a1506c2c57e213701c35d2ff1a528/dialogue-core/src/main/java/com/palantir/dialogue/core/BaseUrl.java#L216

allow `:` in URL paths

carterkozak:
This is required by some GCP APIs, where encoded colons are not respected.

Most places where we'd end up with colons in URIs are within path variable values, where we'd expect values to be handled correctly regardless of percent encoding. For the most part, percent encoding should work fine unless a server uses precise matching on uri segments, where we run into issues.

[rfc3986 section 3.3](https://datatracker.ietf.org/doc/html/rfc3986#section-3.3) clearly allows colon as a pchar, however not all servers are necessarily fully compliant with the rfc. We've had issues with specific sub-delimiters in the past. This shouldn't be terribly risky, esepcially given this implementaiton has never allowed unencoded colons before, and we have no evidence to suggest they should be problematic.

## After this PR
==COMMIT_MSG==
fix #1076: Dialogue allows colons in http paths and query parameters
==COMMIT_MSG==

## Possible downsides?
It's possible that some servers, proxies, etc do not handle unencoded colons well in paths.

## Behavior of common http clients:

Note that the default behavior of these clients doesn't necessarily mean that it's unsafe to do otherwise, however we should tread carefully around nonstandard behavior.

### 🔴 Python 3.9.6 urllib.parse
```python
import urllib.parse
urllib.parse.quote('foo:bar')
'foo%3Abar'
```

### 🔴 Apache HttpComponents (httpcore-5.3) UriBuilder

```java
System.out.println(new org.apache.hc.core5.net.URIBuilder()
        .setScheme("https")
        .setHost("localhost")
        .setPort(8443)
        .appendPathSegments(Collections.singletonList("foo:bar"))
        .toString());
// Produces: https://localhost:8443/foo%3Abar
```

### 🟢 Curl 8.10.1

caveat -- curl sends what you give it with the exception of unicode characters. Not clear that this is a reasonable test, since the raw URI is provided as an input.

```bash
curl 'http://localhost:8443/foo:bar'

GET /foo:bar HTTP/1.1
Host: localhost:8443
User-Agent: curl/8.10.1
Accept: */*
```

### 🟢 OkHttp

similar to curl, the uri is provided as an input, and it doesn't further escape colons

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/BaseUrl.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/BaseUrl.java
@@ -239,16 +239,37 @@ public final class BaseUrl {
     static class UrlEncoder {
         private static final CharMatcher DIGIT = CharMatcher.inRange('0', '9');
         private static final CharMatcher ALPHA = CharMatcher.inRange('a', 'z').or(CharMatcher.inRange('A', 'Z'));
+
+        // unreserved = ALPHA / DIGIT / "-" / "." / "_" / "~"
         private static final CharMatcher UNRESERVED = DIGIT.or(ALPHA).or(CharMatcher.anyOf("-._~"));
+
+        // sub-delims = "!" / "$" / "&" / "'" / "(" / ")" / "*" / "+" / "," / ";" / "="
         private static final CharMatcher SUB_DELIMS = CharMatcher.anyOf("!$&'()*+,;=");
         private static final CharMatcher IS_HOST = UNRESERVED.or(SUB_DELIMS);
-        private static final CharMatcher IS_P_CHAR = UNRESERVED;
-        private static final CharMatcher IS_PATH = UNRESERVED.or(SUB_DELIMS).or(CharMatcher.anyOf("/"));
+
         // The RFC permits percent-encoding any character. We also percent encode sub-delimiters to avoid
         // incompatibilities with http specification beyond the general URI definition per
         // https://tools.ietf.org/html/rfc3986#section-3.3
         // > URI producing applications often use the reserved characters allowed in a segment to
         // > delimit scheme-specific or dereference-handler-specific subcomponents.
+
+        // If memory serves, we've encountered issues with '+' being interpreted by some servers as an
+        // encoded space, and ';' in paths being interpreted as the beginning of http matrix parameters.
+
+        // pchar = unreserved / pct-encoded / sub-delims / ":" / "@"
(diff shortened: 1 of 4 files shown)


dialogue-core/src/main/java/com/palantir/dialogue/core/BaseUrl.java
```diff
@@ -239,16 +239,37 @@ public String toString() {
...
+        // however we have updated it from 'UNRESERVED' to allow colons, and would like to
+        // make changes as slowly as possible.
+        private static final CharMatcher IS_P_CHAR = UNRESERVED.or(CharMatcher.is(':'));
+
+        // For historical reasons, we allow sub-delims in the base uri if provided directly,
+        // but escape them in paths we create/encode. Base-uri paths tend to be simple, so I wouldn't
+        // expect the sub-delims to be relevant here in the vast majority of cases. With sufficient
+        // research and testing, we should incorporate relevant sub-delims into IS_P_CHAR and update this
+        // to: 'IS_P_CHAR.or(CharMatcher.is('/'))'.
+        private static final CharMatcher IS_PATH = IS_P_CHAR.or(SUB_DELIMS).or(CharMatcher.is('/'));
```
carterkozak, line 270: I updated `IS_PATH` to be derived from `IS_P_CHAR` so we don't have to worry about additions to `IS_P_CHAR` getting out of sync. This matcher is only used on the input base-url path component, not subsequent sections.

bjlaub:
Shall we update the `encodesQueryParams` test to ensure colons are not encoded there as well? https://github.com/palantir/dialogue/blob/develop/dialogue-core/src/test/java/com/palantir/dialogue/core/UrlBuilderTest.java#L133-L138

carterkozak:
good call, updated!

bjlaub:
👍 lgtm

PR #2366, opened 2024-09-30
Allow non-constant endpoints to opt-in to timings

pkoenig10:
Related to https://github.com/palantir/dialogue/pull/1414

Motivated by https://github.com/palantir/conjure-java-runtime/pull/2973#discussion_r1781356606

There may be `Endpoint` implementation that are not constant, but whose cardinality is bounded through other means. This allows those `Endpoint` implementation to opt-in to timing metrics.

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueChannel.java
@@ -253,7 +253,8 @@ public final class DialogueChannel implements Channel, EndpointChannelFactory {
                 channel = new RangeAcceptsIdentityEncodingChannel(channel);
                 channel = ContentEncodingChannel.of(channel, endpoint);
                 channel = TracedChannel.create(cf, channel, endpoint);
-                if (ChannelToEndpointChannel.isConstant(endpoint)) {
+                if (ChannelToEndpointChannel.isConstant(endpoint)
+                        || endpoint.tags().contains("dialogue-enable-endpoint-timing")) {
                     // Avoid producing metrics for non-constant endpoints which may produce
                     // high cardinality.
                     channel = TimingEndpointChannel.create(cf, channel, endpoint);


dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueChannel.java
```diff
@@ -253,7 +253,8 @@ private static EndpointChannelFactory createEndpointChannelFactory(Channel multi
                 channel = new RangeAcceptsIdentityEncodingChannel(channel);
                 channel = ContentEncodingChannel.of(channel, endpoint);
                 channel = TracedChannel.create(cf, channel, endpoint);
-                if (ChannelToEndpointChannel.isConstant(endpoint)) {
+                if (ChannelToEndpointChannel.isConstant(endpoint)
+                        || endpoint.tags().contains("dialogue-enable-endpoint-timing")) {
```
carterkozak, line 257: I suppose we could probably update `ChannelToEndpointChannel` to use a caffeine cache with weak keys, and get rid of the `isConstant` check altogether once the feign client implementation reuses endpoint values -- worst case is weakref churn and metric fanout, but we don't expect these things to be entirely unbounded. When Endpoints aren't reused, this method is called for every request, so looking up the timers in a large TaggedMetricRegistry can be nontrivial, but code that expects decent performance shouldn't stray far from the standard path.

pkoenig10: I think I'm in favor of this proposal for a couple reasons:
- Even in the current implementation, a poorly implemented client with unbounded service/endpoint names could blow up this `ChannelToEndpointChannel` cache. Using a Caffeine cache would allow to avoid this by use weak keys and/or bounding the cache.
- `ChannelToEndpointChannel` assumes that non-constant endpoints are uniquely identified by `(service, endpoint)`, but this isn't correct, for the reasons mentioned in https://github.com/palantir/conjure-java-runtime/pull/2973#discussion_r1781373921. It would probably be good to switch to a cache for this reason alone.

pkoenig10: I've put up https://github.com/palantir/dialogue/pull/2367. Let me know if this isn't what you were imagining.

carterkozak:
Closing in favor of #2367

PR #2367, opened 2024-09-30
Use cache with weak keys in ChannelToEndpointChannel

pkoenig10:
See https://github.com/palantir/dialogue/pull/2366#discussion_r1781693515.

This also effectively reverts https://github.com/palantir/dialogue/pull/1414 now that we've removed `isConstant`.

We probably should not merge this until https://github.com/palantir/conjure-java-runtime/pull/2973 is merged.

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/ChannelToEndpointChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/ChannelToEndpointChannel.java
@@ -16,54 +16,25 @@
 
 package com.palantir.dialogue.core;
 
+import com.github.benmanes.caffeine.cache.Caffeine;
+import com.github.benmanes.caffeine.cache.LoadingCache;
 import com.google.common.util.concurrent.ListenableFuture;
 import com.palantir.dialogue.Channel;
 import com.palantir.dialogue.Endpoint;
 import com.palantir.dialogue.Request;
 import com.palantir.dialogue.Response;
-import java.util.Map;
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.function.Function;
+import java.util.function.BiFunction;
 
 final class ChannelToEndpointChannel implements Channel {
 
-    private final Function<Endpoint, Channel> adapter;
-    private final Map<Object, Channel> cache;
+    private final LoadingCache<Endpoint, Channel> cache;
 
-    ChannelToEndpointChannel(Function<Endpoint, Channel> adapter) {
-        this.adapter = adapter;
-        this.cache = new ConcurrentHashMap<>();
+    ChannelToEndpointChannel(Channel channel, BiFunction<Channel, Endpoint, Channel> loader) {
+        this.cache = Caffeine.newBuilder().weakKeys().build(endpoint -> loader.apply(channel, endpoint));
     }
 
     @Override
     public ListenableFuture<Response> execute(Endpoint endpoint, Request request) {
-        return channelFor(endpoint).execute(endpoint, request);
-    }
-
-    private Channel channelFor(Endpoint endpoint) {
-        return cache.computeIfAbsent(key(endpoint), _key -> adapter.apply(endpoint));
-    }
-
-    /**
-     * Constant {@link Endpoint endpoints} may be safely used as cache keys, as opposed to dynamically created
-     * {@link Endpoint} objects which would result in a memory leak.
-     */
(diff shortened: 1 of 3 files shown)


dialogue-core/src/main/java/com/palantir/dialogue/core/ChannelToEndpointChannel.java
```diff
@@ -16,23 +16,21 @@
...
 final class ChannelToEndpointChannel implements Channel {
 
-    private final Function<Endpoint, Channel> adapter;
-    private final Map<Object, Channel> cache;
+    private final LoadingCache<Endpoint, Channel> cache;
 
-    ChannelToEndpointChannel(Function<Endpoint, Channel> adapter) {
-        this.adapter = adapter;
-        this.cache = new ConcurrentHashMap<>();
+    ChannelToEndpointChannel(Channel channel, BiFunction<Channel, Endpoint, Channel> loader) {
```
pkoenig10: I tweaked the loader here to take the delegate channel as an argument rather than capturing it. This makes the usage a bit cleaner by avoiding the need to introduce a final `tracingChannel` variable for capturing.

dialogue-core/src/main/java/com/palantir/dialogue/core/ChannelToEndpointChannel.java
```diff
@@ -16,23 +16,21 @@
...
+        this.cache = Caffeine.newBuilder().weakKeys().build(endpoint -> loader.apply(channel, endpoint));
```
pkoenig10: Do we want to put a large maximum size here for additional protection? Maybe like 10,000?

dialogue-core/src/main/java/com/palantir/dialogue/core/ChannelToEndpointChannel.java
carterkozak: I think I'd prefer to revert this piece -- it's nice to avoid the capture, but it makes it harder to understand what the function represents within this class. The fact the original channel may be returned is an implementation detail of a function passed to this class.

dialogue-core/src/main/java/com/palantir/dialogue/core/ChannelToEndpointChannel.java
carterkozak: That sounds reasonable to me

dialogue-core/src/main/java/com/palantir/dialogue/core/ChannelToEndpointChannel.java
pkoenig10: Yeah this makes sense. I can accomplish what I was trying to here without this change.

dialogue-core/src/main/java/com/palantir/dialogue/core/ChannelToEndpointChannel.java
```diff
@@ -16,54 +16,25 @@
...
 
-    private final Function<Endpoint, Channel> adapter;
-    private final Map<Object, Channel> cache;
+    private final LoadingCache<Endpoint, Channel> cache;
 
-    ChannelToEndpointChannel(Function<Endpoint, Channel> adapter) {
-        this.adapter = adapter;
-        this.cache = new ConcurrentHashMap<>();
+    ChannelToEndpointChannel(Function<Endpoint, Channel> loader) {
+        this.cache = Caffeine.newBuilder().weakKeys().maximumSize(10_000).build(loader::apply);
```
pkoenig10, line 33: I suppose if we were willing to give up the maximum size we could just use a `ConcurrentHashMap` here and avoid some of the `Caffeine` bits. Although maybe it's worth it to have the maximum size.

carterkozak: Also weak keys, which are key to making this somewhat safe

pkoenig10: Ah right, we'd still have to periodically clean up the entries for collected keys.

carterkozak (approved):
Thanks!

carterkozak:
This project depends on a version of CJR which includes https://github.com/palantir/conjure-java-runtime/pull/2973, however it's possible that some projects don't declare a constraint for consistent CJR library versions across modules. The risk is minimal, since we'd only end up with cache churn that shouldn't be any worse than the existing behavior -- this should be safe to merge.

PR #2375, opened 2024-10-08
Handle new QoS Metadata

carterkozak:
See https://github.com/palantir/conjure-java-runtime-api/pull/1231 for more information

## After this PR
==COMMIT_MSG==
Handle new QoS Metadata
==COMMIT_MSG==

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueQosReasonDecoder.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueQosReasonDecoder.java
@@ -1,0 +1,36 @@
+/*
+ * (c) Copyright 2024 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.dialogue.core;
+
+import com.palantir.conjure.java.api.errors.QosReason;
+import com.palantir.conjure.java.api.errors.QosReasons;
+import com.palantir.conjure.java.api.errors.QosReasons.QosResponseDecodingAdapter;
+import com.palantir.dialogue.Response;
+import java.util.Optional;
+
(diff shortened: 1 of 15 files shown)


dialogue-core/src/main/java/com/palantir/dialogue/core/PinUntilErrorNodeSelectionStrategyChannel.java
```diff
@@ -147,7 +147,9 @@ public void onSuccess(Response response) {
                 // workflows where it is important for a large number of requests to all land on the same node,
                 // even if a couple of them get rate limited in the middle.
                 if (Responses.isServerErrorRange(response)
-                        || (Responses.isQosStatus(response) && !Responses.isTooManyRequests(response))) {
+                        || (Responses.isQosStatus(response)
+                                && !Responses.isQosDueToCustom(response)
```
bjlaub, line 151: if i understand this correctly, a `Due-To=CUSTOM` will _not_ count as an error for node pinning, and we will stay on the same node? What's the rationale behind that?

dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueQosReasonDecoder.java
```diff
@@ -0,0 +1,36 @@
...
+
+package com.palantir.dialogue.core;
+
+import com.palantir.conjure.java.api.errors.QosReason;
+import com.palantir.conjure.java.api.errors.QosReasons;
+import com.palantir.conjure.java.api.errors.QosReasons.QosResponseDecodingAdapter;
+import com.palantir.dialogue.Response;
+import java.util.Optional;
+
+enum DialogueQosReasonDecoder implements QosResponseDecodingAdapter<Response> {
```
bjlaub, line 25: any reason this could not live along side `Response` in `dialogue-target` and be shared across projects?

dialogue-core/src/main/java/com/palantir/dialogue/core/PinUntilErrorNodeSelectionStrategyChannel.java
carterkozak: Due-to custom means the client shouldn’t attempt to interpret meaning from the failure, so we don’t route to another node in that case. Changing pinned nodes can have dramatic impacts on perf, some clients use it for cache locality, which we don’t want to give up unless we really need to (instability/etc)

dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueQosReasonDecoder.java
carterkozak: It’s an implementation detail, not part of the codegen target api. I don’t mind duplication for such a simple utility, primary goal is to avoid expanding the public api unnecessarily

bjlaub:
👍 lgtm

PR #2384, opened 2024-10-15
Clarify host vs endpoint concurrency limits in README and code

ash211:
## Before this PR
During a recent P0 we weren't immediately certain how the host and endpoint concurrency limits differed.

## After this PR
==COMMIT_MSG==
Clarify host vs endpoint concurrency limits in README and code
==COMMIT_MSG==

With this PR, I aim to clarify the difference in how these two handle different types of response.  Both in README and in code.

There is no expected behavior change from this PR -- only documentation.

## Possible downsides?
None known

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,12 +175,8 @@ _This API is influenced by gRPC's [Java library](https://github.com/grpc/grpc-ja
 ### Concurrency limits
 
 Every request passes through a pair of [AIMD](https://en.wikipedia.org/wiki/Additive_increase/multiplicative_decrease)
-concurrency limiters.
-There are two types of concurrency limiter: per-host, and per-endpoint. The former is based on failures that
-indicate the target host is in a degraded state, and the latter is based on failures that are coupled to
-an individual endpoint.
-Each concurrency limiter operates in conjunction with a queue to stage pending requests until a
-permit becomes available.
+concurrency limiters. There are two types of concurrency limiter: per-host, and per-endpoint. Each concurrency limiter
+operates in conjunction with a queue to stage pending requests until a permit becomes available.
 
 #### Limiter Diagram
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -175,12 +175,8 @@ _This API is influenced by gRPC's [Java library](https://github.com/grpc/grpc-ja
 ### Concurrency limits
 
 Every request passes through a pair of [AIMD](https://en.wikipedia.org/wiki/Additive_increase/multiplicative_decrease)
-concurrency limiters.
-There are two types of concurrency limiter: per-host, and per-endpoint. The former is based on failures that
```
ash211, line 179: moved this:

> The former is based on failures that
indicate the target host is in a degraded state, and the latter is based on failures that are coupled to an individual endpoint.

below to those specific sections.  This way the `### Concurrency limits` section is in steadily increasing level of detail, instead of putting this up in the summary

README.md
```diff
@@ -207,13 +203,17 @@ permit becomes available.
 
 Each host has a concurrency limiter which protects servers by stopping requests getting out the door on the client-side.
 Permits are decreased after receiving 308 or 501-599 response, or encountering a network error (`IOException`).
-Otherwise, they are increased.
+429 or 500 responses have no change. Otherwise, permits are increased.
+
+Host limits are based on failures that indicate the target host overall is in a degraded state.
```
ash211, line 208: this is an explanation/reason for why the status codes above are handled the way they  are

dialogue-core/src/main/java/com/palantir/dialogue/core/CautiousIncreaseAggressiveDecreaseConcurrencyLimiter.java
```diff
@@ -154,10 +158,22 @@ void onFailure(Throwable _throwable, PermitControl control) {
...
+        /**
+         * Indicates that the request corresponding to this permit was dropped and that the concurrency limit should be
+         * multiplicatively decreased.
+         */
         void dropped();
 
+        /**
+         * Indicates that the request corresponding to this permit was successful and that the concurrency limit should
+         * be additively increased.
+         */
```
ash211, line 176: by putting these Javadoc on the interface instead of the class, they now show in IDE `F1` above

bjlaub:
👍 lgtm, thanks!

PR #2398, opened 2024-10-28
Optimize target selection and URI parsing

schlosna:
## Before this PR

While reviewing some JFRs from dialogue client heavy service, notice that there are a lot of `byte[]` allocations as a result of parsing `String` to `URI`. In most cases, these will be be the same input string, so are good candidates for caching.

image

## After this PR
==COMMIT_MSG==
Optimize target selection and URI parsing
==COMMIT_MSG==

## Possible downsides?
We don't have a great way to inject cache sizing/weight configuration or tagged metrics, so for now have hardcoded max cache weight.


dialogue-clients/src/main/java/com/palantir/dialogue/clients/DnsSupport.java
```diff
@@ -179,10 +196,10 @@ static ImmutableList<TargetUri> getTargetUris(
...
                             UnsafeArg.of("host", host),
                             SafeArg.of("service", serviceNameForLogging));
-                }
-                for (InetAddress addr : resolvedAddresses) {
-                    targetUris.add(
-                            TargetUri.builder().uri(uri).resolvedAddress(addr).build());
+                } else {
+                    TargetUri.Builder builder = TargetUri.builder().uri(uri);
+                    targetUris.addAll(Collections2.transform(resolvedAddresses, addr -> builder.resolvedAddress(addr)
+                            .build()));
```
carterkozak: I'm not a big fan of the partial-builder-as-factory pattern, since sometimes folks add protections later on to prevent interactions after `build()` is called. It shouldn't be an issue here since. the TargetUri type is defined within this repo, so I don't have a terribly strong preference

carterkozak: Perhaps we add an efficient `TargetUri.of(uri, address)` factory method to use instead? That way we bypass the builder entirely.

dialogue-clients/src/main/java/com/palantir/dialogue/clients/DnsSupport.java
```diff
@@ -84,6 +90,19 @@ final class DnsSupport {
...
 
+    /**
+     * Shared cache of string to parsed URI. This avoids excessive allocation overhead when parsing repeated targets.
+     */
+    private static final LoadingCache<String, Parsed> uriCache = CacheStats.of(
+                    SharedTaggedMetricRegistries.getSingleton(), "dialogue-uri")
+            .register(stats -> Caffeine.newBuilder()
+                    .maximumWeight(100_000)
+                    .<String, Parsed>weigher((key, _value) -> key.length())
+                    .expireAfterAccess(Duration.ofMinutes(1))
```
carterkozak, line 101: 👍

dialogue-clients/src/main/java/com/palantir/dialogue/clients/DnsSupport.java
schlosna: good call, updated

carterkozak:
I'm not sure why that check is failing on this PR, but I suspect merging this will fix it: #2403

carterkozak:
#2403 has merged, could you please rebase this to develop when you have a moment?

carterkozak (approved):
thanks!

PR #2403, opened 2024-10-31
simplify jackson version constraints

carterkozak:
==COMMIT_MSG==
simplify jackson version constraints
==COMMIT_MSG==

### DIFF
--- a/versions.props
+++ b/versions.props
@@ -1,4 +1,4 @@
-com.fasterxml.jackson.core:jackson-databind = 2.18.1
+com.fasterxml.jackson.*:* = 2.18.1
 com.google.code.findbugs:jsr305 = 3.0.2
 com.google.guava:guava = 33.3.1-jre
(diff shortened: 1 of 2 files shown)


carterkozak:
#2398 is failing checks complaining about duplicate constraints -- not sure why that's failing on a PR and not wherever introduced the check, though.

versions.props
```diff
@@ -48,7 +48,3 @@ com.palantir.goethe:* = 0.12.0
 
 # Match conjure-java-runtime okhttp version
 com.squareup.okhttp3:* = 3.13.1
-
-# Avoid forcing consumers to upgrade jackson
-com.fasterxml.jackson.*:jackson-* = 2.11.1
-# dependency-upgrader:ON
```
carterkozak, line 54: we manage jackson using its own excavator now, no reason to isolate this.

bjlaub:
👍

PR #2410, opened 2024-11-05
Migrate squareup javapoet to Palantir javapoet

ash211:
## Before this PR
The square version [is deprecated](https://github.com/square/javapoet?tab=readme-ov-file#deprecated) and the Palantir one has fixes, including `https://github.com/square/javapoet/pull/828` which was never released on square's coordinates but was on Palantir's.

## After this PR
==COMMIT_MSG==
Migrate squareup javapoet to Palantir javapoet
==COMMIT_MSG==

## Possible downsides?
None known


carterkozak (approved):
thanks!

PR #2412, opened 2024-11-06
Config.uris uses `toList()` collector

carterkozak:
Minor code cleanup

==COMMIT_MSG==
Config.uris uses `toList()` collector
==COMMIT_MSG==


bjlaub:
👍

PR #2413, opened 2024-11-08
reuse concurrency limiter state when reloading host channels

bjlaub:
## Before this PR
The recent changes to implement DNS node discovery came with the drawback that when domain name resolution changes are detected, we reload channels causing all internal channel state to be discarded and recreated anew. This means that concurrency limiter state (i.e. the current limit values encapsulated within a `CautiousIncreaseAggressiveDecreaseConcurrencyLimiter` instance) for a per-host channel are lost and we need to restart the congestion control algorithm.

This can have some implications in both directions: if the limiter state has reduced the concurrency significantly because the server asked us to slow down, we may end up bombarding the host with more requests than it can handle when a reload happens; if the limiter state has increased concurrency significantly, we will artificially start limiting request concurrency and have to slowly work our way back up to better throughput.

## After this PR
This change introduces a `ChannelState` capable of holding internal state to be retained across channel reloads. In this change, we add a cache mapping `TargetUri` to a `ChannelState` instance, where each `ChannelState` holds an instance of a `CautiousIncreaseAggressiveDecreaseConcurrencyLimiter`. This allows a `ConcurrencyLimitedChannel` for a specific `TargetUri` to be created anew but re-use existing limiter state for that particular host.

Note that this change _only_ affects per-host channels, not per-endpoint channels (which are also concurrency limited, but interact with a `QueuedChannel` in a much different way). A follow-on PR will address retaining limiter state for per-endpoint channels upon reload.

==COMMIT_MSG==
reuse concurrency limiter state when reloading host channels
==COMMIT_MSG==


dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueChannel.java
```diff
@@ -77,6 +84,40 @@ public String toString() {
...
+            return factory;
+        }
+
+        StateHolderKey(final Class<T> valueClass, Supplier<T> factory) {
+            this.valueClass = valueClass;
+            this.factory = factory;
+        }
+    }
+
+    static final class StateHolder {
```
carterkozak: disclaimer: naming is hard, and I'm not very good at it. Take my comments with a grain of salt!

Thoughts on extracting this to its own file, and renaming to `ChannelState`? "Holder" makes me think it's using the old [initialize-on-demand holder pattern](https://en.wikipedia.org/wiki/Initialization-on-demand_holder_idiom).

If we extract to its own file, the key type could make sense as a static nested class, since the outer class provides context as to what it's a key for e.g. `ChannelState.Key`.

dialogue-core/src/main/java/com/palantir/dialogue/core/ConcurrencyLimitedChannel.java
```diff
@@ -178,4 +186,9 @@ public String channelNameForLogging() {
             return channelNameForLogging;
         }
     }
+
+    @Value.Immutable
+    interface ConcurrencyLimitedChannelState {
+        CautiousIncreaseAggressiveDecreaseConcurrencyLimiter hostLimiter();
+    }
```
carterkozak: I think we can put the `CautiousIncreaseAggressiveDecreaseConcurrencyLimiter` directly into the state, since it's the only piece of data we're tracking

dialogue-core/src/main/java/com/palantir/dialogue/core/ConcurrencyLimitedChannel.java
```diff
@@ -39,16 +42,19 @@
 final class ConcurrencyLimitedChannel implements LimitedChannel {
     private static final SafeLogger log = SafeLoggerFactory.get(ConcurrencyLimitedChannel.class);
 
+    private static final StateHolderKey<ConcurrencyLimitedChannelState> STATE_HOLDER_KEY =
+            new StateHolderKey<>(ConcurrencyLimitedChannelState.class, ConcurrencyLimitedChannel::createState);
```
carterkozak: II think we should include `HOST` in the name, since this is creating a host-specific limiter

bjlaub: 👍 done

dialogue-core/src/main/java/com/palantir/dialogue/core/ConcurrencyLimitedChannel.java
bjlaub: done

dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueChannel.java
bjlaub: yep done

dialogue-core/src/main/java/com/palantir/dialogue/core/ChannelState.java
```diff
@@ -0,0 +1,56 @@
...
+        private final Class<T> valueClass;
+        private final Supplier<T> factory;
+
+        T cast(final Object value) {
+            return valueClass.cast(value);
+        }
+
+        Supplier<T> getFactory() {
+            return factory;
+        }
```
carterkozak: These can be private

dialogue-core/src/main/java/com/palantir/dialogue/core/ChannelState.java
```diff
@@ -0,0 +1,56 @@
...
+
+        Key(final Class<T> valueClass, Supplier<T> factory) {
+            this.valueClass = valueClass;
+            this.factory = factory;
+        }
+    }
+
+    @SuppressWarnings("DangerousIdentityKey")
```
carterkozak, line 43: We could implement `equals` and `hashCode` on `Key`, delegating to `super` to make this explicit (and avoid the suppression)

dialogue-core/src/main/java/com/palantir/dialogue/core/ChannelState.java
```diff
@@ -0,0 +1,56 @@
...
+
+    <T> T getState(Key<T> key) {
+        if (state.containsKey(key)) {
+            return key.cast(state.get(key));
+        } else {
+            T value = key.getFactory().get();
+            Preconditions.checkNotNull(value, "state factory cannot produce a null value");
+            state.put(key, value);
+            return value;
+        }
```
carterkozak: I don't think we expect this to be a hot path, so we can use something like:

```java
return key.cast(Preconditions.checkNotNull(
    state.computeIfAbsent(key, keyValue -> keyValue.getFactory().get()),
    "state factory cannot produce a null value"));
```

dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueChannel.java
```diff
@@ -174,18 +181,34 @@ public DialogueChannel build() {
...
-                        UnsafeArg.of("uris", cf.clientConf().uris()),
-                        SafeArg.of("numUris", cf.clientConf().uris().size()),
-                        UnsafeArg.of("targets", targetUris),
-                        SafeArg.of("numTargets", targetUris.size()));
-                ImmutableList<LimitedChannel> targetChannels = createHostChannels(cf, targetUris);
-                return NodeSelectionStrategyChannel.create(cf, targetChannels);
-            }));
+            LimitedChannel nodeSelectionChannel =
+                    new SupplierChannel(cf.uris().map(new Function<List<TargetUri>, LimitedChannel>() {
+                        private final Map<TargetUri, ChannelState> state = new HashMap<>();
```
carterkozak: Let's make this a ConcurrentHashMap for safety, that way we don't need to think about it

dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueChannel.java
```diff
@@ -174,18 +181,34 @@ public DialogueChannel build() {
...
+
+                        @Override
+                        public LimitedChannel apply(List<TargetUri> targetUris) {
+                            // remove state for uris we no longer care about, and create new ChannelStates
+                            // for uris we don't know about yet
+                            Set<TargetUri> toRemove = state.keySet().stream()
+                                    .filter(uri -> !targetUris.contains(uri))
+                                    .collect(Collectors.toSet());
+                            toRemove.forEach(state::remove);
```
carterkozak: I think we can replace this with `state.keySet().retainAll(targetUris);`

dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueChannel.java
```diff
@@ -174,18 +181,34 @@ public DialogueChannel build() {
...
+                            targetUris.forEach(uri -> state.putIfAbsent(uri, new ChannelState()));
```
carterkozak: Perhaps `computeIfAbsent(uri, _uri -> new ChannelState())` to avoid creating `ChannelState` instances for keys that are already in the map?

dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueChannel.java
bjlaub: 👍 this is voodoo that i didn't know existed

dialogue-core/src/main/java/com/palantir/dialogue/core/ConcurrencyLimitedChannel.java
```diff
@@ -39,13 +40,20 @@
 final class ConcurrencyLimitedChannel implements LimitedChannel {
     private static final SafeLogger log = SafeLoggerFactory.get(ConcurrencyLimitedChannel.class);
 
+    @VisibleForTesting
+    static final ChannelState.Key<CautiousIncreaseAggressiveDecreaseConcurrencyLimiter> HOST_SPECIFIC_STATE_KEY =
```
bjlaub, line 44: this is visible for the purposes of `ConcurrencyLimitedChannelTest#testReuseCachedLimiterState_host`, which may not be a particularly useful test. I'm somewhat indifferent, we could delete that test and make this private.

carterkozak (approved):

PR #2418, opened 2024-11-13
Retain endpoint concurrency limits on node refresh

carterkozak:
==COMMIT_MSG==
Retain endpoint concurrency limits on refresh
==COMMIT_MSG==

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueChannel.java
@@ -260,6 +260,9 @@ public final class DialogueChannel implements Channel, EndpointChannelFactory {
                                 uriIndexForInstrumentation,
                                 endpoint,
                                 endpointChannelState.get(endpoint));
+                        // Note that because the queue is recreated when nodes are refreshed, it's critical that
+                        // the queue can force at least one request through at a time using the behavior introduced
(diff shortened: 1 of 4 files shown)


dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueChannel.java
```diff
@@ -252,12 +254,17 @@ private static ImmutableList<LimitedChannel> createHostChannels(
...
                         if (endpoint.tags().contains("dialogue-disable-endpoint-concurrency-limiting")) {
                             return unlimited;
                         }
                         LimitedChannel limited = ConcurrencyLimitedChannel.createForEndpoint(
-                                unlimited, cf.channelName(), uriIndexForInstrumentation, endpoint);
+                                unlimited,
+                                cf.channelName(),
+                                uriIndexForInstrumentation,
+                                endpoint,
+                                endpointChannelState.get(endpoint));
```
carterkozak, line 262: I think this will fail in an edge case where the endpoint limiter is completely full from a previous refresh, so the queued channel created on the next line cannot sense the boundary condition to begin allowing requests through

bjlaub:
> Proof of concept retaining endpoint concurrency limits on refresh

nit: doesn't seem like a proof of concept

bjlaub:
👍

PR #2422, opened 2024-11-14
QueuedChannel enforces its assumption around at least one in-flight request

carterkozak:
WIP, need to clean up tests, and figure out what's up with the simulations

==COMMIT_MSG==
QueuedChannel enforces its assumption around at least one in-flight request
==COMMIT_MSG==

For posterity:

 The changes to the simulations in this PR can be avoided by implementing a recursive-scheduling check like this: #2424
 Are we necessarily better off with that check? I'm not sure, threadlocals are a massive pain, so at the moment I think I'd prefer this behavior.


dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
```diff
@@ -152,8 +157,9 @@ Optional<ListenableFuture<Response>> maybeExecute(Endpoint endpoint, Request req
         // Queuing adds contention between threads and should be avoided unless we need to shed load.
         if (queueSizeEstimate.get() <= 0) {
             Optional<ListenableFuture<Response>> maybeResult =
-                    delegate.maybeExecute(endpoint, request, DO_NOT_SKIP_LIMITS);
+                    delegate.maybeExecute(endpoint, request, limitEnforcement());
```
carterkozak: worth verifying that when we bypass the limiter, we don't receive an empty optional (we should fail with an exception in that case to maintain valid state)

bjlaub (approved):

PR #2430, opened 2024-11-21
Implement support for the Proxy-Upstream-Request-Attempts header

carterkozak:
This aims to provide coordination between clients and reverse proxy servers which implement retries, allowing us to avoid multiplying out the total retry count.

When dialogue is configured with more retries than a proxy, we still may retry more than we want, but not DIALOGUE_ATTEMPTS * PROXY_ATTEMPTS. The worst case scenario should be when the proxy retry attempts don't quite reach the configured dialogue retry limit, and another full round of proxy retries must be accumulated. This should be a reaosnable trade-off because it allows the calling client to be configured with a larger value as needed for stability, without multiplying all retries across the board unnecessarily.

Note that when a Dialogue client receives a tcp error/closure, it can make no assumptions about proxy retry attempts, so this only applies to requests which successfully receive a response.

==COMMIT_MSG==
Implement support for the Proxy-Upstream-Request-Attempts header
==COMMIT_MSG==

Note this feature is meant to be used in tandem with #2434 for flexibility

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/Responses.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/Responses.java
@@ -101,6 +101,9 @@ final class Responses {
                 if (parsed >= 0) {
                     return parsed;
                 }
+                log.warn(
+                        "Received an unexpected negative proxy upstream request attempts value, using zero",
+                        SafeArg.of("proxyUpstreamRequestAttempts", proxyUpstreamRequestAttempts));
             } catch (NumberFormatException e) {
                 log.warn(
                         "Failed to parse proxy upstream request attempts, assuming zero",
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/Responses.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/Responses.java
@@ -15,14 +15,23 @@
  */
 package com.palantir.dialogue.core;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.net.HttpHeaders;
 import com.palantir.conjure.java.api.errors.QosReason;
 import com.palantir.conjure.java.api.errors.QosReason.DueTo;
 import com.palantir.conjure.java.api.errors.QosReason.RetryHint;
 import com.palantir.dialogue.Response;
+import com.palantir.logsafe.SafeArg;
+import com.palantir.logsafe.logger.SafeLogger;
+import com.palantir.logsafe.logger.SafeLoggerFactory;
+import java.util.Optional;
 
 /** Utility functionality for {@link Response} handling. */
 final class Responses {
+    private static final SafeLogger log = SafeLoggerFactory.get(Responses.class);
+
+    @VisibleForTesting
+    static final String PROXY_UPSTREAM_REQUEST_ATTEMPTS = "Proxy-Upstream-Request-Attempts";
 
     static boolean isRetryOther(Response response) {
         // Note that a 308 status may be a non-retryable signal, for instance google sometimes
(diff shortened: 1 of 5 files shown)


dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -307,7 +307,8 @@ private ListenableFuture<Response> handleThrowable(Throwable clientSideThrowable
 
         private ListenableFuture<Response> incrementFailuresAndMaybeRetry(
                 Response response, BiFunction<Endpoint, Response, Throwable> failureSupplier, Meter meter) {
-            if (++failures <= maxRetries) {
+            failures = failures + 1 + Responses.getProxyRetryAttempts(response);
```
carterkozak: note that this retains the `+1` from before in addition to the retry attempts value. I'm not certain if we'd implement this as number of retry attempts (total attempts - 1) or report the total attempts.

I suppose if we're reporting total attempts, perhaps `Proxy-Request-Attempts` may be a more accurate name than `Proxy-Retry-Attempts`, but either model will fit, I'll just need to adjust the floor by 1.

ash211: Fwiw the language I usually use is:
- if 5 total requests are sent:
  - the first one is an attempt, and the latter 4 are "retries" but also attempts
  - so that's "5 attempts" and "4 retries"
- if 1 total request is sent, that's 1 attempt and 0 retries

carterkozak: Yes, it depends where we land on naming the new header :-)

dialogue-core/src/main/java/com/palantir/dialogue/core/Responses.java
```diff
@@ -77,5 +86,30 @@ static boolean isClientError(Response response) {
...
+     * Rejecting negative values is critical because clients mustn't roll back retry counters based on
+     * responses from a remote server.
+     */
+    static int getProxyUpstreamRequestAttempts(Response response) {
+        Optional<String> maybeProxyUpstreamRequestAttempts = response.getFirstHeader(PROXY_UPSTREAM_REQUEST_ATTEMPTS);
+        if (maybeProxyUpstreamRequestAttempts.isPresent()) {
+            String proxyUpstreamRequestAttempts = maybeProxyUpstreamRequestAttempts.get();
+            try {
+                int parsed = Integer.parseInt(proxyUpstreamRequestAttempts.trim());
+                if (parsed >= 0) {
```
mpritham, line 101: Are negative values an indication of something gone wrong? If so, maybe we should we also log a warning when they're encountered as well.

carterkozak: Good idea, updated!

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -307,7 +307,14 @@ private ListenableFuture<Response> handleThrowable(Throwable clientSideThrowable
 
         private ListenableFuture<Response> incrementFailuresAndMaybeRetry(
                 Response response, BiFunction<Endpoint, Response, Throwable> failureSupplier, Meter meter) {
-            if (++failures <= maxRetries) {
+            // We've made a request which has failed, so we apply a floor of a single failure if there are no
+            // proxy upstream request attempts. Otherwise, if the proxy has reported upstream request attempts,
+            // those are considered as well. When the proxy makes a single request, that is considered part of
+            // our initial request to the proxy -- we're counting for total failures sending requests to
+            // the upstream server.
```
mpritham, line 314: > we're counting for total failures sending requests to the upstream server.

If a proxy doesn't actually forward the request to the server, and returns an error response with a  `0` header value, Dialogue will still increment the failures by 1. Given this, I think we also don't expect a `0` value for the header and should log something in that case too?

carterkozak: I think zero is a reasonable value if the server doesn't believe there are any healthy upstreams, I don't think we need to log in that case

mpritham: I see. The logic lgtm. The `failures` count the requests to the upstream server when the header value is positive, and count the requests to the proxy when `0`. Both are failures as far as Dialogue is concerned.

dialogue-core/src/test/java/com/palantir/dialogue/core/ResponsesTest.java
```diff
@@ -0,0 +1,61 @@
...
+    @Test
+    void proxyUpstreamRequestAttempts_malformed() {
+        try (Response response = new TestResponse().withHeader(Responses.PROXY_UPSTREAM_REQUEST_ATTEMPTS, "deadbeef")) {
+            assertThat(Responses.getProxyUpstreamRequestAttempts(response)).isZero();
+        }
+    }
+
+    @Test
+    void proxyUpstreamRequestAttempts_valid() {
+        try (Response response = new TestResponse().withHeader(Responses.PROXY_UPSTREAM_REQUEST_ATTEMPTS, "1")) {
```
mpritham, line 43: Thoughts on also adding some misc spaces to check that it's `trim`'d?

mpritham (dismissed):
LGTM! Removed `merge-when-ready` in case you wanted to make the nit test change.

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -307,7 +307,20 @@ private ListenableFuture<Response> handleThrowable(Throwable clientSideThrowable
...
+            int proxyUpstreamRequestAttempts = Responses.getProxyUpstreamRequestAttempts(response);
+            if (proxyUpstreamRequestAttempts > 1) {
+                log.info(
+                        "Received a Proxy-Upstream-Request-Attempts response header "
+                                + "indicating retries have already occurred",
+                        SafeArg.of("proxyUpstreamRequestAttempts", proxyUpstreamRequestAttempts));
+            }
```
carterkozak, line 321: I added an info-level log statement here as well

dialogue-core/src/test/java/com/palantir/dialogue/core/ResponsesTest.java
carterkozak: :+1:

carterkozak:
Thanks, this needs another +1 since I've pushed again!

mpritham (approved):

PR #2432, opened 2024-11-22
Cache parsed URIs throughout Dialogue

schlosna:
## Before this PR

Noticed `new URI(String)` and some of its side effects in JFRs from services making high volumes of Dialogue requests.

## After this PR
==COMMIT_MSG==
Parsing String to URI can be expensive in terms of CPU and allocations for high throughput services. This generalizes the caching previously added to DnsSupport in https://github.com/palantir/dialogue/pull/2398 to also cover ApacheHttpClientBlockingChannel requests and HttpsProxyDefaultRoutePlanner proxy parsing to leverage a shared parsed URI cache.
==COMMIT_MSG==


dialogue-core/src/main/java/com/palantir/dialogue/core/Uris.java
```diff
@@ -0,0 +1,142 @@
...
+import org.immutables.value.Value;
+
+public final class Uris {
+    private static final SafeLogger log = SafeLoggerFactory.get(Uris.class);
+
+    /**
+     * This prefix may reconfigure several aspects of the client to work better in a world where requests are routed
+     * through a service mesh like istio/envoy.
+     */
+    private static final String MESH_PREFIX = "mesh-";
```
carterkozak: At this point the `mesh-` prefix is no longer used (or supported) in any environment -- it should make things simpler if we introduce another change first to remove support

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientBlockingChannel.java
```diff
@@ -100,8 +102,9 @@ final class ApacheHttpClientBlockingChannel implements BlockingChannel {
     public Response execute(Endpoint endpoint, Request request) throws IOException {
         // Create base request given the URL
         URL target = baseUrl.render(endpoint, request);
+        URI targetUri = Uris.tryParse(target.toString()).uriOrThrow();
```
carterkozak: Unlike the base URIs, this value will have much higher cardinality any time an endpoint includes path parameters. I'm a little bit uneasy about caching.

We could bypass the URL and URI types entirely, setting specific elements directly on the request builder instead. That would require expanding the BaseUrl API a bit, but may produce better results. What do you think?

dialogue-core/src/main/java/com/palantir/dialogue/core/Uris.java
schlosna: I took a look at removing mesh mode, and it has tentacles across a bunch of modules & tests, so might hold off on removing that until after this PR if that's ok.

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientBlockingChannel.java
schlosna: good call, updated to just use the URL protocol, authority, and file which avoids some of the more expensive `new URI(String)` parsing and not flooding the URI parsing cache with high cardinality URIs

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/HttpsProxyDefaultRoutePlanner.java
```diff
@@ -49,12 +50,8 @@ final class HttpsProxyDefaultRoutePlanner extends DefaultRoutePlanner {
     @Override
     @CheckForNull
     public HttpHost determineProxy(final HttpHost target, final HttpContext _context) throws HttpException {
-        final URI targetUri;
-        try {
-            targetUri = new URI(target.toURI());
-        } catch (final URISyntaxException ex) {
-            throw new HttpException("Cannot convert host to URI: " + target, ex);
-        }
+        final URI targetUri = parseTargetUri(target);
```
carterkozak: I suspect the cardinality is similarly high here, however unlike the per-request parse, this should only occur once per new connection we create, which should be an order or two lower magnitude than requests.

My preference is not to use the cache here. I suspect the perf cost of uri parsing is insignificant compared to tls handshake overhead. If that turns out not to be the case, It think we could plumb through the original base uri via the `HttpContext` so we can reuse the same value for each connection.

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientBlockingChannel.java
schlosna: split this portion out to a separate PR https://github.com/palantir/dialogue/pull/2437 as this is portion is the biggest win by avoiding a `URI` parse per request.

dialogue-core/src/main/java/com/palantir/dialogue/core/Uris.java
schlosna: mesh removed in https://github.com/palantir/dialogue/pull/2474

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/HttpsProxyDefaultRoutePlanner.java
schlosna: I would expect proxy URI cardinality to be generally small (on the order of single digits in most cases where proxies are in use), and this path will be hit for every request as it determines the route plan.

image

PR #2434, opened 2024-11-25
Implement the `Client-Can-Retry: true` hint header for reverse proxies

carterkozak:
This feature functions in combination with `Proxy-Upstream-Request-Attempts` (#2430) to allow negotiation for retries to occur based on proxy configuration, without requiring clients to be updated or reconfigured.

==COMMIT_MSG==
Implement the `Client-Can-Retry: true` hint header for reverse proxies
==COMMIT_MSG==


bjlaub:
👍 lgtm

PR #2437, opened 2024-11-26
ApacheHttpClientBlockingChannel uses URL to build request

schlosna:
## Before this PR

Noticed new URI(String) and some of its side effects in JFRs from services making high volumes of Dialogue requests.

Splitting out of PR https://github.com/palantir/dialogue/pull/2432

## After this PR
==COMMIT_MSG==
Parsing String to URI can be expensive in terms of CPU and allocations for high throughput services. 
==COMMIT_MSG==

### DIFF
--- a/dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientBlockingChannel.java
+++ b/dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientBlockingChannel.java
@@ -15,6 +15,7 @@
  */
 package com.palantir.dialogue.hc5;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ListMultimap;
 import com.google.common.collect.MultimapBuilder;
@@ -33,7 +34,9 @@ import com.palantir.logsafe.Arg;
 import com.palantir.logsafe.Preconditions;
 import com.palantir.logsafe.SafeArg;
 import com.palantir.logsafe.SafeLoggable;
+import com.palantir.logsafe.UnsafeArg;
 import com.palantir.logsafe.exceptions.SafeExceptions;
+import com.palantir.logsafe.exceptions.SafeIllegalArgumentException;
 import com.palantir.logsafe.exceptions.SafeRuntimeException;
 import com.palantir.logsafe.logger.SafeLogger;
 import com.palantir.logsafe.logger.SafeLoggerFactory;
@@ -45,6 +48,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.net.InetAddress;
+import java.net.URISyntaxException;
 import java.net.URL;
 import java.util.Arrays;
 import java.util.Collections;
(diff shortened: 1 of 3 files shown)


dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientBlockingChannel.java
```diff
@@ -167,6 +154,42 @@ public Response execute(Endpoint endpoint, Request request) throws IOException {
...
+        return builder.build();
+    }
+
+    @VisibleForTesting
+    static URIAuthority parseAuthority(URL url) {
+        try {
+            return URIAuthority.create(url.getAuthority());
+        } catch (URISyntaxException e) {
+            throw new SafeIllegalArgumentException("Invalid URI authority", e, UnsafeArg.of("url", url));
+        }
```
carterkozak: I wonder if it's worth rolling the dice with something along these lines to avoid string tokenizing/manipulation overhead? We have most of the data parsed out via URL already, I believe it's just ipv6 brackets that URL doesn't handle well.

```java
String host = url.getHost();
if (host != null && host.startsWith("[") && host.endsWith("]")) {
    host = host.substring(1, host.length() - 1);
}
return new URIAuthority(url.getUserInfo(), host, url.getPort());
```

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientBlockingChannel.java
```diff
@@ -167,6 +154,42 @@ public Response execute(Endpoint endpoint, Request request) throws IOException {
...
+        } else if (requiresEmptyBody(endpoint)) {
+            builder.setEntity(EmptyHttpEntity.INSTANCE);
+        }
+        return builder.build();
+    }
+
+    @VisibleForTesting
+    static URIAuthority parseAuthority(URL url) {
+        try {
+            return URIAuthority.create(url.getAuthority());
```
carterkozak, line 192: +1 I've verified that `url.getAuthority()` is equivalent to `URI.getRawAuthority()`, which is used internally within apache httpclient.

dialogue-apache-hc5-client/src/test/java/com/palantir/dialogue/hc5/ApacheHttpClientBlockingChannelTest.java
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    @ParameterizedTest
+    @CsvSource({
+        "http://example.com, example.com, -1,",
+        "https://www.example.com, www.example.com, -1,",
+        "https://www.example.com:443, www.example.com, 443,",
+        "https://www.example.com/path/to/foo/bar, www.example.com, -1,",
+        "https://www.example.com/path/to/foo/bar?baz=quux&hello=world#hash-octothorpe, www.example.com, -1,",
+        "https://user@www.example.com:8443/path/to/foo/bar?baz=quux&hello=world#hash-octothorpe ,"
```
carterkozak: Let's add a couple test cases for encoding edge cases:

* A URL which includes percent-escaped values in the users password: `https://user:slash%2Fslash@www.example.com`
* A URL which includes bracket-formatted ipv6 address: `https://user@[::1]/path` (perhaps with and without a port)

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientBlockingChannel.java
carterkozak: Alternatively, it may be safer if we check `host != null && host.startsWith("[")` and in that case, we use the parsing `URIAuthority.create` path, otherwise (in the common case) use `new URIAuthority(url.getUserInfo(), host, url.getPort());`

dialogue-apache-hc5-client/src/test/java/com/palantir/dialogue/hc5/ApacheHttpClientBlockingChannelTest.java
schlosna: Thanks, added some more test coverage.

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientBlockingChannel.java
schlosna: Specifically we're computing the authority similar to what is done in:

https://github.com/apache/httpcomponents-core/blob/rel/v5.3.1/httpcore5/src/main/java/org/apache/hc/core5/http/support/AbstractRequestBuilder.java#L176-L186

carterkozak (approved):
thanks!

PR #2443, opened 2024-12-05
Support endpoint error deserialization in `ConjureBodySerDe`

mpritham:
## Before this PR
We'd like to support the a new method in `BodySerDe` that allows the creation of deserializers given an expected result type, and a set of expected error types.

A [ConjureError](https://github.com/palantir/conjure-java/blob/develop/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureError.java#L54) is serialized and sent by servers. Dialogue deserializes the errors to a `SerializableError` which loses the type information of the parameters. By providing the type information of the parameters, Dialogue can deserialize the received `ConjureError` into the well-typed objects (instead of Strings). 

## After this PR
==COMMIT_MSG==
Support endpoint error deserialization in `ConjureBodySerDe`
==COMMIT_MSG==


dialogue-serde/src/test/java/com/palantir/conjure/java/dialogue/serde/EndpointErrorsConjureBodySerDeTest.java
```diff
@@ -0,0 +1,185 @@
...
+        }
+    }
+
+    @Generated("by conjure-java")
+    @JsonDeserialize(using = ErrorForEndpoint.Deserializer.class)
+    record ErrorForEndpoint(
+            @Safe String errorCode,
+            @Safe String errorName,
+            @Safe String errorInstanceId,
+            @Safe String arg,
```
mpritham: - can put the args into a separate record which includes matches the `parameters` map for deserialization so we don't need a custom deserializer.
- let's annotate each of the parameter fields with a `@JsonProperty`

dialogue-serde/src/test/java/com/palantir/conjure/java/dialogue/serde/EndpointErrorsConjureBodySerDeTest.java
```diff
@@ -0,0 +1,185 @@
...
+    record StringReturn(String value) implements EndpointReturnBaseType {
+        @JsonCreator
+        public static StringReturn create(String value) {
+            return new StringReturn(Preconditions.checkArgumentNotNull(value, "value cannot be null"));
+        }
+    }
+
+    @Generated("by conjure-java")
+    @JsonDeserialize(using = ErrorForEndpoint.Deserializer.class)
+    record ErrorForEndpoint(
```
mpritham: We're going to have `errorCode`, `errorName`, `errorInstanceId` every error, so we can create `GenericError<ArgsT>`.

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ConjureBodySerDe.java
```diff
@@ -74,6 +86,7 @@ final class ConjureBodySerDe implements BodySerDe {
         this.encodingsSortedByWeight = sortByWeight(encodings);
         Preconditions.checkArgument(encodings.size() > 0, "At least one Encoding is required");
         this.defaultEncoding = encodings.get(0).encoding();
+        this.emptyContainerDeserializer = emptyContainerDeserializer;
         this.binaryInputStreamDeserializer = new EncodingDeserializerRegistry<>(
```
mpritham: We cache b/c some services return + consume the same type from many diff endpoints. 

Each endpoint will have it's own type so the benefits of caching aren't as clear.

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ConjureBodySerDe.java
```diff
@@ -67,31 +75,49 @@ final class ConjureBodySerDe implements BodySerDe {
      */
     ConjureBodySerDe(
             List<WeightedEncoding> rawEncodings,
-            ErrorDecoder errorDecoder,
+            ErrorDecoder _errorDecoder,
```
mpritham: We don't need to pass this in. `DefaultConjureRuntime` is the public API where this ctor is created, and it passes in a hardcoded `ErrorDecoder.INSTANCE`.

dialogue-serde/src/test/java/com/palantir/conjure/java/dialogue/serde/ConjureBodySerDeTest.java
```diff
@@ -137,14 +141,12 @@ public void testRequestUnknownContentType() throws IOException {
...
-    public void testErrorsDecoded() {
-        TestResponse response = new TestResponse().code(400);
-
+    public void testErrorsDecoded() throws JsonProcessingException {
         ServiceException serviceException = new ServiceException(ErrorType.INVALID_ARGUMENT);
-        SerializableError serialized = SerializableError.forException(serviceException);
-        errorDecoder = mock(ErrorDecoder.class);
-        when(errorDecoder.isError(response)).thenReturn(true);
-        when(errorDecoder.decode(response)).thenReturn(new RemoteException(serialized, 400));
+        TestResponse response = TestResponse.withBody(
```
mpritham: This is to have an "e2e" test where the error decoder is actually called (instead of mocked) - to validate that replacing the `EncodingDeserializerRegistry` with `EncodingDeserializerForEndpointRegistry` doesn't introduce a regression.

I will also port over the tests from `ErrorDecoderTest` to test the new `EndpointErrorDecoder`

dialogue-annotations/src/main/java/com/palantir/dialogue/annotations/ConjureErrorDecoder.java
```diff
@@ -18,6 +18,7 @@
 
 import com.palantir.dialogue.Response;
 
+// TODO(pm): use the new EndpointErrorDecoder
```
mpritham: Replace impl in `ErrorDecoder` to delegate to the empty-map initialized EndpointErrorDecoder

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/EndpointErrorDecoder.java
```diff
@@ -0,0 +1,253 @@
...
+
+    public T decode(Response response) {
+        if (log.isDebugEnabled()) {
+            log.debug("Received an error response", diagnosticArgs(response));
+        }
+        try {
+            // this both returns an error and throws a runtime exception.
+            return decodeInternal(response);
+        } catch (Exception e) {
+            e.addSuppressed(diagnostic(response));
```
mpritham, line 96: This diagnostic information is not present in the result type. I think this is reasonable: when the response body is one of the expected error types, the client should have the information it needs to handle the endpoint response. The diagnostic info is available in a debug log if needed.

carterkozak: 👍

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ConjureBodySerDe.java
```diff
@@ -67,31 +75,48 @@ final class ConjureBodySerDe implements BodySerDe {
      */
     ConjureBodySerDe(
             List<WeightedEncoding> rawEncodings,
-            ErrorDecoder errorDecoder,
             EmptyContainerDeserializer emptyContainerDeserializer,
             CaffeineSpec cacheSpec) {
         List<WeightedEncoding> encodings = decorateEncodings(rawEncodings);
         this.encodingsSortedByWeight = sortByWeight(encodings);
         Preconditions.checkArgument(encodings.size() > 0, "At least one Encoding is required");
+        // note(pm): why do the weighted encoding thing? can we just pass in the default encoding?
```
mpritham: Maybe a FLUP. Going to remove this comment on in this PR.

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ConjureBodySerDe.java
```diff
@@ -42,12 +43,18 @@
...
+import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
 import java.util.Optional;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
 
+/**
+ * items:
+ * - we don't want to use `String` for the error identifier. Let's create an `ErrorName` class.
```
mpritham: `ErrorType#name` is sent over the wire, which is a string. It seems fine to keep this as a string.

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ConjureBodySerDe.java
```diff
@@ -58,7 +60,8 @@ final class ConjureBodySerDe implements BodySerDe {
     private final Deserializer<Optional<InputStream>> optionalBinaryInputStreamDeserializer;
     private final Deserializer<Void> emptyBodyDeserializer;
     private final LoadingCache<Type, Serializer<?>> serializers;
-    private final LoadingCache<Type, Deserializer<?>> deserializers;
+    private final LoadingCache<Type, EncodingDeserializerForEndpointRegistry<?>> deserializers;
+    private final EmptyContainerDeserializer emptyContainerDeserializer;
```
mpritham, line 68: The `emptyContainerDeserializer` is needed to created the `EncodingDeserializerForEndpointRegistry`. Previously `EncodingDeserializerRegistry` was only created in the constructor, but not it's created in the `deserializers(DeserializerArgs<>)` method as well. I added `emptyContainerDeserializer` as a field so I can access it there.

mpritham:
@carterkozak whenever you're free, this is good for another look.

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/EndpointErrorDecoder.java
```diff
@@ -0,0 +1,250 @@
...
+ * @param <T> the base type of the endpoint response. It's a union of the result type and all the error types.
+ */
+final class EndpointErrorDecoder<T> {
+    private static final SafeLogger log = SafeLoggerFactory.get(EndpointErrorDecoder.class);
+    private static final ObjectMapper MAPPER = ObjectMappers.newClientObjectMapper();
+    private final Map<String, Encoding.Deserializer<? extends T>> errorNameToJsonDeserializerMap;
+
+    EndpointErrorDecoder(
+            Map<String, TypeMarker<? extends T>> errorNameToTypeMap, Optional<Encoding> maybeJsonEncoding) {
+        this.errorNameToJsonDeserializerMap = maybeJsonEncoding
```
carterkozak: I wonder if the JSON `Encoding` should be required, and used for the RemoteException `SerializableError` deserialization as well?

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/EndpointErrorDecoder.java
```diff
@@ -0,0 +1,250 @@
...
+                .<Map<String, Encoding.Deserializer<? extends T>>>map(
+                        jsonEncoding -> errorNameToTypeMap.entrySet().stream()
+                                .collect(Collectors.toMap(
+                                        Map.Entry::getKey, entry -> jsonEncoding.deserializer(entry.getValue()))))
+                .orElseGet(Collections::emptyMap);
```
carterkozak: I might simplify this to avoid streams, for readability using `ImmutableMap.copyOf(Maps.transformValues(map, jsonEncoding::deserializer))` (immutable map copy is required because `transformValues` does a lazy transformation)

dialogue-target/src/main/java/com/palantir/dialogue/DeserializerArgs.java
```diff
@@ -0,0 +1,108 @@
...
+    public static final class Builder<T> {
+        private boolean buildInvoked = false;
+        private TypeMarker<T> baseType;
+        private boolean baseTypeSet = false;
+        private TypeMarker<? extends T> expectedResultType;
+        private boolean expectedResultSet = false;
+        private final Map<String, TypeMarker<? extends T>> errorNameToTypeMarker;
+
+        @SuppressWarnings("NullAway")
+        // We ensure that the baseType and expectedResultType are set before building.
```
carterkozak: You can annotate those builder fields with `@Nullable` instead of suppressing `NullAway` here, this way nullaway helps us remember to check for nullness when they're used within the builder. `baseTypeSet` be removed in favor of comparing `baseType` with `null`

dialogue-target/src/main/java/com/palantir/dialogue/DeserializerArgs.java
```diff
@@ -0,0 +1,108 @@
...
+        }
+
+        public Builder<T> withBaseType(@Nonnull TypeMarker<T> base) {
+            checkNotBuilt();
+            this.baseType = Preconditions.checkNotNull(base, "base type must be non-null");
+            this.baseTypeSet = true;
+            return this;
+        }
+
+        public Builder<T> withExpectedResult(TypeMarker<? extends T> expectedResultT) {
```
carterkozak: perhaps we use "success" here instead of "expected", for uniformity with "error" naming?

dialogue-target/src/main/java/com/palantir/dialogue/DeserializerArgs.java
```diff
@@ -0,0 +1,108 @@
...
+        private boolean expectedResultSet = false;
+        private final Map<String, TypeMarker<? extends T>> errorNameToTypeMarker;
+
+        @SuppressWarnings("NullAway")
+        // We ensure that the baseType and expectedResultType are set before building.
+        private Builder() {
+            this.errorNameToTypeMarker = new HashMap<>();
+        }
+
+        public Builder<T> withBaseType(@Nonnull TypeMarker<T> base) {
```
carterkozak: We tend not to use `with` prefixes on builders, removing the prefix keeps things concise.

`with` is more common as a utility function to make a copy of an existing immutable object, where one field replaced with a new value (I think immutables `Immutable*` generated types follow this convention when the immutable types are used directly)

dialogue-target/src/main/java/com/palantir/dialogue/DeserializerArgs.java
```diff
@@ -0,0 +1,108 @@
...
+            Preconditions.checkState(!buildInvoked, "Build has already been called");
+        }
+
+        private void checkRequiredArgsSet() {
+            Preconditions.checkState(baseTypeSet, "base type must be set");
+            Preconditions.checkState(expectedResultSet, "expected result type must be set");
+        }
+    }
+
+    public TypeMarker<? extends T> baseType() {
```
carterkozak: Are there cases where this isn't `TypeMarker<T>` but a subtype of T?

dialogue-serde/src/test/java/com/palantir/conjure/java/dialogue/serde/ErrorDecoderTest.java
```diff
@@ -54,24 +60,26 @@ public final class ErrorDecoderTest {
...
 
     private static final ErrorDecoder decoder = ErrorDecoder.INSTANCE;
+    private static final EndpointErrorDecoder<?> endpointErrorDecoder =
+            new EndpointErrorDecoder<>(Collections.emptyMap(), Optional.empty());
 
-    @Test
-    public void extractsRemoteExceptionForAllErrorCodes() {
+    @ParameterizedTest
+    @ValueSource(booleans = {true, false})
+    public void extractsRemoteExceptionForAllErrorCodes(boolean isLegacyErrorDecoder) {
```
carterkozak: Using an enum source makes the test results a bit clearer

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/EndpointErrorDecoder.java
mpritham: On second thought, I don't think we need to have an encoding passed in. We expect the error to be JSON and can use `Encodings.json()` for both the error types and `SerializableError`, whether it's passed into `ConjureBodySerDe` as a supported encoder or not.

My thinking is: the purpose of the encodings passed into the constructor of `ConjureBodySerDe` is to deserialize the response, but when we _know_ the error response will be a certain format, we can attempt to deserialize with that format directly.

Let me know if that tracks with you. I put up a diff with showing what I mean.

dialogue-serde/src/test/java/com/palantir/conjure/java/dialogue/serde/ErrorDecoderTest.java
mpritham: It does. Updated.

dialogue-target/src/main/java/com/palantir/dialogue/DeserializerArgs.java
mpritham: I don't think so. The `baseType` variable is also constrained to `TypeMarker<T>`. Updated.

dialogue-target/src/main/java/com/palantir/dialogue/DeserializerArgs.java
mpritham: I see. I do notice that as well. Updated.

dialogue-target/src/main/java/com/palantir/dialogue/DeserializerArgs.java
mpritham: Makes sense! Updated.

dialogue-target/src/main/java/com/palantir/dialogue/DeserializerArgs.java
mpritham: That's cleaner! Updated.

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/EndpointErrorDecoder.java
carterkozak: That makes sense to me!

One minor tweak I'd suggest is that we should still evaluate to a `Map<String, Encoding.Deserializer<? extends T>>` in the constructor, since the evaluation isn't guaranteed to be inexpensive, this way we don't have to produce a new deserializer every time it's needed.

mpritham: Yep that's true. Updated.

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/EndpointErrorDecoder.java
```diff
@@ -0,0 +1,250 @@
...
+            body = toString(response.body());
+        } catch (NullPointerException | IOException e) {
+            UnknownRemoteException exception = new UnknownRemoteException(code, "<unparseable>");
+            exception.initCause(e);
+            throw exception;
+        }
+
+        Optional<String> contentType = response.getFirstHeader(HttpHeaders.CONTENT_TYPE);
+        String jsonContentType = "application/json";
+        if (contentType.isPresent() && Encodings.matchesContentType(jsonContentType, contentType.get())) {
```
carterkozak: Perhaps we update this to leverage the `Encoding` with something along these lines?
```java
Encodings.matchesContentType(JSON_ENCODING.getContentType(), contentType.get())
```

If it's not terribly troublesome, we may want to take the existing json encoding from the configured set of encodings when it's present, and create a new one as a fallback. This way if the client is configured with encodings wrapped to produce metrics/etc, we can retain that functionality.

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/EndpointErrorDecoder.java
```diff
@@ -0,0 +1,250 @@
...
+            try {
+                JsonNode node = MAPPER.readTree(body);
+                JsonNode errorNameNode = node.get("errorName");
```
carterkozak: As written we're fully transforming the input json into a JsonNode tree, and reading at most a single string element.
We can define something like this to allow jackson to extract only the pieces of data we're interested in:

```java
private record NamedError(@JsonProperty("errorName") String errorName) {}
```

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/EndpointErrorDecoder.java
```diff
@@ -0,0 +1,268 @@
...
+        SerializableError serializableError = JSON_ENCODING
+                .deserializer(new TypeMarker<SerializableError>() {})
+                .deserialize(new ByteArrayInputStream(body));
+        return new RemoteException(serializableError, code);
+    }
+
+    private static byte[] toByteArray(InputStream body) throws IOException {
+        try (Reader reader = new InputStreamReader(body, StandardCharsets.UTF_8)) {
+            return CharStreams.toString(reader).getBytes(StandardCharsets.UTF_8);
+        }
```
carterkozak: We don't need to convert the bytestream into a charstream (reader) to then convert back again in the other direction!

```suggestion
        try (body) {
            return body.readAllBytes();
        }
```

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/EndpointErrorDecoder.java
```diff
@@ -0,0 +1,268 @@
...
+                throw new UnknownRemoteException(code, new String(body, StandardCharsets.UTF_8));
+            }
+        }
+
+        throw new UnknownRemoteException(code, new String(body, StandardCharsets.UTF_8));
+    }
+
+    private static RemoteException createRemoteException(byte[] body, int code) throws IOException {
+        SerializableError serializableError = JSON_ENCODING
+                .deserializer(new TypeMarker<SerializableError>() {})
```
carterkozak: Can we initialize this `Deserializer<SerializableError>` once above, rather than creating it anew each time it's needed?

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/EndpointErrorDecoder.java
mpritham: 🤦 This was dumb. Fixed.

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/EndpointErrorDecoder.java
```diff
@@ -0,0 +1,266 @@
...
+        this(errorNameToTypeMap, Optional.empty());
+    }
+
+    EndpointErrorDecoder(
+            Map<String, TypeMarker<? extends T>> errorNameToTypeMap, Optional<Encoding> maybeJsonEncoding) {
+        this.errorNameToJsonDeserializerMap = ImmutableMap.copyOf(
+                Maps.transformValues(errorNameToTypeMap, maybeJsonEncoding.orElse(JSON_ENCODING)::deserializer));
+    }
+
+    public boolean isError(Response response) {
```
carterkozak: we can remove `public ` from all the utility methods here, since the class itself is package private

carterkozak:
Looks solid -- want to start on the codegen piece using an RC?

mpritham:
Yep. Added `do-not-merge`. Need a +1 to cut the RC.

carterkozak (dismissed):
Approving to cut an RC for iteration with conjure-java codegen 👍

dialogue-target/src/main/java/com/palantir/dialogue/DeserializerArgs.java
```diff
@@ -0,0 +1,98 @@
...
+            this.errorNameToTypeMarker.put(
+                    Preconditions.checkNotNull(errorName, "error name must be non-null"),
+                    Preconditions.checkNotNull(errorT, "error type must be non-null"));
+            return this;
+        }
+
+        public DeserializerArgs<T> build() {
+            checkNotBuilt();
+            Preconditions.checkNotNull(baseType, "base type must be set");
+            Preconditions.checkNotNull(successType, "success type must be set");
```
mpritham, line 82: fysa @carterkozak, I'm thinking through how we can handle endpoints that don't return anything but can return errors. This `success` type doesn't always need to be set.

carterkozak: Wouldn’t it need to be an empty type or similar? I wouldn’t want to use null, since that’s a recipe for NPEs

mpritham: Yeah. I was initially thinking that if it's an empty type, it wouldn't need to be provided by the caller (via `DeserializerArgs`). That doesn't work though: the caller needs to get _some_ object to indicate a successful result.

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/JacksonEmptyContainerLoader.java
```diff
@@ -63,6 +63,11 @@ private Optional<Object> constructEmptyInstance(Type type, TypeMarker<?> origina
             return jacksonInstance;
         }
 
+        Method noArgStaticFactoryMethod = getFirstNoArgCreatorStaticMethod(type);
```
mpritham: Actually instead of relying on reflection, we could use a custom deserializer while overriding `getNullValue`. I prefer this over using reflection. We could define the custom deserializer once in Conjure-Java in a shared utilities class:

```java
abstract static class CustomNullDeserializer<T> extends JsonDeserializer<T> {
    public abstract T create();

    @Override
    public T deserialize(JsonParser _parser, DeserializationContext _ctxt) {
         throw ...
    }

    @Override
    public T getNullValue(DeserializationContext _ctxt) {
        return create();
    }
}

static class EmptyReturnValueDeserializer extends CustomNullDeserializer<EmptyReturnValue> {
    @Override
    public EmptyReturnValue create() {
        return new EmptyReturnValue();
    }
}

@Generated("by conjure-java")
@JsonDeserialize(using = EmptyReturnValueDeserializer.class)
record EmptyReturnValue() implements EmptyBodyEndpointReturnBaseType {}
```

mpritham: Updated.

carterkozak: This has quite a bit in common with our handling of null-to-empty coercion used for empty optionals (and aliases of optionals) when endpoints return 204 with no content -- we have to handle explicit `null` tokens and empty streams the same way.
I wonder if we could reuse any of that code here?

carterkozak: There's a trade-off between reflection, and a great deal of additional codegen we have to format/ship/compile/etc. In practice this deserializer codegen asks jackson to use reflection in a way that jackson understands, which may be much more verbose than writing something more specific into the deserializer.

mpritham: > I wonder if we could reuse any of that code here?

I realize I didn't make a new commit when I replaced the old version with the custom deserializer approach. Updated with an implementation that refactors `JacksonEmptyContainerLoader#constructEmptyInstance`.

> In practice this deserializer codegen asks jackson to use reflection in a way that jackson understands, which may be much more verbose than writing something more specific into the deserializer.

I agree. It is more generated code that needs to be packaged and compiled. I was initially hesitant on adding additional special handling here (i.e. do X if there is a 0-parameter static factory annotated with JsonCreator) because I thought it'd be nice to separate the concerns - the client code (Conjure-Java) that calls into Dialogue would have to be responsible for specifying how null values should be deserialized. But, ultimately this is used in the implementation of **Conjure**BodySerDe: the coupling is expected, and as you pointed out, there is some special handling for optionals and collections already.

dialogue-target/src/main/java/com/palantir/dialogue/BodySerDe.java
```diff
@@ -41,9 +43,13 @@ public interface BodySerDe {
      */
     Deserializer<InputStream> inputStreamDeserializer();
 
+    <T> Deserializer<T> inputStreamDeserializer(DeserializerArgs<T> deserializerArgs);
```
mpritham, line 54: I also added methods to deserialize input streams and optional input streams. The idea is that the type `T.Success` here has a constructor with `InputStream` as the only parameter, and on a successful binary response body, a `SuccessT(response.body())` will be returned.

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ConjureBodySerDe.java
```diff
@@ -365,4 +447,35 @@ public String toString() {
...
+        if (successT.getType() instanceof Class<?>) {
+            try {
+                return (T) ((Class<?>) successT.getType())
+                        .getConstructor(InputStream.class)
+                        .newInstance(inputStream);
+            } catch (ReflectiveOperationException e) {
+                throw new SafeRuntimeException("Failed to create success type", e);
+            }
+        }
+        return null;
```
carterkozak: in the else case, we should throw rather than returning a null ref that will cause something else to throw in a more confusing way later on!

I might suggest splitting this up a bit as well: We should be able to do the validation on `successT.getType()` once when the deserializer is first created, as well as the `.getConstructor(<arg>)`, that way we only need to execute `newInstance` on a per-response basis.

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/JacksonEmptyContainerLoader.java
```diff
@@ -152,6 +165,21 @@ private static Method getJsonCreatorStaticMethod(Type type) {
...
+    @Nullable
+    private static Constructor<?> getEmptyRecordCanonicalConstructor(Type type) {
+        if (type instanceof Class) {
+            Class<?> clazz = (Class<?>) type;
+            if (clazz.isRecord()) {
+                for (Constructor<?> ctor : clazz.getDeclaredConstructors()) {
+                    if (0 == ctor.getParameterCount()) {
+                        return ctor;
+                    }
+                }
```
carterkozak, line 181: Perhaps we check `clazz.getRecordComponents().length == 0` to ensure we're not seeing a record with fields and a special no-arg ctor?

I prefer to be as precise as possible for these sorts of things, otherwise we can allow unexpected behavior

dialogue-target/src/main/java/com/palantir/dialogue/ConjureErrors.java
```diff
@@ -0,0 +1,59 @@
...
+ */
+
+package com.palantir.dialogue;
+
+import com.palantir.logsafe.Safe;
+
+public final class ConjureErrors {
+    private ConjureErrors() {}
+
+    public abstract static class BaseEndpointError<T> {
```
carterkozak: I don't think the `ConjureErrors` type buys us anything here, can we remove that and make `BaseEndpointError` the top-level class? Might consider renaming it to `EndpointError`, I don't think `Base` adds a ton

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/JacksonEmptyContainerLoader.java
mpritham: Makes sense. I didn't think of that case.

carterkozak (dismissed):
approving for an RC

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/JacksonEmptyContainerLoader.java
```diff
@@ -84,9 +85,22 @@ private Optional<Object> constructEmptyInstance(Type type, TypeMarker<?> origina
             }
         }
 
+        Constructor<?> emptyRecordConstructor = getEmptyRecordCanonicalConstructor(type);
```
carterkozak, line 88: Not exactly related to this PR, but we'll need to remember to register these with reachability-metadata!

dialogue-target/src/main/java/com/palantir/dialogue/BodySerDe.java
```diff
@@ -28,6 +28,8 @@ public interface BodySerDe {
     /** Creates a {@link Deserializer} for the requested type. Deserializer instances should be reused. */
     <T> Deserializer<T> deserializer(TypeMarker<T> type);
 
+    <T> Deserializer<T> deserializer(DeserializerArgs<T> deserializerArgs);
```
bjlaub, line 36: nit: should these additions have some javadoc?

mpritham: Added some docs to the `DeserializerArgs<T>` class specifying what `baseType`, `successType` and `errorNameToTypeMarker` are

bjlaub (dismissed):
looks like the latest commit just needs some fixes on the javadoc, but otherwise lgtm

carterkozak (approved):

PR #2459, opened 2025-01-08
Don't mock `ErrorDecoder` in `ConjureBodySerDeTest`

mpritham:
## Before this PR
When testing `ConjureBodySerDe`, we mock the function calls to `ErrorEncoder`. When implementing #2443, I introduce a replacement for `ErrorDecoder`. I would like to be able to validate that we don't have regressions in the current error handling logic and not mocking `ErrorDecoder` will help: I can validate that the logic in `ErrorDecoder` is still valid with my changes in #2443 where I replace logic in `ErrorDecoder` with calls to a delegate `EndpointErrorDecoder`.

## After this PR
==COMMIT_MSG==
Don't mock `ErrorDecoder` in `ConjureBodySerDeTest`
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-2459.v2.yml
+++ b/changelog/@unreleased/pr-2459.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Don't mock `ErrorDecoder` in `ConjureBodySerDeTest`
+  links:
+  - https://github.com/palantir/dialogue/pull/2459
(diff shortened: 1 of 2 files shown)


changelog/@unreleased/pr-2459.v2.yml
```diff
@@ -0,0 +1,5 @@
+type: improvement
+improvement:
+  description: Don't mock `ErrorDecoder` in `ConjureBodySerDeTest`
+  links:
+  - https://github.com/palantir/dialogue/pull/2459
```
carterkozak, line 5: no need to add changelogs for changes that only update tests, users won't be impacted. Doesn't hurt anything, no need to revert this, just something for the future :-)

carterkozak (approved):

changelog/@unreleased/pr-2459.v2.yml
mpritham: Sounds good!

PR #2474, opened 2025-01-22
Remove unsupported `mesh-` URIs

carterkozak:
We deprecated support for mesh mode about a year ago, and have removed support from our core infrastructure. The code is no longer used, and can be removed for simplicity.

This should have no impact for users.

==COMMIT_MSG==
Remove unsupported `mesh-` URI support. This should have no impact for users.
==COMMIT_MSG==

## Possible downsides?
Possible the prefix could be hard-coded somewhere we don't expect, however if so, it's certainly not being used in a supported or intended way! We have not produced such URIs since early last year, and the infrastructure that handled such requests is _gone_.


mpritham (approved):
Thanks!

mpritham:
I notice it's also in [`conjure-go-runtime`](https://github.com/palantir/conjure-go-runtime/blob/db1a104c3b006009a16131077a29a2371459ebca/conjure-go-client/httpclient/internal/request_retrier.go#L26)

PR #2483, opened 2025-01-30
[RC] Use Guava instead of Caffeine caches for Android

ellisjoe:
## Before this PR

Caffeine caches run into issues when running on Android, minimally related to its usage of `java.lang.System$Logger` which isn't present.
## After this PR
Swap over to using Guava caches which should be fully compatible on Android devices.
==COMMIT_MSG==

==COMMIT_MSG==

## Possible downsides?

Ideally this change would be done in a way where we pick Caffeine caches when running on standard JVMs and Guava caches when running on Android devices.


dialogue-clients/src/main/java/com/palantir/dialogue/clients/DnsSupport.java
```diff
@@ -93,15 +93,19 @@ final class DnsSupport {
     /**
      * Shared cache of string to parsed URI. This avoids excessive allocation overhead when parsing repeated targets.
      */
-    private static final LoadingCache<String, MaybeUri> uriCache = CacheStats.of(
-                    SharedTaggedMetricRegistries.getSingleton(), "dialogue-uri")
```
carterkozak, line 98: I've dropped this instrumentation since we don't have a 1-1 replacement using the same timeseries names for guava caches. If this unblocks android stuff and we move forward, we can look into adding some guava cache utility functionality.

carterkozak (dismissed):
Approving to create an RC release for testing on Android

PR #2485, opened 2025-01-30
set deadlines on outbound requests

bjlaub:
## Before this PR
No `Expect-Within` header is added to outbound requests

## After this PR
Adds the [deadlines-java](https://github.com/palantir/deadlines-java) library and utilizes it to set `Expect-Within` headers on outbound requests.

This change adds a `DeadlineAdvertisementChannel` which calls `Deadlines.encodeToRequest` to set an `Expect-Within` header equal to the configured socket read timeout for the client, or a remaining deadline read from a `TraceLocal` per the implementation in deadlines-java (whichever is smaller).

deadlines-java does not currently doe any enforcement on deadline expiration, and instead only reports metrics when a deadline is exceeded before sending an outbound request.

==COMMIT_MSG==
set deadlines on outbound requests
==COMMIT_MSG==


dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueChannel.java
```diff
@@ -242,6 +242,8 @@ private static ImmutableList<LimitedChannel> createHostChannels(
                 channel = HostMetricsChannel.create(cf, channel, targetUri.uri());
                 channel =
                         new TraceEnrichingChannel(channel, DialogueTracing.tracingTags(cf, uriIndexForInstrumentation));
+                channel = new DeadlineAdvertisementChannel(
```
bjlaub, line 245: are there other places we need to create one of these?

carterkozak: There shouldn't be -- this should do the trick! This position looks great because executes after the queues and retrying channels so that it will be re-applied on each request with an updated `Expect-With` value

dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueChannel.java
```diff
@@ -242,6 +242,8 @@ private static ImmutableList<LimitedChannel> createHostChannels(
...
+                        channel, cf.clientConf().readTimeout());
```
carterkozak, line 246: There are a couple places which set a read timeout value of zero expecting it to be infinite. We handle this case in the apache httpclient shim here: https://github.com/palantir/dialogue/blob/f4915fc9eee84d202ed1102c41af31995594d670/dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientChannels.java#L641-L648
I think we'll need to handle that case in the DeadlineAdvertisementChannel

bjlaub: 👍 fixed!

carterkozak (approved):

PR #2489, opened 2025-02-11
Suppress deprecation warning in generated Dialogue factory if needed

kevinychen:
## Before this PR
Declaring a Dialogue service with a deprecated method like
```
@DialogueService(MyBlockingDialogueServiceFactory.class)
public interface MyServiceBlocking {
    @Request(method = HttpMethod.GET, path = "/path")
    @Deprecated
    void foo();
```
will fail compilation with `-Xlint:deprecation` for the generated DialogueServiceFactory:
```
...Factory.java: warning: [deprecation] foo(AuthHeader) in MyServiceBlocking has been deprecated
            public void foo(
                    ^
```

## After this PR
`@SuppressWarnings("deprecation")` is now included in the generated DialogueServiceFactory:
```
@Override
@SuppressWarnings("deprecation")
public void foo() {
    Request.Builder _request = Request.builder();
    ...
}
```
The implementation in this PR is basically the same as for Conjure, [here](https://github.com/palantir/conjure-java/blob/4114c4142b2d513e988640629c4a3cc12a0aaf6a/conjure-undertow-processor/src/main/java/com/palantir/conjure/java/undertow/processor/data/EndpointDefinitions.java#L142-L143) and [here](https://github.com/palantir/conjure-java/blob/4114c4142b2d513e988640629c4a3cc12a0aaf6a/conjure-java-core/src/main/java/com/palantir/conjure/java/services/UndertowServiceHandlerGenerator.java#L248-L253).

==COMMIT_MSG==
Suppress deprecation warning in generated Dialogue factory if needed
==COMMIT_MSG==

### DIFF
--- a/dialogue-annotations-processor/src/test/java/com/palantir/dialogue/annotations/processor/DialogueRequestAnnotationsProcessorTest.java
+++ b/dialogue-annotations-processor/src/test/java/com/palantir/dialogue/annotations/processor/DialogueRequestAnnotationsProcessorTest.java
@@ -112,8 +112,14 @@ public final class DialogueRequestAnnotationsProcessorTest {
         String generatedFqnClassName = clazz.getPackage().getName() + "." + generatedClassName;
         String generatedClassFileRelativePath = generatedFqnClassName.replaceAll("\\.", "/") + ".java";
         assertThat(compilation.generatedFile(StandardLocation.SOURCE_OUTPUT, generatedClassFileRelativePath))
-                .hasValueSatisfying(
-                        javaFileObject -> assertContentsMatch(javaFileObject, generatedClassFileRelativePath));
+                .hasValueSatisfying(javaFileObject -> {
+                    assertContentsMatch(javaFileObject, generatedClassFileRelativePath);
+
+                    assertThat(Compiler.javac()
+                                    .withOptions("-source", "11", "-Werror", "-Xlint:deprecation", "-Xlint:unchecked")
+                                    .compile(javaFileObject))
+                            .succeededWithoutWarnings();
+                });
     }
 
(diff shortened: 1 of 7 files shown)


dialogue-annotations-processor/src/test/java/com/palantir/dialogue/annotations/processor/DialogueRequestAnnotationsProcessorTest.java
```diff
@@ -123,9 +129,7 @@ private Compilation compileTestClass(Path basePath, Class<?> clazz) {
         try {
             return Compiler.javac()
                     // This is required because this tool does not know about our gradle setting.
-                    .withOptions("-source", "11")
-                    .withOptions("-Werror")
-                    .withOptions("-Xlint:unchecked")
```
kevinychen, line 128: `.withOptions(...)` overwrites previous options, so fixed this

carterkozak (commented):
Would it be sufficient to add a matching deprecation annotation to the override instead of suppressing? I think that may be technically more accurate. Ideally we preserve the “for removal” flag/etc

kevinychen:
That makes sense, can update this. Admittedly I had just copied the logic from the Conjure Dialogue generator.
Do you care at all about preserving any of the `@Deprecated` fields? (`@Deprecated(since = "since", forRemoval = true)`)

carterkozak:
> Do you care at all about preserving any of the @Deprecated fields? (@Deprecated(since = "since", forRemoval = true))

it would be nice to have, but it's not worth blocking on! Thanks for the contribution!

carterkozak (approved):

PR #2537, opened 2025-03-24
HTTP/1.0 does not support keep-alive

schlosna:
## Before this PR

Dialogue would attempt to use default Apache HTTP client connection reuse timeout of 3 minutes for an HTTP/1.0 response. As HTTP/1.0 closes the connection after each request/response interaction by default, this would result in connection reuse failures resulting in retries and exponential backoff.

## After this PR
==COMMIT_MSG==

See https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Keep-Alive

> HTTP/1.0 closes the connection after each request/response interaction by
> default, so persistent connections in HTTP/1.0 must be explicitly negotiated.
> Some clients and servers might wish to be compatible with previous approaches
> to persistent connections, and can do this with a Connection: keep-alive
> request header. Additional parameters for the connection can be requested with
> the Keep-Alive header.
>
> Warning: Connection-specific header fields such as Connection and Keep-Alive
> are prohibited in HTTP/2 and HTTP/3. Chrome and Firefox ignore them in HTTP/2
> responses, but Safari conforms to the HTTP/2 specification requirements and
> does not load any response that contains them.

==COMMIT_MSG==

### DIFF
--- a/dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/InactivityValidationAwareConnectionKeepAliveStrategy.java
+++ b/dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/InactivityValidationAwareConnectionKeepAliveStrategy.java
@@ -30,6 +30,8 @@ import org.apache.hc.client5.http.protocol.HttpClientContext;
 import org.apache.hc.core5.http.HeaderElement;
 import org.apache.hc.core5.http.HeaderElements;
 import org.apache.hc.core5.http.HttpResponse;
+import org.apache.hc.core5.http.HttpVersion;
+import org.apache.hc.core5.http.ProtocolVersion;
 import org.apache.hc.core5.http.message.MessageSupport;
 import org.apache.hc.core5.http.protocol.HttpContext;
 import org.apache.hc.core5.util.TimeValue;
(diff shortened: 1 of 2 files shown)


dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/InactivityValidationAwareConnectionKeepAliveStrategy.java
```diff
@@ -86,6 +88,15 @@ public TimeValue getKeepAliveDuration(HttpResponse response, HttpContext context
...
             }
         }
+
+        ProtocolVersion responseVersion = response.getVersion();
+        if (responseVersion != null && responseVersion.lessEquals(HttpVersion.HTTP_1_0)) {
+            log.info(
+                    "HTTP/1.0 requires keep-alive negotiation, but no timeout was found",
+                    SafeArg.of("protocolVersion", responseVersion));
+            return TimeValue.ZERO_MILLISECONDS;
+        }
```
carterkozak, line 98: I think this suggests that `ConnectionReuseStrategy.keepAlive` incorrectly returned `true` -- I think the expectation is for `getKeepAliveDuration` to be called after a `ConnectionReuseStrategy.keepAlive` check

schlosna: yeah, this case should already be handled by [DefaultConnectionReuseStrategy.keepAlive](https://github.com/apache/httpcomponents-core/blob/rel/v5.3.4/httpcore5/src/main/java/org/apache/hc/core5/http/impl/DefaultConnectionReuseStrategy.java#L107-L160) returning false, I'm trying to run down a reproducer to test with full HTTP headers.

carterkozak: We may be able to adapt something like this to reproduce the observed behavior: #2538

schlosna:
Closing as we're not moving forward with this

PR #2542, opened 2025-04-02
Reduce deprecation logging to debug

pkoenig10:
These logs are quite noisy given that:
- We have a metric
- Clients already know where they are using deprecated endpoints because the endpoints are marked `@Deprecated` (we should encourage repos to enable `-Xlint:deprecation`).

These logs do not proactively alert service owners. I posit that no one actively looks at these logs.

This PR reduces the log level to `DEBUG`. If clients want to get more information about exactly what requests are triggering the deprecation metric, they can enable debug logs.

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/DeprecationWarningChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/DeprecationWarningChannel.java
@@ -88,8 +88,8 @@ final class DeprecationWarningChannel implements EndpointChannel {
             }
 
             meterSupplier.get().mark();
-            if (log.isWarnEnabled() && tryAcquire(channelName, endpoint)) {
-                log.warn(
+            if (log.isDebugEnabled() && tryAcquire(channelName, endpoint)) {
+                log.debug(
                         "Using a deprecated endpoint when connecting to service",
                         SafeArg.of("channelName", channelName),
                         SafeArg.of("serviceName", endpoint.serviceName()),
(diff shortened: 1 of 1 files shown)


dialogue-core/src/main/java/com/palantir/dialogue/core/DeprecationWarningChannel.java
```diff
@@ -111,11 +111,7 @@ private FutureCallback<Response> createCallback(String channelName, Endpoint end
     private boolean tryAcquire(String channelName, Endpoint endpoint) {
         LoggingRateLimiterKey key =
                 LoggingRateLimiterKey.of(channelName, endpoint.serviceName(), endpoint.endpointName());
-        if (loggingRateLimiter.getIfPresent(key) == null) {
-            loggingRateLimiter.put(key, SENTINEL);
-            return true;
-        }
-        return false;
+        return loggingRateLimiter.asMap().putIfAbsent(key, SENTINEL) == null;
```
pkoenig10, line 114: Drive-by fix to make this cache update atomic.

invliD:
> * Clients already know where they are using deprecated endpoints because the endpoints are marked `@Deprecated` (we should encourage repos to enable `-Xlint:deprecation`).

That's not entirely true here. This log line / metric are emitted when the _server_ says the endpoint is deprecated. The client may be using an outdated client that does not know about the deprecation yet.

pkoenig10:
Ah right. But we do upgrade client libraries somewhat aggressively, which makes this less of a concern.

mpritham:
> But we do upgrade client libraries somewhat aggressively, which makes this less of a concern.

I think for clients built with dialogue-annotations-processor, servers could return a HTTP `deprecation` header and clients wouldn't know at compile time.

mpritham:
But also, as you said, folks can enable debug logging if they'd like to investigate which of their endpoints are deprecated. 

Separately: Perhaps we should add some (very low cadence) monitoring for this. Would not like to train folks to ignore the monitor exactly, but it'd be good for devs to ack their use of deprecated endpoints.


Curious for your thoughts @bjlaub.

Shall we also update the javadoc and the class name to reflect that this is no longer a `warning`, but debug information?

pkoenig10:
> Shall we also update the javadoc and the class name to reflect that this is no longer a warning, but debug information?

I think the class name is fine. I interpret the name as a lower-case "warning" rather than "`WARN` log level".

pkoenig10:
@bjlaub do you have any additional thoughts here?

bjlaub:
apologies, this slipped through the cracks on my end. will try to take a look shortly

pkoenig10:
@bjlaub Gentle ping on this

mpritham (approved):
Sorry for the delay here. Thanks!

PR #2547, opened 2025-04-07
Copy over existing request query params when creating a new request

rbapat:
## Before this PR
Request objects do not copy over query parameters when the builder is created from an existing object

## After this PR
==COMMIT_MSG==
When request objects are created using a builder on an existing object, query parameters will be copied over from the previous object to match the headers and path params implementation 
==COMMIT_MSG==

### DIFF
--- a/dialogue-target/src/main/java/com/palantir/dialogue/Request.java
+++ b/dialogue-target/src/main/java/com/palantir/dialogue/Request.java
@@ -307,7 +307,12 @@ public final class Request {
         private ListMultimap<String, String> mutableQueryParams() {
             if (!isQueryMutable()) {
                 setQueryMutable();
-                queryParams = Multimaps.newListMultimap(new LinkedHashMap<>(), MAP_VALUE_FACTORY);
+                ListMultimap<String, String> mutable =
+                        Multimaps.newListMultimap(new LinkedHashMap<>(), MAP_VALUE_FACTORY);
+                if (!queryParams.isEmpty()) {
+                    queryParams.forEach(mutable::put);
+                }
+                queryParams = mutable;
             }
             return queryParams;
         }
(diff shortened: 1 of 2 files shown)


dialogue-target/src/main/java/com/palantir/dialogue/Request.java
```diff
@@ -307,7 +307,12 @@ private ListMultimap<String, String> mutableHeaderParams() {
         private ListMultimap<String, String> mutableQueryParams() {
             if (!isQueryMutable()) {
                 setQueryMutable();
-                queryParams = Multimaps.newListMultimap(new LinkedHashMap<>(), MAP_VALUE_FACTORY);
+                ListMultimap<String, String> mutable =
+                        Multimaps.newListMultimap(new LinkedHashMap<>(), MAP_VALUE_FACTORY);
+                if (!queryParams.isEmpty()) {
+                    queryParams.forEach(mutable::put);
+                }
+                queryParams = mutable;
```
ash211, line 315: please add a test case that would fail before and now succeeds as a result of this change

rbapat: Just added!

dialogue-target/src/test/java/com/palantir/dialogue/RequestTest.java
```diff
@@ -88,4 +92,38 @@ void from_method_query_param_mutation() {
...
+                .putHeaderParams("header2", "headerVal2")
+                .putPathParams("path2", "pathVal2")
+                .build();
+
+        // make sure new query/header/path parameters were added in addition to whatever was in `oldRequest`
+        assertThat(newRequest)
+                .extracting(Request::queryParams)
+                .extracting(Multimap::entries, collection(Map.Entry.class))
+                .containsExactlyInAnyOrder(
+                        MapEntry.entry("query1", "queryVal1"), MapEntry.entry("query2", "queryVal2"));
```
schlosna, line 116: would be good to also test unique query param in each of old & new requests as well as duplicate query param names (since they're a `ListMultimap`)

PR #2572, opened 2025-05-09
BaseUrl$UrlEncoder uses precomputed CharMatchers

schlosna:
## Before this PR

In some profiles, noticed that `BaseUrl$UrlEncoder.encodeQueryNameOrValue` was using `CharMatcher$Or.matches(char)` and not an optimized precomputed bit set `CharMatcher`

https://github.com/palantir/dialogue/blob/f93d327dc9dad09d0457a404e2bf3a857ffc72e2/dialogue-core/src/main/java/com/palantir/dialogue/core/BaseUrl.java#L239-L273

https://github.com/palantir/dialogue/blob/f93d327dc9dad09d0457a404e2bf3a857ffc72e2/dialogue-core/src/main/java/com/palantir/dialogue/core/BaseUrl.java#L299-L300

## After this PR
==COMMIT_MSG==
Precompute optimized CharMatchers for BaseUrl$UrlEncoder
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-2572.v2.yml
+++ b/changelog/@unreleased/pr-2572.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Precompute optimized CharMatchers for BaseUrl$UrlEncoder
+  links:
+  - https://github.com/palantir/dialogue/pull/2572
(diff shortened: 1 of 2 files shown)


mpritham (dismissed):
Thanks!

mpritham (approved):

PR #2604, opened 2025-07-10
Return a failed future on deadline expiration

kunalrkak:
## Before this PR

From the `Channel` Javadoc: 
<h4>Behavior</h4>
Implementations of {@link Channel#execute(Endpoint, Request)} must never throw. A failed {@link ListenableFuture} must be returned instead.

The DeadlineExpiredException is logged when we handle throwables [here](https://github.com/palantir/conjure-java/blob/6ba3a53d2a04b2a1e65b0bf32f43e2c7ff799bb4/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptions.java#L205-L209), so we should just be able to return a failed future. 

## After this PR
==COMMIT_MSG==
Makes `DeadlineAdvertisementChannel` adhere to the `Channel` spec by not throwing an exception.
==COMMIT_MSG==


aldexis (dismissed):

versions.props
```diff
@@ -8,7 +8,7 @@ com.palantir.conjure.java.api:* = 2.59.0
 com.palantir.conjure.java.runtime:* = 8.18.0
 com.palantir.conjure.java:* = 8.45.0
 com.palantir.conjure:conjure = 4.51.0
-com.palantir.deadlines:* = 0.12.0
+com.palantir.deadlines:* = 0.14.0
```
schlosna: I think you'll need to merge `develop` as excavator bumped this and the `versions.props` & `versions.lock` changes should drop out of this PR.

aldexis (approved):

PR #2628, opened 2025-07-29
[WIP] Accept conjure error params as json

bjlaub:
## After this PR
==COMMIT_MSG==

==COMMIT_MSG==


dialogue-clients/src/main/java/com/palantir/dialogue/clients/ReloadingClientFactory.java
```diff
@@ -135,6 +136,11 @@ default boolean dnsNodeDiscovery() {
         }
 
         Optional<ExecutorService> blockingExecutor();
+
+        @Value.Default
```
mpritham, line 140: I need access to this value in `ConjureRuntime` so I'm able to use it in generated code. 

The generated Dialogue services have a factory method providing two arguments: an `EndpointChannelFactory` (which is an interface with a method providing an `EndpointChannel`, and `ConjureRuntime`.

PR #2629, opened 2025-07-29
More descriptive channel names for enabled/disabled params

ash211:
## Before this PR
Users were seeing dialogue channels with names like `dialogue-actions-enabled` and were unsure what the `-enabled` meant. [Internal slack](https://palantir.slack.com/archives/C5MKHTMD4/p1753800307213799).

## After this PR
==COMMIT_MSG==
More descriptive channel names for enabled/disabled params
==COMMIT_MSG==

With this change, configuration options with generic names enabled/disabled have the name of the field prepended.

So now instead of `dialogue-actions-enabled`, they would see `dialogue-actions-client_qos_enabled`

## Possible downsides?
Any code parsing the format of these channel names may not be able to handle the new format. However, I don't believe the channel names are considered part of the stable public API of dialogue.

### DIFF
--- a/dialogue-clients/src/test/java/com/palantir/dialogue/clients/ChannelNamesTest.java
+++ b/dialogue-clients/src/test/java/com/palantir/dialogue/clients/ChannelNamesTest.java
@@ -34,6 +34,32 @@ class ChannelNamesTest {
     }
 
     @Test
+    void prepends_enabled_disabled_with_field_name() {
+        String channelName = ChannelNames.reloading(
+                "multipass",
+                ImmutableReloadingParams.builder()
+                        .scb(Refreshable.only(null))
+                        .clientQoS(ClientConfiguration.ClientQoS.ENABLED)
+                        .retryOnTimeout(ClientConfiguration.RetryOnTimeout.DISABLED)
+                        .build());
+        assertThat(channelName).isEqualTo("dialogue-multipass-CLIENT_QOS_ENABLED-RETRY_ON_TIMEOUT_DISABLED");
+    }
+
+    @Test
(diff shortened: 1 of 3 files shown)


dialogue-clients/src/test/java/com/palantir/dialogue/clients/ChannelNamesTest.java
```diff
@@ -33,6 +33,32 @@ void name_is_concise_by_default() {
...
+    @Test
+    void prepends_enabled_disabled_with_field_name() {
+        String channelName = ChannelNames.reloading(
+                "multipass",
+                ImmutableReloadingParams.builder()
+                        .scb(Refreshable.only(null))
+                        .clientQoS(ClientConfiguration.ClientQoS.ENABLED)
+                        .retryOnTimeout(ClientConfiguration.RetryOnTimeout.DISABLED)
+                        .build());
+        assertThat(channelName).isEqualTo("dialogue-multipass-CLIENT_QOS_ENABLED-RETRY_ON_TIMEOUT_DISABLED");
```
ash211, line 45: this would have been `dialogue-multipass-ENABLED-DISABLED` before, which is incomprehensible

aldexis (approved):
Definitely in favor of this, but wouldn't this risk breaking existing metrics and some dashboards? I'm skeptical anybody would make a dashboard that relies on filtering to a specific channel with these settings, so this is probably fine, but this could theoretically break these, correct?

ash211:
Yes this could theoretically break these. As mentioned in the "Possible downsides?" section, I don't think this is considered part of the stable public API of dialogue, and that any actual breakage would be minimal

ash211:
Good to merge?

aldexis:
Yes, let's merge 👍

PR #2643, opened 2025-08-06
Pick up `backoffFactor` and `maxBackoff` parameters

fsamuel-bs:
## Before this PR

Pick up changes described in https://github.com/palantir/conjure-java-runtime/pull/3186.

Note that there's a time-window where if a client picks up 3186, but not this PR, the curve could be adjusted down to `[0-10ms]`, `[0-20ms]`, `[0-40ms]`, `[0-80ms]`. We find this unlikely, since all clients internally pick up `conjure-java-runtime` via the dialogue transitive dependency anyway.

## After this PR
==COMMIT_MSG==

==COMMIT_MSG==


dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -376,8 +388,9 @@ private long getBackoffNanoseconds() {
             if (failures == 0) {
                 return 0L;
             }
-            int upperBound = (int) Math.pow(2, failures - 1);
-            return Math.round(backoffSlotSize.toNanos() * jitter.getAsDouble() * upperBound);
+            int upperBound = (int) Math.pow(backoffFactor, failures - 1);
+            long backoffNanos = Math.round(backoffSlotSize.toNanos() * jitter.getAsDouble() * upperBound);
+            return Math.min(backoffNanos, maxBackoff.toNanos());
```
pkoenig10, line 393: This meaningfully complicates the backoff logic, making it harder to users to understand how the various configuration options affect behavior. I think it is important to keep the semantics of retry backoffs simple so it can be easily understood.

Why is not sufficient to modify the `backoffSlotSize` configuration?

fsamuel-bs: I'd like the first retry to be fired immediately, as I said in the description, in the interval `[0 - 10ms]`. With this in mind, I tried to match the existing backoff semantics with the other parameters.

Other solutions I thought of were:
- decrease slotSize + increase numRetries to achieve the same "total retry period" - this feels like requests could be retrying for longer (because we're tuning only delay between requests, the requests themselves can take a long time, making the total retry behavior be quite long)
- just reduce default slot size to 10 - I'm happy with this, but this would mean we'd exhaust retries after < 100ms, potentially not enough for downstream services to recover
- hardcode factor = 10 - we could do this, but feels like clients could want to reduce the factor too
- hardcode max = 2s - this feels arbitrary, and if we had a max we probably should also make it configurable

pkoenig10: This backoff logic is applied for both exceptions and explicit failure HTTP responses. I don't think we want retries to be fired immediately for at least some of the latter cases - explicit QoS responses, for example.

fsamuel-bs: I think 250ms is still too high, even for QoS. But agree that 10ms is too little. So we maybe we could:
1. move default 100ms~150ms
2. just retry immediately on first retry of failure? Then on further retries could move back to normal logic

I think this would be a good middle-ground.

pkoenig10: > just retry immediately on first retry of failure?

I'm not convinced this is a good idea.
- For the majority of failures, you don't want this.
- I don't think there is any guarantee that a retried request will go to a different node. We probably want some amount of backoff time for the scores to updated so that it becomes less likely we send requests to the unavailable node.

PR #2644, opened 2025-08-06
Retry immediately on first failure

fsamuel-bs:
## Before this PR

We run services on environments which are quite hostile - servers restarts every X hours, which causes `IOException` errors for clients. Our retry strategy does make these requests eventually succeed, but at the cost of a first retry, which, with default parameters takes from 0 to `250 ms` ([ref](https://github.com/palantir/conjure-java-runtime/blob/db9a10d5942aa79e6d68a204b2a1dd79db9c5ac1/client-config/src/main/java/com/palantir/conjure/java/client/config/ClientConfigurations.java#L54)).

This PR attempts to solve this common pattern (retries due to service restarts) by immediately retrying on the 1st failure.

From internal numbers for a service on a test stack:
- 50% of requests succeed on the first retry
- 50% of requests fail after exhausting all four retries
- the vast majority of exceptions are of type `HttpHostConnectException`, with the 2 most common being `NoHttpResponseException`

## After this PR
==COMMIT_MSG==
If a request failed, the call is immediately retried on a second host.
==COMMIT_MSG==

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
@@ -276,7 +276,7 @@ final class RetryingChannel implements EndpointChannel {
                 if (requestCanBeRetried() && shouldAttemptToRetry(clientSideThrowable)) {
                     callsiteStacktrace.ifPresent(clientSideThrowable::addSuppressed);
                     Meter retryReason = retryDueToThrowable.apply(clientSideThrowable);
-                    long backoffNanoseconds = getBackoffNanoseconds();
+                    long backoffNanoseconds = failures <= 1 ? 0 : getBackoffNanoseconds();
                     infoLogRetry(backoffNanoseconds, OptionalInt.empty(), clientSideThrowable);
                     return scheduleRetry(retryReason, backoffNanoseconds);
                 } else if (log.isDebugEnabled()) {
--- a/changelog/@unreleased/pr-2644.v2.yml
+++ b/changelog/@unreleased/pr-2644.v2.yml
@@ -1,0 +1,6 @@
+type: improvement
+improvement:
+  description: On request failures due to a retryable exception, the call is immediately
+    retried on a second host.
+  links:
+  - https://github.com/palantir/dialogue/pull/2644
--- a/dialogue-core/src/test/java/com/palantir/dialogue/core/RetryingChannelTest.java
+++ b/dialogue-core/src/test/java/com/palantir/dialogue/core/RetryingChannelTest.java
@@ -170,6 +170,31 @@ public class RetryingChannelTest {
     }
 
     @Test
+    public void retriesFirstFailureImmediately() throws Exception {
+        when(channel.execute(any())).thenReturn(FAILED).thenReturn(SUCCESS);
+
+        // One retry allows an initial request (not a retry) and a single retry.
+        long startTime = System.nanoTime();
+        Duration backoffSlotSize = Duration.ofSeconds(10);
+        EndpointChannel retryer = new RetryingChannel(
+                channel,
+                TestEndpoint.POST,
+                "my-channel",
+                1,
+                Duration.ZERO,
+                ClientConfiguration.ServerQoS.AUTOMATIC_RETRY,
+                ClientConfiguration.RetryOnTimeout.DISABLED);
+        ListenableFuture<Response> response = retryer.execute(REQUEST);
+        assertThat(response).isDone();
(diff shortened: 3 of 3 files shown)


pkoenig10:
> If a request failed, the call is immediately retried on a second host.

To be clear, there is no guarantee that the retried will requests will be routed to a different node.

But in practice we expect that a second attempt will typically succeed because the `IOException` from the first attempt [updates the balanced score tracker](https://github.com/palantir/dialogue/blob/1f21912856b571a71a3a896e61f2a65efec760d5/dialogue-core/src/main/java/com/palantir/dialogue/core/BalancedScoreTracker.java#L191) in a way that makes it much less likely that the original node will be selected for the second attempt.

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -276,7 +276,7 @@ private ListenableFuture<Response> handleThrowable(Throwable clientSideThrowable
                 if (requestCanBeRetried() && shouldAttemptToRetry(clientSideThrowable)) {
                     callsiteStacktrace.ifPresent(clientSideThrowable::addSuppressed);
                     Meter retryReason = retryDueToThrowable.apply(clientSideThrowable);
-                    long backoffNanoseconds = getBackoffNanoseconds();
+                    long backoffNanoseconds = failures <= 1 ? 0 : getBackoffNanoseconds();
```
pkoenig10, line 279: We should have a test for this, similar to the [existing test](https://github.com/palantir/dialogue/blob/1f21912856b571a71a3a896e61f2a65efec760d5/dialogue-core/src/test/java/com/palantir/dialogue/core/RetryingChannelTest.java#L347-L349) for the [similar logic that handles 308 responses](https://github.com/palantir/dialogue/blob/1f21912856b571a71a3a896e61f2a65efec760d5/dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java#L315).

fsamuel-bs: Added.

pkoenig10 (approved):

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
aldexis, line 279: @fsamuel-bs I know I arrive after the battle and this has already merged (I was ooto), but would it be possible to inline some comment/explanation as to 
- why we're doing this
- why we're only doing it here and not on the `handleHttpResponse` code-path
?

As is, even reading the PR description, it's not entirely obvious, so a quick comment might go a long way

pkoenig10: We should definitely update the PR description, but responding to the specific points:

It's relatively common for requests to fail when a node that is not live, often because it is restarting. In these cases we want to quickly try again on another node. The request didn't fail because the server didn't have capacity, etc. so there's no reason to wait before trying again.

We don't want to do this on the `handleHttpResponse` code path because we received a response from the server, so we know that node is at least live.

pkoenig10: > > If a request failed, the call is immediately retried on a second host.
>
> To be clear, there is no guarantee that the retried will requests will be routed to a different node.
>
> But in practice we expect that a second attempt will typically succeed because the IOException from the first attempt [updates the balanced score tracker](https://github.com/palantir/dialogue/blob/1f21912856b571a71a3a896e61f2a65efec760d5/dialogue-core/src/main/java/com/palantir/dialogue/core/BalancedScoreTracker.java#L191) in a way that makes it much less likely that the original node will be selected for the second attempt.

I do think it would be prudent for us to try an validation this implicit assumption being made here. If this assumption is not true, then this change may more more harmful than beneficial.

aldexis: The rationale makes sense to me, but I think it should be documented in the code, e.g.
```suggestion
                    // Retry immediately when we failed to reach the server, since it wasn't a capacity issue
                    // Note that this should retry on another node
                    long backoffNanoseconds = failures <= 1 ? 0 : getBackoffNanoseconds();
```
I'm not sure what validation we could make to verify we didn't reach the server though 🤔 (not familiar enough with the types of throwables we can get here)

fsamuel-bs: https://github.com/palantir/dialogue/pull/2657

PR #2647, opened 2025-08-07
Cache `EndpointChannel` across sticky clients

fsamuel-bs:
## Before this PR

Internally, we're seeing clients that use the sticky client incur in significant GC pressure because they re-create the client sticky per-session, and re-creating the client causes all metrics in all channels to be re-created.

Attempts to solve this, e.g. https://github.com/palantir/dialogue/pull/2026, had limited success.

This PR attempts a new approach: the stateful bits of the sticky client are in `StickyChannels2`:
- the `StickyRouter`, which adds `stickyTarget` to requests upon the first request succeeding + adds them to following requests
- `StickyEndpointChannels2EndpointFactorySupplier` (what a name) that created `queueOverrides`.

We attempt to cache the inner `EndpointChannel` blobs across clients, while preserving the per-client stateful messages.

See prior art in https://github.com/palantir/dialogue/pull/1334 + https://github.com/palantir/dialogue/pull/1336

## After this PR
==COMMIT_MSG==

==COMMIT_MSG==


dialogue-core/src/main/java/com/palantir/dialogue/core/StickyEndpointChannelCache.java
```diff
@@ -0,0 +1,37 @@
...
+import com.palantir.dialogue.Endpoint;
+import com.palantir.dialogue.EndpointChannel;
+import com.palantir.dialogue.HttpMethod;
+import java.util.function.Function;
+
+public class StickyEndpointChannelCache {
+    record StickyCacheKey(String serviceName, HttpMethod httpMethod, String endpointName) {}
+
+    private Cache<StickyCacheKey, EndpointChannel> cache =
+            Caffeine.newBuilder().maximumSize(1000).weakValues().build();
```
fsamuel-bs: Not fully happy with this, but `Endpoint` didn't have better uniqueness elements from a first glance + this works for a proof-of-concept.

pkoenig10: These values aren't guaranteed to be unique, so we have to use something else.

We can just use `Endpoint`. In practice, these should always be either be enum values (for interfaces generated by conjure or dialogue-annotations) or `FeignEndpoint` values, which are cached since https://github.com/palantir/conjure-java-runtime/pull/2973.

We used `Endpoint` as a cache key in a couple other places.
- https://github.com/palantir/dialogue/blob/a86b91e131f62b647f90b580aba9f6e94f97ca34/dialogue-core/src/main/java/com/palantir/dialogue/core/ChannelToEndpointChannel.java#L30
- https://github.com/palantir/dialogue/blob/a86b91e131f62b647f90b580aba9f6e94f97ca34/dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueChannel.java#L294

dialogue-core/src/main/java/com/palantir/dialogue/core/StickyEndpointChannelCache.java
```diff
@@ -0,0 +1,37 @@
...
+
+    public EndpointChannel getChannel(Endpoint endpoint, Function<Endpoint, EndpointChannel> channelFactory) {
+        return cache.get(
+                new StickyCacheKey(endpoint.serviceName(), endpoint.httpMethod(), endpoint.endpointName()),
+                _unused -> channelFactory.apply(endpoint));
+    }
```
pkoenig10: We should probably make this a loading cache, so we guarantee the `channelFactory` is always the same (it is, but that's not enforced by this API).

dialogue-core/src/main/java/com/palantir/dialogue/core/StickyChannelFactory.java
```diff
@@ -0,0 +1,23 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.dialogue.core;
+
+import com.palantir.dialogue.Channel;
+
+public interface StickyChannelFactory {
```
pkoenig10: nit: Can we give this a different name to disambiguate from `DialogueClients.StickyChannelFactory`? The `DialgoueClients` interfaces are already confusing enough...

dialogue-core/src/main/java/com/palantir/dialogue/core/StickyChannelFactory.java
```diff
@@ -0,0 +1,23 @@
...
+    Channel stickyChannel(StickyEndpointChannelCache cache);
```
pkoenig10: It's weird to pass the cache as a parameter - the `StickyChannelFactory` should own the cache.

```suggestion
    Channel stickyChannel();
```

dialogue-core/src/main/java/com/palantir/dialogue/core/StickyChannelFactory.java
fsamuel-bs: Agreed, but I think we can now remove `StickyChannelFactory` and leave only `StickyChannelFactory2` - `StickyChannelFactory` is not referenced by any code internally but this repo.

dialogue-core/src/main/java/com/palantir/dialogue/core/StickyEndpointChannelCache.java
fsamuel-bs: I'll address this comment + https://github.com/palantir/dialogue/pull/2647#discussion_r2261259251 here.

`channelFactory` it's potentially different. The issue is that the lifecycle of the cache is associated with the factory that can construct any kind of client (i.e. the same class constructs `FooBlocking.class` or `BarBlocking.class`). And the `channelFactory` that we're getting here is an endpointChannel factory (not the outer channel factory), which is constructed per `DialogueChannel` - so we don't have it at the time of cache creation.

Note that, for a given `endpoint` - it's likely that the `channelFactory` will have the same configs, so this is not an issue in practice. But it's hard to make the code apply this restriction.

fsamuel-bs: As a middle-ground I've moved the sticky endpoints cache to `ChannelCache`, which is constructed once per `DialogueFactory`, thus making it clearer that this is related to the factory lifecycle.

PR #2649, opened 2025-08-12
Add methods in `BodySerDe` to allow deserializing errors into exceptions

mpritham:
## Before this PR
Clients are not able to deserialize errors sent by servers into custom exception types. 

As part of the migration from using `Objects.toString` serialized error parameter values to using `JSON`, we'd like to enable clients to create custom exceptions that are a sub-type of `RemoteException`, maintaining backward compatibility. These new exceptions will also provide a method to access a record object holding the parameters deserialized into rich objects: clients will be able to access parameters via fields on the parameters record instead of relying on errorprone and hacky `Objects.toString` parsing.
 
## After this PR
==COMMIT_MSG==
Add methods in `BodySerDe` to allow deserializing errors into exceptions
==COMMIT_MSG==


dialogue-clients/src/main/java/com/palantir/dialogue/clients/ReloadingClientFactory.java
```diff
@@ -137,6 +148,111 @@ default boolean dnsNodeDiscovery() {
...
+
+        ErrorSerializationFormatSettingConjureRuntime(
+                ConjureRuntime delegate, ConjureErrorParameterFormat errorParameterSerializationFormat) {
+            this.delegate = delegate;
+            this.errorParameterSerializationFormat = errorParameterSerializationFormat;
+        }
+
+        @Override
+        public BodySerDe bodySerDe() {
+            return new ErrorParameterSerializationFormatSettingBodySerDe(
```
bjlaub: do we want to make this a private member variable?

mpritham: Makes sense. Updated.

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ConjureBodySerDe.java
```diff
@@ -104,6 +106,23 @@ final class ConjureBodySerDe implements BodySerDe {
...
+    private <T> Deserializer<T> buildExceptionCacheEntry(ExceptionDeserializerCacheKey<T> key) {
+        ExceptionDeserializerArgs<T> args = key.args();
+        return switch (key.type()) {
+            case STANDARD ->
+                EncodingDeserializerForEndpointRegistry.createExceptionThrowing(
+                        encodingsSortedByWeight, emptyContainerDeserializer, args);
+            case EMPTY_BODY ->
+                (Deserializer<T>) new EmptyBodyDeserializer(
+                        new ExceptionThrowingErrorDecoder<>(args.errorNameToExceptionTypeMarkers()));
+            case INPUT_STREAM, OPTIONAL_INPUT_STREAM ->
```
mpritham: The cache key contains the entire `ExceptionDeserializerArgs` value, so we shouldn't need to specify if the type is `InputStream` or `Optional<InputStream>`.

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ConjureBodySerDe.java
```diff
@@ -501,4 +582,49 @@ private static <S, T> Function<S, T> successTypeCreatorFactory(
...
     }
+
+    private enum CacheEntryType {
+        STANDARD,
+        EMPTY_BODY,
+        INPUT_STREAM,
+        OPTIONAL_INPUT_STREAM
+    }
+
+    private static final class ExceptionDeserializerCacheKey<T> {
```
mpritham: This can be a record.

.palantir/revapi.yml
```diff
@@ -325,3 +325,22 @@ acceptedBreaks:
       new: "method <T> com.palantir.dialogue.Deserializer<T> com.palantir.dialogue.BodySerDe::optionalInputStreamDeserializer(com.palantir.dialogue.DeserializerArgs<T>)"
       justification: "Adding new methods to create deserializers in support of endpoint\
         \ associated error deserialization"
+  "6.5.0":
+    com.palantir.dialogue:dialogue-target:
+    - code: "java.method.addedToInterface"
```
mpritham, line 330: These are new methods added to the interfaces. This will result in compile breaks for consumers that implement these interfaces, but there should not be subtle ABI breaks where clients silently (without a compile-time break) call a method that no longer exists or call the wrong method.

dialogue-target/src/main/java/com/palantir/dialogue/BodySerDe.java
```diff
@@ -35,11 +34,23 @@ public interface BodySerDe {
...
+
     /**
      * Returns a {@link Deserializer} that fails if a non-empty reponse body is presented and returns null otherwise.
      */
     Deserializer<Void> emptyBodyDeserializer();
 
+    /**
+     * Similar to {@link #emptyBodyDeserializer()} but allows for specifying error type to exception mapping.
+     */
+    Deserializer<Void> emptyBodyDeserializer(ExceptionDeserializerArgs<Void> exceptionDeserializerArgs);
```
mpritham, line 52: The existing `DeserializerArgs` argument provides a way to deserialize the response as a success type or one of many error types, all of which are sub-types of a base type. One alternative to adding a new argument type was to extend the `DeserializerArgs` type to accept exceptions - if there are no entries in the mapping from error name to error type, the deserialization logic could try searching for entries in the mapping from error name to exception type. 

I opt'd to create a separate `ExceptionDeserializerArgs` type for endpoints that can only return one type successfully (as opposed to one of many sub-types of a base type). This will be used for endpoints that do not have associated errors.

I like having them be separate because it's simpler to reason about what to expect when using the deserializer: either a return type `T` or one of the exceptions provided in the `ExceptionDeserializerArgs`.

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/AbstractErrorDecoder.java
```diff
@@ -0,0 +1,193 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.time.Duration;
+import java.util.List;
+import java.util.Optional;
+import org.jspecify.annotations.Nullable;
+
+abstract class AbstractErrorDecoder<T> {
```
mpritham, line 49: `ExceptionThrowingErrorDecoder` and `EndpointErrorDecoder` share a lot of the same logic. I moved the things in common into this package-private abstract class.

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ConjureBodySerDe.java
```diff
@@ -104,6 +107,28 @@ final class ConjureBodySerDe implements BodySerDe {
         this.serializers = Caffeine.from(cacheSpec)
                 .build(type -> new EncodingSerializerRegistry<>(defaultEncoding, TypeMarker.of(type)));
         this.deserializers = Caffeine.from(cacheSpec).build(type -> buildCacheEntry(TypeMarker.of(type)));
+        // Do not use weakKeys. Weak keys leads to equality checks to use == instead of equals().
```
mpritham, line 110: maybe 🌶️ ?

> Caffeine.weakKeys() stores keys using weak references. This allows entries to be garbage-collected if there are no other strong references to the keys. Since garbage collection depends only on identity equality, this causes the whole cache to use identity (==) equality to compare keys, instead of equals().

[source](https://github.com/ben-manes/caffeine/wiki/Eviction)

By not using weak keys, the entries will not be garbage collected when there are no strong references to the keys. I'm using weak values: entries will be garbage collected when there are strong references to the value.

The key is a `ExceptionDeserializerCacheKey`. This is private to this class: there can't be any class outside of this one holding strong references to the cache keys.

kunalrkak (dismissed):
👍🏼  for RC

kunalrkak (dismissed):
Approving for RC

.palantir/revapi.yml
aldexis: There _are_ cases where such a change may break, such as calling the new method on the interface, when it is in fact absent from the implementation.
It shouldn't break _just_ because an implementation exists that doesn't implement the method (as long as the method isn't called), as shown by [this test](https://github.com/palantir/gradle-transitive-abi-checker/blob/68ecf801f54f874190662b80ea9a3ba2ac11b084/abi-check-core/src/test/java/com/palantir/abi/checker/integration/MethodConflictCheckerIntegrationTest.java#L1275-L1318)

versions.props
```diff
@@ -3,9 +3,9 @@ com.google.code.findbugs:jsr305 = 3.0.2
 com.google.guava:guava = 33.4.8-jre
 com.google.testing.compile:compile-testing = 0.21.0
 com.palantir.common:streams = 2.4.0
-com.palantir.conjure.java.api:* = 2.62.0
-com.palantir.conjure.java.runtime:* = 8.23.0
 com.palantir.conjure.java:* = 8.52.0
+com.palantir.conjure.java.api:* = 2.63.0-rc1
```
aldexis, line 7: Let's not forget to bump this to non-rc release

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/DefaultClients.java
```diff
@@ -120,7 +120,8 @@ public <T> T block(ListenableFuture<T> future) {
 
             // TODO(jellis): can consider propagating other relevant exceptions (eg: HttpConnectTimeoutException)
             // see HttpClientImpl#send(HttpRequest req, BodyHandler<T> responseHandler)
-            if (cause instanceof RemoteException remoteException) {
+            if (cause instanceof RemoteException remoteException
+                    && cause.getClass().equals(RemoteException.class)) {
                 throw newRemoteException(remoteException);
             }
```
aldexis, line 126: Why add this check?

dialogue-target/src/main/java/com/palantir/dialogue/BodySerDe.java
aldexis: Considering we want to get rid of the dialogue result-type code-gen (for now), can we simply do that first and simplify this PR? It's super confusing to me to have both of these and I'd rather remove the old things first than have to mentally juggle with both (or just update the existing one)

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ConjureBodySerDe.java
```diff
@@ -501,4 +588,41 @@ private static <S, T> Function<S, T> successTypeCreatorFactory(
...
+
+        @Override
+        public int hashCode() {
+            return Objects.hash(args.returnType(), args.errorNameToExceptionTypeMarkers(), type);
+        }
+
+        @Override
+        public String toString() {
+            return "ExceptionDeserializerCacheKey{" + "args.returnType=" + args.returnType() + ", type=" + type + '}';
+        }
```
aldexis, line 626: Why do you need to override these? Why not make the return type and map as fields of the record and have a static builder that takes the args instead? Or just override the comparisons methods in `ExceptionDeserializerArgs` (better yet, make it an @Value.Immutable class, with custom helper builder methods?)

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ConjureBodySerDe.java
```diff
@@ -501,4 +588,41 @@ private static <S, T> Function<S, T> successTypeCreatorFactory(
...
+
+        @Override
+        public int hashCode() {
+            return Objects.hash(args.returnType(), args.errorNameToExceptionTypeMarkers(), type);
+        }
+
+        @Override
+        public String toString() {
+            return "ExceptionDeserializerCacheKey{" + "args.returnType=" + args.returnType() + ", type=" + type + '}';
```
aldexis, line 625: Should this maybe show the exception mapping too (since they're part of the comparisons)?

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ExceptionThrowingErrorDecoder.java
```diff
@@ -0,0 +1,135 @@
...
+import com.palantir.logsafe.logger.SafeLoggerFactory;
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.lang.reflect.Constructor;
+import java.lang.reflect.Type;
+import java.nio.charset.StandardCharsets;
+import java.util.Map;
+import java.util.Optional;
+
+final class ExceptionThrowingErrorDecoder<T> extends AbstractErrorDecoder<T> {
```
aldexis, line 38: Might be worth a javadoc to explain what "exception-throwing" means here (I imagine it's throwing rich exceptions, not just saying it can throw an exception while decoding?)

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ExceptionThrowingErrorDecoder.java
```diff
@@ -0,0 +1,135 @@
...
+                    throw createRemoteException(body, code);
+                }
+
+                // Attempt to deserialize the error using the deserializer for the specific exception type.
+                AbstractSerializableError<?> error;
+                try {
+                    error = deserializerExceptionPair.deserializer().deserialize(new ByteArrayInputStream(body));
+                } catch (Exception e) {
+                    // If we're unable to deserialize the error as JSON, throw a RemoteException.
+                    throw createRemoteException(body, code);
```
aldexis, line 103: You should probably log something here?

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ExceptionThrowingErrorDecoder.java
```diff
@@ -0,0 +1,135 @@
...
+                }
+
+                Type exceptionType = deserializerExceptionPair.exceptionType().getType();
+                @SuppressWarnings("unchecked")
+                Class<? extends RemoteException> exceptionClass =
+                        (Class<? extends RemoteException>) Class.forName(exceptionType.getTypeName());
+                Constructor<? extends RemoteException> exceptionConstructor =
+                        exceptionClass.getConstructor(error.getClass(), int.class);
+                throw exceptionConstructor.newInstance(error, code);
```
aldexis, line 112: Should we fall back to legacy remote exception if any of these lines fail? (e.g. can't find the exception class, or the appropriate constructor, or the call to create it failed)

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ExceptionThrowingErrorDecoder.java
```diff
@@ -0,0 +1,135 @@
...
+            body = toByteArray(response.body());
+        } catch (NullPointerException | IOException e) {
+            UnknownRemoteException exception = new UnknownRemoteException(code, "<unparseable>");
+            exception.initCause(e);
+            throw exception;
+        }
+
+        Optional<String> contentType = response.getFirstHeader(HttpHeaders.CONTENT_TYPE);
+        if (contentType.isPresent()
+                && Encodings.matchesContentType(JSON_ENCODING.getContentType(), contentType.get())) {
```
aldexis, line 85: nit: maybe flip this such that the main logic isn't nested?

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/DefaultClients.java
mpritham: This was needed I didn't want the custom generated exceptions to be wrapped in another `RemoteException` when `newRemoteException` was called. I will move this logic to a different PR.

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ConjureBodySerDe.java
mpritham: I didn't want to use an `Immutable` because I didn't want the generated code to also have to use Immutables. I think we'll require an API dependency on Immutables for anything that uses the generated Dialogue interfaces.

In #2666, I updated the `ExceptionDeserializerArgs` to have a custom `equals()` override.

mpritham:
This has been superseded by #2666.

PR #2652, opened 2025-08-15
Support `withConjureErrorParameterFormat` in ReloadingClientFactory

mpritham:
## Before this PR
We'd like to allow clients to specify the Conjure error parameter serialization format they expect from servers. This is done via a header. This header is sent in Conjure-Java generated code: the generated code needs to know the error parameter format the client has selected.

## After this PR
`BodySerDe` is the data structure that is provided to the generated code. This PR:
- adds support for the the newly added `withConjureErrorParameterSerializationFormat` method in `ReloadingClientFactory`
- plumbs the selected format into `BodySerDe`

==COMMIT_MSG==
Add `errorParameterDeserializationFormat` to `BodySerDe` and support `withConjureErrorParameterFormat` in `ReloadingClientFactory` 
==COMMIT_MSG==


versions.props
```diff
@@ -4,7 +4,7 @@ com.google.guava:guava = 33.4.8-jre
 com.google.testing.compile:compile-testing = 0.21.0
 com.palantir.common:streams = 2.4.0
 com.palantir.conjure.java.api:* = 2.62.0
-com.palantir.conjure.java.runtime:* = 8.23.0
+com.palantir.conjure.java.runtime:* = 8.24.0-rc1
```
mpritham: noting.

dialogue-clients/src/main/java/com/palantir/dialogue/clients/ErrorSerializationFormatSettingConjureRuntime.java
```diff
@@ -0,0 +1,119 @@
...
+import com.palantir.dialogue.Deserializer;
+import com.palantir.dialogue.DeserializerArgs;
+import com.palantir.dialogue.PlainSerDe;
+import com.palantir.dialogue.RequestBody;
+import com.palantir.dialogue.Serializer;
+import com.palantir.dialogue.TypeMarker;
+import java.io.InputStream;
+import java.util.Optional;
+
+final class ErrorSerializationFormatSettingConjureRuntime implements ConjureRuntime {
```
aldexis, line 33: Do we need a new one, as opposed to just having a new parameter with a default fallback (or optional) in the existing implementation?

mpritham: Synced on a call: `ConjureRuntime` and `BodySerDe` are interfaces so it's not possible for me to set a value on a field on `BodySerDe` directly.

dialogue-clients/src/main/java/com/palantir/dialogue/clients/ReloadingClientFactory.java
```diff
@@ -377,6 +378,12 @@ public DialogueClients.ReloadingFactory withHostEventsSink(HostEventsSink value)
         return new ReloadingClientFactory(params.withHostEventsSink(value), cache);
     }
 
+    @Override
+    public ReloadingFactory withConjureErrorParameterSerializationFormat(ConjureErrorParameterFormat format) {
```
aldexis: Is this supposed to be `withConjureErrorParameterFormat` based on https://github.com/palantir/conjure-java-runtime/pull/3192/files#diff-f3739f131aec4d56d379c2f53bd699c3eefde39ca6b51c12ab4253317e293b2dR93?

dialogue-clients/src/test/java/com/palantir/dialogue/clients/ReloadingClientFactoryTest.java
```diff
@@ -91,4 +99,42 @@ void live_reloading_wrapper_still_uses_the_EndpointChannelFactory_factory() {
...
+
+        // Set the error parameter format to JSON
+        ReloadingFactory modifiedFactory =
+                factory.withConjureErrorParameterSerializationFormat(ConjureErrorParameterFormat.JSON_FORMAT);
+
+        // Call get
+        RuntimeCapturingService service = modifiedFactory.get(RuntimeCapturingService.class, "foo");
+
+        assertThat(service.runtime().bodySerDe().errorParameterFormat())
+                .contains(ConjureErrorParameterFormat.JSON_FORMAT);
```
aldexis, line 122: I would also verify that we didn't change anything on the original factory and runtime, provided that is our intent
```suggestion
        // Call get
        RuntimeCapturingService service = modifiedFactory.get(RuntimeCapturingService.class, "foo");

        assertThat(service.runtime().bodySerDe().errorParameterFormat())
                .contains(ConjureErrorParameterFormat.JSON_FORMAT);
        
        // Verify we didn't impact the original factory
        RuntimeCapturingService originalFactoryService = factory.get(RuntimeCapturingService.class, "foo");
        
        assertThat(originalFactoryService.runtime().bodySerDe().errorParameterFormat()).isEmpty();
```

dialogue-clients/src/test/java/com/palantir/dialogue/clients/ReloadingClientFactoryTest.java
```diff
@@ -91,4 +99,42 @@ void live_reloading_wrapper_still_uses_the_EndpointChannelFactory_factory() {
...
+    void withConjureErrorParameterSerializationFormat_wraps_runtime_with_specified_format() {
+        ChannelCache mockCache = mock(ChannelCache.class);
+        ImmutableReloadingParams params = ImmutableReloadingParams.builder()
+                .scb(Refreshable.only(ServicesConfigBlock.builder().build()))
+                .build();
+        ReloadingClientFactory factory = new ReloadingClientFactory(params, mockCache);
+
+        // Get the original runtime
+        ConjureRuntime originalRuntime = params.runtime();
+        assertThat(originalRuntime.bodySerDe().errorParameterFormat()).isEqualTo(Optional.empty());
```
aldexis: nit: I think you can simply do
```suggestion
        assertThat(originalRuntime.bodySerDe().errorParameterFormat()).isEmpty();
```

aldexis (approved):

PR #2657, opened 2025-08-19
Add comment for #2644

fsamuel-bs:
## Before this PR
See https://github.com/palantir/dialogue/pull/2644#discussion_r2284660456

## After this PR
==COMMIT_MSG==

==COMMIT_MSG==

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
@@ -276,6 +276,10 @@ final class RetryingChannel implements EndpointChannel {
                 if (requestCanBeRetried() && shouldAttemptToRetry(clientSideThrowable)) {
                     callsiteStacktrace.ifPresent(clientSideThrowable::addSuppressed);
                     Meter retryReason = retryDueToThrowable.apply(clientSideThrowable);
+                    // Retry immediately when we failed to reach the server, since it wasn't a capacity issue
+                    // Note that this should retry on another node with DialogueNodeSelectionStrategy.BALANCED,
+                    // since this failure since this failure should be recorded in BalancedScoreTracker and
+                    // the inner-most NodeSelectionStrategyChannel should select a new node.
                     long backoffNanoseconds = failures <= 1 ? 0 : getBackoffNanoseconds();
                     infoLogRetry(backoffNanoseconds, OptionalInt.empty(), clientSideThrowable);
                     return scheduleRetry(retryReason, backoffNanoseconds);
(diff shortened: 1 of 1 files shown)


dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -276,6 +276,10 @@ private ListenableFuture<Response> handleThrowable(Throwable clientSideThrowable
                 if (requestCanBeRetried() && shouldAttemptToRetry(clientSideThrowable)) {
                     callsiteStacktrace.ifPresent(clientSideThrowable::addSuppressed);
                     Meter retryReason = retryDueToThrowable.apply(clientSideThrowable);
+                    // Retry immediately when we failed to reach the server, since it wasn't a capacity issue
+                    // Note that this should retry on another node with DialogueNodeSelectionStrategy.BALANCED,
+                    // since this failure since this failure should be recorded in BalancedScoreTracker and
+                    // the inner-most NodeSelectionStrategyChannel should select a new node.
```
aldexis: nit:
```suggestion
                    // Retry immediately when we failed to reach the server, since it wasn't a capacity issue
                    // Note that this should retry on another node with DialogueNodeSelectionStrategy.BALANCED,
                    // since this failure should be recorded in BalancedScoreTracker and
                    // the inner-most NodeSelectionStrategyChannel should select a new node.
```

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -276,6 +276,10 @@ private ListenableFuture<Response> handleThrowable(Throwable clientSideThrowable
                 if (requestCanBeRetried() && shouldAttemptToRetry(clientSideThrowable)) {
                     callsiteStacktrace.ifPresent(clientSideThrowable::addSuppressed);
                     Meter retryReason = retryDueToThrowable.apply(clientSideThrowable);
+                    // Retry immediately when we failed to reach the server, since it wasn't a capacity issue
```
pkoenig10, line 279: Just because we are in this code path does not mean we failed to reach the server.

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -276,6 +276,10 @@ private ListenableFuture<Response> handleThrowable(Throwable clientSideThrowable
...
+                    // Note that this should retry on another node with DialogueNodeSelectionStrategy.BALANCED,
```
pkoenig10, line 280: "should" feels like too strong of language here. There is no guarantee that we will retry on another node.

I would phrase this like "It is expected that most of the time the request will be retried on another node..."

PR #2661, opened 2025-08-20
Revert "Support endpoint error deserialization in `ConjureBodySerDe` (#2443)"

mpritham:
This reverts commit 88176ae93df91e474dc3ec01f427818624ff280f.

## Before this PR
Following [this PR in `conjure-java` merging](https://github.com/palantir/conjure-java/pull/2618), I'd like to remove support for deserializing responses as result types.

This PR reverts #2443.

## After this PR
==COMMIT_MSG==
Revert "Support endpoint error deserialization in `ConjureBodySerDe` (#2443)"
==COMMIT_MSG==


dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ErrorDecoder.java
```diff
@@ -45,32 +59,58 @@ public enum ErrorDecoder {
...
     private RuntimeException decodeInternal(Response response) {
         // TODO(rfink): What about HTTP/101 switching protocols?
         // TODO(rfink): What about HEAD requests?
 
-        Optional<RuntimeException> maybeQosException = ENDPOINT_ERROR_DECODER.checkCode(response);
-        if (maybeQosException.isPresent()) {
-            return maybeQosException.get();
+        int code = response.code();
+        switch (code) {
+            case 308 -> {
```
mpritham, line 82: I used the arrow-style switch case here because of a build failure from [the StatementSwitchToExpressionSwitch check](https://errorprone.info/bugpattern/StatementSwitchToExpressionSwitch)

kunalrkak (dismissed):
+1 for RC

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/JacksonEmptyContainerLoader.java
```diff
@@ -85,22 +84,9 @@ private Optional<Object> constructEmptyInstance(Type type, TypeMarker<?> origina
...
-        if (emptyRecordConstructor != null) {
-            try {
-                return Optional.of(emptyRecordConstructor.newInstance());
-            } catch (InstantiationException | IllegalAccessException | InvocationTargetException e) {
-                if (log.isDebugEnabled()) {
-                    log.debug("Empty record construction failed", SafeArg.of("type", type), e);
-                }
-                return Optional.empty();
-            }
-        }
```
aldexis, line 98: Should we maybe keep this piece of code around? (I'm not sure exactly in which case we needed it, but it seems potentially separately useful from the rest of the original PR itself)

aldexis (dismissed):

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/JacksonEmptyContainerLoader.java
mpritham: This is also related to results. This is used to ensure that we're able to deserialize empty response bodies into empty success records. This isn't needed because we use[ a separate deserializer (`emptyBodyDeserializer`) for responses with empty bodies.](https://github.com/palantir/dialogue/blob/develop/dialogue-target/src/main/java/com/palantir/dialogue/BodySerDe.java#L42)

aldexis (approved):

PR #2666, opened 2025-08-21
Add methods in BodySerDe to allow deserializing errors into exceptions

mpritham:
## Before this PR
Replaces #2649. 

Clients are not able to deserialize errors sent by servers into custom exception types. 

As part of the migration from using `Objects.toString` serialized error parameter values to using `JSON`, we'd like to enable clients to create custom exceptions that are a sub-type of `RemoteException`, maintaining backward compatibility. These new exceptions will also provide a method to access a record object holding the parameters deserialized into rich objects: clients will be able to access parameters via fields on the parameters record instead of relying on errorprone and hacky `Objects.toString` parsing.
 
## After this PR
==COMMIT_MSG==
Add methods in `BodySerDe` to allow deserializing errors into exceptions
==COMMIT_MSG==


dialogue-target/src/main/java/com/palantir/dialogue/ExceptionDeserializerArgs.java
```diff
@@ -0,0 +1,123 @@
...
+    @Override
+    public int hashCode() {
+        return Objects.hash(returnType, errorNameToExceptionTypeMarkers);
+    }
+
+    public static <T> Builder<T> builder() {
+        return new Builder<>();
+    }
+
+    public static final class Builder<T> {
```
aldexis, line 82: Might be worth adding a comment explaining this isn't an immutable because we don't want to force immutables on the code-generated libraries' classpath

dialogue-target/src/main/java/com/palantir/dialogue/ExceptionDeserializerArgs.java
```diff
@@ -0,0 +1,123 @@
...
+        private boolean buildInvoked = false;
+        private @Nullable TypeMarker<T> returnType;
+        private final Map<String, ErrorExceptionPair<?>> errorNameToExceptionTypeMarkers;
```
aldexis: nit: thoughts on making this an immutable map builder to avoid the final `ImmutableMap.copyOf` on build?

dialogue-target/src/main/java/com/palantir/dialogue/ExceptionDeserializerArgs.java
```diff
@@ -0,0 +1,123 @@
...
+            return false;
+        }
+        return Objects.equals(returnType, other.returnType)
+                && Objects.equals(errorNameToExceptionTypeMarkers, other.errorNameToExceptionTypeMarkers);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(returnType, errorNameToExceptionTypeMarkers);
+    }
```
aldexis, line 74: Maybe worth adding some tests to validate the hand-written equals and hashcode?

dialogue-target/src/main/java/com/palantir/dialogue/ExceptionDeserializerArgs.java
```diff
@@ -0,0 +1,123 @@
...
+    public TypeMarker<T> returnType() {
+        return returnType;
+    }
+
+    public ImmutableMap<String, ErrorExceptionPair<?>> errorNameToExceptionTypeMarkers() {
+        return errorNameToExceptionTypeMarkers;
+    }
+
+    public record ErrorExceptionPair<U extends AbstractSerializableError<?>>(
+            TypeMarker<U> errorType, TypeMarker<? extends RemoteException> exceptionType) {}
```
aldexis: nit: would you mind adding a quick example of what such a pair looks like? I can somewhat imagine it, but it requires some thinking to remember every time.

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ErrorDecoder.java
```diff
@@ -59,58 +45,32 @@ public enum ErrorDecoder {
...
-            case 503 -> {
-                return QosException.unavailable(qosReason(response));
-            }
+        Optional<RuntimeException> maybeException = EXCEPTION_DESERIALIZING_ERROR_DECODER.checkCode(response);
+        if (maybeException.isPresent()) {
+            return maybeException.get();
         }
 
+        int code = response.code();
         String body;
```
aldexis: Why isn't the entire method not just `EXCEPTION_DESERIALIZING_ERROR_DECODER.decode`? It seems to end up doing the same logic when we have an empty exception map? (sounds like this class could and possible should just entirely delegate to `EXCEPTION_DESERIALIZING_ERROR_DECODER`?)

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ExceptionDeserializingErrorDecoder.java
```diff
@@ -0,0 +1,299 @@
...
+    public RuntimeException decode(Response response) {
+        if (log.isDebugEnabled()) {
+            log.debug("Received an error response", diagnosticArgs(response));
+        }
+        try {
+            return decodeInternal(response);
+        } catch (Exception e) {
+            e.addSuppressed(diagnostic(response));
+            throw e;
+        }
```
aldexis: There are two differences here with the original ErrorDecoder:
- We catch thrown exceptions and add the diagnostics (I think that's good)
- We don't add the diagnostics anymore to the main exception - We might want to add that back?

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ExceptionDeserializingErrorDecoder.java
```diff
@@ -0,0 +1,299 @@
...
+        try {
+            NamedError namedError = NAMED_ERROR_DESERIALIZER.deserialize(new ByteArrayInputStream(body));
+            if (namedError == null) {
+                return null;
+            }
+            return namedError.errorName();
+        } catch (IOException | RuntimeException e) {
+            return null;
+        }
+    }
```
aldexis: nit: Move this down, next to `NamedError`

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ExceptionDeserializingErrorDecoder.java
```diff
@@ -0,0 +1,299 @@
...
+    private RuntimeException decodeInternal(Response response) {
+        // Check if the response code indicates an error, and if so, handle it accordingly.
+        Optional<RuntimeException> maybeException = checkCode(response);
+        if (maybeException.isPresent()) {
+            return maybeException.get();
+        }
+        int code = response.code();
+        byte[] body;
+        try {
+            body = toByteArray(response.body());
```
aldexis: style nit:
```suggestion
        byte[] body;
        try (InputStream bodyStream = response.body()) {
            body = bodyStream.readAllBytes();
```
avoids an extra method and doesn't really change readability

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ExceptionDeserializingErrorDecoder.java
```diff
@@ -0,0 +1,299 @@
...
+    private enum DialogueQosResponseDecodingAdapter implements QosResponseDecodingAdapter<Response> {
+        INSTANCE;
+
+        @Override
+        public Optional<String> getFirstHeader(Response response, String headerName) {
+            return response.getFirstHeader(headerName);
+        }
+    }
+
+    record NamedError(@JsonProperty("errorName") String errorName) {}
```
aldexis, line 276: Does this deserialize `errorName` as `null` if it's absent? Or does it throw? (if it throws, we could switch to `Optional<String>` to avoid having to go through exception handling for this edge case)

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ExceptionDeserializingErrorDecoder.java
```diff
@@ -0,0 +1,299 @@
...
+    }
+
+    private enum DialogueQosResponseDecodingAdapter implements QosResponseDecodingAdapter<Response> {
+        INSTANCE;
+
+        @Override
+        public Optional<String> getFirstHeader(Response response, String headerName) {
+            return response.getFirstHeader(headerName);
+        }
+    }
```
aldexis: style nit: Considering how you've moved this a couple of times (from `ErrorDecoder` to `EndpointErrorDecoder`, back to `ErrorDecoder` and now here), maybe it ought to be extracted as a package-private class?

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ExceptionDeserializingErrorDecoder.java
```diff
@@ -0,0 +1,299 @@
...
+            try {
+                AbstractSerializableError<?> error =
+                        deserializerExceptionPair.deserializer().deserialize(new ByteArrayInputStream(body));
+                Type exceptionType = deserializerExceptionPair.exceptionType().getType();
+                @SuppressWarnings("unchecked")
+                Class<? extends RemoteException> exceptionClass =
+                        (Class<? extends RemoteException>) Class.forName(exceptionType.getTypeName());
+                Constructor<? extends RemoteException> exceptionConstructor =
+                        exceptionClass.getConstructor(error.getClass(), int.class);
+                return exceptionConstructor.newInstance(error, code);
```
aldexis: This makes an assumption that a constructor taking `AbstractSerializableError<?>` (presumably of a specific type?) and `int` parameters exists on the rich exception in `ErrorExceptionPair`. You should probably document that in that class? Maybe even move the logic within `ErrorExceptionPair`? (encapsulated in a method `getExceptionFromSerializableError(U error, int code)`)

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ConjureBodySerDe.java
```diff
@@ -77,22 +80,55 @@ final class ConjureBodySerDe implements BodySerDe {
...
+                (Deserializer<T>) new EmptyBodyDeserializer(
+                        new ExceptionDeserializingErrorDecoder(args.errorNameToExceptionTypeMarkers()));
+            case INPUT_STREAM_OR_OPTIONAL_INPUT_STREAM ->
+                new EncodingDeserializerRegistry<>(
+                        ImmutableList.of(BinaryEncoding.INSTANCE),
+                        new ExceptionDeserializingErrorDecoder(args.errorNameToExceptionTypeMarkers()),
+                        emptyContainerDeserializer,
+                        args.returnType());
+        };
     }
```
aldexis: style nit: we could make this be
```suggestion
    @SuppressWarnings("unchecked")
    private <T> Deserializer<T> deserializerFor(
            DeserializerType type, TypeMarker<T> returnType, Map<String, ErrorExceptionPair<?>> errorNameToExceptionTypeMarkers) {
        ExceptionDeserializingErrorDecoder errorDecoder = new ExceptionDeserializingErrorDecoder(errorNameToExceptionTypeMarkers);
        return switch (key.type()) {
            case STANDARD ->
                new EncodingDeserializerRegistry<>(
                        encodingsSortedByWeight,
                        errorDecoder,
                        emptyContainerDeserializer,
                        returnType);
            case EMPTY_BODY ->
                (Deserializer<T>) new EmptyBodyDeserializer(errorDecoder);
            case INPUT_STREAM_OR_OPTIONAL_INPUT_STREAM ->
                new EncodingDeserializerRegistry<>(
                        ImmutableList.of(BinaryEncoding.INSTANCE),
                        errorDecoder,
                        emptyContainerDeserializer,
                        returnType);
        };
    }
```
(by also renaming `CacheEntryType` -> `DeserializerType`) then this method could be used to build each of `binaryInputStreamDeserializer`, `optionalBinaryInputStreamDeserializer`, `emptyBodyDeserializer`, `deserializers` as well as `exceptionDeserializers` (rather than have multiple places call into the constructors)

dialogue-serde/src/test/java/com/palantir/conjure/java/dialogue/serde/ExceptionDeserializationTestUtils.java
```diff
@@ -0,0 +1,170 @@
...
+import java.util.OptionalLong;
+import java.util.function.Function;
+
+final class ExceptionDeserializationTestUtils {
+    record ConjureError(
+            @JsonProperty("errorCode") String errorCode,
+            @JsonProperty("errorName") String errorName,
+            @JsonProperty("errorInstanceId") String errorInstanceId,
+            @JsonProperty("parameters") Map<String, Object> parameters,
+            @JsonIgnore Optional<Map<String, String>> legacyParameters) {
```
aldexis: `legacyParameters` isn't going to be a thing anymore, is it?

dialogue-serde/src/test/java/com/palantir/conjure/java/dialogue/serde/ExceptionDeserializingDecoderTest.java
```diff
@@ -0,0 +1,431 @@
...
+    private static final ObjectMapper MAPPER = ObjectMappers.newServerObjectMapper();
+    public static final ErrorType TEST_ERROR_TYPE =
+            ErrorType.create(ErrorType.Code.INVALID_ARGUMENT, "Conjure:TestError");
+
+    public static ServiceException testError(@Safe String stringArg, @Unsafe ComplexArg complexArg) {
+        return new ServiceException(
+                TEST_ERROR_TYPE, SafeArg.of("stringArg", stringArg), UnsafeArg.of("complexArg", complexArg));
+    }
+
+    @Generated("by conjure-java")
```
aldexis: Just out of curiosity, why do we need the Generated annotations here?

dialogue-serde/src/test/java/com/palantir/conjure/java/dialogue/serde/ExceptionDeserializingDecoderTest.java
```diff
@@ -0,0 +1,431 @@
...
+        // Then
+        assertThat(value).isEqualTo(expectedString);
+    }
+
+    /**
+     * This test validates that when a server sends the toString representation of parameters over the wire, the
+     * deserializer falls back to throwing a RemoteException and fails to create the TestErrorException.
+     */
+    @Test
+    public void testDeserializationFallsBackToRemoteExceptionWhenToStringParamsAreSent() throws IOException {
```
aldexis, line 76: Should we also check this falls back for
- Is a rich error, but not one we have a mapping for
- error name is not present (kind of an edge case, so maybe does not matter too much)

aldexis (commented):
Mostly have style nits/thoughts, which I defer to you on what you prefer. Higher priority comments are:
- Diagnostics not being added as suppressed to the normal remote exceptions anymore
- We should probably have ErrorDecoder entirely delegate to the exception one
- Adding a test around deserializing an exception we don't have the mapping for

dialogue-target/src/main/java/com/palantir/dialogue/ExceptionDeserializerArgs.java
mpritham: I think we can't avoid copying over the entries if we made this an `ImmutableMap`. In the method that adds a new `exception`, we'd have to create a copy of the map stored here, and add the new entry. This would result in more copies: e.g. 4 exceptions = 1 (second exception added) + 2 (third exception added)  + 3 (fourth exception added) copies.

One alternative is to _only_ have a setter for `exception` that takes a `Map`.

Let me know if you had something different in mind.

aldexis: I meant an immutable map _builder_ (`ImmutableMap.Builder`). This doesn't have to be an actual map at all, does it?

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ErrorDecoder.java
mpritham: You're right, it can. Updated

fwiw, I originally wanted the origin of the log to appear as `ErrorDecoder` (because maybe someone relies on searching for logs with that origin in some dashboard). It felt to drop a log with `ErrorDecoder` as the origin then use the decoding logic in the `ExceptionDeserializingErrorDecoder`.

But let's just delegate the entire call.

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ExceptionDeserializingErrorDecoder.java
mpritham: 🤦 thanks. This was a mistake. I should add a test to check that diagnostics are added so we catch things like this.

dialogue-target/src/main/java/com/palantir/dialogue/ExceptionDeserializerArgs.java
mpritham: > I meant an immutable map builder

Sorry yes, that makes sense. Will update.

> This doesn't have to be an actual map at all, does it?

I want to look up errors by their error names in the decoder.

aldexis: Definitely - I meant that the field in the `Builder` doesn't have to be an actual map, but can be a map builder :)

dialogue-serde/src/test/java/com/palantir/conjure/java/dialogue/serde/ExceptionDeserializingDecoderTest.java
mpritham: We don't. It was really just documentation to point out that those classes are expected to be generated.

dialogue-serde/src/test/java/com/palantir/conjure/java/dialogue/serde/ExceptionDeserializationTestUtils.java
mpritham: removed.

dialogue-serde/src/test/java/com/palantir/conjure/java/dialogue/serde/ExceptionDeserializingDecoderTest.java
mpritham: > Is a rich error, but not one we have a mapping for

Added a test for this.

> error name is not present (kind of an edge case, so maybe does not matter too much)

I can add a test to ensure this fails to throw a rich exception. We expect `errorName` to be present (per the Conjure wire spec)

dialogue-target/src/main/java/com/palantir/dialogue/ExceptionDeserializerArgs.java
mpritham:
```
ErrorExceptionPair[
    errorType=TypeMarker{type=class ExampleSerializableError}, 
    exceptionType=TypeMarker{type=class ExampleException}]
```

This stores the implementation of the `AbstractSerializableError` which should be what we deserialize the JSON response to, and the exception which should be constructed and thrown.

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ExceptionDeserializingErrorDecoder.java
mpritham: I added some documentation to `ExceptionDeserializerArgs`.

> Maybe even move the logic within ErrorExceptionPair? (encapsulated in a method getExceptionFromSerializableError(U error, int code))

Sorry, what did you have in mind?

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ExceptionDeserializingErrorDecoder.java
mpritham: hm, this is only used in the decoder class. I don't think anything else should be using this for now.

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ExceptionDeserializingErrorDecoder.java
aldexis: I meant adding a method in `ErrorExceptionPair` like
```
public record ErrorExceptionPair<U extends AbstractSerializableError<?>, V extends RemoteException>(
            TypeMarker<U> errorType, TypeMarker<V> exceptionType) {

    // TODO: add some docs / on the class explaining that there is an implicit contract
    //   that V has a public constructor taking U and an int
    //   (maybe also a concrete example of a pair)
    public V getExceptionFromSerializableError(U error, int code) {
        Class<? extends RemoteException> exceptionClass =
                        (Class<? extends RemoteException>) Class.forName(exceptionType.getTypeName());
        Constructor<? extends RemoteException> exceptionConstructor =
                        exceptionClass.getConstructor(error.getClass(), int.class);
        return exceptionConstructor.newInstance(error, code);
    }
}
```

.palantir/revapi.yml
```diff
@@ -347,3 +347,18 @@ acceptedBreaks:
...
+    - code: "java.method.addedToInterface"
+      new: "method com.palantir.dialogue.Deserializer<java.io.InputStream> com.palantir.dialogue.BodySerDe::inputStreamDeserializer(com.palantir.dialogue.ExceptionDeserializerArgs<java.io.InputStream>)"
+      justification: "Adding methods to support deserialization of errors as exceptions"
+    - code: "java.method.addedToInterface"
+      new: "method com.palantir.dialogue.Deserializer<java.lang.Void> com.palantir.dialogue.BodySerDe::emptyBodyDeserializer(com.palantir.dialogue.ExceptionDeserializerArgs<java.lang.Void>)"
+      justification: "Adding methods to support deserialization of errors as exceptions"
+    - code: "java.method.addedToInterface"
+      new: "method com.palantir.dialogue.Deserializer<java.util.Optional<java.io.InputStream>>\
+        \ com.palantir.dialogue.BodySerDe::optionalInputStreamDeserializer(com.palantir.dialogue.ExceptionDeserializerArgs<java.util.Optional<java.io.InputStream>>)"
+      justification: "Adding methods to support deserialization of errors as exceptions"
```
aldexis, line 364: As a reminder: we'll need to go and fix the internal consumers that have implemented their own `BodySerDe` (if any) to avoid blocking excavators

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ConjureBodySerDe.java
```diff
@@ -77,22 +82,58 @@ final class ConjureBodySerDe implements BodySerDe {
...
                 BinaryEncoding.MARKER);
         this.optionalBinaryInputStreamDeserializer = new EncodingDeserializerRegistry<>(
                 ImmutableList.of(BinaryEncoding.INSTANCE),
-                errorDecoder,
+                new ExceptionDeserializingErrorDecoder(Collections.emptyMap()),
                 emptyContainerDeserializer,
                 BinaryEncoding.OPTIONAL_MARKER);
-        this.emptyBodyDeserializer = new EmptyBodyDeserializer(errorDecoder);
+        this.emptyBodyDeserializer =
+                new EmptyBodyDeserializer(new ExceptionDeserializingErrorDecoder(Collections.emptyMap()));
```
aldexis: Let's also move these to use `deserializerFor`, to ensure we have a consistent logic (we can maybe make a convenience method that avoids us having to pass in an empty map as well)

Alternatively, let's at least not create three times the same decoder

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ExceptionDeserializingErrorDecoder.java
```diff
@@ -0,0 +1,293 @@
...
+    public RuntimeException decode(Response response) {
+        if (log.isDebugEnabled()) {
+            log.debug("Received an error response", diagnosticArgs(response));
+        }
+        RuntimeException result = decodeInternal(response);
+        result.addSuppressed(diagnostic(response));
+        return result;
+    }
+
+    Optional<RuntimeException> checkCode(Response response) {
```
aldexis: nit: this method can now be inlined

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ExceptionDeserializingErrorDecoder.java
```diff
@@ -0,0 +1,293 @@
...
+                        exceptionClass.getConstructor(error.getClass(), int.class);
+                return exceptionConstructor.newInstance(error, code);
+            } catch (Exception e) {
+                // If we're unable to deserialize the error as JSON, throw a RemoteException.
+                log.error("Failed to deserialize error response as exception", SafeArg.of("errorName", errorName), e);
+                return createRemoteException(body, code);
+            }
+        } catch (RemoteException remoteException) {
+            // rethrow the created remote exception
+            return remoteException;
```
aldexis: Out of curiosity, when do we expect this to happen? Nothing in the try block should _throw_ a RemoveException, should it?

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ExceptionDeserializingErrorDecoder.java
```diff
@@ -0,0 +1,293 @@
...
+                return createRemoteException(body, code);
+            }
+            DeserializerExceptionPair<?> deserializerExceptionPair = errorNameToExceptionDeserializerMap.get(errorName);
+            if (deserializerExceptionPair == null) {
+                return createRemoteException(body, code);
+            }
+            // Attempt to deserialize the error using the deserializer for the specific exception type.
+            try {
+                AbstractSerializableError<?> error =
+                        deserializerExceptionPair.deserializer().deserialize(new ByteArrayInputStream(body));
```
aldexis, line 175: I'm not sure what the generated code will look like, but assuming the exception constructor won't take `AbstractSerializableError`, but a specific subclass, shouldn't we cast this to the exact error type? Or is that actually going to be done at runtime regardless? (if so, this is probably worth a comment)

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ExceptionDeserializingErrorDecoder.java
aldexis: Ping on this question

dialogue-serde/src/test/java/com/palantir/conjure/java/dialogue/serde/ExceptionDeserializingDecoderTest.java
```diff
@@ -0,0 +1,386 @@
...
+        ExceptionDeserializerArgs<String> exceptionDeserializerArgs = ExceptionDeserializerArgs.<String>builder()
+                .returnType(new TypeMarker<>() {})
+                .build();
+        // When
+        try {
+            bodySerDe.deserializer(exceptionDeserializerArgs).deserialize(response);
+        } catch (RemoteException e) {
+            // Deserialization should have failed because `ComplexArg[foo=1, bar=bar]` is not a valid JSON
+            // representation of a ComplexArg. We should not throw a TestErrorException here, but rather fallback to
+            // throwing a RemoteException.
```
aldexis: nit: this comment seems incorrect for this test

dialogue-serde/src/test/java/com/palantir/conjure/java/dialogue/serde/ExceptionDeserializingDecoderTest.java
aldexis: Re the absence of `errorName`, there is one case where we don't currently fail (if `"message" is present as a field), because `SerializableError` should be able to deserialize the body.

It might be worth a test that just tries to deserialize such a hard-coded json string into a `RemoteException` rather than `UnknownRemoteException`?

aldexis (commented):
Lgtm, with two main comments/questions I'd like to discuss/address before approving:
- [Moving the serializable error -> exception conversion to the ErrorExceptionPair class](https://github.com/palantir/dialogue/pull/2666#discussion_r2301248334), since this is the thing that most closely associates the two and can describe the contract it expects the two to follow
- Verifying whether we need to cast the deserialized `AbstractSerializableError` before passing it to the constructor (I think maybe not, but this is worth verifying and adding a comment about)

dialogue-target/src/main/java/com/palantir/dialogue/ExceptionDeserializerArgs.java
```diff
@@ -0,0 +1,132 @@
...
+                            Preconditions.checkNotNull(errorType, "error type must be non-null"),
+                            Preconditions.checkNotNull(exceptionType, "exception type must be non-null")));
+            return this;
+        }
+
+        public ExceptionDeserializerArgs<T> build() {
+            checkNotBuilt();
+            Preconditions.checkNotNull(returnType, "base type must be set");
+            buildInvoked = true;
+            return new ExceptionDeserializerArgs<>(returnType, errorNameToExceptionTypeMarkers.buildKeepingLast());
```
aldexis: What do you think about
```suggestion
            return new ExceptionDeserializerArgs<>(returnType, errorNameToExceptionTypeMarkers.buildOrThrow());
```
? (to avoid silently dropping values, since it's never correct to register multiple bindings for a give error)

.palantir/revapi.yml
mpritham: There are none.

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ExceptionDeserializingErrorDecoder.java
mpritham: Removed.

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ExceptionDeserializingErrorDecoder.java
mpritham: Looks like this is done at runtime. I was able to get an e2e unit test in conjure-java.

dialogue-target/src/main/java/com/palantir/dialogue/ExceptionDeserializerArgs.java
mpritham: updated.

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ConjureBodySerDe.java
```diff
@@ -68,31 +74,62 @@ final class ConjureBodySerDe implements BodySerDe {
...
         this.encodingsSortedByWeight = sortByWeight(encodings);
         Preconditions.checkArgument(encodings.size() > 0, "At least one Encoding is required");
         this.defaultEncoding = encodings.get(0).encoding();
-        this.binaryInputStreamDeserializer = new EncodingDeserializerRegistry<>(
-                ImmutableList.of(BinaryEncoding.INSTANCE),
-                errorDecoder,
+        this.binaryInputStreamDeserializer = deserializerFor(
+                DeserializerType.STANDARD, emptyContainerDeserializer, BinaryEncoding.MARKER, Collections.emptyMap());
+        this.optionalBinaryInputStreamDeserializer = deserializerFor(
+                DeserializerType.STANDARD,
```
aldexis: I think these ought to be 
```suggestion
        this.binaryInputStreamDeserializer = deserializerFor(
                DeserializerType. INPUT_STREAM_OR_OPTIONAL_INPUT_STREAM, emptyContainerDeserializer, BinaryEncoding.MARKER, Collections.emptyMap());
        this.optionalBinaryInputStreamDeserializer = deserializerFor(
                DeserializerType. INPUT_STREAM_OR_OPTIONAL_INPUT_STREAM,
```
?

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ExceptionDeserializingErrorDecoder.java
```diff
@@ -0,0 +1,296 @@
...
+        @SuppressWarnings("unchecked")
+        public V getExceptionFromSerializableError(AbstractSerializableError<?> error, int code)
+                throws InvocationTargetException, InstantiationException, IllegalAccessException, NoSuchMethodException,
+                        ClassNotFoundException {
+            Class<? extends RemoteException> exceptionClass = (Class<? extends RemoteException>)
+                    Class.forName(exceptionType.getType().getTypeName());
+            Constructor<? extends RemoteException> exceptionConstructor =
+                    exceptionClass.getConstructor(error.getClass(), int.class);
+            return (V) exceptionConstructor.newInstance(error, code);
+        }
```
aldexis: Thoughts on moving this into `ErrorExceptionPair`? This allows for a somewhat stronger typing between the serializable error subclass and the remote exception type (especially as this is the record that effectively defines the contract between these two classes).

Alternatively, this _might_ just be
```suggestion
        @SuppressWarnings("unchecked")
        public V getExceptionFromSerializableError(U error, int code)
                throws InvocationTargetException, InstantiationException, IllegalAccessException, NoSuchMethodException,
                        ClassNotFoundException {
            Class<V> exceptionClass = (Class<V>)
                    Class.forName(exceptionType.getType().getTypeName());
            Constructor<V> exceptionConstructor =
                    exceptionClass.getConstructor(error.getClass(), int.class);
            return (V) exceptionConstructor.newInstance(error, code);
        }
```

aldexis (commented):
Lgtm - thanks for the iterations! Just one thing which I think is a bug and might be the cause of the failing tests

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/DefaultClients.java
```diff
@@ -151,6 +162,57 @@ private static RemoteException newRemoteException(RemoteException remoteExceptio
...
+                    continue;
+                }
+                // Found the interface, now we can proceed to find the constructor
+                foundSerializableErrorProviderInterface = true;
+                break;
+            }
+            if (!foundSerializableErrorProviderInterface) {
+                log.warn("RemoteException subclass does not implement SerializableErrorProvider");
+                return newRemoteException(remoteException);
+            }
```
aldexis: This seems to be simpler?
```suggestion
            Class<?> clazz = remoteException.getClass();
            if (!remoteException instanceOf SerializableErrorProvider) {
                log.warn("RemoteException subclass does not implement SerializableErrorProvider - this is unexpected and will be wrapped and rethrown as a legacy RemoteException", SafeArg.of("class", clazz.getName()));
                return newRemoteException(remoteException);
            }
```

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ConjureBodySerDe.java
mpritham: Fixed!

dialogue-serde/src/test/java/com/palantir/conjure/java/dialogue/serde/ExceptionDeserializingDecoderTest.java
mpritham: It looks like the `message` field is deprecated was used for `remoting2` servers [(src)](https://github.com/palantir/conjure-java-runtime-api/blob/464e37742f3094aaa061619a71a58351ec57c9ed/errors/src/main/java/com/palantir/conjure/java/api/errors/SerializableError.java#L104-L113). 

Nonetheless, I've added a test that checks that we are able to create a `RemoteException` when `message` is present but `errorName` is not, for completeness.

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ExceptionDeserializingErrorDecoder.java
mpritham: I made the method static in `ErrorExceptionPair`. I didn't want to keep track of the `errorName` to `ErrorExceptionPair` mapping.

dialogue-target/src/main/java/com/palantir/dialogue/ExceptionDeserializerArgs.java
```diff
@@ -128,6 +130,22 @@ public TypeMarker<T> returnType() {
...
+            U extends AbstractSerializableError<?>, V extends RemoteException & SerializableErrorProvider<?>>(
+            TypeMarker<U> errorType, TypeMarker<V> exceptionType) {
+        @SuppressWarnings("unchecked")
+        public static <U extends AbstractSerializableError<?>, V extends RemoteException & SerializableErrorProvider<?>>
+                V getExceptionFromSerializableError(U error, TypeMarker<V> exceptionType, int code)
+                        throws InvocationTargetException, InstantiationException, IllegalAccessException,
+                                NoSuchMethodException, ClassNotFoundException {
+            Class<V> exceptionClass =
+                    (Class<V>) Class.forName(exceptionType.getType().getTypeName());
+            Constructor<V> exceptionConstructor = exceptionClass.getConstructor(error.getClass(), int.class);
```
aldexis: vnit: We could add a comment explaining that 
```suggestion
            // RemoteExceptions constructible from a serializable error MUST have a constructor that takes such an error as well as the http error code
            Constructor<V> exceptionConstructor = exceptionClass.getConstructor(error.getClass(), int.class);
```
and maybe add a try-catch that throws a more explicit exception saying something along the lines of "could not find a valid constructor for RemoteException subclass to construct from provided serializable error" with both class names as argument, which will make it easier to debug if this ever happens

That would also let us encapsulate the `NoSuchMethodException` and `ClassNotFoundException` cases here 🤷‍♂️

aldexis (dismissed):

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/DefaultClients.java
```diff
@@ -151,6 +162,47 @@ private static RemoteException newRemoteException(RemoteException remoteExceptio
...
+                // Found a matching constructor, use it to create a new instance
+                AbstractSerializableError<?> abstractSerializableError =
+                        (AbstractSerializableError<?>) clazz.getMethod("error").invoke(remoteException);
+                return (RemoteException)
+                        constructor.newInstance(abstractSerializableError, remoteException.getStatus());
+            }
+            log.warn("Did not find a constructor on "
+                    + "RemoteException subclass with parameters (AbstractSerializableError, int)");
+        } catch (Exception e) {
+            log.warn("Failed to create new RemoteException with reflection, falling back to base type", e);
```
aldexis: nit: include the remote exception class as an argument here for both of these

dialogue-target/src/main/java/com/palantir/dialogue/ExceptionDeserializerArgs.java
mpritham: Added. Also I previously added a comment to the method in the builder.

dialogue-target/src/main/java/com/palantir/dialogue/ExceptionDeserializerArgs.java
```diff
@@ -140,12 +142,21 @@ public record ErrorExceptionPair<
...
+                Class<V> exceptionClass =
+                        (Class<V>) Class.forName(exceptionType.getType().getTypeName());
+                Constructor<V> exceptionConstructor = exceptionClass.getConstructor(error.getClass(), int.class);
+                return exceptionConstructor.newInstance(error, code);
+            } catch (ClassNotFoundException | NoSuchMethodException e) {
+                throw new SafeRuntimeException(
+                        "Failed to find constructor for exception which has an AbstractSerializableError as the first "
+                                + "argument and an integer http status code as the second argument",
+                        e,
+                        SafeArg.of("exceptionType", exceptionType));
```
aldexis: vnit: Let's also add `error.getClass().getSimpleName()` as an argument here?

mpritham: Added.

aldexis (approved):

PR #2670, opened 2025-08-26
Reduce Optional allocations from limiter permit acquire()

ash211:
## Before this PR
We are allocating enough instances of Optional in this method that it's showing up in JFRs on prod stacks (barely):

CleanShot 2025-08-25 at 23 26 12

## After this PR
==COMMIT_MSG==
Reduce Optional allocations from limiter permit acquire()
==COMMIT_MSG==

Switch from `Optional<Permit>` to `@Nullable Permit` in `CautiousIncreaseAggressiveDecreaseConcurrencyLimiter.acquire(LimitEnforcement)`

## Possible downsides?
- reduces code legibility and increases the risk of NullPointerExceptions if calling code doesn't have NullAway style protections running.  I'm not sure if these allocations are frequent enough that we should try to make improvements, or if we should just leave the code alone.
- risk of API break from changing the method signature, but I think this being package-private and internal API to dialogue means it won't break any consumers

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/StickyConcurrencyLimitedChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/StickyConcurrencyLimitedChannel.java
@@ -22,11 +22,13 @@ import com.palantir.dialogue.Endpoint;
 import com.palantir.dialogue.Request;
 import com.palantir.dialogue.Response;
 import com.palantir.dialogue.core.CautiousIncreaseAggressiveDecreaseConcurrencyLimiter.Behavior;
+import com.palantir.dialogue.core.CautiousIncreaseAggressiveDecreaseConcurrencyLimiter.Permit;
 import com.palantir.dialogue.futures.DialogueFutures;
 import com.palantir.logsafe.SafeArg;
 import com.palantir.logsafe.logger.SafeLogger;
 import com.palantir.logsafe.logger.SafeLoggerFactory;
 import java.util.Optional;
+import org.jspecify.annotations.Nullable;
 
 final class StickyConcurrencyLimitedChannel implements LimitedChannel {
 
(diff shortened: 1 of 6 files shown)


dialogue-core/src/main/java/com/palantir/dialogue/core/StickyConcurrencyLimitedChannel.java
```diff
@@ -68,7 +69,7 @@ public Optional<ListenableFuture<Response>> maybeExecute(
                 DialogueFutures.addDirectCallback(result.get(), permit);
                 return result;
             } else {
-                maybePermit.get().dropped();
+                permit.dropped();
```
schlosna:
```suggestion
                maybePermit.dropped();
```

dialogue-core/src/main/java/com/palantir/dialogue/core/StickyConcurrencyLimitedChannel.java
```diff
@@ -49,10 +51,9 @@ final class StickyConcurrencyLimitedChannel implements LimitedChannel {
     @Override
     public Optional<ListenableFuture<Response>> maybeExecute(
             Endpoint endpoint, Request request, LimitEnforcement limitEnforcement) {
-        Optional<CautiousIncreaseAggressiveDecreaseConcurrencyLimiter.Permit> maybePermit =
-                limiter.acquire(limitEnforcement);
-        if (maybePermit.isPresent()) {
-            CautiousIncreaseAggressiveDecreaseConcurrencyLimiter.Permit permit = maybePermit.get();
+        @Nullable Permit maybePermit = limiter.acquire(limitEnforcement);
+        if (maybePermit != null) {
+            CautiousIncreaseAggressiveDecreaseConcurrencyLimiter.Permit permit = maybePermit;
```
schlosna: do we need this local var? I'd suggest removing it completely or if you want to keep it move it inside the else block scope

```suggestion
```

schlosna (dismissed):
overall looks good and I'm always happy to remove some allocations from hot paths. one minor suggestion for consideration

dialogue-core/src/main/java/com/palantir/dialogue/core/StickyConcurrencyLimitedChannel.java
ash211: I thought it was a good way to remove the `@Nullable` annotation after we check for null, but happy to inline it -- pushed a commit with it inlined

schlosna (approved):

PR #2681, opened 2025-09-03
Configure deadline enforcement from dialogue clients

bjlaub:
Allows configuration of deadline enforcement on a per-client basis.
This allows for rolling out deadlines in a more selective and controlled way.


dialogue-core/src/main/java/com/palantir/dialogue/core/DeadlineAdvertisementChannel.java
```diff
@@ -30,22 +30,35 @@ final class DeadlineAdvertisementChannel implements Channel {
...
     private final Channel delegate;
     private final Duration readTimeout;
+    private final Deadlines.Enforcement enforcement;
 
-    DeadlineAdvertisementChannel(Channel delegate, Duration readTimeout) {
+    static DeadlineAdvertisementChannel create(Channel delegate, Duration readTimeout) {
+        return new DeadlineAdvertisementChannel(delegate, readTimeout, Deadlines.Enforcement.DEFER);
+    }
+
+    static DeadlineAdvertisementChannel create(Channel delegate, Duration readTimeout, boolean enforcementEnabled) {
```
kunalrkak: Did we structure this to pass enforcementEnabled as a boolean to abstract away the `Enforcement` class from non-deadlines related Dialogue classes? just wondering why we don't just keep a single constructor that takes the enforcement type so we can change the following 

```
if (cf.deadlineEnforcementEnabled().isPresent()) {
                    channel = DeadlineAdvertisementChannel.create(
                            channel,
                            cf.clientConf().readTimeout(),
                            cf.deadlineEnforcementEnabled().get());
                } else {
                    channel = DeadlineAdvertisementChannel.create(
                            channel, cf.clientConf().readTimeout());
                }
```
to something like this
```
channel = DeadlineAdvertisementChannel.create(
                            channel,
                            cf.clientConf().readTimeout(),
                            cf.deadlineEnforcementEnabled()
                                .map(e -> e ? Enforcement.ENFORCE : Enforcement:DISABLE)
                                .orElse(Enforcement.DEFER)
);
```

dialogue-core/src/main/java/com/palantir/dialogue/core/Config.java
```diff
@@ -79,6 +80,11 @@ default int maxQueueSize() {
 
     OptionalInt overrideSingleHostIndex();
 
+    @Value.Default
+    default Optional<Boolean> deadlineEnforcementEnabled() {
+        return Optional.empty();
+    }
```
kunalrkak, line 86: I think we said we wanted to add the config here to avoid modifying conjure-java-runtime, but doesn't this make it difficult for a witchcraft service to set this config when creating clients using `conjureClients()`?

kunalrkak: Adding a passthrough on witchcraft side to be able to set this easily - spoke offline and bluf of not adding to CJR is since deadlines conceptually is abstracted away from conjure at the moment

dialogue-core/src/main/java/com/palantir/dialogue/core/DeadlineAdvertisementChannel.java
kunalrkak: Think I'm going to leave this as is, both for the abstraction benefit, but also since using an if statement should be slightly more performant than an optional#map call here

dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueChannel.java
```diff
@@ -242,8 +250,15 @@ private static ImmutableList<LimitedChannel> createHostChannels(
                 channel = HostMetricsChannel.create(cf, channel, targetUri.uri());
                 channel =
                         new TraceEnrichingChannel(channel, DialogueTracing.tracingTags(cf, uriIndexForInstrumentation));
-                channel = new DeadlineAdvertisementChannel(
-                        channel, cf.clientConf().readTimeout());
+                if (cf.deadlineEnforcementEnabled().isPresent()) {
```
bjlaub: rethinking this a bit, i wonder if the `Optional<Boolean>` is really necessary here? we have a ctor overload for DeadlineAdvertisementChannel such that passing a boolean either enables/disables enforcement, and the alternative just defaults to DEFER, so perhaps the Optional isn't necessary to expose to callers?

kunalrkak: I think I'm missing something here, how would a caller select "defer" if we just used a boolean in the config? that would only allow for representing two states

bjlaub: doh, i think you're right - disregard.

dialogue-clients/src/main/java/com/palantir/dialogue/clients/DialogueClients.java
```diff
@@ -202,6 +202,9 @@ public interface ReloadingFactory
         /** Feature flag to opt into or out of the default dns-based node discovery behavior. */
         ReloadingFactory withDnsNodeDiscovery(boolean dnsNodeDiscovery);
 
+        /** Configures whether deadline enforcement should be enabled for outbound requests. */
```
bjlaub: nit: it might be worth clarifying that this explicitly **enables** or **disables** enforcement, but it is not required, e.g. the default factory will just use "defer".

dialogue-clients/src/main/java/com/palantir/dialogue/clients/AugmentClientConfig.java
```diff
@@ -55,6 +55,8 @@ default TaggedMetricRegistry taggedMetrics() {
 
     Optional<ClientConfiguration.RetryOnTimeout> retryOnTimeout();
 
+    Optional<Boolean> deadlineEnforcement();
```
bjlaub: i think i'd prefer for this to be added to `ReloadingClientFactory.ReloadingParams` instead. `AugmentClientConfig` seems primarily concerned with converting a cjr `ServiceConfiguration` -> `ClientConfiguration`, with optional overrides, whereas this feels purely like a dialogue implementation detail.

We'll have to also add `Optional<Boolean> deadlineEnforcement()` to `ChannelCache.ChannelCacheKey`.

dialogue-core/src/main/java/com/palantir/dialogue/core/Config.java
```diff
@@ -79,6 +80,11 @@ default int maxQueueSize() {
 
     OptionalInt overrideSingleHostIndex();
 
+    @Value.Default
+    default Enforcement deadlineEnforcement() {
```
bjlaub: thoughts on modeling this just as `Optional<Boolean>` instead? i think i'd maybe prefer to push the usage of the `Deadlines.Enforcement` enum as close to the actual usage as possible (e.g. into `DeadlineAdvertisementChannel`).

dialogue-core/src/main/java/com/palantir/dialogue/core/DeadlineAdvertisementChannel.java
```diff
@@ -30,22 +31,32 @@ final class DeadlineAdvertisementChannel implements Channel {
...
     private final Duration readTimeout;
+    private final Deadlines.Enforcement enforcement;
 
-    DeadlineAdvertisementChannel(Channel delegate, Duration readTimeout) {
+    static DeadlineAdvertisementChannel create(Channel delegate, Duration readTimeout) {
+        return new DeadlineAdvertisementChannel(delegate, readTimeout, Deadlines.Enforcement.DEFER);
+    }
+
+    static DeadlineAdvertisementChannel create(Channel delegate, Duration readTimeout, Enforcement enforcement) {
+        return new DeadlineAdvertisementChannel(delegate, readTimeout, enforcement);
```
bjlaub: see my [other comment](https://github.com/palantir/dialogue/pull/2681/files#r2414110093) re: modeling as optional.
```suggestion
    static DeadlineAdvertisementChannel create(Channel delegate, Duration readTimeout, Optional<Boolean> maybeEnforced) {
        Deadlines.Enforcement enforcement = maybeEnforced.map(v -> v ? Deadlines.Enforcement.ENFORCE : Deadlines.Enforcement.DISABLE).orElse(Deadlines.Enforcement.DEFER);
        return new DeadlineAdvertisementChannel(delegate, readTimeout, enforcement);
```

dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueChannel.java
```diff
@@ -141,6 +145,14 @@ public Builder overrideHostIndex(OptionalInt maybeUriIndex) {
             return this;
         }
 
+        public Builder deadlineEnforcement(Optional<Boolean> deadlineEnforcement) {
+            Enforcement enforcement = deadlineEnforcement
+                    .map(enforce -> enforce ? Enforcement.ENFORCE : Enforcement.DISABLE)
+                    .orElse(Enforcement.DEFER);
+            builder.deadlineEnforcement(enforcement);
```
bjlaub:
```suggestion
            builder.deadlineEnforcement(deadlineEnforcement);
```

dialogue-clients/src/main/java/com/palantir/dialogue/clients/ChannelCache.java
bjlaub, line 139: we'll need to add a line here to set the `deadlineEnforcement` on the `ChannelCacheKey`:
```suggestion
                .dnsNodeDiscovery(overrideHostIndex.isEmpty() && reloadingParams.dnsNodeDiscovery())
                .deadlineEnforcement(reloadingParams.deadlineEnforcement())
```

kunalrkak (approved):

PR #2710, opened 2025-09-29
Fix deadlines deprecation

aldexis:
## Before this PR
InlineMeInliner does not get auto-applied yet. Instead, it gets suppressed, as in https://github.com/palantir/dialogue/pull/2704, which means that the annotation added in https://github.com/palantir/deadlines-java/pull/188/files#diff-f8f90c9951b97fc3abe992cb927e3601222e7e3ce25efb471821039feab57432R179-R181 doesn't get auto-fixed

## After this PR
==COMMIT_MSG==
Fix deadlines deprecation
==COMMIT_MSG==


kunalrkak:
@aldexis this will be updated to the new method in #2681 when that PR merges shortly / it should be okay to just leave in the suppression until then. A larger question is should we keep the `InlineMeInliner` annotation in deadlines-java if it doesn't get auto-applied?

PR #2723, opened 2025-10-01
[Alternate] Configure deadline enforcement for dialogue clients

kunalrkak:
## Before this PR

Proposes an alternative implementation to #2681 

## After this PR

==COMMIT_MSG==
==COMMIT_MSG==


dialogue-clients/src/main/java/com/palantir/dialogue/clients/ReloadingClientFactory.java
```diff
@@ -592,4 +599,29 @@ public String toString() {
...
+            ConjureRuntime runtime = delegate.params.runtime();
+            return Reflection.callStaticFactoryMethod(serviceClass, channel, runtime);
+        }
+
+        private Channel getDeadlineEnforcedChannel(String serviceName) {
+            Channel internalDialogueChannel =
+                    delegate.getInternalDialogueChannel(serviceName).get();
+
+            Channel deadlineEnforcingChannel = (endpoint, request) -> {
+                request.attachments().put(RequestAttachmentKey.create(Boolean.class), enforce);
```
kunalrkak: I don't think this should be an issue but ideally the RequestAttachmentKey would be something more specific (would only be an issue if some other codepath tried to add a boolean request attachment)? We could potentially use `Enforcement.class` here or expose some wrapper class in dialogue-core for this.

dialogue-clients/src/test/java/com/palantir/dialogue/clients/DeadlineEnforcementFactoryTest.java
```diff
@@ -0,0 +1,116 @@
...
+
+        Channel channel1 = getChannelFromDeadlineEnforcementFactory(factoryWithEnforcement, "test-service");
+        Channel channel2 = getChannelFromDeadlineEnforcementFactory(factoryWithEnforcement, "test-service");
+
+        assertThat(channel1).isNotNull();
+        assertThat(channel2).isNotNull();
+        assertThat(getCacheSize(cache)).isEqualTo(1);
+    }
+
+    // Reflection hackery below
```
kunalrkak, line 80: Basically wanted to test things related to the underlying channels and confirm we were using a cached channel for new instances with the same config - let me know if this is overkill for tests though / if I'm overcomplicating

dialogue-clients/src/main/java/com/palantir/dialogue/clients/ReloadingClientFactory.java
bjlaub: > I don't think this should be an issue but ideally the RequestAttachmentKey would be something more specific (would only be an issue if some other codepath tried to add a boolean request attachment)?

I don't think this will work as written. The attachment key needs to be an instance of `RequestAttachmentKey`, and the same instance must be used when querying the attachments. The type parameter on `RequestAttachmentKey` is used to cast the resulting value, which internally is stored in a `Map<AttachmentKey<?>, Object>`.

I think the simplest way to achieve this is to make a `static final RequestAttachmentKey<Boolean>` instance somewhere that is accessible both to this code and within `DeadlineAdvertisementChannel`. Might be easier just to create a concrete `DeadlineEnforcingChannel` within dialogue-core to house that logic?

kunalrkak: I have no idea why I thought the wrapped type class was what the map would be keyed by 🤦🏼. Will fix to use a shared instance of the key in both places.

kunalrkak:
Ended up sticking with the original implementation

PR #2745, opened 2025-10-16
Clarify client response metric includes retries and queued time

aldexis:
## Before this PR
It is not clear, and possibly surprising, that the client.response timer metric also includes the retries as well as the queued time (i.e. it measures the time from the method call, not the individual time each network request took).

## After this PR
==COMMIT_MSG==
Clarify client response metric includes retries and queued time
==COMMIT_MSG==

This also adds tests to clearly verify this behavior


bjlaub (approved):
lgtm!

PR #2749, opened 2025-10-20
Client should not await the readTimeout twice when server is unresponsive

aldexis:
## Before this PR
Dialogue uses the configured `readTimeout` as the underlying Apache HTTP client's socket timeout.

Getting a `SocketTimeoutException` upon reading from the socket will end up going through [`HttpRequestExecutor`'s cleanup, which attempts to close the connection by calling `close(CloseMode.IMMEDIATE)`](https://github.com/apache/httpcomponents-core/blob/1ee9e0875fca9f5cb369a54f23ae8aa41b1f2327/httpcore5/src/main/java/org/apache/hc/core5/http/impl/io/HttpRequestExecutor.java#L221) (since https://github.com/apache/httpcomponents-core/pull/508).

However, when used with an SSL connection, especially when using http-client 5.3 (which might get fixed once https://github.com/palantir/dialogue/pull/2348 is picked back up and merged), because the [baseSocket here](https://github.com/apache/httpcomponents-core/blob/1ee9e0875fca9f5cb369a54f23ae8aa41b1f2327/httpcore5/src/main/java/org/apache/hc/core5/http/impl/io/BHttpConnectionBase.java#L235-L236) is [actually an `SSLSocket`](https://github.com/apache/httpcomponents-client/blob/rel/v5.3.1/httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/DefaultHttpClientConnectionOperator.java#L189-L190), this causes [the `Closer.closeQuietly(baseSocket);` call](https://github.com/apache/httpcomponents-core/blob/1ee9e0875fca9f5cb369a54f23ae8aa41b1f2327/httpcore5/src/main/java/org/apache/hc/core5/http/impl/io/BHttpConnectionBase.java#L243) to try to clean up the SSL state, which will potentially hang up to the configured `socketTimeout`.

## After this PR
==COMMIT_MSG==
Client should not await the readTimeout twice when server is unresponsive
==COMMIT_MSG==

## Possible downsides?
We should probably not put this in `InstrumentedManagedHttpClientConnection` and should instead put it in another intermediary layer, but this is simpler and hopefully becomes unnecessary once either https://github.com/apache/httpcomponents-core/pull/573 gets merged, or we bump the http-client dependency through something akin to https://github.com/palantir/dialogue/pull/2348 (at which point we can probably get rid of this)

### DIFF
--- a/dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/InstrumentedManagedHttpClientConnection.java
+++ b/dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/InstrumentedManagedHttpClientConnection.java
@@ -179,6 +179,19 @@ final class InstrumentedManagedHttpClientConnection implements ManagedHttpClient
 
     @Override
     public void close(CloseMode closeMode) {
+        if (closeMode == CloseMode.IMMEDIATE) {
+            // When the connection is being closed in IMMEDIATE mode, we want to avoid any blocking read calls.
+            // In particular, SSLSocketImpl's close method tries to read any remaining data from the input stream
+            //   to complete the SSL/TLS closure handshake. By setting a very short socket timeout,
+            //   we ensure that any read operations will time out almost immediately, allowing for a prompt closure
+            //   of the connection without unnecessary delays.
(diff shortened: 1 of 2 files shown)


aldexis:
Closing as unneeded now that https://github.com/palantir/dialogue/pull/2751 merged

PR #2751, opened 2025-10-21
Bump httpclient5 to 5.4.4

aldexis:
## Before this PR
https://github.com/palantir/dialogue/pull/2348 stalled out and we didn't bump httpclient since. We should get it bumped again, as a separate fix for https://github.com/palantir/dialogue/pull/2749.

Note that just bumping doesn't seem to be enough to make the test pass, and also fails some of our tests, so fixing the changes is needed.

## After this PR
==COMMIT_MSG==
Bump httpclient5 to 5.4.4
==COMMIT_MSG==

Note that this only bump to 5.4. I will look in further bumping to 5.5 afterwards.


dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientChannels.java
```diff
@@ -394,31 +394,6 @@ public static ClientBuilder clientBuilder() {
 
     public static final class ClientBuilder {
 
-        // Most of our servers use a keep-alive timeout of one minute, by using a slightly lower value on the
-        // client side we can avoid unnecessary retries due to race conditions when servers close idle connections
-        // as clients attempt to use them.
-        // Note that pooled idle connections use an infinite socket timeout so there is no reason to scale
-        // this value with configured timeouts.
-        private static final Timeout IDLE_CONNECTION_TIMEOUT = Timeout.ofSeconds(50);
```
aldexis, line 402: Moved to `InactivityValidationAwareConnectionKeepAliveStrategy`

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientChannels.java
```diff
@@ -394,31 +394,6 @@ public static ClientBuilder clientBuilder() {
...
-        // https://httpd.apache.org/docs/2.2/mod/core.html#keepalivetimeout
-        // nodejs http server: 5 seconds
-        // https://nodejs.org/api/http.html#http_server_keepalivetimeout
-        // nginx: 75 seconds (good)
-        // https://nginx.org/en/docs/http/ngx_http_core_module.html#keepalive_timeout
-        // dropwizard: 30 seconds (see idleTimeout in the linked docs)
-        // https://www.dropwizard.io/en/latest/manual/configuration.html#Connectors
-        // wc: 60 seconds (internal)
-        private static final TimeValue CONNECTION_INACTIVITY_CHECK = TimeValue.ofMilliseconds(
-                Integer.getInteger("dialogue.experimental.inactivity.check.threshold.millis", 4_000));
```
aldexis, line 420: Moved to `DialogueConnectionConfigResolver`

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientChannels.java
```diff
@@ -486,40 +460,34 @@ public CloseableClient build() {
 
             InetSocketAddress socksProxyAddress = getSocksProxyAddress(conf);
             SSLSocketFactory rawSocketFactory = conf.sslSocketFactory();
-            Supplier<Socket> simpleSocketCreator = socksProxyAddress == null
-                    ? () -> new Socket(Proxy.NO_PROXY)
-                    : () -> new Socket(new Proxy(Proxy.Type.SOCKS, socksProxyAddress));
+            // TODO(aldexis): replace this by just relying on the provided proxy set in setDefaultConfig below
+            DetachedSocketFactory simpleSocketCreator = socksProxyAddress == null
+                    ? proxy -> new Socket(proxy == null ? Proxy.NO_PROXY : proxy)
+                    : _proxy -> new Socket(new Proxy(Proxy.Type.SOCKS, socksProxyAddress));
```
aldexis, line 466: I believe this can be simplified to
```suggestion
            // TODO(aldexis): replace this by just relying on the provided proxy set in setDefaultConfig below
            DetachedSocketFactory simpleSocketCreator = proxy -> new Socket(proxy == null ? Proxy.NO_PROXY : proxy);
```
because of [this code](https://github.com/apache/httpcomponents-client/blob/f0afb8e6b73ed74c785747dea4dec43383c3d5db/httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/DefaultHttpClientConnectionOperator.java#L189-L190) which creates and passes the proxy configured below in the default config with `.setSocksProxyAddress(socksProxyAddress)`. I want to make sure we have a test for this first though. For now, this ensures we stay compatible with prior logic

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientChannels.java
```diff
@@ -486,40 +460,34 @@ public CloseableClient build() {
...
-                                            connectInstrumentation,
-                                            MetricRegistries.instrument(
-                                                    conf.taggedMetricRegistry(), rawSocketFactory, name),
-                                            TlsProtocols.get(),
-                                            supportedCipherSuites(
-                                                    CipherSuites.allCipherSuites(), rawSocketFactory, name),
-                                            new InstrumentedHostnameVerifier(
-                                                    new DefaultHostnameVerifier(), name, conf.taggedMetricRegistry()),
-                                            simpleSocketCreator))
-                            .build(),
```
aldexis, line 514: The `ConnectionSocketFactory` implementations are replaced by
- `TlsSocketStrategy` for the actual TLS socket layering
  - This is where most of the arguments to `InstrumentedSslConnectionSocketFactory` get moved
- `InstrumentedHttpClientConnectionOperator` which is where we've moved the socket connection instrumentation

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientChannels.java
```diff
@@ -486,40 +460,34 @@ public CloseableClient build() {
...
+                    operator,
                     PoolConcurrencyPolicy.LAX,
                     // Allow unnecessary connections to time out reducing system load.
                     PoolReusePolicy.LIFO,
                     // No maximum time to live
                     TimeValue.NEG_ONE_MILLISECOND,
-                    null,
-                    new InstrumentedDnsResolver(
-                            SystemDefaultDnsResolver.INSTANCE, dnsResolver, name, conf.taggedMetricRegistry()),
```
aldexis, line 522: This got moved to `InstrumentedHttpClientConnectionOperator`

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientChannels.java
```diff
@@ -528,10 +496,11 @@ public CloseableClient build() {
                     // By using a more specific timeout here, we bound the handshake in addition to the
                     // socket.connect call.
                     .setSoTimeout(handshakeTimeout)
-                    // Doesn't appear to do anything in this release
                     .setSocksProxyAddress(socksProxyAddress)
                     .build());
-            internalConnectionManager.setValidateAfterInactivity(CONNECTION_INACTIVITY_CHECK);
+            DialogueConnectionConfigResolver connectionConfigResolver =
+                    new DialogueConnectionConfigResolver(connectTimeout, socketTimeout);
+            internalConnectionManager.setConnectionConfigResolver(connectionConfigResolver);
```
aldexis, line 503: [`setValidateAfterInactivity` is deprecated and meant to be replaced by a `ConnectionConfigResolver`](https://github.com/apache/httpcomponents-client/blob/f0afb8e6b73ed74c785747dea4dec43383c3d5db/httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/PoolingHttpClientConnectionManager.java#L732-L734)

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientChannels.java
```diff
@@ -541,10 +510,8 @@ public CloseableClient build() {
                     new InstrumentedPoolingHttpClientConnectionManager(
                             internalConnectionManager, conf.taggedMetricRegistry(), name);
 
-            @SuppressWarnings("for-rollout:deprecation")
             HttpClientBuilder builder = HttpClients.custom()
                     .setDefaultRequestConfig(RequestConfig.custom()
-                            .setConnectTimeout(connectTimeout)
```
aldexis, line 547: This is deprecated and [moved to `DialogueConnectionConfigResolver`](https://github.com/palantir/dialogue/pull/2751/files#diff-760a100c02e4cdc96d025e2e3886066feaadb54effce609f20417d23c4f8bd33R61) (it even has now been removed in 5.5)

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/DialogueTlsSocketStrategy.java
```diff
@@ -0,0 +1,119 @@
...
+import org.apache.hc.core5.util.Timeout;
+
+/**
+ * {@link DialogueTlsSocketStrategy} is based closely on
+ * {@code org.apache.hc.client5.http.ssl.AbstractClientTlsStrategy}, except that it only requires a
+ * {@link SSLSocketFactory} rather than an {@link javax.net.ssl.SSLContext}.
+ * We only implement the minimal required {@link TlsSocketStrategy} interface rather than
+ * {@link org.apache.hc.core5.http.nio.ssl.TlsStrategy}, which isn't required by socket-based clients.
+ */
+final class DialogueTlsSocketStrategy implements TlsSocketStrategy {
```
aldexis, line 45: Unfortunately, the implementations in httpclient5 [require us to pass an `SSLContext`](https://github.com/apache/httpcomponents-client/blob/f0afb8e6b73ed74c785747dea4dec43383c3d5db/httpclient5/src/main/java/org/apache/hc/client5/http/ssl/AbstractClientTlsStrategy.java#L88-L89), because of `TlsStrategy`. We don't need `TlsStrategy` and we don't _have_ an `SSLContext` here, so we have no choice but to reimplement [this](https://github.com/apache/httpcomponents-client/blob/f0afb8e6b73ed74c785747dea4dec43383c3d5db/httpclient5/src/main/java/org/apache/hc/client5/http/ssl/AbstractClientTlsStrategy.java#L200-L316).

I'm not a big fan of it, since our implementation could deviate from the upstream's, but I don't think we have much of a choice here, unless we manage to convince the maintainers to provide us with a dedicated `TlsSocketStrategy` implementation.

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/DialogueTlsSocketStrategy.java
```diff
@@ -0,0 +1,119 @@
...
+        }
+    }
+
+    private void executeHandshake(SSLSocket upgradedSocket, String target, Object attachment) throws IOException {
+        SSLParameters sslParameters = upgradedSocket.getSSLParameters();
+        sslParameters.setProtocols(supportedProtocols);
+        sslParameters.setCipherSuites(supportedCipherSuites);
+
+        // If we want to enable the builtin hostname verification support:
+        // sslParameters.setEndpointIdentificationAlgorithm(URIScheme.HTTPS.id);
```
aldexis, line 83: Might want to get rid of this, which was copied from https://github.com/palantir/dialogue/pull/2348

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/InstrumentedHttpClientConnectionOperator.java
```diff
@@ -0,0 +1,65 @@
...
+import org.apache.hc.core5.http.protocol.HttpContext;
+
+final class InstrumentedHttpClientConnectionOperator extends DefaultHttpClientConnectionOperator {
+
+    private static final String CONNECT_BEGAN_ATTRIBUTE = "onBeforeSocketConnectNanoTime";
+
+    private final ConnectInstrumentation connectInstrumentation;
+
+    InstrumentedHttpClientConnectionOperator(
+            DetachedSocketFactory detachedSocketFactory,
```
aldexis, line 36: `HttpClientConnectionOperator` and `DetachedSocketFactory` are both marked as `@Internal` upstream, but the current API leaves us with little choice in the matter of subclassing it if we want to keep the socket connection metrics.

An alternative approach would be to get the maintainers to give us a way to pass a socket factory as well as a callback for before/after socket connection that would be set in [PoolingHttpClientConnectionManagerBuilder](https://github.com/apache/httpcomponents-client/blob/master/httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/PoolingHttpClientConnectionManagerBuilder.java) and used in [DefaultHttpClientConnectionOperator](https://github.com/apache/httpcomponents-client/blob/f0afb8e6b73ed74c785747dea4dec43383c3d5db/httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/DefaultHttpClientConnectionOperator.java#L189-L190)

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/InstrumentedManagedHttpClientConnection.java
```diff
@@ -49,6 +50,11 @@ public void bind(Socket socket) throws IOException {
         delegate.bind(socket);
     }
 
+    @Override
+    public void bind(SSLSocket sslSocket, Socket socket) throws IOException {
+        delegate.bind(sslSocket, socket);
+    }
```
aldexis, line 56: This is actually an important delegate we were missing

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/InstrumentedHttpClientConnectionOperator.java
```diff
@@ -0,0 +1,65 @@
...
+
+    @Override
+    protected void onAfterSocketConnect(HttpContext httpContext, HttpHost endpointHost) {
+        Object value = httpContext.getAttribute(CONNECT_BEGAN_ATTRIBUTE);
+        if (value instanceof Long) {
+            long duration = System.nanoTime() - (long) value;
+            connectInstrumentation.timer(true, httpContext).update(duration, TimeUnit.NANOSECONDS);
+        }
+        super.onAfterSocketConnect(httpContext, endpointHost);
+    }
```
aldexis, line 64: This is _slightly_ different compared to what we measured before. This specifically measures the time between [these two points](https://github.com/apache/httpcomponents-client/blob/f0afb8e6b73ed74c785747dea4dec43383c3d5db/httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/DefaultHttpClientConnectionOperator.java#L208-L221), which includes:
- `detachedSocketFactory.create(endpointHost.getSchemeName(), socksProxy);`
- configuring the socket
- binding the socket to the connection
- actually connecting the socket

whereas previously this would measure
- For plain sockets
  - The time spent [here](https://github.com/apache/httpcomponents-client/blob/f0afb8e6b73ed74c785747dea4dec43383c3d5db/httpclient5/src/main/java/org/apache/hc/client5/http/socket/PlainConnectionSocketFactory.java#L90-L112), which is essentially just the socket connection
- For SSL sockets
  - The time spent [here](https://github.com/apache/httpcomponents-client/blob/f0afb8e6b73ed74c785747dea4dec43383c3d5db/httpclient5/src/main/java/org/apache/hc/client5/http/ssl/SSLConnectionSocketFactory.java#L277-L294), which is also essentially just the socket connection

Another approach we could take would be to create an instrumented socket instead, by subclassing Socket and adding the connection metrics in there. I'm not sure that's much better though, and I don't imagine creating and configuring the socket object should take long comparatively to the socket connection itself?

dialogue-clients/src/test/java/com/palantir/dialogue/otherpackage/DialogueClientsIntegrationTest.java
```diff
@@ -128,6 +135,40 @@ void throws_if_user_agent_is_missing() {
                 .hasMessageContaining("userAgent must be specified");
     }
 
+    @Test
+    void timeout_is_respected() {
```
aldexis, line 139: This is a new test, taken from https://github.com/palantir/dialogue/pull/2749

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/InstrumentedHttpClientConnectionOperator.java
aldexis: They were also used directly in https://github.com/docker-java/docker-java/pull/2364/files#diff-6008de797a2952a755f012235847dcb00a03881e6e89c16a14c09f399fa9b2c3R148 by the httpcleitn5 maintainer themselves, so I'm guessing it's not a huge deal

mpritham (dismissed):
approving for an rc!

bjlaub (dismissed):
approving for rc

dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientChannels.java
```diff
@@ -486,40 +460,34 @@ public CloseableClient build() {
...
+                    new InstrumentedHostnameVerifier(new DefaultHostnameVerifier(), name, conf.taggedMetricRegistry()));
+
+            InstrumentedDnsResolver instrumentedDnsResolver = new InstrumentedDnsResolver(
+                    SystemDefaultDnsResolver.INSTANCE, dnsResolver, name, conf.taggedMetricRegistry());
+            HttpClientConnectionOperator operator = new InstrumentedHttpClientConnectionOperator(
+                    simpleSocketCreator, instrumentedDnsResolver, tlsStrategy, connectInstrumentation);
+
             PoolingHttpClientConnectionManager internalConnectionManager = new PoolingHttpClientConnectionManager(
-                    RegistryBuilder.<ConnectionSocketFactory>create()
-                            .register(
```
mpritham, line 498: I see that we've previously registered `URIScheme.HTTP.id` here. I think we only register `HTTPS` in the new code? Unless there is some other path I'm missing here (not yet super familiar with this side of Dialogue).

We've changed the constructor used to create a `PoolingHttpClientConnectionManager`. [This one (in the old code)](https://github.com/apache/httpcomponents-client/blob/06634831effa5c83536f4ed035fb13af8ffc78ea/httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/PoolingHttpClientConnectionManager.java#L145-L146) ultimately creates a `DefaultHttpClientConnectionOperator` with the passed in `ConnectionSocketFactory` [(src)](https://github.com/apache/httpcomponents-client/blob/06634831effa5c83536f4ed035fb13af8ffc78ea/httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/PoolingHttpClientConnectionManager.java#L209) which is converted to an instance of `Lookup<TlsSocketStrategy>` [here](https://github.com/apache/httpcomponents-client/blob/06634831effa5c83536f4ed035fb13af8ffc78ea/httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/DefaultHttpClientConnectionOperator.java#L102-L110). 

Our implementation of `InstrumentedHttpClientConnectionOperator`[ only has `HTTPS` registered.](https://github.com/palantir/dialogue/pull/2751/files#diff-50a3cd834c2c76cb38f0cbb334a4de143e0d6068691b42e26168ff83f9f404d6R44)

mpritham: So reading the code in `DefaultHttpClientConnectionOperator` I see that we only use the `tlsSocketStrategyLookup` to ["upgrade" the connection to a TLS connection](https://github.com/apache/httpcomponents-client/blob/06634831effa5c83536f4ed035fb13af8ffc78ea/httpclient5/src/main/java/org/apache/hc/client5/http/impl/io/DefaultHttpClientConnectionOperator.java#L225), so it doesn't make sense to add `HTTP` here.

aldexis: Yup, exactly :)

mpritham (approved):

PR #2760, opened 2025-10-23
Bump httpclient5 to 5.5

aldexis:
## Before this PR
Similar to https://github.com/palantir/dialogue/pull/2751 - trying to bump one step further to 5.5 so we can reenable auto-upgrades.

## After this PR
==COMMIT_MSG==
Bump httpclient5 to 5.5
==COMMIT_MSG==


dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientChannels.java
```diff
@@ -492,12 +499,9 @@ public CloseableClient build() {
                             ManagedHttpClientConnectionFactory.INSTANCE, conf.taggedMetricRegistry(), name));
             internalConnectionManager.setDefaultSocketConfig(SocketConfig.custom()
                     .setSoKeepAlive(true)
-                    // The default socket configuration socket timeout only applies prior to request execution.
-                    // By using a more specific timeout here, we bound the handshake in addition to the
-                    // socket.connect call.
-                    .setSoTimeout(handshakeTimeout)
```
aldexis, line 498: https://github.com/apache/httpcomponents-client/commit/8094bac428ad61622530d216ee763724985f612f#diff-c8d113aa21d5fe2678c4c36f731cda6ed01d172d7e47e4d9fc527a5d4a921077R230-R233 changed the behavior of the underlying connection operator to actually rely on the configured handshake timeout, and default it to the connect timeout if not explicitly set, which is not the exact same semantics we want https://github.com/palantir/dialogue/blob/d51be1bf65fa17e4b164e62a41b3735f559bc7f5/dialogue-apache-hc5-client/src/main/java/com/palantir/dialogue/hc5/ApacheHttpClientChannels.java#L655-L672

aldexis: This also happens to make it be used here https://github.com/palantir/dialogue/pull/2751/files#diff-3a67d92618f4d3ef40e83b23222c66adab3933341207ca48a4958a086f2db03dR87-R92 so we could probably push this change down into https://github.com/palantir/dialogue/pull/2751, though this was moved off the upstream version in https://github.com/apache/httpcomponents-client/commit/8094bac428ad61622530d216ee763724985f612f#diff-a5e74a0fa48dd91a1e1bac65ff84722564da9e8b73fa835551c811a845aa2f2dL241-L244

mpritham (dismissed):
approving for an RC!

mpritham (approved):
Thank you! Sorry for the delay here.

PR #2810, opened 2025-12-05
Pass `ConjureRuntime` object to ser/de classes when using dialogue-annotation-processor

mpritham:
## Before this PR
Currently, the Dialogue client interfaces generated when using dialogue-annotation-processor do not use the `ConjureRuntime` object provided in the constructor of the interface, but use the `DefaultConjureRuntime` object, when constructor serializers and deserializers. Some consumers may want to create a custom `ConjureRuntime` implementation with its own ser/de logic to be used when using clients generated with dialogue-annotation-processor.

## After this PR

==COMMIT_MSG==
Pass `ConjureRuntime` object to ser/de classes when using dialogue-annotation-processor
==COMMIT_MSG==


dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/data/ParameterType.java
```diff
@@ -26,7 +27,7 @@ interface Cases<R> {
 
         R rawBody();
 
-        R body(TypeName serializerFactory, String serializerFieldName);
+        R body(TypeName serializerFactory, TypeMirror serializerType, String serializerFieldName);
```
pkoenig10: This design requires a conditional logic in `ServiceInterfaceGenerator` that couples it pretty tightly with `ParamTypesResolver`.

Instead, I'd suggest we keep this concern contained within `ParamTypesResolver` by using `CodeBlock` values as the interface between these components.

This is what we do in conjure-undertow-annotations, for example:

https://github.com/palantir/conjure-java/blob/777ab3e4aa5570719b3ca7876d6b052f28070462/conjure-undertow-processor/src/main/java/com/palantir/conjure/java/undertow/processor/data/ParameterType.java#L36

https://github.com/palantir/conjure-java/blob/777ab3e4aa5570719b3ca7876d6b052f28070462/conjure-undertow-processor/src/main/java/com/palantir/conjure/java/undertow/processor/data/ParamTypesResolver.java#L373-L397

dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/data/ReturnType.java
```diff
@@ -18,15 +18,19 @@
 
 import com.palantir.javapoet.TypeName;
 import java.util.Optional;
+import javax.lang.model.type.TypeMirror;
 import org.immutables.value.Value;
 
 @Value.Immutable
 @StagedBuilder
+// TODO(pm): is this actually used outside dialogue or just public to use in a diff package?
```
pkoenig10: The latter. Nothing should be depending on `dialogue-annotation-processor` (semantically, it's not a library). If they are, we shouldn't care about breaking them.

dialogue-annotations-example/src/test/java/com/palantir/myservice/example/MyServiceIntegrationTest.java
```diff
@@ -117,7 +117,7 @@ public void testGreet() {
         undertowHandler = exchange -> {
             exchange.assertMethod(HttpMethod.POST);
             exchange.assertPath("/greet");
-            exchange.assertAccept().isEqualTo("application/json");
+            exchange.assertAccept().contains("application/json");
```
bjlaub, line 122: discussed offline, but one side-effect of this change is that clients _may_ (in most cases) now send something like
```
Accept: application/x-jackson-smile, application/json, application/cbor
```
as opposed to just
```
Accept: application/json
```
as this is the [default behavior in DefaultConjureRuntime](https://github.com/palantir/dialogue/blob/develop/dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/DefaultConjureRuntime.java#L47) unless encodings are set explicitly in the builder, which they [previously were](https://github.com/palantir/dialogue/blob/develop/dialogue-annotations/src/main/java/com/palantir/dialogue/annotations/BodySerDeSingleton.java#L26).

This feels strictly better I think, but it is different.

dialogue-annotations-example/src/test/java/com/palantir/myservice/example/MyServiceIntegrationTest.java
bjlaub, line 108: perhaps we should write a test that also constructs a client using `.withRuntime(<other ConjureRuntime>)`?

mpritham: Added a test using a custom runtime with a custom deserializer.

dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/data/Types.java
```diff
@@ -0,0 +1,57 @@
...
+    public static Optional<ExecutableElement> findConstructorWithConjureRuntimeParameter(TypeMirror typeMirror) {
+        DeclaredType declaredType = MoreTypes.asDeclared(typeMirror);
+        TypeElement typeElement = (TypeElement) declaredType.asElement();
+        return typeElement.getEnclosedElements().stream()
+                .filter(element -> element.getKind() == ElementKind.CONSTRUCTOR)
+                .map(ExecutableElement.class::cast)
+                .filter(element -> !element.getModifiers().contains(Modifier.PRIVATE)
+                        && element.getParameters().size() == 1
+                        && TypeName.get(element.getParameters().get(0).asType())
+                                .equals(ClassName.get(ConjureRuntime.class)))
```
pkoenig10, line 56: We should also check thrown types here. [Example](https://github.com/palantir/conjure-java/blob/c1ffbd1cb5f2fe59d070f14158db8566af163c22/conjure-undertow-processor/src/main/java/com/palantir/conjure/java/undertow/processor/data/ParamTypesResolver.java#L363-L365).

mpritham: I see: if a constructor were to throw a checked exception, we'd need to ensure the generated code handles it.

I think this is confusing for devs who may write a constructor taking a `ConjureRuntime` argument and throw a checked exception in the constructor - their constructor would just not be used, and the generated code would either fallback to the no-arg constructor, or if that doesn't exist, fail to compile.

I wonder if we should just wrap the call to custom ser/de class constructors with some logic that catches all `Exception`s, and throw a `SafeRuntimeException` with the exception thrown by the constructor.

pkoenig10: I'm not a fan of that. If something is going to fail I think it's better to force the author of the deserializer to acknowledge that can fail and handle the error in a way they determine is appropriate.

Also, I think there is a lot value in being consistent with the behavior/capabilities of conjure-undertow-annotations.

pkoenig10: In practice, I don't think this will be an issue. Constructors throwing checking exceptions are pretty uncommon. Especially given that none of the `ConjureRuntime` methods throw checked exceptions.

I can't recall a case where this requirement was a problem for users of conjure-undertow-annotations.

mpritham: Sounds good. I also think it'd be quite rare.

dialogue-annotations-example/src/main/java/com/palantir/myservice/example/CustomRuntimeAwareDeserializer.java
```diff
@@ -0,0 +1,53 @@
...
+        runtime = null;
+    }
+
+    @Override
+    public MySerializableType deserialize(Response response) {
+        if (runtime == null) {
+            throw new SafeIllegalStateException(
+                    "Wrong constructor for CustomRuntimeAwareDeserializer used in generated code.");
+        }
+        Deserializer<SafeLong> deserializer = runtime.bodySerDe().deserializer(new TypeMarker<>() {});
```
pkoenig10: This is a somewhat bad example because it create a new deserializer on every request to `deserialize`. We wouldn't want people following this pattern when writing their own deserializers.

This should not implement `StdDeserializer`. Can we change this?

dialogue-annotations-example/src/main/java/com/palantir/myservice/example/CustomRuntimeAwareDeserializer.java
```diff
@@ -0,0 +1,63 @@
...
+import com.palantir.dialogue.Response;
+import com.palantir.dialogue.TypeMarker;
+import com.palantir.dialogue.annotations.DeserializerFactory;
+import com.palantir.logsafe.Preconditions;
+import com.palantir.logsafe.exceptions.SafeIllegalStateException;
+import java.util.Optional;
+import javax.annotation.Nullable;
+
+public final class CustomRuntimeAwareDeserializer
+        implements DeserializerFactory<MySerializableType>, Deserializer<MySerializableType> {
```
pkoenig10: I think this still isn't quite right - we shouldn't be instantiating deserializers in the `DeserializerFactory` constructor.

We also should update the test cases in `dialogue-annotations-processor`, since those example are still using hard-coded object mappers instead of using the provided `ConjureRuntime` (this PR adds a constructor with an unused `ConjureRuntime` parameter, which is a bit strange).

I think we want something like this:

https://github.com/palantir/dialogue/commit/4a6e6a1298daefe9a45e0d114a1747148c29c0e6

It would be good to maintain the consistency between the examples in  `dialogue-annotations-processor` tests and `dialogue-annotations-example`.

dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/data/ReturnTypesResolver.java
```diff
@@ -84,15 +90,16 @@ private Optional<TypeName> orDefaultDeserializerFactory(
...
             if (!hasMustBeClosed) {
                 context.reportError("When returning InputStream, remember to add @MustBeClosed annotation", element);
                 return Optional.empty();
             }
-            return Optional.of(context.getTypeName(InputStreamDeserializer.class));
+            return Optional.of(
+                    CodeBlock.of("new $T(runtime.bodySerDe())", context.getTypeName(InputStreamDeserializer.class)));
         }
-        return Optional.of(context.getTypeName(Json.class));
+        return Optional.of(CodeBlock.of("new $T(runtime.bodySerDe())", context.getTypeName(Json.class)));
```
pkoenig10: nit: Can we add constructor overloads that take `ConjureRuntime` so that instantiation of the built-in deserializer factories is consistent with custom ones?


```suggestion
        return Optional.of(CodeBlock.of("new $T(runtime)", context.getTypeName(Json.class)));
```

dialogue-annotations-processor/src/main/java/com/palantir/dialogue/annotations/processor/data/ParamTypesResolver.java
```diff
@@ -115,25 +116,32 @@ public Optional<ParameterType> getParameterType(EndpointName endpointName, Varia
...
                     && context.isSameTypes(serializer, Json.class)) {
                 context.reportError(
                         "BinaryRequestBody is not supported by default, prefer the more expressive RequestBody type",
                         variableElement);
                 return Optional.empty();
             }
-
+            CodeBlock serializerFactory;
+            if (customSerializer.isEmpty()) {
+                serializerFactory = CodeBlock.of("new $T(runtime.bodySerDe())", serializerType);
```
pkoenig10: nit: Same here, can we add a `Json(ConjureRuntime)` constructor overload?


```suggestion
                serializerFactory = CodeBlock.of("new $T(runtime)", serializerType);
```

dialogue-annotations-processor/src/test/java/com/palantir/myservice/service/CustomInputStreamDeserializer.java
```diff
@@ -27,6 +28,10 @@ public CustomInputStreamDeserializer() {
         super("application/octet-stream");
     }
 
+    public CustomInputStreamDeserializer(ConjureRuntime _runtime) {
+        super("application/octet-stream");
+    }
```
pkoenig10: We should not add these duplicative constructors with unused parameters.


```suggestion
```

dialogue-annotations-example/src/main/java/com/palantir/myservice/example/CustomRuntimeAwareDeserializer.java
pkoenig10: For instance, this allows you to do more interesting things, like deserialize into a generic type while using the provided `ConjureRuntime` deserializer to deserialize the response into an instance of the generic type parameter.

https://github.com/palantir/dialogue/commit/abd0c7c26ee39c088a19fdfbcda7c3efe746e4f7

mpritham: I see. Thanks for putting that together. I've cherry picked these commits.

pkoenig10 (commented):
Approving for RC

pkoenig10 (dismissed):
Approving for RC

PR #2820, opened 2025-12-16
[DNM] hack: add Do-Not-Retry hint when retries are exhausted

bjlaub:
## After this PR

==COMMIT_MSG==
==COMMIT_MSG==


dialogue-core/src/main/java/com/palantir/dialogue/core/RetriesExhaustedChannel.java
```diff
@@ -0,0 +1,58 @@
...
+import com.google.common.util.concurrent.ListenableFuture;
+import com.palantir.dialogue.DialogueRetries;
+import com.palantir.dialogue.DialogueRetries.DialogueRetriesResponseDecodingAdapter;
+import com.palantir.dialogue.EndpointChannel;
+import com.palantir.dialogue.Request;
+import com.palantir.dialogue.Response;
+import com.palantir.dialogue.futures.DialogueFutures;
+import java.util.Optional;
+
+// a channel that detects if upstream dialogue retries were exhausted, and fails fast if so
```
bjlaub: oops, this comment is wrong - don't fail fast here, instead set a response attachment that causes RetryingChannel to short-circuit

dialogue-target/src/main/java/com/palantir/dialogue/DialogueRetries.java
```diff
@@ -0,0 +1,64 @@
...
+package com.palantir.dialogue;
+
+import java.util.Optional;
+
+public final class DialogueRetries {
+    static final ResponseAttachmentKey<Boolean> RETRIES_EXHAUSTED_TOKEN = ResponseAttachmentKey.create(Boolean.class);
+
+    private static final String DIALOGUE_RETRIES_EXHAUSTED_HEADER = "Dialogue-Retries-Exhausted";
+
+    public static boolean isRetriesExhausted(Response response) {
```
bjlaub, line 26: we should separate the response attachment logic from the header parsing logic into two separate classes, the former being package-private

PR #2823, opened 2025-12-18
Improve ExceptionDeserializingErrorDecoder logs and remove unused constructors

aldexis:
## Before this PR
The `ExceptionDeserializingErrorDecoder` decode logic logs an error when using a the client has registered rich exception types, but didn't receive json errors from the server.

This is desirable _only_ if the client actually _requested_ the server to send json errors, rather than just being able to deserialize them, which isn't always the case, and can lead to confusion.

## After this PR
==COMMIT_MSG==
Improve ExceptionDeserializingErrorDecoder logs and remove unused constructors
==COMMIT_MSG==

This
- Adds a boolean to `ExceptionDeserializingErrorDecoder` to indicate whether we should expect to receive back json errors for rich exceptions we know about (i.e. if we have sent the json error format header)
  - Updates the error log to log as debug if we didn't send that header
- Adds some additional debug logs to facilitate debugging this workflow in the future if needed
- Refactors out by splitting into two methods to reduce cyclomatic complexity

Another approach _could_ be to directly deserialize as a `RemoteException` if we didn't send the header, but this is better since it lets us support the server sending json errors regardless of the request headers.

### DIFF
--- a/dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ConjureBodySerDe.java
+++ b/dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ConjureBodySerDe.java
@@ -107,7 +107,9 @@ final class ConjureBodySerDe implements BodySerDe {
                         key.type(),
                         emptyContainerDeserializer,
                         key.args().returnType(),
-                        new ExceptionDeserializingErrorDecoder(key.args().errorNameToExceptionTypeMarkers())));
+                        new ExceptionDeserializingErrorDecoder(
+                                key.args().errorNameToExceptionTypeMarkers(),
+                                errorParameterFormat().isPresent())));
     }
 
     @SuppressWarnings("unchecked")
--- a/dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ConjureBodySerDe.java
+++ b/dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ConjureBodySerDe.java
@@ -23,6 +23,7 @@ import com.google.common.base.Suppliers;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Lists;
 import com.google.common.net.HttpHeaders;
+import com.palantir.conjure.java.api.errors.ConjureErrorParameterFormat;
 import com.palantir.dialogue.BinaryRequestBody;
 import com.palantir.dialogue.BodySerDe;
 import com.palantir.dialogue.Deserializer;
@@ -108,8 +109,7 @@ final class ConjureBodySerDe implements BodySerDe {
                         emptyContainerDeserializer,
                         key.args().returnType(),
                         new ExceptionDeserializingErrorDecoder(
-                                key.args().errorNameToExceptionTypeMarkers(),
-                                errorParameterFormat().isPresent())));
+                                key.args().errorNameToExceptionTypeMarkers(), expectJsonErrors())));
     }
 
     @SuppressWarnings("unchecked")
@@ -234,6 +234,11 @@ final class ConjureBodySerDe implements BodySerDe {
                 }
             }
         };
+    }
+
+    private boolean expectJsonErrors() {
+        return errorParameterFormat().isPresent()
+                && errorParameterFormat().get().equals(ConjureErrorParameterFormat.JSON_FORMAT);
     }
 
     private static final class EncodingSerializerRegistry<T> implements Serializer<T> {
--- a/dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ExceptionDeserializingErrorDecoder.java
+++ b/dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ExceptionDeserializingErrorDecoder.java
@@ -172,7 +172,7 @@ final class ExceptionDeserializingErrorDecoder {
         }
     }
 
-    private RuntimeException jsonExceptionFromBody(byte[] body, int code) throws IOException {
+    private RemoteException jsonExceptionFromBody(byte[] body, int code) throws IOException {
         String errorName = extractErrorName(body);
         if (errorName == null) {
             // Per the Conjure spec, errorName is required. We fall back to creating a RemoteException to handle
--- a/dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ExceptionDeserializingErrorDecoder.java
+++ b/dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ExceptionDeserializingErrorDecoder.java
@@ -143,7 +143,7 @@ final class ExceptionDeserializingErrorDecoder {
         try (InputStream bodyStream = response.body()) {
             body = bodyStream.readAllBytes();
         } catch (NullPointerException | IOException e) {
-            UnknownRemoteException exception = new UnknownRemoteException(code, "<unreadable>");
+            UnknownRemoteException exception = new UnknownRemoteException(code, "<unparseable>");
             exception.initCause(e);
             return exception;
         }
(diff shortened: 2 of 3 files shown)


dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ExceptionDeserializingErrorDecoder.java
```diff
@@ -145,41 +143,27 @@ private RuntimeException decodeInternal(Response response) {
...
         }
 
+        if (log.isDebugEnabled()) {
+            log.debug("Read error response body", UnsafeArg.of("body", new String(body, StandardCharsets.UTF_8)));
+        }
+
         Optional<String> contentType = response.getFirstHeader(HttpHeaders.CONTENT_TYPE);
         if (contentType.isEmpty() || !Encodings.matchesContentType(JSON_ENCODING.getContentType(), contentType.get())) {
-            return new UnknownRemoteException(code, new String(body, StandardCharsets.UTF_8));
+            String stringBody = new String(body, StandardCharsets.UTF_8);
```
mpritham: Can we create this variable on line 150 just above the debug log so we don't have to allocate another `String` object here?

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ExceptionDeserializingErrorDecoder.java
```diff
@@ -145,41 +143,27 @@ private RuntimeException decodeInternal(Response response) {
...
+            log.debug(
```
mpritham: Should we wrap this in a `if (log.isDebugEnabled())`?

aldexis: It's not necessary, since we're always going to create the stringBody (to put it in the `UnknownRemoteException`)

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ExceptionDeserializingErrorDecoder.java
aldexis: I would rather not, because in most cases we don't need to allocate the string. We'd need to allocate it if we are in this case _or_ if we are in debug, and twice if we are in this case _and_ in debug, but that last case should be pretty rare, so we should not allocate it for all the other cases

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ExceptionDeserializingErrorDecoder.java
```diff
@@ -188,6 +172,46 @@ private RuntimeException decodeInternal(Response response) {
         }
     }
 
+    private RuntimeException jsonExceptionFromBody(byte[] body, int code) throws IOException {
```
mpritham: nit:
```suggestion
    private RemoteException jsonExceptionFromBody(byte[] body, int code) throws IOException {
```

Just in case we mess with this implementation later, would be good to have the type-checker warn us if we try to throw anything other than a `RemoteException`.

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ExceptionDeserializingErrorDecoder.java
aldexis: Or did you mean creating a potentially null string above, and only creating it as needed? (seems somewhat overkill for what is a rare case)

mpritham: It was for the last case, and I also assumed that this branch would be hit most often (i.e. that we expect a content-type to be present and set to JSON). But maybe that's not true 🤷. Happy to keep as-is.

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ConjureBodySerDe.java
```diff
@@ -107,7 +107,9 @@ final class ConjureBodySerDe implements BodySerDe {
                         key.type(),
                         emptyContainerDeserializer,
                         key.args().returnType(),
-                        new ExceptionDeserializingErrorDecoder(key.args().errorNameToExceptionTypeMarkers())));
+                        new ExceptionDeserializingErrorDecoder(
+                                key.args().errorNameToExceptionTypeMarkers(),
+                                errorParameterFormat().isPresent())));
```
mpritham: This will work but I think we should check that this is exactly `ConjureErrorParameterFormat.JSON_FORMAT`, in case we end up adding some sort of way for clients to explicitly request the `toString` representation (or some other formatting?).

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ExceptionDeserializingErrorDecoder.java
```diff
@@ -74,22 +74,20 @@ final class ExceptionDeserializingErrorDecoder {
             JSON_ENCODING.deserializer(new TypeMarker<>() {});
     private static final Deserializer<SerializableError> SERIALIZABLE_ERROR_DESERIALIZER =
             JSON_ENCODING.deserializer(new TypeMarker<>() {});
+
     private final Map<String, DeserializerExceptionPair<?, ?>> errorNameToExceptionDeserializerMap;
+    private final boolean expectRichErrors;
```
mpritham:
```suggestion
    private final boolean expectJsonErrors;
```

v nit: because I suggested we check that the format is explicitly `JSON_FORMAT` when populating this variable in `ConjureBodySerde`

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ExceptionDeserializingErrorDecoder.java
mpritham: Sorry not sure I follow: I'm suggesting wrapping this debug log in a block that checks that debug logging is enabled as not to spend time constructing the arguments (e.g. `UnsafeArg.of`).

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ConjureBodySerDe.java
aldexis: Ah yes, that's a good point

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ExceptionDeserializingErrorDecoder.java
aldexis: Oh I thought it was about the string construction, as I assumed the UnsafeArg.of was a small enough allocation that it didn't matter. I'm happy to wrap in this case.

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ExceptionDeserializingErrorDecoder.java
aldexis: I've added some logic so we only convert to string maximum once. It's not super pretty, but using a memoized supplier would require using some classes that add locking (unless we build our own) and I didn't want to add perf impacts here.

This branch is also only hit if the content type is _not_ json, so I assumed it would be rather rare (though I'm unsure whether where smile goes in this? I guess we don't use it for errors). As far as I can tell, most of the time, we should not need to put the error into a string at all

mpritham (approved):
Thanks!

PR #2857, opened 2026-01-29
Add metrics to track exhausted retries and retry counts

mpritham:
## Before this PR
We'd like to gain better visibility into Dialogue's retrying behavior. Initially, we'd like to gather some data on which services consistently exhaust their retries, and the distributions of retries needed for requests to succeed.

## After this PR

==COMMIT_MSG==
Add metrics to track exhausted retries and retry counts
==COMMIT_MSG==

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
@@ -287,13 +287,16 @@ final class RetryingChannel implements EndpointChannel {
             DialogueFutures.addDirectCallback(result, new FutureCallback<>() {
                 @Override
                 public void onSuccess(@Nullable Response response) {
-                    if (Responses.isSuccess(response)) {
-                        retryCountSuccessHistogram.get().update(failures);
-                    } else {
-                        retryCountFailureHistogram.get().update(failures);
-                    }
                     if (failures > 0) {
                         span.complete(RetryingCallbackTranslator.INSTANCE, RetryingCallback.this);
+                        if (Responses.isSuccess(response)) {
+                            retryCountSuccessHistogram.get().update(failures);
+                        } else {
+                            retryCountFailureHistogram.get().update(failures);
+                        }
+                    } else if (requestCanBeRetried()) {
+                        // Only update the success metric if the request was retry-able.
+                        retryCountSuccessHistogram.get().update(failures);
                     }
                 }
 
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
@@ -157,6 +157,29 @@ final class RetryingChannel implements EndpointChannel {
                 () -> ThreadLocalRandom.current().nextDouble());
     }
 
+    @VisibleForTesting
+    RetryingChannel(
+            EndpointChannel delegate,
+            Endpoint endpoint,
+            String channelName,
+            TaggedMetricRegistry taggedMetricRegistry,
+            int maxRetries,
+            Duration backoffSlotSize,
+            ClientConfiguration.ServerQoS serverQoS,
+            ClientConfiguration.RetryOnTimeout retryOnTimeout) {
+        this(
+                delegate,
+                endpoint,
+                channelName,
+                taggedMetricRegistry,
+                maxRetries,
+                backoffSlotSize,
+                serverQoS,
+                retryOnTimeout,
+                sharedScheduler.get(),
+                () -> ThreadLocalRandom.current().nextDouble());
+    }
+
     private RetryingChannel(
             EndpointChannel delegate,
             Endpoint endpoint,
@@ -263,10 +286,14 @@ final class RetryingChannel implements EndpointChannel {
             ListenableFuture<Response> result = wrap(delegate.execute(request));
             DialogueFutures.addDirectCallback(result, new FutureCallback<>() {
                 @Override
-                public void onSuccess(@Nullable Response _response) {
+                public void onSuccess(@Nullable Response response) {
                     if (failures > 0) {
                         span.complete(RetryingCallbackTranslator.INSTANCE, RetryingCallback.this);
-                        retryCountSuccessHistogram.get().update(failures);
+                        if (Responses.isSuccess(response)) {
+                            retryCountSuccessHistogram.get().update(failures);
+                        } else {
+                            retryCountFailureHistogram.get().update(failures);
+                        }
                     }
                 }
 
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
@@ -287,21 +287,21 @@ final class RetryingChannel implements EndpointChannel {
             DialogueFutures.addDirectCallback(result, new FutureCallback<>() {
                 @Override
                 public void onSuccess(@Nullable Response response) {
+                    if (Responses.isSuccess(response)) {
+                        retryCountSuccessHistogram.get().update(failures);
+                    } else {
+                        retryCountFailureHistogram.get().update(failures);
+                    }
                     if (failures > 0) {
                         span.complete(RetryingCallbackTranslator.INSTANCE, RetryingCallback.this);
-                        if (Responses.isSuccess(response)) {
-                            retryCountSuccessHistogram.get().update(failures);
-                        } else {
-                            retryCountFailureHistogram.get().update(failures);
-                        }
                     }
                 }
 
                 @Override
                 public void onFailure(Throwable _throwable) {
+                    retryCountFailureHistogram.get().update(failures);
                     if (failures > 0) {
                         span.complete(RetryingCallbackTranslator.INSTANCE, RetryingCallback.this);
-                        retryCountFailureHistogram.get().update(failures);
                     }
                 }
             });
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
@@ -250,14 +250,22 @@ final class RetryingChannel implements EndpointChannel {
 
         ListenableFuture<Response> execute() {
             ListenableFuture<Response> result = wrap(delegate.execute(request));
-            result.addListener(
-                    () -> {
-                        if (failures > 0) {
-                            span.complete(RetryingCallbackTranslator.INSTANCE, this);
-                            retryCountHistogram.get().update(failures);
-                        }
-                    },
-                    DialogueFutures.safeDirectExecutor());
+            DialogueFutures.addDirectCallback(result, new FutureCallback<>() {
+                @Override
+                public void onSuccess(@Nullable Response _response) {
+                    if (failures > 0) {
+                        span.complete(RetryingCallbackTranslator.INSTANCE, RetryingCallback.this);
+                        retryCountHistogram.get().update(failures);
+                    }
+                }
+
+                @Override
+                public void onFailure(Throwable _throwable) {
+                    if (failures > 0) {
+                        span.complete(RetryingCallbackTranslator.INSTANCE, RetryingCallback.this);
+                    }
+                }
+            });
             return result;
         }
 
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
@@ -297,7 +297,7 @@ final class RetryingChannel implements EndpointChannel {
                         span.complete(RetryingCallbackTranslator.INSTANCE, RetryingCallback.this);
                     }
 
-                    if (Responses.isSuccess(response)) {
+                    if (requestCanBeRetried() && Responses.isSuccess(response)) {
                         retryCountSuccessHistogram.get().update(failures);
                     } else if (failures > maxRetries) {
                         // Means we got a retryable failure response, but retries were exhausted
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
@@ -158,29 +158,6 @@ final class RetryingChannel implements EndpointChannel {
                 () -> ThreadLocalRandom.current().nextDouble());
     }
 
-    @VisibleForTesting
-    RetryingChannel(
-            EndpointChannel delegate,
-            Endpoint endpoint,
-            String channelName,
-            TaggedMetricRegistry taggedMetricRegistry,
-            int maxRetries,
-            Duration backoffSlotSize,
-            ClientConfiguration.ServerQoS serverQoS,
-            ClientConfiguration.RetryOnTimeout retryOnTimeout) {
-        this(
-                delegate,
-                endpoint,
-                channelName,
-                taggedMetricRegistry,
-                maxRetries,
-                backoffSlotSize,
-                serverQoS,
-                retryOnTimeout,
-                sharedScheduler.get(),
-                () -> ThreadLocalRandom.current().nextDouble());
-    }
-
     private RetryingChannel(
             EndpointChannel delegate,
             Endpoint endpoint,
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
@@ -35,6 +35,7 @@ import com.palantir.dialogue.HttpMethod;
 import com.palantir.dialogue.Request;
 import com.palantir.dialogue.RequestBody;
 import com.palantir.dialogue.Response;
+import com.palantir.dialogue.core.DialogueClientMetrics.RequestRetryCount_Result;
 import com.palantir.dialogue.futures.DialogueFutures;
 import com.palantir.logsafe.SafeArg;
 import com.palantir.logsafe.exceptions.SafeRuntimeException;
@@ -108,7 +109,8 @@ final class RetryingChannel implements EndpointChannel {
     private final Supplier<Meter> retryDueToServerError;
     private final Supplier<Meter> retryDueToQosResponse;
     private final Function<Throwable, Meter> retryDueToThrowable;
-    private final Supplier<Histogram> retryCountHistogram;
+    private final Supplier<Histogram> retryCountSuccessHistogram;
+    private final Supplier<Histogram> retryCountFailureHistogram;
     private final Supplier<Counter> exhaustedDueToServerError;
     private final Supplier<Counter> exhaustedDueToQosResponse;
     private final Function<Throwable, Counter> exhaustedDueToThrowable;
@@ -191,7 +193,16 @@ final class RetryingChannel implements EndpointChannel {
                 .channelName(channelName)
                 .reason(throwable.getClass().getSimpleName())
                 .build();
-        this.retryCountHistogram = Suppliers.memoize(() -> dialogueClientMetrics.requestRetryCount(channelName));
+        this.retryCountSuccessHistogram = Suppliers.memoize(() -> dialogueClientMetrics
+                .requestRetryCount()
+                .channelName(channelName)
+                .result(RequestRetryCount_Result.SUCCESS)
+                .build());
+        this.retryCountFailureHistogram = Suppliers.memoize(() -> dialogueClientMetrics
+                .requestRetryCount()
+                .channelName(channelName)
+                .result(RequestRetryCount_Result.FAILURE)
+                .build());
         this.exhaustedDueToServerError = Suppliers.memoize(() -> dialogueClientMetrics
                 .requestRetryExhausted()
                 .channelName(channelName)
@@ -255,7 +266,7 @@ final class RetryingChannel implements EndpointChannel {
                 public void onSuccess(@Nullable Response _response) {
                     if (failures > 0) {
                         span.complete(RetryingCallbackTranslator.INSTANCE, RetryingCallback.this);
-                        retryCountHistogram.get().update(failures);
+                        retryCountSuccessHistogram.get().update(failures);
                     }
                 }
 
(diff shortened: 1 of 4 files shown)


dialogue-core/src/main/metrics/dialogue-core-metrics.yml
```diff
@@ -34,6 +34,16 @@ namespaces:
         type: meter
         tags: [channel-name, reason]
         docs: Rate at which the RetryingChannel retries requests (across all endpoints).
+      request.retry.count:
+        type: histogram
+        tags: [channel-name]
```
mpritham: This is only tagged with `channel-name`. I think for both of these metrics, having an `endpoint` tag would greatly increase the cardinality of the metrics. But without an `endpoint` tag, are these metrics still valuable? Perhaps initially we can use these metrics to identify services with outlier exhausted retries, and use logs to figure out the offending endpoints.

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -290,12 +315,18 @@ private ListenableFuture<Response> handleThrowable(Throwable clientSideThrowable
                                 clientSideThrowable);
                     }
                 }
+            } else {
+                // Retries exhausted due to throwable
+                exhaustedDueToThrowable.apply(clientSideThrowable).inc();
```
mpritham, line 356: I think there are a handful of throwables (< 10?) and this should not blow up cardinality.

mpritham: We also use the same logic (tagging the `simpleName` of the throwable class) with the `retryDueToThrowable` metric.

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -228,13 +250,22 @@ private RetryingCallback(
...
-                            span.complete(RetryingCallbackTranslator.INSTANCE, this);
-                        }
-                    },
-                    DialogueFutures.safeDirectExecutor());
+            DialogueFutures.addDirectCallback(result, new FutureCallback<>() {
+                @Override
+                public void onSuccess(@Nullable Response _response) {
+                    if (failures > 0) {
+                        span.complete(RetryingCallbackTranslator.INSTANCE, RetryingCallback.this);
+                        retryCountHistogram.get().update(failures);
```
mpritham: [Initially](https://github.com/palantir/dialogue/pull/2857/changes/839092ce18121c60047bb8b43694d5a116603113), I had `retryCountHistogram` update the failure count for all responses, but for our purposes I think it'd be nicer to have this metric only count the retries required to get a successful response.

dialogue-core/src/main/metrics/dialogue-core-metrics.yml
aldexis: @mpritham Wondering whether it makes sense to add a comment here to explain we aren't adding endpoint due to fear of exploding metrics cardinality

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -185,6 +191,22 @@ private RetryingChannel(
...
+        this.exhaustedDueToQosResponse = Suppliers.memoize(() -> dialogueClientMetrics
+                .requestRetryExhausted()
+                .channelName(channelName)
+                .reason("qosResponse")
+                .build());
+        this.exhaustedDueToThrowable = throwable -> dialogueClientMetrics
+                .requestRetryExhausted()
+                .channelName(channelName)
+                .reason(throwable.getClass().getSimpleName())
+                .build();
```
aldexis, line 226: ~Do we have a sense about the cardinality of such throwables?~ Edit: Seeing as how we already have similar metrics for `retryDueToThrowable`, the cardinality isn't a problem (and I also just saw you mentioned the same thing below)

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
aldexis: I'm curious if it would make sense to just have both, with an additional tag (success/failure).
I think we can derive the failure from the exhausted + configured number of retries, but since this is configurable, it might be annoying to have to go and figure it out, whereas we could get the metric directly here

aldexis (commented):
Generally looks good to me - just a couple of questions

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
kunalrkak: I'm a bit confused on what exactly is being tracked by the success/failure in this context - is it a successful response in terms of HTTP code or is it just successful in the sense the request completed without something like a connection failure? And in either case what's the importance of making the distinction?

mpritham: > I'm curious if it would make sense to just have both, with an additional tag (success/failure).

I think this makes sense, I've updated the PR.

> is it a successful response in terms of HTTP code or is it just successful in the sense the request completed without something like a connection failure?

The latter. [The request isn't always retried when there is an exception.](https://github.com/palantir/dialogue/blob/ec0c908b723519ddd9946106641bd2e8457c744a/dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java#L394-L410) This bit was added to the docs to make it clear for folks:

"The result tag of "success" indicates that the request eventually received a response, while "failure" indicates that there was no response received from the server (e.g. socket read timeout)."

> And in either case what's the importance of making the distinction?

Some of the questions we're interested in exploring are: "when can we reduce the retry count to less than the default (4)?" and "is there an argument to reduce Dialogue's default retries to something below 4?". To help answer these questions, separating the retry distribution for requests that ultimately succeeded from those that failed is useful. For example, if we noticed a very right tailed distribution (most observations around 1-2 retries), and we know that the request eventually succeeded, we have some argument that we can drop the retry count. Without the distinction, it could be that several requests just fail (maybe due to a node rolling, or whatever) and they would have otherwise succeeded with more retries.

aldexis: @mpritham What about requests that get Qos'ed? These are considered a failure (and maybe retried, with a potential eventual success), correct?

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -228,13 +261,23 @@ private RetryingCallback(
...
-                            span.complete(RetryingCallbackTranslator.INSTANCE, this);
-                        }
-                    },
-                    DialogueFutures.safeDirectExecutor());
+            DialogueFutures.addDirectCallback(result, new FutureCallback<>() {
+                @Override
+                public void onSuccess(@Nullable Response _response) {
+                    if (failures > 0) {
+                        span.complete(RetryingCallbackTranslator.INSTANCE, RetryingCallback.this);
+                        retryCountSuccessHistogram.get().update(failures);
```
aldexis, line 278: Do we have a sample bias here? This only counts requests that succeeded after a retry, right? Should we maybe also include the succeeded-in-the-first-try?

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -228,13 +261,23 @@ private RetryingCallback(
...
+                    }
+                }
+
+                @Override
+                public void onFailure(Throwable _throwable) {
+                    if (failures > 0) {
+                        span.complete(RetryingCallbackTranslator.INSTANCE, RetryingCallback.this);
+                        retryCountFailureHistogram.get().update(failures);
```
aldexis: A bit similar to the above, though I'd expect failures to always be >0 here

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
mpritham: > These are considered a failure (and maybe retried, with a potential eventual success), correct?

These will be considered a successful response: QOS response will hit the `onSuccess` path here.

I've added some tests to show this is true, and adding some branching logic here checking the response code so that we increment the "success" tagged metric only on HTTP 2xx responses.

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
mpritham: Sure, I can update. I think the exclusion of 0-retry values doesn't affect the analysis we'd like to do i.e. whether it makes sense to adjust the max retry count, but including the 0-values will not hurt.

dialogue-core/src/main/metrics/dialogue-core-metrics.yml
```diff
@@ -34,6 +34,24 @@ namespaces:
...
+      # Note: endpoint is not added as a tag to the retry metrics below because we do not want to greatly increase
+      # the cardinality of tags: doing so adds load to the observability pipeline.
+      request.retry.count:
+        type: histogram
+        tags:
+          - name: channel-name
+          - name: result
+            values: [ success, failure ]
+        docs: |
+          Distribution of retry counts per request for requests that required at least one retry. The result tag of
```
aldexis: We aren't requiring at least one retry anymore, correct?
```suggestion
          Distribution of retry counts per request for requests. The result tag of 
```

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -228,13 +284,27 @@ private RetryingCallback(
...
+                    }
+                }
+
+                @Override
+                public void onFailure(Throwable _throwable) {
+                    retryCountFailureHistogram.get().update(failures);
+                    if (failures > 0) {
+                        span.complete(RetryingCallbackTranslator.INSTANCE, RetryingCallback.this);
+                    }
+                }
```
aldexis, line 295: Just to make sure I understand correctly, if we exhaust retries, we will also increment the retryCountFailureHistogram, correct?

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -228,13 +284,27 @@ private RetryingCallback(
...
-                        }
-                    },
-                    DialogueFutures.safeDirectExecutor());
+            DialogueFutures.addDirectCallback(result, new FutureCallback<>() {
+                @Override
+                public void onSuccess(@Nullable Response response) {
+                    if (Responses.isSuccess(response)) {
+                        retryCountSuccessHistogram.get().update(failures);
+                    } else {
+                        retryCountFailureHistogram.get().update(failures);
```
aldexis, line 281: I'm wondering how these will appear/how we want them to appear for non-retryable errors. These are not technically "success", but since they're not retryable, we'll look like we failed after a single attempt, which might bias our results?
Should we distinguish between "success", "retryable-failure" and "non-retryable-failure"? If I understand correctly, what we care about is measuring the number of attempts we gave before we got something we didn't need to/couldn't retry, as well as whether we also exhausted retries retrying things that could be retried? (so splitting all three of these, then having a dashboard that maybe coalesces both success and non-retryable failures might be what we want?)

dialogue-core/src/main/metrics/dialogue-core-metrics.yml
```diff
@@ -34,6 +34,24 @@ namespaces:
...
+      request.retry.count:
+        type: histogram
+        tags:
+          - name: channel-name
+          - name: result
+            values: [ success, failure ]
+        docs: |
+          Distribution of retry counts per request for retryable requests. The result tag of "success" indicates that 
+          the request eventually received a response, while "failure" indicates that there was no response received 
+          from the server (e.g. socket read timeout).
```
bjlaub: Does this metric only apply to scenarios where no response is received (like a timeout)? There are many (maybe most) scenarios where we get a response, it's just not the one we wanted, and we retry.

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -149,6 +157,29 @@ static EndpointChannel create(Config cf, EndpointChannel channel, Endpoint endpo
                 () -> ThreadLocalRandom.current().nextDouble());
     }
 
+    @VisibleForTesting
+    RetryingChannel(
```
bjlaub: nit: the constructor [here](https://github.com/palantir/dialogue/pull/2857/changes#diff-0fc6bbc06785e70f7ea8aa8c1a0cdd305eecf9718a966e57ecf937fe22520544R138-R146) appears to only be used in test code - why not just update it to accept a `TaggedMetricRegistry` and fix up the tests, rather than add a new constructor?

dialogue-core/src/main/metrics/dialogue-core-metrics.yml
```diff
@@ -34,6 +34,24 @@ namespaces:
...
         tags: [channel-name, reason]
         docs: Rate at which the RetryingChannel retries requests (across all endpoints).
+      # Note: endpoint is not added as a tag to the retry metrics below because we do not want to greatly increase
+      # the cardinality of tags: doing so adds load to the observability pipeline.
+      request.retry.count:
+        type: histogram
+        tags:
+          - name: channel-name
+          - name: result
+            values: [ success, failure ]
```
bjlaub: this metric is a little confusing to me. what is the difference between a retry count tagged with `success` versus `failure`? How is the latter different from the `request.retry.exhausted` counter?

mpritham: > what is the difference between a retry count tagged with success versus failure? How is the latter different from the request.retry.exhausted counter?

"success" would indicate that the request eventually received a 2xx response, and "failure" indicates that there was no response received from the server (some failure mode like a timeout/other exception thrown) or, all retries were exhausted and Dialogue received a non-2xx response.

So for a request that got a 2xx response on the 3rd retry, we expect to have a `(3, 'success')` logged. For a request that got a 429 response after the final retry, we expect a `(final_retry_count, 'failure')` logged. For a request that timed out on the 2nd retry, we expect a `(2, 'failure')` logged.   

The "success" case is useful for us: it'll give us the histogram of # of retries before a request returns a 2xx.

> How is the latter different from the request.retry.exhausted counter?

`request.retry.exhausted` is incremented by 1 for every request that fails due to retry exhaustion. It also has a tag for the reason the last response failed (e.g. qos reason, or a throwable, or a server error). The `failure` tag on this histogram will count the number of retries before failure. So for the cases where retries have been exhausted, we should see a `(max_num_retries, 'failure')` metric logged. We would expect this to be the default max number of retries for most clients but some clients do override this value.

dialogue-core/src/main/metrics/dialogue-core-metrics.yml
mpritham: The 'failure' case also applies to cases where the response received on the final retry is not a 2xx.

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
mpritham: Yes, I'm proposing that the `success`/`failure` tag represent whether the response (after all retries) is "successful" in the sense that it is usable by a client (a 2xx).

mpritham (commented):
comments

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
mpritham: I think tracking the non-retryable case may not be interesting to us? If a request is not retryable (e.g. it's a POST request), we wouldn't have _any_ retries, and I think this metric shouldn't be updated for such requests.

What are your thoughts on the "success"/"failure" tag representing whether the response received by the client (after all/any retries) is a 2xx?

aldexis: > What are your thoughts on the "success"/"failure" tag representing whether the response received by the client (after all/any retries) is a 2xx?

I think this would hide the cases where we retried (e.g. due to network blips) and _then_ got e.g. a 4xx?

The information we want afaik is "Can we safely reduce the number of retries?". Errors can be meaningful. If we want to make sure we don't change the system's behavior, we'd want to determine "how many retries did we take before we got to an end state?" where end state can be
- We got a successful response
- We got a non-retryable error
- We exhausted our retries

_Currently_, we don't need to distinguish successful responses from non-retryable errors in our analysis (we probably do need _both_ though, as per the above comment). However, it seems pretty confusing to me to mark these as success in our metrics (since they're not a 2xx response), which is why I was proposing we have the three possible states in our metrics (and our analysis would consider success and non-retryable as one and the same)

mpritham: > I think this would hide the cases where we retried (e.g. due to network blips) and then got e.g. a 4xx?

I think this previously wasn't the case because this logic is run after all of the retries. I have some tests that attempt to validate this.

Ok this breakdown of 3 states makes sense to me as well and will give us more info which doesn't hurt. I pushed a change doing what I believe you're proposing.

> our analysis would consider success and non-retryable as one and the same

fwiw I think we may not want to do this. Dropping retry count when we've observed that we're receiving failures (e.g. 500s) after say N retries may not be what we want? I think that's unrelated to this change though: can discuss elsewhere/later.

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -228,13 +290,34 @@ private RetryingCallback(
...
+                        } else if (failures > maxRetries) {
+                            retryCountFailureHistogram.get().update(failures);
+                        } else {
+                            retryCountNonRetryableHistogram.get().update(failures);
                         }
-                    },
-                    DialogueFutures.safeDirectExecutor());
+                    } else if (requestCanBeRetried() && Responses.isSuccess(response)) {
+                        // Only update the success metric if the request was retry-able and succeeded.
+                        retryCountSuccessHistogram.get().update(failures);
```
aldexis: nit: might be a bit unexpected to not see non-retryable counts as 0 if they failed immediately? (I mentally model these as just different terminal states and think the metric about how many retries it took to reach a terminal state to be as valuable, regardless of the terminal state we reached)

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -228,13 +290,34 @@ private RetryingCallback(
...
+                public void onSuccess(@Nullable Response response) {
+                    if (failures > 0) {
+                        span.complete(RetryingCallbackTranslator.INSTANCE, RetryingCallback.this);
+                        if (Responses.isSuccess(response)) {
+                            retryCountSuccessHistogram.get().update(failures);
+                        } else if (failures > maxRetries) {
+                            retryCountFailureHistogram.get().update(failures);
+                        } else {
+                            retryCountNonRetryableHistogram.get().update(failures);
                         }
```
aldexis: Does this update the correct metric if we receive a non-retryable response as the last retry? (It seems like it would, but I'm fuzzy as to why we'd end up with more failures than max retries)

Similarly, shouldn't we check the response for whether it is retryable in the last branch? Otherwise, aren't we updating the histogram multiple times? (or is this the "terminal" future already, after all possible retries?)

dialogue-core/src/test/java/com/palantir/dialogue/core/RetryingChannelTest.java
```diff
@@ -854,6 +856,361 @@ public void close() {}
...
+                        .getValues())
+                .as("Success histogram should record 2 failures (the 429s) before success")
+                .containsExactly(2L);
+        assertThat(metrics.requestRetryCount()
+                        .channelName("my-channel")
+                        .result(RequestRetryCount_Result.FAILURE)
+                        .build()
+                        .getCount())
+                .as("Failure histogram should not be updated when request eventually succeeds")
+                .isEqualTo(0);
```
aldexis: nit: seems like this is boilerplate common to all the tests and could be extracted to a verification method?

aldexis: See https://github.com/palantir/dialogue/pull/2877 for some proposed changes to reduce the boilerplate and make tests more readable

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
aldexis: Added a test for this in https://github.com/palantir/dialogue/pull/2877

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
aldexis: For instance, when sharing the verification logic in the tests as I propose in https://github.com/palantir/dialogue/pull/2877, it doesn't make sense to single out this case. Instead it makes more sense to have consistent behavior, and this is what I'd expect when interacting with the system. I made some proposed changes to cover this

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
mpritham: This logic is run after the retries so I expect this to update the correct metric and only once.

dialogue-core/src/test/java/com/palantir/dialogue/core/RetryingChannelTest.java
```diff
@@ -889,6 +901,261 @@ public void close() {}
...
+        EndpointChannel retryer = channel(4);
+
+        ListenableFuture<Response> response = retryer.execute(requestWithNonRepeatableBody());
+        assertThat(response.get().code()).isEqualTo(200);
+        verify(channel, times(1)).execute(any());
+
+        verifyMetrics(
+                new ExpectedMetrics(RequestRetryCount_Result.NON_RETRYABLE, 0),
+                "non-retryable request success should not be counted in success histogram");
+    }
```
aldexis, line 1036: I think this might be slightly confusing (to me at least) between non-retryable requests which succeed and responses that fail in a non-retryable manner 🤔 But this is probably fine for now anyway and we don't need to add more complexity until we've tested the workflows

aldexis (approved):

aldexis:
I think you might need to self-approve (or get someone else to approve) since I've now also committed to the branch

mpritham:
👍

PR #2877, opened 2026-02-13
Proposed change for #2857 tests + code

aldexis:
Proposing some changes per my review at https://github.com/palantir/dialogue/pull/2857#pullrequestreview-3796631574


mpritham (approved):
thanks!

PR #2881, opened 2026-02-17
Preserve stacktrace for thrown QosExceptions, same as RemoteExceptions

ash211:
## Before this PR
When QosExceptions are thrown from dialogue blocking clients, they have little useful stacktrace because they're run on a separate executor service from the calling thread.

Example:
```
com.palantir.conjure.java.api.errors.QosException$Throttle: Suggested request throttling
	at com.palantir.conjure.java.api.errors.QosException.throttle(QosException.java:77)
	at com.palantir.conjure.java.dialogue.serde.ExceptionDeserializingErrorDecoder.lambda$decodeInternal$1(ExceptionDeserializingErrorDecoder.java:135)
	at java.base/java.util.Optional.orElseGet(Optional.java:364)
	at com.palantir.conjure.java.dialogue.serde.ExceptionDeserializingErrorDecoder.decodeInternal(ExceptionDeserializingErrorDecoder.java:135)
	at com.palantir.conjure.java.dialogue.serde.ExceptionDeserializingErrorDecoder.decode(ExceptionDeserializingErrorDecoder.java:101)
	at com.palantir.conjure.java.dialogue.serde.ConjureBodySerDe$EncodingDeserializerRegistry.deserialize(ConjureBodySerDe.java:322)
	at com.palantir.dialogue.futures.DialogueDirectTransformationFuture.onSuccess(DialogueDirectTransformationFuture.java:109)
	at com.google.common.util.concurrent.Futures$CallbackListener.run(Futures.java:1132)
	at com.palantir.dialogue.futures.SafeDirectExecutor.execute(SafeDirectExecutor.java:32)
	at com.google.common.util.concurrent.AbstractFuture.executeListener(AbstractFuture.java:1004)
	at com.google.common.util.concurrent.AbstractFuture.complete(AbstractFuture.java:767)
	at com.google.common.util.concurrent.AbstractFuture.set(AbstractFuture.java:491)
	at com.google.common.util.concurrent.SettableFuture.set(SettableFuture.java:48)
	at com.palantir.dialogue.futures.DialogueDirectTransformationFuture.onSuccess(DialogueDirectTransformationFuture.java:110)
	at com.google.common.util.concurrent.Futures$CallbackListener.run(Futures.java:1132)
	at com.palantir.dialogue.futures.SafeDirectExecutor.execute(SafeDirectExecutor.java:32)
	at com.google.common.util.concurrent.AbstractFuture.executeListener(AbstractFuture.java:1004)
	at com.google.common.util.concurrent.AbstractFuture.complete(AbstractFuture.java:767)
	at com.google.common.util.concurrent.AbstractFuture.set(AbstractFuture.java:491)
	at com.google.common.util.concurrent.AbstractCatchingFuture.run(AbstractCatchingFuture.java:122)
	at com.palantir.dialogue.futures.SafeDirectExecutor.execute(SafeDirectExecutor.java:32)
	at com.google.common.util.concurrent.MoreExecutors.lambda$rejectionPropagatingExecutor$0(MoreExecutors.java:1063)
	at com.google.common.util.concurrent.AbstractFuture.executeListener(AbstractFuture.java:1004)
	at com.google.common.util.concurrent.AbstractFuture.complete(AbstractFuture.java:767)
	at com.google.common.util.concurrent.AbstractFuture.set(AbstractFuture.java:491)
	at com.google.common.util.concurrent.AbstractCatchingFuture.run(AbstractCatchingFuture.java:122)
	at com.palantir.dialogue.futures.SafeDirectExecutor.execute(SafeDirectExecutor.java:32)
	at com.google.common.util.concurrent.MoreExecutors.lambda$rejectionPropagatingExecutor$0(MoreExecutors.java:1063)
	at com.google.common.util.concurrent.AbstractFuture.executeListener(AbstractFuture.java:1004)
	at com.google.common.util.concurrent.AbstractFuture.complete(AbstractFuture.java:767)
	at com.google.common.util.concurrent.AbstractFuture.set(AbstractFuture.java:491)
	at com.google.common.util.concurrent.AbstractCatchingFuture.run(AbstractCatchingFuture.java:122)
	at com.palantir.dialogue.futures.SafeDirectExecutor.execute(SafeDirectExecutor.java:32)
	at com.google.common.util.concurrent.MoreExecutors.lambda$rejectionPropagatingExecutor$0(MoreExecutors.java:1063)
	at com.google.common.util.concurrent.AbstractFuture.executeListener(AbstractFuture.java:1004)
	at com.google.common.util.concurrent.AbstractFuture.complete(AbstractFuture.java:767)
	at com.google.common.util.concurrent.AbstractFuture.set(AbstractFuture.java:491)
	at com.google.common.util.concurrent.AbstractCatchingFuture.run(AbstractCatchingFuture.java:122)
	at com.palantir.dialogue.futures.SafeDirectExecutor.execute(SafeDirectExecutor.java:32)
	at com.google.common.util.concurrent.MoreExecutors.lambda$rejectionPropagatingExecutor$0(MoreExecutors.java:1063)
	at com.google.common.util.concurrent.AbstractFuture.executeListener(AbstractFuture.java:1004)
	at com.google.common.util.concurrent.AbstractFuture.complete(AbstractFuture.java:767)
	at com.google.common.util.concurrent.AbstractFuture.set(AbstractFuture.java:491)
	at com.google.common.util.concurrent.AbstractCatchingFuture.run(AbstractCatchingFuture.java:122)
	at com.palantir.dialogue.futures.SafeDirectExecutor.execute(SafeDirectExecutor.java:32)
	at com.google.common.util.concurrent.MoreExecutors.lambda$rejectionPropagatingExecutor$0(MoreExecutors.java:1063)
	at com.google.common.util.concurrent.AbstractFuture.executeListener(AbstractFuture.java:1004)
	at com.google.common.util.concurrent.AbstractFuture.complete(AbstractFuture.java:767)
	at com.google.common.util.concurrent.AbstractFuture.setFuture(AbstractFuture.java:560)
	at com.google.common.util.concurrent.AbstractTransformFuture$AsyncTransformFuture.setResult(AbstractTransformFuture.java:237)
	at com.google.common.util.concurrent.AbstractTransformFuture$AsyncTransformFuture.setResult(AbstractTransformFuture.java:213)
	at com.google.common.util.concurrent.AbstractTransformFuture.run(AbstractTransformFuture.java:172)
	at com.palantir.dialogue.futures.SafeDirectExecutor.execute(SafeDirectExecutor.java:32)
	at com.google.common.util.concurrent.MoreExecutors.lambda$rejectionPropagatingExecutor$0(MoreExecutors.java:1063)
	at com.google.common.util.concurrent.AbstractFuture.executeListener(AbstractFuture.java:1004)
	at com.google.common.util.concurrent.AbstractFuture.complete(AbstractFuture.java:767)
	at com.google.common.util.concurrent.AbstractFuture.set(AbstractFuture.java:491)
	at com.google.common.util.concurrent.SettableFuture.set(SettableFuture.java:48)
	at com.palantir.dialogue.core.RetryingChannel$RetryingCallback$1.onSuccess(RetryingChannel.java:350)
	at com.palantir.dialogue.core.RetryingChannel$RetryingCallback$1.onSuccess(RetryingChannel.java:347)
	at com.google.common.util.concurrent.Futures$CallbackListener.run(Futures.java:1132)
	at com.palantir.dialogue.futures.SafeDirectExecutor.execute(SafeDirectExecutor.java:32)
	at com.google.common.util.concurrent.AbstractFuture.executeListener(AbstractFuture.java:1004)
	at com.google.common.util.concurrent.AbstractFuture.complete(AbstractFuture.java:767)
	at com.google.common.util.concurrent.AbstractFuture.setFuture(AbstractFuture.java:560)
	at com.google.common.util.concurrent.AbstractTransformFuture$AsyncTransformFuture.setResult(AbstractTransformFuture.java:237)
	at com.google.common.util.concurrent.AbstractTransformFuture$AsyncTransformFuture.setResult(AbstractTransformFuture.java:213)
	at com.google.common.util.concurrent.AbstractTransformFuture.run(AbstractTransformFuture.java:172)
	at com.palantir.dialogue.futures.SafeDirectExecutor.execute(SafeDirectExecutor.java:32)
	at com.google.common.util.concurrent.MoreExecutors.lambda$rejectionPropagatingExecutor$0(MoreExecutors.java:1063)
	at com.google.common.util.concurrent.AbstractFuture.executeListener(AbstractFuture.java:1004)
	at com.google.common.util.concurrent.AbstractFuture.complete(AbstractFuture.java:767)
	at com.google.common.util.concurrent.AbstractFuture.set(AbstractFuture.java:491)
	at com.google.common.util.concurrent.SettableFuture.set(SettableFuture.java:48)
	at com.palantir.dialogue.blocking.BlockingChannelAdapter$BlockingChannelAdapterChannel$BlockingChannelAdapterTask.run(BlockingChannelAdapter.java:141)
	at com.palantir.tracing.Tracers$TracingAwareRunnable.run(Tracers.java:617)
	at com.palantir.tritium.metrics.TaggedMetricsExecutorService$TaggedMetricsRunnable.run(TaggedMetricsExecutorService.java:134)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:572)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:317)
	at com.palantir.nylon.threads.RenamingExecutorService$RenamingRunnable.run(RenamingExecutorService.java:92)
	at org.jboss.threads.EnhancedViewExecutor$EnhancedViewExecutorRunnable.run(EnhancedViewExecutor.java:496)
	at org.jboss.threads.ContextHandler$1.runWith(ContextHandler.java:18)
	at org.jboss.threads.EnhancedQueueExecutor$Task.doRunWith(EnhancedQueueExecutor.java:2651)
	at org.jboss.threads.EnhancedQueueExecutor$Task.run(EnhancedQueueExecutor.java:2630)
	at org.jboss.threads.EnhancedQueueExecutor.runThreadBody(EnhancedQueueExecutor.java:1622)
	at org.jboss.threads.EnhancedQueueExecutor$ThreadBody.run(EnhancedQueueExecutor.java:1589)
	at com.palantir.tritium.metrics.TaggedMetricsThreadFactory$InstrumentedTask.run(TaggedMetricsThreadFactory.java:94)
	at java.base/java.lang.Thread.run(Thread.java:1583)
	Suppressed: com.palantir.conjure.java.dialogue.serde.ExceptionDeserializingErrorDecoder$ResponseDiagnostic: Response Diagnostic Information
```

## After this PR
==COMMIT_MSG==
Preserve stacktrace for thrown QosExceptions, same as RemoteExceptions
==COMMIT_MSG==

With this change, we apply the same technique that `RemoteExceptions` and `UnknownRemoteExceptions` get in the two clauses above.  Those blocks re-create the exception, but now with the current thread's stacktrace. This will now include the service's callsite, making it easier to determine what codepath is triggering the QosExceptions. It also includes the service + method in the stacktrace (e.g. `DeletionOrchestrationServiceBlocking$1.initializeAndGetActiveServiceDeletionContexts`) making it easier to determine which endpoint is responding with QosExceptions.

## Possible downsides?
- this is now more code to execute during QosExceptions, which happen when the service is already under load.  Ideally we wouldn't put more work on the "service is overloaded" codepath, but I think the improvement in debuggability is worth it to me as a service owner.

### DIFF
--- a/dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/DefaultClients.java
+++ b/dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/DefaultClients.java
@@ -21,6 +21,7 @@ import com.google.common.util.concurrent.FutureCallback;
 import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
 import com.palantir.conjure.java.api.errors.AbstractSerializableError;
+import com.palantir.conjure.java.api.errors.QosException;
 import com.palantir.conjure.java.api.errors.RemoteException;
 import com.palantir.conjure.java.api.errors.SerializableErrorProvider;
 import com.palantir.conjure.java.api.errors.UnknownRemoteException;
@@ -139,6 +140,10 @@ enum DefaultClients implements Clients {
                 throw newUnknownRemoteException(unknownRemoteException);
             }
 
+            if (cause instanceof QosException qosException) {
+                throw newQosException(qosException);
+            }
+
             // In this case we provide a suppressed exception to mark the site where the failure was rethrown
             // to avoid losing data while retaining the original failure information.
             if (cause instanceof RuntimeException runtimeException) {
@@ -213,6 +218,30 @@ enum DefaultClients implements Clients {
     private static UnknownRemoteException newUnknownRemoteException(UnknownRemoteException cause) {
         UnknownRemoteException newException = new UnknownRemoteException(cause.getStatus(), cause.getBody());
         newException.initCause(cause);
+        return newException;
+    }
+
+    // Re-create QosException so our current stacktrace is included, similar to RemoteException handling
+    private static QosException newQosException(QosException qosException) {
+        QosException newException = qosException.accept(new QosException.Visitor<>() {
+            @Override
+            public QosException visit(QosException.Throttle exception) {
+                return exception.getRetryAfter().isPresent()
+                        ? QosException.throttle(
+                                exception.getReason(), exception.getRetryAfter().get(), qosException)
+                        : QosException.throttle(exception.getReason(), qosException);
+            }
+
+            @Override
+            public QosException visit(QosException.RetryOther exception) {
+                return QosException.retryOther(exception.getReason(), exception.getRedirectTo(), qosException);
+            }
+
+            @Override
+            public QosException visit(QosException.Unavailable exception) {
+                return QosException.unavailable(exception.getReason(), qosException);
+            }
+        });
         return newException;
     }
 
--- a/.gitignore
+++ b/.gitignore
@@ -60,3 +60,5 @@ __init__.pyc
 
 # Gradle JDKs setup
 !gradle/*
+
+.claude/settings.local.json
--- a/dialogue-serde/src/test/java/com/palantir/conjure/java/dialogue/serde/DefaultClientsBlockingTest.java
+++ b/dialogue-serde/src/test/java/com/palantir/conjure/java/dialogue/serde/DefaultClientsBlockingTest.java
@@ -26,6 +26,8 @@ import com.google.common.util.concurrent.ListenableFuture;
 import com.google.common.util.concurrent.SettableFuture;
 import com.palantir.conjure.java.api.errors.AbstractSerializableError;
 import com.palantir.conjure.java.api.errors.ErrorType;
+import com.palantir.conjure.java.api.errors.QosException;
+import com.palantir.conjure.java.api.errors.QosReason;
 import com.palantir.conjure.java.api.errors.RemoteException;
 import com.palantir.conjure.java.api.errors.SerializableError;
 import com.palantir.conjure.java.api.errors.SerializableErrorProvider;
@@ -35,6 +37,7 @@ import com.palantir.dialogue.DialogueException;
 import com.palantir.logsafe.exceptions.SafeRuntimeException;
 import java.io.IOException;
 import java.io.InputStream;
+import java.time.Duration;
 import java.util.Map;
 import java.util.Optional;
 import org.assertj.core.api.Assertions;
@@ -70,6 +73,62 @@ public class DefaultClientsBlockingTest {
                 .satisfies(exception -> {
                     assertThat(((UnknownRemoteException) exception).getBody()).isEqualTo("Nginx broke");
                 });
+    }
+
+    @Test
+    public void testQosExceptionThrottle() {
+        QosException qosException = QosException.throttle(QosReason.of("test-reason"));
+        ListenableFuture<Object> failedFuture = Futures.immediateFailedFuture(qosException);
+
+        assertThatThrownBy(() -> DefaultClients.INSTANCE.block(failedFuture))
+                .isInstanceOf(QosException.Throttle.class)
+                .isNotSameAs(qosException)
+                .hasCause(qosException)
+                .satisfies(exception ->
+                        assertThat(((QosException) exception).getReason()).isEqualTo(QosReason.of("test-reason")));
+    }
+
+    @Test
+    public void testQosExceptionThrottleWithRetryAfter() {
+        QosException qosException = QosException.throttle(QosReason.of("test-reason"), Duration.ofSeconds(30));
+        ListenableFuture<Object> failedFuture = Futures.immediateFailedFuture(qosException);
+
+        assertThatThrownBy(() -> DefaultClients.INSTANCE.block(failedFuture))
+                .isInstanceOf(QosException.Throttle.class)
+                .isNotSameAs(qosException)
+                .hasCause(qosException)
+                .satisfies(exception -> {
+                    QosException.Throttle throttle = (QosException.Throttle) exception;
+                    assertThat(throttle.getReason()).isEqualTo(QosReason.of("test-reason"));
+                    assertThat(throttle.getRetryAfter()).hasValue(Duration.ofSeconds(30));
+                });
+    }
+
+    @Test
+    public void testQosExceptionUnavailable() {
+        QosException qosException = QosException.unavailable(QosReason.of("test-unavailable"));
+        ListenableFuture<Object> failedFuture = Futures.immediateFailedFuture(qosException);
+
+        assertThatThrownBy(() -> DefaultClients.INSTANCE.block(failedFuture))
+                .isInstanceOf(QosException.Unavailable.class)
+                .isNotSameAs(qosException)
+                .hasCause(qosException)
+                .satisfies(exception ->
+                        assertThat(((QosException) exception).getReason()).isEqualTo(QosReason.of("test-unavailable")));
+    }
+
+    @Test
+    public void testQosExceptionStackTraceIncludesBlockCallSite() {
+        QosException qosException = QosException.throttle(QosReason.of("test-reason"));
+        // Clear the stack trace to simulate an exception created on a remote/async thread
+        qosException.setStackTrace(new StackTraceElement[0]);
+        ListenableFuture<Object> failedFuture = Futures.immediateFailedFuture(qosException);
+
+        assertThatThrownBy(() -> DefaultClients.INSTANCE.block(failedFuture))
+                .isInstanceOf(QosException.Throttle.class)
+                .satisfies(exception -> assertThat(exception.getStackTrace())
+                        .extracting(StackTraceElement::getMethodName)
+                        .contains("testQosExceptionStackTraceIncludesBlockCallSite"));
     }
 
     @Test


dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/DefaultClients.java
```diff
@@ -216,6 +221,30 @@ private static UnknownRemoteException newUnknownRemoteException(UnknownRemoteExc
         return newException;
     }
 
+    // Re-create QosException so our current stacktrace is included, similar to RemoteException handling
+    private static QosException newQosException(QosException qosException) {
```
ash211, line 225: I looked for an equivalent method to this, but I don't see one elsewhere. The closest is in an old gotham StacktracePreservingProxy class that doesn't copy over all the properties.

A new method `QosException.copy(QosException)` might be appropriate here longer term, to reduce the risk that new fields are added in the future and not copied over.

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/DefaultClients.java
```diff
@@ -139,6 +140,10 @@ public <T> T block(ListenableFuture<T> future) {
                 throw newUnknownRemoteException(unknownRemoteException);
             }
 
+            if (cause instanceof QosException qosException) {
+                throw newQosException(qosException);
+            }
```
pkoenig10, line 145: I would expect the thrown exception to include the caller's stacktrace in a suppressed exception per the `RuntimeException` case below. Is that not happening? If so, why?

Recreating the exception is pretty brittle - if we add new functionality to `QosException`, we need to remember to update this logic so we don't lose the intention of the caller. It's also a bit surprising for callers to not get back the original `QosException`. Suppressed exceptions are the more correct way to capture this information.

aldexis: +1 to this - I'd like to confirm why the suppressed exception isn't enough first (and if we _must_ do this, it seems like a good idea to add the original as a suppressed exception instead?)

pkoenig10: > it seems like a good idea to add the original as a suppressed exception instead

If we were to create a new exception, then the original exception should probably just be the cause.

ash211: Is that the same pattern as the existing code for RemoteExceptions?

https://github.com/palantir/dialogue/blob/ce70d95373f8e328fbcce788740ffe19a442925b/dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/DefaultClients.java#L158-L169

I think you're maybe suggesting we add

```
newException.initCause(qosException);
```

in the second to last line of the `newQosException()` method.

aldexis: > If we were to create a new exception, then the original exception should probably just be the cause.

Oh, yes, of course 🤦 

@ash211 Yes, I think the initCause would be necessary here, but going back to @pkoenig10 's original question - why isn't the suppressed exception working?

pkoenig10: Just because existing code does something a certain way doesn't mean it's the correct thing to do.

This code blames all the way back to https://github.com/palantir/dialogue/pull/115, with no indication why we create a new exception other than to record the caller's stacktrace. If that is the only goal, then using a suppressed exception is the more appropriate mechanism for doing so.

I would propose we change this implementation to use suppressed exception for all these cases (this would likely simplify this code quite a bit).

pkoenig10: To make this concrete, this is the change I am proposing: https://github.com/palantir/dialogue/commit/ad1720da282d4ad984f02341e4d750e34fcc7120

ash211: The change in that commit is a redefinition of what `DialogueException` is supposed to represent.

https://github.com/palantir/dialogue/blob/1232fdf97b84581d8251c1c5668885b438db7c6f/dialogue-target/src/main/java/com/palantir/dialogue/DialogueException.java#L25

When we get RemoteExceptions or QosExceptions back, we do have a response, whereas DialogueException is supposed to be `where no response is returned`.

As I look more closely, it seems like we're currently adding a suppressed exception before rethrowing it here:

https://github.com/palantir/dialogue/blob/1232fdf97b84581d8251c1c5668885b438db7c6f/dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/DefaultClients.java#L142-L147

But it doesn't seem to be included in the logging point where I see in my logs: https://pl.ntr/2zO.  Do you think the suppressed exception is getting lost somewhere?

pkoenig10: > The change in that commit is a redefinition of what `DialogueException` is supposed to represent.

That commit doesn't throw `DialogueException` in any cases that don't throw `DialogueException` in the current implementation. It simply avoids special casing certain `RuntimeException` types. It doesn't throw `DialogueException` for `RemoteException` or `QosException` - you can see this in the tests.

> But it doesn't seem to be included in the logging point where I see in my logs: https://pl.ntr/2zO. Do you think the suppressed exception is getting lost somewhere?

There's not enough information in those logs alone to know where that exception came from. The origin of that logs appears to be some sort of logging utility that is used by multiple code paths. My best guess is that this exception came from an async client and we pass the exception from the failed future directly to that utility.

pkoenig10: Looking at the other logs on this trace ID, I see [this `RetryingChannel` log](https://github.com/palantir/dialogue/blob/1232fdf97b84581d8251c1c5668885b438db7c6f/dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java#L415-L423) `Exhausted {} retries, returning last received response with status {}` for `ActionsService.applyActionV2`. And looking at the code for this service, the only non-test usage of this endpoint is [this code](https://pl.ntr/2zP) which does use an async client.

pkoenig10: And [here's the code](https://pl.ntr/2zQ) where we handle the failure of this future by passing the exception directly into the logging utility.

So the code being modified here isn't relevant for the error that motivated this PR.

ash211: Looks like you verified the hypothesis that the callsite is an Async conjure client with a common logging method on any thrown Throwables.

Do you think the fix I'm looking for (useful stacktraces) would actually come from switching the conjure client from Async->Blocking?  I have a PR up [here](https://pl.ntr/2zT) with discussion on that, and I'd appreciate your thoughts there.

pkoenig10: > Do you think the fix I'm looking for (useful stacktraces) would actually come from switching the conjure client from Async->Blocking?

That would fix the immediate problem, but isn't strictly necessary. There's nothing about this usage that is specific to Dialogue clients - the same thing would happen with any future task whose failure you are logging.

While we do generally recommend using blocking clients with thread pools, there are cases where you want to use async clients.

Instead of switching the clients, I'd recommend refactoring this code to capture the caller's stacktrace at the point you consume the future. Concretely, instead of logging the failure using `exceptionally(...)`, do the logging at the point where you consume the future using `get()` and attach the caller's stacktrace at that point. For example, with a utility like this:

```java
public static <T> T get(Future<T> future) {
    try {
        return future.get();
    } catch (ExecutionException e) {
        switch (e.getCause()) {
            case RuntimeException runtimeException -> {
                runtimeException.addSuppressed(new SafeRuntimeException("Future failed"));
                throw runtimeException;
            }
            case Error error -> {
                error.addSuppressed(new SafeRuntimeException("Future failed"));
                throw error;
            }
            default -> {
                throw new RuntimeException(e.getCause());
            }
        }
    } catch (InterruptedException e) {
        Thread.currentThread().interrupt();
        throw new RuntimeException(e);
    }
}
```

pkoenig10:
Going to close this out, as it seems like we misdiagnosed the problem that motivated this PR.

I've put up https://github.com/palantir/dialogue/pull/2885 to improve/simplify the way we wrap exceptions here.

PR #2885, opened 2026-02-18
Don't wrap thrown exceptions in blocking client

pkoenig10:
## Before this PR

When handling failure in `DefaultClients`, some types of `RuntimeException` are reconstructed and re-thrown in order to capture the caller's stacktrace. This approach is quite brittle because it requires us to update this implementation every time new capabilities are added to `RemoteException` or we risk failing to propagate meaningful error information.

## After this PR

We record the caller's stacktrace using suppressed exceptions. This is a more standard and robust way to capture this information that doesn't require this code to make any assumptions about the exceptions and ensure that caller's receive the exact same exception instance that was thrown.

See discussion in https://github.com/palantir/dialogue/pull/2881 for more context.

### DIFF
--- a/dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/DefaultClients.java
+++ b/dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/DefaultClients.java
@@ -20,10 +20,6 @@ import com.google.common.net.HttpHeaders;
 import com.google.common.util.concurrent.FutureCallback;
 import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
-import com.palantir.conjure.java.api.errors.AbstractSerializableError;
-import com.palantir.conjure.java.api.errors.RemoteException;
-import com.palantir.conjure.java.api.errors.SerializableErrorProvider;
-import com.palantir.conjure.java.api.errors.UnknownRemoteException;
 import com.palantir.dialogue.Channel;
 import com.palantir.dialogue.Clients;
 import com.palantir.dialogue.Deserializer;
@@ -44,7 +40,6 @@ import com.palantir.logsafe.logger.SafeLogger;
 import com.palantir.logsafe.logger.SafeLoggerFactory;
 import java.io.Closeable;
 import java.io.IOException;
-import java.lang.reflect.Constructor;
 import java.util.Optional;
 import java.util.concurrent.ExecutionException;
 import org.jspecify.annotations.Nullable;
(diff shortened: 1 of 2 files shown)


dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/DefaultClients.java
```diff
@@ -121,99 +116,21 @@ public <T> T block(ListenableFuture<T> future) {
...
-            // TODO(jellis): can consider propagating other relevant exceptions (eg: HttpConnectTimeoutException)
-            // see HttpClientImpl#send(HttpRequest req, BodyHandler<T> responseHandler)
-            if (cause instanceof RemoteException remoteException) {
-                if (cause.getClass() == RemoteException.class) {
-                    throw newRemoteException(remoteException);
-                } else {
-                    // This is a user-defined subclass of RemoteException, so we want to preserve the type. But we also
-                    // want to propagate the current stacktrace. Another (arguably better) option is to add a suppressed
-                    // RuntimeException. In order to preserve backwards compatibility for clients that may inspect the
-                    // cause of a remote exception, we take this approach.
```
pkoenig10, line 133: I'm not sure what backward compatibility we are concerned about here. I would hope we don't have clients that are looking at the _cause_ of the caught exception to make decisions. The current behavior is somewhat weird/unintuitive in that we throw a `RemoteException` whose cause is a basically identical `RemoteException`.

aldexis: iirc we wanted to be conservative here because it was part of a larger change, and we didn't want to break things as part of that change

Specifically, we were targeting full code backwards-compatibility for clients that were catching and using RemoteException when changing the error parameters from toString to json, and using rich exception types instead of a plain RemoteException, and guarding against people doing things like relying on the specific cause chain (yes, it's a terrible idea, but we couldn't guarantee nobody was doing this)

I prefer this approach as well though, so probably worth doing 🤞

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/DefaultClients.java
```diff
@@ -121,99 +116,21 @@ public <T> T block(ListenableFuture<T> future) {
...
             if (cause instanceof RuntimeException runtimeException) {
                 cause.addSuppressed(new SafeRuntimeException("Rethrown by dialogue"));
                 throw runtimeException;
             }
 
-            // This matches the behavior in Futures.getUnchecked(Future)
             if (cause instanceof Error error) {
                 cause.addSuppressed(new SafeRuntimeException("Rethrown by dialogue"));
                 throw error;
             }
```
aldexis, line 130: Is there any reason not to just do
```
cause.addSuppressed(new SafeRuntimeException("Rethrown by dialogue"));
throw cause;
```
for all cases here? 🤔

aldexis (approved):
I'm wondering whether we could/should improve some of the error handling for the async codepaths, but I imagine it's just part of how async clients work?

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/DefaultClients.java
pkoenig10: We can't rethrow a checked exception.

And if the exception is a checked exception, then wrapping exception will contain the callers stacktrace, so there's no need for the suppressed exception.

This is a very standard pattern for propagating failed futures as exceptions. We use this pattern it in a bunch of places.

pkoenig10:
> I'm wondering whether we could/should improve some of the error handling for the async codepaths, but I imagine it's just part of how async clients work?

I'm not sure what you're imagining. The async code path has no knowledge of knowing the caller's stacktrace.

This isn't even particularly well-defined - what is the "caller's stacktrace" for any async request? The only "caller's stacktrace" the async code path could know of is the place the request is submitted, but that's probably not what we want for a couple reasons:
- It would require unconditionally creating an exception and populating a stacktrace for every RPC (most of which will succeed), which is very expensive.
- It's semantically different from the sync behavior, where we capture the caller's stacktrace at the place where the response is "consumed" (although for sync clients this ends up being the same as the place as where the request is submitted).

dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/DefaultClients.java
mpritham: > I would hope we don't have clients that are looking at the cause of the caught exception to make decisions.

I think there are places where clients look at the cause of the thrown exception. E.g. [1](https://pl.ntr/2zZ), [2](https://pl.ntr/2z-).

PR #2926, opened 2026-03-27
Queue timeout

mpritham:
## Current state
When a Dialogue client submits a request, it may be queued waiting for a concurrency permit without a timeout. On the blocking path, this manifests as callers stuck with no way to time out. On the async path, the ListenableFuture returned to the caller doesn’t complete.

There are 2 related concepts in Dialogue:
1. read timeout - applies only to requests on the wire, after they've completed queueing and acquired a permit from concurrency limiters
2. deadlines - propagated from upstream callers, checked in the `DeadlineAdvertisementChannel` which sits below the host-level concurrency limiter in the per-host pipeline.

Neither bounds the time a request spends waiting in the queue. A request waiting for a concurrency permit has not yet reached the wire (so read timeout hasn't started) and has not yet reached `DeadlineAdvertisementChannel` (so deadline expiry isn't checked). 

Under sustained load where all hosts' AIMD limits are saturated, requests can queue for a long time.

## Queue timeouts

It'd be useful to have requests in the queues (channel and endpoint) timeout after a certain duration: clients will have predictable upper bounds on the time spent waiting for a response. 

To be more precise:
- Per-client, disabled by default via Optional<Duration> queueTimeout() in Config
- Proactive eviction using per-request ScheduledFuture tasks on a dedicated single-threaded ScheduledExecutorService
- Shared timeout budget across both queue layers (channel-level and endpoint-level) via a request attachment (QueueTimeoutAttachments)
- Timeout cancelled on dispatch (when a request leaves the queue, its ScheduledFuture is cancelled to avoid wasting work on already-dispatched requests)
- Span and timer eagerly closed on timeout via `IdempotentTimerContext` wrapper (makes `timer.stop()` safe to call multiple times)
- Not retried: SafeRuntimeException is not IOException, so RetryingChannel won't retry
  - if a request is retried (for some other reason: e.g. 429), the request will be re-queued and a fresh queue timeout will be applied
- New metric: dialogue.client.request.queue.timeout (counter, tagged by channel-name)

### Which queues does this apply to?

Dialogue has two layers of queues:

1. Channel queue ([created here](https://github.com/palantir/dialogue/blob/9f84b55cd1d4094b0118e139513634ef4fef16f2/dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueChannel.java#L220)). One per DialogueChannel. Requests wait for any host to have capacity. When a request is dequeud, the request goes through node selection fresh and can land on any host.
2. Endpoint queue ([created here](https://github.com/palantir/dialogue/blob/9f84b55cd1d4094b0118e139513634ef4fef16f2/dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueChannel.java#L276) when concurrency limiting is enabled). One per (host, endpoint) pair. It sits below the host-level concurrency limiter. Requests have already been assigned to a specific host and are waiting for that endpoint's concurrency permit on that host. A request in this queue holds a host-level concurrency permit.

A request is never in both queues simultaneously. When a host-level concurrency limit is acquired, the request can transition to the endpoint queue (if there is no permits available for the endpoint on the host).

The same queue timeout is shared between both queues.

### Usage

Clients can be configured to use a queue timeout as follows:

```java
 DialogueClients.create(servicesConfigBlock)
          .withQueueTimeout(Duration.ofSeconds(10))
          .get(MyService.class);
```

### Downsides
There is added memory usage: one new `ScheduledFuture` is created per enqueued request. (todo: use JOL to find how many bytes exactly).

_Some discussion on resetting the timeout on retry_: When a request is retried, we reset the queue timeout. Arguably this is semantically confusing - the maximum time spent in a queue with a queue timeout can be `retries * queue timeout`. However, I think resetting the queue timeout on retry is favorable: 
- Resetting the timeout on each retry keeps the setting narrowly defined: a bound on the amount of time a single attempt can wait in the queue.
- Cumulative queued time is harder to reason about: a later attempt can fail due to queue time consumed by earlier attempts, possibly in different queues.
  - Also, retries and queue timeouts become coupled: if a user increases the number of retries, they should know to increase the queue timeout as well.

### Test plan

I would like to cut an RC and experiment on a real service to see the behavior in production end-to-end before merging.

==CHANGELOG_MSG==
Add configurable queue timeout to QueuedChannel that proactively fails requests waiting for concurrency permits after a configured duration, preventing unbounded queueing under load.
==CHANGELOG_MSG==


dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
```diff
@@ -190,12 +244,18 @@ Optional<ListenableFuture<Response>> maybeExecute(Endpoint endpoint, Request req
 
         shouldRecordQueueMetrics = true;
 
+        SettableFuture<Response> responseFuture = SettableFuture.create();
+        DetachedSpan span = DetachedSpan.start("Dialogue-request-enqueued");
+        IdempotentTimerContext timer = new IdempotentTimerContext(queuedTime.time());
```
mpritham, line 270: The purpose of creating this thin wrapper around the timer is so that we are able to call `.stop()` both when the timeout task runs and when `isDone()` is checked in `scheduleNextTask`. Some alternatives I considered were to have `timer.stop()` run only once by either 1) adding a direct callback to the request future which closes the span and stops the timer removing the logic in `isDone` that does the same thing, 2) attempting to read the exception on the future in `isDone` before calling `stop()` on the timer.

(1) does not work: the time measured will include the time on the wire which we do not want.
(2) seems clunky: we will need to call `get()` on the future and catch an exception which we probably do not want to do on the hot path here.

dialogue-core/src/main/java/com/palantir/dialogue/core/Config.java
```diff
@@ -86,6 +86,16 @@ default Optional<Boolean> deadlineEnforcement() {
         return Optional.empty();
     }
 
+    @Value.Default
+    default Optional<Duration> queueTimeout() {
+        return Optional.empty();
+    }
+
+    @Value.Default
+    default ScheduledExecutorService queueTimeoutScheduler() {
```
parker89: Why's this in Config.java? Could it instead be in QueuedChannel.java?

parker89: Then we could only instantiate it if `queueTimeout` is non-empty

dialogue-core/src/main/java/com/palantir/dialogue/core/QueueTimeoutAttachments.java
```diff
@@ -0,0 +1,47 @@
...
+    static long setExpirationIfAbsent(Request request, long expirationNanos) {
+        Long existing = request.attachments().getOrDefault(QUEUE_EXPIRATION_NANOS, null);
+        if (existing != null) {
+            return existing;
+        }
+        request.attachments().put(QUEUE_EXPIRATION_NANOS, expirationNanos);
+        return expirationNanos;
+    }
+
+    @com.google.common.annotations.VisibleForTesting
```
parker89: Can we import these 2 instead of fully qualifying

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
```diff
@@ -190,12 +246,18 @@ Optional<ListenableFuture<Response>> maybeExecute(Endpoint endpoint, Request req
...
+        @Nullable ScheduledFuture<?> timeoutFuture = scheduleQueueTimeout(request, responseFuture, span, timer);
```
parker89: Should we add a 
```java
    if (responseFuture.isDone()) {
        return Optional.of(responseFuture);
    }
```
immediately after running `scheduleQueueTimeout`? Would prevent it from creating a DeferredCall if the request has already timed out (eg due to having already gone through the first queue)

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
```diff
@@ -392,6 +392,8 @@ private ListenableFuture<Response> incrementFailuresAndMaybeRetry(
         @SuppressWarnings({"FutureReturnValueIgnored", "CheckReturnValue"})
         private ListenableFuture<Response> scheduleRetry(Meter meter, long backoffNanoseconds) {
             meter.mark();
+            // Clear the queue timeout expiration so the retry gets a fresh queue timeout.
+            QueueTimeoutAttachments.clearExpiration(request);
```
mpritham: I don't love that `RetryingChannel` needs to know about queue timeouts. The alternative approach I thought of was worse:

- when reading the timeout attachment in `QueuedChannel`, consider the `queueType` to decide whether to overwrite or read the existing expiration. 
- The channel-level queue and sticky queue are "entry points" (retries re-enter them), so they'd always stamp fresh. The endpoint queue is downstream, so it'd read the remaining budget. 
 
This works but requires `QueuedChannel` to know its role in the pipeline, which is clunky and breaks down if the pipeline ever changes (e.g. a new queue type is added without updating the check).

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
mpritham: Yes, good idea.

dialogue-core/src/main/java/com/palantir/dialogue/core/Config.java
mpritham: Yeah it could be in `QueuedChannel.java`. I was following the same pattern used for the `RetryingChannel.sharedScheduler`. In theory, putting it in the config allows tests to pass in a custom scheduler but I didn't add any such tests that use a custom scheduler.

No strong feeling here. Will update.

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
```diff
@@ -94,29 +114,55 @@ final class QueuedChannel implements Channel {
     // Tracks requests that are current executing in delegate and are not tracked in queuedCalls
     private final AtomicInteger inFlight = new AtomicInteger();
 
+    // The timeout budget is shared across both queue layers via a Request attachment (see QueueTimeoutAttachments).
+    private final OptionalLong queueTimeoutNanos;
```
mpritham, line 130: This is an `OptionalLong` while the queue timeout in the config is `Optional<Duration>`. Duration is user-friendly and using an `OptionalLong` avoids the call to `toNanos()` on duration 1x/queued request.

dialogue-clients/src/main/java/com/palantir/dialogue/clients/AugmentClientConfig.java
```diff
@@ -65,6 +66,8 @@ default TaggedMetricRegistry taggedMetrics() {
 
     Optional<HostEventsSink> hostEventsSink();
 
+    Optional<Duration> queueTimeout();
```
mpritham, line 73: The flow here is a bit confusing and WIP. There are two related changes in other repos: [1](https://github.com/palantir/conjure-java-runtime-api/compare/pm/queue-timeout?expand=1), [2](https://github.com/palantir/conjure-java-runtime/compare/pm/add-queue-timeout?expand=1)

`ReloadingClientFactory.withQueueTimeout(30s)`
  → `params.withQueueTimeout(30s)`  (stored on `ReloadingParams` which extends `AugmentClientConfig`)

When a client is created:
    → `ChannelCache` line 212: `AugmentClientConfig.getClientConf(serviceConf, request)`
      → `builder.queueTimeout(30s)`  (from augment, line 93)
      → returns `ClientConfiguration` with queueTimeout=30s
    → `ChannelCache` line 178: `ClientConfiguration.builder().from(apacheClient.conf())`
      → copies queueTimeout=30s into the new `ClientConfiguration`
    → `DialogueChannel.builder().clientConfiguration(conf)`
      → `Config.rawConfig()` = conf
      → `Config.queueTimeout()` = `rawConfig().queueTimeout()` = `Optional.of(30s)`
    → `QueuedChannel.create(cf, delegate)`
      → `cf.queueTimeout()` = `Optional.of(30s)`
      → timeout enabled

dialogue-core/src/main/java/com/palantir/dialogue/core/QueueTimeoutAttachments.java
```diff
@@ -0,0 +1,59 @@
...
+
+    private QueueTimeoutAttachments() {}
+
+    /**
+     * Clears the expiration attachment by setting it to a sentinel value that {@link #setExpirationIfAbsent} treats as
+     * absent.
+     */
+    static void clearExpiration(Request request) {
+        request.attachments().put(QUEUE_EXPIRATION_NANOS, CLEARED_SENTINEL);
+    }
```
aldexis: Why are we not allowing to remove attachments? It seems like this would make this code simpler

dialogue-core/src/main/java/com/palantir/dialogue/core/QueueTimeoutAttachments.java
```diff
@@ -0,0 +1,59 @@
...
+
+    /** Sets the expiration only if one hasn't been set yet (or was cleared). */
+    static long setExpirationIfAbsent(Request request, long expirationNanos) {
+        Long existing = request.attachments().getOrDefault(QUEUE_EXPIRATION_NANOS, null);
+        if (existing != null && existing != CLEARED_SENTINEL) {
+            return existing;
+        }
+        request.attachments().put(QUEUE_EXPIRATION_NANOS, expirationNanos);
```
aldexis: There is technically a race condition here where the attachment would have been updated in between. This is probably not a huge deal, but I'm wondering whether we should add `putIfAbsent` to `RequestAttachments` and `Attachments` (since it's ultimately backed by a `ConcurrentHashMap`)

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
```diff
@@ -261,10 +384,12 @@ private boolean scheduleNextTask() {
...
-        // If the future has been completed (most likely via cancel) the call should not be queued.
+        // If the future has been completed (most likely via cancel or queue timeout) the call should
+        // not be dispatched. Cancel the timeout task to free the ScheduledFuture from the heap.
         // There's a race where cancel may be invoked between this check and execution, but the scheduled
         // request will be quickly cancelled in that case.
         if (queueHead.response().isDone()) {
+            cancelTimeoutFuture(queueHead);
             decrementQueueSize();
             queueHead.span().complete(QueuedChannelTagTranslator.INSTANCE, this);
             queueHead.timer().stop();
```
aldexis: There are 4 places that seem to all do this same block, or something very close to it. Would it make sense to extract it to a dedicated method?

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
```diff
@@ -282,6 +407,8 @@ private boolean scheduleTaskFromQueue(DeferredCall queueHead) {
                     delegate.maybeExecute(endpoint, queueHead.request(), limitEnforcement);
 
             if (maybeResponse.isPresent()) {
+                // Request is leaving the queue — cancel the timeout task.
+                cancelTimeoutFuture(queueHead);
```
aldexis: You mention it is possible to have a race condition where we dispatch the request, but the response is discarded. Can't we cancel the timeout future _before_ dispatching?
Since the timeout is for bounding the queued duration, it would make sense to order operations such as:
- Pop the queue head
- Cancel the queue head timeout
- Check if it was done already
That way, it can't be timed out after we have already checked (and since we popped it from the queue, it shouldn't accept a timeout anymore at that point)

dialogue-core/src/main/java/com/palantir/dialogue/core/QueueTimeoutAttachments.java
mpritham: Opened #2957

dialogue-core/src/main/java/com/palantir/dialogue/core/QueueTimeoutAttachments.java
mpritham: Opened #2957. Initially I had a slight preference for not supporting `remove` because it would force every user of `Attachments` to reason about the case where an attachment is removed, but I don't think this is a huge deal.

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
mpritham: > You mention it is possible to have a race condition where we dispatch the request, but the response is discarded.

Yeah. This happens because the current ordering is `isDone()` -> `delegate.execute` -> cancel timeout if `delegate.execute` successfully got a concurrency permit (otherwise re-queue the request with the timeout unchanged)

The race is when the timeout runs while `delegate.execute` is running but has not returned: the blocking logic that runs here is the acquisition of a concurrency permit (a CAS):

https://github.com/palantir/dialogue/blob/2cd59db9dd6c63c7d8f9a25bf095944fb9ea2aff/dialogue-core/src/main/java/com/palantir/dialogue/core/ConcurrencyLimitedChannel.java#L91-L105

FWIW I think this race will be rare.

> Can't we cancel the timeout future before dispatching?

Yes good idea. I agree this better represents that the queue timeout is strictly counted against the time the request is in the queue. Updated the logic.

I think one trade-off here is that, when the delegate responds with `Optional.empty` (i.e. we are not able to get a concurrency permit), we need to re-queue the request (and re-schedule the same timeout). The way I've written it requires an allocation of a `DeferredCall` each time `delegate.execute` returns `Optional.empty`. This code is also a bit more complex.

I think it's fine to have the extra short-lived allocation + complexity here for the sake of correctness.

dialogue-core/src/main/java/com/palantir/dialogue/core/QueueTimeoutAttachments.java
aldexis: Are there places we'd need to update right now, so they don't break in the future for a removed attachment?

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
aldexis: I agree that correctness seems worth it, though I didn't realize the race condition was only about acquiring the permit, and not waiting for the response.

dialogue-core/src/main/java/com/palantir/dialogue/core/QueueTimeoutAttachments.java
mpritham: What do you mean? None of the other attachments are ever removed. In the future, if they are, the author will need to reason about the impact of the removal and audit the pipeline. We don't need to think through that/update anything now.

aldexis: I was thinking about places that may consider absent as "never present" rather than "potentially removed", and whether that may be a concern, but I'm fine putting the onus on the person starting to remove attachments

dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueChannel.java
```diff
@@ -75,6 +82,11 @@ public Supplier<Channel> stickyChannels() {
         return stickyChannelSupplier;
     }
 
+    @VisibleForTesting
+    QueuedChannel getMultiHostQueuedChannelForTesting() {
+        return multiHostQueuedChannel;
+    }
```
aldexis: It sounds like we're only using this to get access to the queue size in tests. Why are we not relying on the metrics instead?

dialogue-core/src/test/java/com/palantir/dialogue/core/QueueTimeoutEteTest.java
```diff
@@ -0,0 +1,315 @@
...
+            // Wait for timeout
+            Thread.sleep(QUEUE_TIMEOUT.toMillis() + 100);
+
+            for (int i = 0; i < 5; i++) {
+                assertThat(queued[i]).isDone();
+                assertThat(queued[i])
+                        .failsWithin(Duration.ZERO)
+                        .withThrowableThat()
+                        .withMessageContaining("queue timeout");
+            }
```
aldexis: Isn't there a race condition here which could make the test flaky? How about waiting up to twice the timeout using awaitility (so we don't wait longer than needed, but always wait enough)?

dialogue-core/src/test/java/com/palantir/dialogue/core/QueueTimeoutEteTest.java
```diff
@@ -0,0 +1,315 @@
...
+
+            // Trigger scheduleNextTask to clean up queue entries
+            wire.completeOldest();
+            // The queue should now be empty because scheduleNextTask will clean up the queue entries that have timed
+            // out
+            assertThat(outerQueue.getQueueSizeForTesting()).isEqualTo(0);
```
aldexis: This made me realize that the queue size metrics will look larger than necessary when there are expired requests in there.

Do you think it would make sense to proactively decrement the queue size counter when we timeout? (and check when we pop whether it was already removed)

dialogue-core/src/test/java/com/palantir/dialogue/core/QueueTimeoutEteTest.java
```diff
@@ -0,0 +1,315 @@
...
+class QueueTimeoutEteTest {
+
+    private static final UserAgent USER_AGENT = UserAgent.of(UserAgent.Agent.of("foo", "1.0.0"));
+    private static final SslConfiguration SSL_CONFIG = SslConfiguration.of(
+            Paths.get("src/test/resources/trustStore.jks"), Paths.get("src/test/resources/keyStore.jks"), "keystore");
+    private static final Endpoint ENDPOINT = TestEndpoint.POST;
+    private static final String HOST_A = "http://hostA:8080";
+    private static final String HOST_B = "http://hostB:8080";
+    private static final String HOST_C = "http://hostC:8080";
+    private static final Duration QUEUE_TIMEOUT = Duration.ofSeconds(5);
```
aldexis: 5 seconds feels a bit long if it's going to be applied to a number of the below tests 🤔 Maybe one or two seconds?

dialogue-core/src/test/java/com/palantir/dialogue/core/QueueTimeoutEteTest.java
```diff
@@ -0,0 +1,315 @@
...
+                    .as("Queued request should have been dispatched")
+                    .hasSize(20);
+
+            // Wait past the original timeout
+            Thread.sleep(QUEUE_TIMEOUT.toMillis() + 100);
+
+            // The dispatched request should NOT have timed out — timeout was cancelled on dispatch
+            assertThat(queued)
+                    .as("Timeout should have been cancelled on dispatch")
+                    .isNotDone();
```
aldexis: For the same reason as above, this _could_ pass even if the timeout wasn't cancelled, since it may end up taking a bit longer. Unfortunately, I'm not sure how to have better guarantees. I'd probably just add a bit more delay past the expected timeout (and maybe reduce the timeout as well?)

dialogue-core/src/test/java/com/palantir/dialogue/core/QueueTimeoutEteTest.java
```diff
@@ -0,0 +1,315 @@
...
+    }
+
+    @Nested
+    class EndpointQueueTimeout {
+
+        @Test
+        void request_times_out_in_endpoint_queue() throws Exception {
+            // Single host, drive endpoint AIMD limit down with 429s
+            ConcurrentLinkedQueue<SettableFuture<Response>> pending = new ConcurrentLinkedQueue<>();
+            boolean[] returnImmediate429 = {true};
```
aldexis: nit: I've seen AtomicBoolean generally used rather than a boolean array, for when we need a mutable boolean. Don't have strong feelings - just seemed unusual

dialogue-core/src/test/java/com/palantir/dialogue/core/QueueTimeoutEteTest.java
```diff
@@ -0,0 +1,315 @@
...
+            DialogueChannel channel = DialogueChannel.builder()
+                    .channelName("cancel-on-dispatch")
+                    .clientConfiguration(singleHostConfig(HOST_A))
+                    .factory(_args -> wire)
+                    .maxQueueSize(100)
+                    .queueTimeout(QUEUE_TIMEOUT)
+                    .build();
+
+            // Saturate the host-level limit (AIMD starts at 20)
+            for (int i = 0; i < 20; i++) {
```
aldexis: For all these tests, can we maybe get the actual values from the `CautiousIncreaseAggressiveDecreaseConcurrencyLimiter` class? If we ever change them, it's likely going to be confusing to debug issues in these tests

dialogue-core/src/test/java/com/palantir/dialogue/core/QueueTimeoutEteTest.java
```diff
@@ -0,0 +1,315 @@
...
+            DialogueChannel channel = DialogueChannel.builder()
+                    .channelName("endpoint-timeout")
+                    .clientConfiguration(singleHostConfig(HOST_A))
+                    .factory(_args -> wire)
+                    .maxQueueSize(100)
+                    .queueTimeout(QUEUE_TIMEOUT)
+                    .build();
+
+            // Drive endpoint AIMD limit down to 1
+            for (int i = 0; i < 50; i++) {
```
aldexis: Why 50? Is that just a large number, or is there a specific reason this value works better than others?

dialogue-core/src/test/java/com/palantir/dialogue/core/QueueTimeoutEteTest.java
```diff
@@ -0,0 +1,315 @@
...
+    class EndpointQueueTimeout {
+
+        @Test
+        void request_times_out_in_endpoint_queue() throws Exception {
+            // Single host, drive endpoint AIMD limit down with 429s
+            ConcurrentLinkedQueue<SettableFuture<Response>> pending = new ConcurrentLinkedQueue<>();
+            boolean[] returnImmediate429 = {true};
+            Channel wire = (_endpoint, _request) -> {
+                if (returnImmediate429[0]) {
+                    return Futures.immediateFuture(new TestResponse().code(429));
```
aldexis, line 180: I can never remember - does this only decrease the endpoint limit and not the host limit?

dialogue-core/src/test/java/com/palantir/dialogue/core/QueueTimeoutEteTest.java
```diff
@@ -0,0 +1,315 @@
...
+            ListenableFuture<Response> request1 =
+                    channel.execute(ENDPOINT, Request.builder().build());
+            assertThat(request1).isNotDone();
+            assertThat(pending).hasSize(1);
+
+            // Second request: endpoint limiter full
+            ListenableFuture<Response> request2 =
+                    channel.execute(ENDPOINT, Request.builder().build());
+            assertThat(request2).isNotDone();
+            assertThat(pending).as("request2 is in endpoint queue, not on wire").hasSize(1);
```
aldexis, line 215: Technically, this only checks that the request is queued, but doesn't confirm that it's not queued in the host queue, right? Could we check the queue metric to confirm it's in the endpoint queue/not in the host queue?

dialogue-core/src/test/java/com/palantir/dialogue/core/QueueTimeoutEteTest.java
```diff
@@ -0,0 +1,315 @@
...
+            Thread.sleep(QUEUE_TIMEOUT.toMillis() + 100);
+
+            assertThat(queued).isDone();
+            assertThat(queued).failsWithin(Duration.ZERO).withThrowableThat().withMessageContaining("queue timeout");
+
+            // The wire should NOT have been hit again — the timeout exception is
+            // SafeRuntimeException (not IOException), so RetryingChannel doesn't retry
+            assertThat(wireHitCount.get())
+                    .as("Wire should not be hit after timeout — timeout is not retried")
+                    .isEqualTo(wireHitsAfterSaturation);
```
aldexis: _If_ we were retrying on timeout, wouldn't the request be queued anyway, and thus not change the wireHitCount anyway? 🤔 

Shouldn't you instead finish all pending requests and verify there is no more request making it to the wire?

dialogue-core/src/test/java/com/palantir/dialogue/core/RetryingChannelTest.java
```diff
@@ -1156,6 +1157,45 @@ public void close() {}
...
+                // First attempt: expiration is set
+                assertThat(QueueTimeoutAttachments.getExpiration(request))
+                        .as("First attempt should see the expiration")
+                        .isEqualTo(12345L);
+                return Futures.immediateFuture(mockResponse(429));
+            }
+            // Second attempt: expiration should have been cleared by RetryingChannel
+            assertThat(QueueTimeoutAttachments.getExpiration(request))
+                    .as("RetryingChannel should clear the expiration before retrying")
+                    .isNull();
```
aldexis: I'm not clear on the expected behavior here. We're still going to have a queue timeout upon retries, correct? (and it's going to be reset to the original value)

dialogue-core/src/test/java/com/palantir/dialogue/core/QueuedChannelTest.java
```diff
@@ -550,4 +561,385 @@ public void testInitialRequestIsIllegallyLimited_queuedRequest() {
...
+        private ScheduledExecutorService scheduler;
+
+        @BeforeEach
+        void beforeEach() {
+            ticker = new ManualTicker();
+            request = Request.builder().build();
+            delegate = new ToggleableDelegate();
+            // Scheduler is required when timeout is configured. The 10-hour timeout means scheduled tasks won't fire
+            // during the test, and we call failWithQueueTimeout directly.
+            scheduler = Executors.newSingleThreadScheduledExecutor();
```
aldexis: Isn't there a better executor service that you can control granularly? (e.g. something made specifically for tests)
This would also avoid having to fake the timeout behavior with the below `simulateTimeout` 

I see `DeterministicScheduler` from jmock used in various places (at least internally)

dialogue-core/src/test/java/com/palantir/dialogue/core/QueuedChannelTest.java
```diff
@@ -550,4 +561,385 @@ public void testInitialRequestIsIllegallyLimited_queuedRequest() {
...
+            SettableFuture<Response> wireFuture = delegate.lastDispatched();
+            assertThat(wireFuture).isNotNull();
+
+            // callerFuture is now failed with timeout, but ForwardAndSchedule will
+            // close the response when the wire completes
+            TestResponse wireResponse = new TestResponse().code(200);
+            wireFuture.set(wireResponse);
+            assertThat(wireResponse.isClosed())
+                    .as("Response must be closed since caller future was already failed")
+                    .isTrue();
```
aldexis, line 768: I'm confused at how this is behaving. If we are dispatching the request first, then shouldn't `callerFuture is now failed with timeout` be false?

dialogue-core/src/test/java/com/palantir/dialogue/core/QueuedChannelTest.java
```diff
@@ -550,4 +561,385 @@ public void testInitialRequestIsIllegallyLimited_queuedRequest() {
...
+
+            ListenableFuture<Response> callerFuture = queuedChannel.execute(TestEndpoint.POST, request);
+            assertThat(instrumentation.requestsQueued().getCount()).isEqualTo(1);
+
+            // schedule() is called ->
+            // The delegate rejects request ->
+            // assert that the request is re-queued
+            queuedChannel.schedule();
+            assertThat(instrumentation.requestsQueued().getCount()).isEqualTo(1);
+            assertThat(callerFuture).isNotDone();
```
aldexis, line 809: Should we have some test that verifies the eventual timeout stays the same in this scenario? (i.e. we aren't resetting the queue timeout)

dialogue-core/src/test/java/com/palantir/dialogue/core/QueuedChannelTest.java
```diff
@@ -550,4 +561,385 @@ public void testInitialRequestIsIllegallyLimited_queuedRequest() {
...
+            assertThat(future2).as("Second request should still be alive").isNotDone();
+
+            // scheduleNextTask should skip future1 and dispatch future2
+            delegate.setAccepting(true);
+            queuedChannel.schedule();
+
+            assertThat(instrumentation.requestsQueued().getCount()).isEqualTo(0);
+            assertThat(future2)
+                    .as("Second request should have been dispatched, not timed out")
+                    .isNotDone();
```
aldexis: Is there a way we could test that future2 is actually in-flight, rather than just not done? We're testing the same state for future2 as before calling `queuedChannel.schedule();`

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
aldexis: That's probably due to how complex our layering of the various channels is. Having multiple queues doesn't help either. I agree that this is probably the simplest, though it may be interesting to have an ete test for this behavior, as this seems easy to break through a refactor

aldexis (approved):
Have various comments and questions, but approving for RC

parker89 (dismissed):

dialogue-clients/src/main/java/com/palantir/dialogue/clients/ReloadingClientFactory.java
```diff
@@ -393,6 +393,11 @@ public DialogueClients.ReloadingFactory withHostEventsSink(HostEventsSink value)
         return new ReloadingClientFactory(params.withHostEventsSink(value), cache);
     }
 
+    @Override
+    public DialogueClients.ReloadingFactory withQueueTimeout(Duration duration) {
+        return new ReloadingClientFactory(params.withQueueTimeout(duration), cache);
+    }
```
aldexis, line 413: There are so many different configs in different classes that it becomes easily confusing. Could you clarify (also in the PR description) what are the expected ways for users to configure the timeout? As far as I can tell, this is the external entrypoint? (and we'd use this internally to allow configuring all the clients used by a service with a single config?)

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
aldexis: @mpritham Just so I'm sure I understand: when a request gets retried, we're going to reset the entire queue timeout, correct? (rather than e.g. using the "left-over" queue timeout from previous queued attempts)

I think we should document this behavior somewhere, as it's not necessarily obvious unless someone is reading the code (and even then, considering the complexity of the dialogue codebase)

aldexis: @mpritham A thought I had: how about we add a callback parameter to `RetryingChannel`, which triggers upon retries? This way, we could instead have `DialogueChannel` pass it in https://github.com/palantir/dialogue/blob/babb46f37261edef140364040aa3a9757b83c1c0/dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueChannel.java#L316-L336

It's also where it makes the most sense to put it, because `DialogueChannel` is the one responsible for the ordering of the various channel layers, and thus it's what defines that you want to clear the expiration upon retrying. This should be a fairly simple change, and allows `RetryingChannel` not to have to know about queue timeouts, or how/where it's used

dialogue-core/src/main/java/com/palantir/dialogue/core/StickyEndpointChannels2.java
```diff
@@ -63,21 +67,38 @@ private static final class QueueOverrideSupplier implements Supplier<Channel> {
...
             this.maxQueueSize = cf.maxQueueSize();
             this.queuedChannelInstrumentation = QueuedChannel.stickyInstrumentation(
                     DialogueClientMetrics.of(cf.clientConf().taggedMetricRegistry()), channelName);
             this.nodeSelectionChannel = nodeSelectionChannel;
+            this.queueTimeoutNanos = cf.queueTimeout()
+                    .map(Duration::toNanos)
+                    .map(OptionalLong::of)
+                    .orElseGet(OptionalLong::empty);
+            this.clock = cf.ticker();
+            this.scheduler = queueTimeoutNanos.isPresent() ? QueuedChannel.sharedTimeoutScheduler.get() : null;
```
aldexis: Since this is all private anyway, is there any reason not to keep only the `queueTimeoutNanos` field and derive the scheduler in `QueuedChannel.createForSticky` like we're already doing for `QueuedChannel.create`?

(separately, I'll admit I'm confused why we don't simply keep a reference to `Config` and pass it down to `QueuedChannel.createForSticky`, which would allow to share more of the logic with `QueuedChannel.create`)

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
aldexis: Also, do we have a test somewhere that confirms the behavior of channels created by `DialogueChannel` is correct with regards to the interaction between retries and queue timeouts? We should probably have tests in `DialogueChannelTest`

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
```diff
@@ -69,6 +79,17 @@
  */
 final class QueuedChannel implements Channel {
     private static final SafeLogger log = SafeLoggerFactory.get(QueuedChannel.class);
+    private static final String TIMEOUT_SCHEDULER_NAME = "dialogue-QueuedChannel-timeout-scheduler";
+
+    @SuppressWarnings("deprecation") // Singleton registry for a singleton executor
+    static final Supplier<ScheduledExecutorService> sharedTimeoutScheduler =
+            Suppliers.memoize(() -> DialogueExecutors.newSharedSingleThreadScheduler(MetricRegistries.instrument(
```
aldexis, line 86: Since we're using a single thread, have you thought about the risk of contention if we have lots of clients and requests being queued and expired? I don't think this is likely to be problematic, since I imagine each task should be particularly fast (I imagine something on the order of milliseconds, so we should be able to handle tens of thousands of expiring requests per second?), but I haven't tested this theory.

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
```diff
@@ -94,24 +116,55 @@ final class QueuedChannel implements Channel {
...
     private final AtomicInteger inFlight = new AtomicInteger();
 
+    // The timeout budget is shared across both queue layers via a Request attachment (see QueueTimeoutAttachments).
+    private final OptionalLong queueTimeoutNanos;
+    private final Ticker clock;
+
+    @Nullable
+    private final ScheduledExecutorService scheduler;
+
     QueuedChannel(
```
aldexis, line 137: nit: Seems like we only keep this for tests, so we could potentially mark it as such, or even just remove it entirely and have a util in the test class?
```suggestion
    @VisibleForTesting
    QueuedChannel(
```

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
```diff
@@ -124,21 +177,39 @@ static QueuedChannel createForSticky(
...
+                queuedChannelInstrumentation,
+                maxQueueSize,
+                queueTimeoutNanos,
+                clock,
+                scheduler);
     }
 
     static QueuedChannel create(Config cf, LimitedChannel delegate) {
+        OptionalLong timeoutNanos =
+                cf.queueTimeout().map(Duration::toNanos).map(OptionalLong::of).orElseGet(OptionalLong::empty);
```
aldexis: nit: We could potentially pass in `Optional<Duration>` in the constructor and have the constructor itself compute the `timeoutNanos`, as well as determine the scheduler (which could delegate to a second constructor, used only for testing).
This would avoid having to duplicate logic in multiple places (twice here in the two create methods, and one in `StickyEndpointChannels2`)

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
```diff
@@ -190,12 +264,35 @@ Optional<ListenableFuture<Response>> maybeExecute(Endpoint endpoint, Request req
...
+        if (responseFuture.isDone()) {
+            return Optional.of(responseFuture);
+        }
+
+        // The queue-timeout expiration is stamped on a Request attachment and shared across queue layers for the
+        // duration of a single execution. Clear it once this execution terminates (success, failure, timeout, or
+        // cancellation). This is mostly defensive: we don't expect requests to be re-used.
+        if (queueTimeoutNanos.isPresent()) {
+            DialogueFutures.addDirectListener(responseFuture, () -> QueueTimeoutAttachments.clearExpiration(request));
+        }
```
aldexis, line 290: Wouldn't this actually cover the retry usecase? 🤔

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
```diff
@@ -344,6 +475,86 @@ private boolean scheduleTaskFromQueue(DeferredCall queueHead) {
...
+        if (expirationNanos == null) {
+            return null;
+        }
+        long delayNanos = expirationNanos - clock.read();
+        if (delayNanos <= 0) {
+            // The timeout is already reached, so we fail immediately. We don't decrement the queue size here, because
+            // in both cases that reach this branch the count is already correct: on initial enqueue the entry has not
+            // been counted yet (maybeExecute returns early, before incrementing), and on re-queue the entry is still
+            // counted and gets decremented when a drain next pops it or when the timeout is hit.
+            failWithQueueTimeout(responseFuture, span, timer);
```
aldexis: This feels pretty brittle to be honest 🤔 Could we maybe instead of the `AtomicBoolean queueSizeDecremented` have a _slightly_ more involved mechanism which effectively does "decrement the queue size if and only if it has been incremented, and not already decremented"? (e.g. a `QueueCounter` or something, which would contain two volatile booleans) That way, we can simply always call it. This would also potentially allow us to share the logic between `failWithQueueTimeoutAndDecrementQueueSize` and `cleanupDeferredCall` which should effectively be going through the same steps (with the difference that the former needs to set the exception on the response, but both could even take an optional `timeoutFuture` to cancel, where it would be empty in the former case)

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
```diff
@@ -344,6 +475,86 @@ private boolean scheduleTaskFromQueue(DeferredCall queueHead) {
...
+                .from(original)
+                .timeoutFuture(Optional.ofNullable(newTimeout))
+                .build();
+    }
+
+    /**
+     * Schedules a timeout task based on the expiration already stamped on the request attachment.
+     */
+    @Nullable
+    private ScheduledFuture<?> scheduleTimeoutFromExpiration(
```
aldexis: nit: This should live just below `scheduleQueueTimeout`, to avoid jumping back and forth between code

dialogue-core/src/test/java/com/palantir/dialogue/core/RetryingChannelTest.java
```diff
@@ -1156,6 +1157,48 @@ public void close() {}
                 .build();
     }
 
+    @Test
+    void retry_clears_queue_timeout_attachment() throws Exception {
```
aldexis: Mentioned in other comments, but I think this is more a factor of the `DialogueChannel` implementation, rather than `RetryingChannel` and we should probably architect it as such, with a callback mechanism (and this should test the callback)

dialogue-core/src/test/java/com/palantir/dialogue/core/QueueTimeoutEteTest.java
aldexis: Ping on this (`can we maybe get the actual values from the CautiousIncreaseAggressiveDecreaseConcurrencyLimiter class`)

dialogue-core/src/test/java/com/palantir/dialogue/core/QueueTimeoutEteTest.java
```diff
@@ -0,0 +1,357 @@
...
+            Awaitility.waitAtMost(QUEUE_TIMEOUT.multipliedBy(2)).untilAsserted(() -> {
+                for (ListenableFuture<Response> future : queued) {
+                    assertThat(future).isDone();
+                }
+            });
+            for (ListenableFuture<Response> future : queued) {
+                assertThat(future)
+                        .failsWithin(Duration.ZERO)
+                        .withThrowableThat()
+                        .withMessageContaining("queue timeout");
```
aldexis: nit: thoughts on making this a custom `SafeQueueTimeoutException`?

dialogue-core/src/test/java/com/palantir/dialogue/core/QueueTimeoutEteTest.java
```diff
@@ -0,0 +1,357 @@
...
+            }
+
+            // The queue size is decremented proactively when each timeout fires, without waiting for the next
+            // scheduleNextTask drain to pop the timed-out entries.
+            Awaitility.waitAtMost(QUEUE_TIMEOUT.multipliedBy(2))
+                    .untilAsserted(() -> assertThat(queueSize.getCount()).isEqualTo(0));
```
aldexis: Why are we waiting again if the queue should be proactively decremented?

dialogue-core/src/test/java/com/palantir/dialogue/core/QueueTimeoutEteTest.java
```diff
@@ -0,0 +1,357 @@
...
+                    .queueTimeout(QUEUE_TIMEOUT)
+                    .build();
+
+            // Drive the endpoint AIMD limit down to its floor of 1. Each non-custom 429 multiplies the limit by 0.9,
+            // so dropping from the initial 20 to 1 takes ~28 steps; 50 is comfortably more than enough.
+            for (int i = 0; i < 50; i++) {
+                ListenableFuture<Response> resp =
+                        channel.execute(ENDPOINT, Request.builder().build());
+                assertThat(resp).isDone();
+            }
```
aldexis: If we're going to refactor how AIMD works, this may end up breaking. Would it be possible to confirm we are in the expected state? I guess this is validated by the below logic, which verifies requests are queued after the first?

dialogue-core/src/test/java/com/palantir/dialogue/core/QueueTimeoutEteTest.java
```diff
@@ -0,0 +1,357 @@
...
+                    .isEqualTo(1);
+
+            // Wait for timeout, request2 should be failed
+            Awaitility.waitAtMost(QUEUE_TIMEOUT.multipliedBy(2))
+                    .untilAsserted(() -> assertThat(request2).isDone());
+            assertThat(request2).failsWithin(Duration.ZERO).withThrowableThat().withMessageContaining("queue timeout");
+
+            // The endpoint queue size is decremented proactively when the timeout fires.
+            Awaitility.waitAtMost(QUEUE_TIMEOUT.multipliedBy(2))
+                    .untilAsserted(() -> assertThat(endpointQueued.getCount()).isEqualTo(0));
```
aldexis: Same question regarding waiting when the comment claims it's done proactively

dialogue-core/src/test/java/com/palantir/dialogue/core/QueueTimeoutEteTest.java
```diff
@@ -0,0 +1,357 @@
...
+            // Queue a request — it will time out while every permit is held.
+            ListenableFuture<Response> queued =
+                    channel.execute(ENDPOINT, Request.builder().build());
+
+            // The queue-timeout failure is a SafeRuntimeException (not an IOException or QoS response), so
+            // RetryingChannel does not retry it: the request fails outright. If it were retried, this future would
+            // instead stay pending while the retried attempt re-queued, and this await would time out.
+            Awaitility.waitAtMost(QUEUE_TIMEOUT.multipliedBy(2))
+                    .untilAsserted(() -> assertThat(queued).isDone());
+            assertThat(queued).failsWithin(Duration.ZERO).withThrowableThat().withMessageContaining("queue timeout");
```
aldexis: nit: we could potentially also observe the retry metrics to confirm we aren't retrying?

dialogue-core/src/test/java/com/palantir/dialogue/core/QueuedChannelTest.java
```diff
@@ -550,4 +559,480 @@ public void testInitialRequestIsIllegallyLimited_queuedRequest() {
...
+                    .withThrowableThat()
+                    .withMessageContaining("queue timeout");
+            assertThat(instrumentation.requestsQueued().getCount())
+                    .as("the timeout task proactively decrements the queue size")
+                    .isEqualTo(0);
+
+            // The deque still holds the timed-out entry; the next drain pops it
+            delegate.setAccepting(true);
+            queuedChannel.schedule();
+            assertThat(instrumentation.requestsQueued().getCount()).isEqualTo(0);
```
aldexis: We don't seem to be observing any difference here (the count was already 0) - how could we confirm something did happen?

dialogue-core/src/test/java/com/palantir/dialogue/core/QueuedChannelTest.java
```diff
@@ -550,4 +559,480 @@ public void testInitialRequestIsIllegallyLimited_queuedRequest() {
...
+            // Dispatch the queued request, which cancels its timeout task.
+            delegate.setAccepting(true);
+            queuedChannel.schedule();
+
+            // Advance past the timeout. The cancelled task must not fire.
+            scheduler.tick(QUEUE_TIMEOUT_NANOS * 2, TimeUnit.NANOSECONDS);
+
+            assertThat(callerFuture)
+                    .as("dispatch cancelled the timeout, so firing the scheduler is a no-op")
+                    .isNotDone();
```
aldexis, line 680: nit: should we confirm the queue size in this test?

dialogue-core/src/test/java/com/palantir/dialogue/core/QueuedChannelTest.java
```diff
@@ -550,4 +559,480 @@ public void testInitialRequestIsIllegallyLimited_queuedRequest() {
...
+            wire.set(new TestResponse().code(200));
+            assertThat(first).succeedsWithin(Duration.ZERO);
+
+            assertThat(QueueTimeoutAttachments.getExpiration(reused))
+                    .as("expiration is cleared on terminal completion so the request can be safely reused")
+                    .isNull();
+
+            // Reuse the same request for a second execution after the clock has advanced past the old timeout. It must
+            // not inherit the stale deadline (which would make it time out immediately). It gets a new budget instead
+            ticker.advance(Duration.ofNanos(QUEUE_TIMEOUT_NANOS * 2));
```
aldexis, line 722: The ticker advance doesn't actually matter here, correct? (this is for my understanding)

dialogue-core/src/test/java/com/palantir/dialogue/core/QueuedChannelTest.java
```diff
@@ -550,4 +559,480 @@ public void testInitialRequestIsIllegallyLimited_queuedRequest() {
...
+            // Dispatch the queued request.
+            delegate.setAccepting(true);
+            queuedChannel.schedule();
+
+            // Model the lost race: the timeout fires after dispatch. failWithQueueTimeout (invoked directly here,
+            // bypassing the dispatch-time cancel) fails the already-dispatched caller future.
+            simulateTimeout(callerFuture);
+            assertThat(callerFuture)
+                    .as("Timeout won the race, so the caller future is failed")
+                    .isDone();
```
aldexis: Is this even possible? Aren't we cancelling the timeout upon dispatch?

dialogue-core/src/test/java/com/palantir/dialogue/core/QueuedChannelTest.java
```diff
@@ -550,4 +559,480 @@ public void testInitialRequestIsIllegallyLimited_queuedRequest() {
...
+                    rejectingDelegate,
+                    "shared",
+                    "channel",
+                    sharedInstrumentation,
+                    QUEUE_SIZE,
+                    OptionalLong.of(QUEUE_TIMEOUT_NANOS),
+                    ticker,
+                    scheduler);
+            QueuedChannel queue2 = new QueuedChannel(
+                    rejectingDelegate,
```
aldexis, line 853: If we're modelling how it works in the actual case, shouldn't queue2's delegate be queue1 (or vice versa)? I guess it makes it a bit easier to manually control the state of both queues this way? (although we could possibly still do it by confirming the queue sizes?)

dialogue-core/src/test/java/com/palantir/dialogue/core/QueuedChannelTest.java
```diff
@@ -550,4 +559,480 @@ public void testInitialRequestIsIllegallyLimited_queuedRequest() {
...
+                    .as("request was re-queued, not dispatched or failed")
+                    .isNotDone();
+            assertThat(instrumentation.requestsQueued().getCount()).isEqualTo(1);
+            assertThat(QueueTimeoutAttachments.getExpiration(req))
+                    .as("re-queue must preserve the original expiration, not reset the queue-timeout budget")
+                    .isEqualTo(originalExpiration);
+        }
+
+        @Test
+        void endpoint_queue_fails_immediately_when_budget_exhausted() {
```
aldexis: This isn't unique to the endpoint queue, right? This is simply a factor of "if there is already an attached expiration, and it's expired, the queue should fail immediately". Similarly, the previous test about reading from the existing expiration could probably just set an expiration and pass it to a queue, rather than simulate multiple queues (as long as we have another test which confirms that the delegate sees the expiration set on the request)

dialogue-core/src/test/java/com/palantir/dialogue/core/QueuedChannelTest.java
```diff
@@ -550,4 +559,480 @@ public void testInitialRequestIsIllegallyLimited_queuedRequest() {
...
+            // ForwardAndSchedule and complete future2 with the response.
+            delegate.lastDispatched();
+            SettableFuture<Response> future2Wire = delegate.lastDispatched();
+            assertThat(future2Wire)
+                    .as("future2 should have been handed to the delegate")
+                    .isNotNull();
+            future2Wire.set(new TestResponse().code(200));
+            assertThat(future2)
+                    .as("future2 completes via its dispatched wire call, confirming it was dispatched")
+                    .succeedsWithin(Duration.ZERO);
```
aldexis: nit: could we somehow test the `future2` response itself, rather than just testing it succeeds? (e.g. reading its code?)

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
mpritham: Since the queue timeout value is a field on `Config`, I think it's cleaner to just pass in the `Config` and read the value from that.

We can avoid duplication of the scheduler selection logic by having a helper method. Will add that.

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
mpritham: You're right, this does over the retry case. I've removed the logic in `RetryingChannel`, which is nice because now only this class cares about the request attachment.

dialogue-core/src/main/java/com/palantir/dialogue/core/RetryingChannel.java
mpritham: We've removed the logic in this class dealing with queue timeouts so this wouldn't be needed anymore.

I added another test in `QueueTimeoutEteTest#RetryResetsBudget` to check that retries clear the queue timeout.

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
mpritham: Sounds good. Introduced `QueueSizeAccounting` to encapsulate this logic in one place.

dialogue-core/src/test/java/com/palantir/dialogue/core/QueuedChannelTest.java
mpritham: Added `requeue_preserves_original_expiration`

dialogue-core/src/test/java/com/palantir/dialogue/core/QueuedChannelTest.java
mpritham: The race I'm worried about here is:
- `scheduleNextTask` cancels via `queueHead.timeoutFuture().cancel(false)`. 
- If the timeout body starts running concurrently just before the cancel in `scheduleNextTask`, the call to cancel returns false and `failWithQueueTimeout` can fail the already-dispatched caller future.

mpritham (commented):
WIP addressing some of the review comments.

dialogue-core/src/test/java/com/palantir/dialogue/core/QueueTimeoutEteTest.java
mpritham: This was just a large number to ensure that our AIMD limit dropped down to 1. I replaced this (and the other constants) with references to the values (or computations on values) in the CIAD file.

dialogue-core/src/test/java/com/palantir/dialogue/core/QueueTimeoutEteTest.java
mpritham: A 429 only decreases the endpoint limit and does not decrease the host limit.

dialogue-core/src/test/java/com/palantir/dialogue/core/QueueTimeoutEteTest.java
mpritham: Sounds good. Updated the test to ensure that the wireHitCount is only equal to the wireHitCount after saturation (but before making the request that we have timeout)

dialogue-core/src/test/java/com/palantir/dialogue/core/QueuedChannelTest.java
mpritham: [I think this comment is helpful?](https://github.com/palantir/dialogue/pull/2926#discussion_r3537632772)

dialogue-core/src/test/java/com/palantir/dialogue/core/QueuedChannelTest.java
mpritham: Yeah. I think this assertion is a bit useless.

I updated the test to add a new request, dispatch it, and validate that we remove the timed-out request from the queue. This at least validates that the metric is incremented to 1 even when there are 2 entries in the queue (the timed-out entry and the live one).

dialogue-core/src/test/java/com/palantir/dialogue/core/QueuedChannelTest.java
mpritham: In theory, I could've not advanced the ticker here, and just asserted that the expiration saved in the request attachment was `QUEUE_TIMEOUT_NANOS`. I thought advancing the ticker and asserting that the request attachment is `tickerValue + QUEUE_TIMEOUT_NANOS` made it really clear that the saved request attachment was newly computed.

dialogue-core/src/test/java/com/palantir/dialogue/core/QueuedChannelTest.java
mpritham: Sorry this is a bit of a confusing test. What I'm testing here is:
- the timeout is hit and the logic to throw a "queue timeout" exception runs concurrently with `scheduleNextTask`.
- the `cancel` call in `scheduleNextTask` fails (returns false) but the timeout task has not actually cancelled the request yet, so `queueHead.response().isDone()` returns `false` (and so we continue delegating the request)
- the timeout task cancels the request and sets the exception
- the caller future should now be failed with the timeout

dialogue-core/src/test/java/com/palantir/dialogue/core/QueuedChannelTest.java
mpritham: In `QueueTimeoutEteTest#EndpointQueueTimeout`, we use the `DialogueChannel` to run tests which will replicate how the delegation is setup. In this test, I just wanted to validate that the queue will read an existing request attachment when present, so I think it's ok to have them be independent.

dialogue-core/src/test/java/com/palantir/dialogue/core/QueuedChannelTest.java
mpritham: Yes this isn't unique to the endpoint queue. Updated the test.

> Similarly, the previous test about reading from the existing expiration could probably just set an expiration and pass it to a queue

Yep we can do this. I slightly prefer the approach above because we also just ensure that we've stamped the expiration when enqueuing the first time. We definitely do that in other tests too, so no strong opinion.

dialogue-core/src/test/java/com/palantir/dialogue/core/QueueTimeoutEteTest.java
mpritham: We need to wait once (for the timeout to elapse) but not twice as written here. Updated.

dialogue-clients/src/main/java/com/palantir/dialogue/clients/ReloadingClientFactory.java
mpritham: I updated the PR description with a "usage" section.

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
mpritham: [I ran a JMH locally.](https://pl.ntr/2CJ)

There are 2 tests:
- "scheduleAndCancel" which schedules a timeout future then cancels it
- "expiryStorm" which schedules 50k timeouts and "expires" all of them (the timeout task takes in a CountdownLatch which is triggered after scheduling the timeouts to force an expiration). 


```
 ./gradlew :dialogue-jmh:jmh -PjmhOptions='QueueTimeoutSchedulerBenchmark'
```

```
Benchmark                                          Mode  Cnt        Score        Error  Units
QueueTimeoutSchedulerBenchmark.expiryStorm        thrpt   10  1592905.929 ±  29082.606  ops/s
QueueTimeoutSchedulerBenchmark.scheduleAndCancel  thrpt   10  3587984.232 ± 107135.892  ops/s
```

schedule and cancel for 8 threads (` ./gradlew :dialogue-jmh:jmh -PjmhOptions='scheduleAndCancel -t 8'`):
```
Benchmark                                          Mode  Cnt        Score       Error  Units
QueueTimeoutSchedulerBenchmark.scheduleAndCancel  thrpt   10  3821842.201 ± 77757.651  ops/s
```

schedule and cancel for 32 threads (` ./gradlew :dialogue-jmh:jmh -PjmhOptions='scheduleAndCancel -t 8'`):
```
Benchmark                                          Mode  Cnt        Score       Error  Units
QueueTimeoutSchedulerBenchmark.scheduleAndCancel  thrpt   10  3764749.514 ± 29603.978  ops/s
```

From these figures, I don't think this should be an issue.

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
aldexis: `new QueuedChannel(...)` below already extracts a bunch of fields from the `Config` object. I'm not proposing we change the signature of this `create` method, but instead pass in `cf.queueTimeout()` to the `QueuedChannel` constructor and have that constructor convert that to `OptionalLong queueTimeoutNanos`

This way, we can also pass around just the `queueTimeout()` value in e.g. `StickyEndpointChannels2`, as well as the second `create` method.

An alternative would be to actually pass in the entire `Config` object, but this may make less sense and require more changes which don't belong in this already large PR

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
```diff
@@ -251,6 +408,24 @@ private void decrementQueueSize() {
...
 
+    final class QueueSizeAccounting {
+        private final AtomicBoolean incremented = new AtomicBoolean(false);
+        private final AtomicBoolean decremented = new AtomicBoolean(false);
+
+        int incrementAndGet() {
+            int newSize = incrementQueueSize();
+            incremented.set(true);
+            return newSize;
+        }
```
aldexis, line 432: We should probably throw if we already decremented, as this shouldn't happen

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
```diff
@@ -251,6 +408,24 @@ private void decrementQueueSize() {
         queueSizeCounter.get().dec();
     }
 
+    final class QueueSizeAccounting {
+        private final AtomicBoolean incremented = new AtomicBoolean(false);
+        private final AtomicBoolean decremented = new AtomicBoolean(false);
```
aldexis: I'm trying to get more used to working with AtomicX vs volatile, as I can never remember when one is needed vs the other. Would volatile fields here not have been enough? (I'm not sure whether the `compareAndSet` below is actually needed for correctness, as opposed to non-atomic comparison and set)

dialogue-core/src/test/java/com/palantir/dialogue/core/QueuedChannelTest.java
```diff
@@ -550,4 +577,489 @@ public void testInitialRequestIsIllegallyLimited_queuedRequest() {
...
+            assertThat(instrumentation.requestsQueued().getCount())
+                    .as("dispatch dequeued the request")
+                    .isEqualTo(0);
+
+            // Advance past the timeout. The cancelled task must not fire.
+            scheduler.tick(QUEUE_TIMEOUT_NANOS * 2, TimeUnit.NANOSECONDS);
+
+            assertThat(callerFuture)
+                    .as("dispatch cancelled the timeout, so firing the scheduler is a no-op")
+                    .isNotDone();
```
aldexis, line 680: We should possibly also check that when the server responds, we actually see that response, and not some delayed timeout

dialogue-core/src/test/java/com/palantir/dialogue/core/QueuedChannelTest.java
aldexis: Ah I see - I hadn't thought of the race condition where the cancel call returns false. I imagine we could somehow handle it, but I'm not even sure what the correct behavior should be 🤔 (since I guess there could be other reasons we failed to cancel it, and it doesn't mean the timeout will always execute before we get the response either)

aldexis (commented):
Thanks for all the iterations! Approving for RC - if you want to merge this, you'll want to remove the build.gradle changes as well as move to non-RC versions

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
mpritham: For `decremented` at least, using `compareAndSet` avoids this race:
- suppose there is a thread that is in the timeout task (the queue timeout has been triggered) and we're decrementing the metric in `decrementIfCounted`
- suppose there's another thread in the `scheduleNextTask` -> `cleanupDeferredCall` path.
- both can read `false` from a volatile `decremented` and decrement the metric, but with a `compareAndSet` only one can read `false`

For `incremented` we can probably use `volatile` since we don't make use of the `compareAndSet` for correctness.

README.md
```diff
@@ -215,6 +215,33 @@ backpressure in the form of 429 status QoS responses. Permits are decreased afte
...
+a permit becomes available. By default a request stays queued until a permit frees up or its overall deadline / read
+timeout expires. The **queue timeout** places a separate bound on how long a request may wait *in the queue* — distinct
+from the read timeout, which only starts once a request has been dispatched to a host.
+
+When a request exceeds the queue timeout it fails immediately to the caller (it is **not** retried), and the
+`dialogue.client.request.queue.timeout` counter (tagged `channel-name`) is incremented. Time spent waiting in the queue
+is recorded by the `dialogue.client.request.queued.time` timer.
+
+Configure it in code via `withQueueTimeout(Duration)` on the client factory, or per service via the `queue-timeout` field
+of `ServiceConfiguration` (the `queue-timeout` key in service-discovery config). When both are set, the per-service
```
aldexis: I think the config mention is specific to internal?
```suggestion
of `ServiceConfiguration`. When both are set, the per-service
```

README.md
```diff
@@ -215,6 +215,33 @@ backpressure in the form of 429 status QoS responses. Permits are decreased afte
...
+`queue-timeout` wins; the code-supplied value is used only when the service-configuration field is absent.
+
+There is no default queue timeout. Setting one trades tolerance of bursts for faster failure:
```
aldexis, line 233: Do you think there is a world where we can add a reasonable default? (e.g. 1h would likely be pretty conservative)
Or would we just eventually integrate with deadlines and use that instead?

mpritham: I think after ruling out outlier cases, we can introduce a conservative default. I'm slightly worried about doing that as part of this change in case there are edge cases where a long queuing time is not actually a problem (e.g. misc background jobs). Something really conservative can work as well, but I think the feature would be most useful when the value isn't so large.

> Or would we just eventually integrate with deadlines and use that instead?

The integration with deadlines here should be reading the deadline value to configure the queue timeout. I think these two knobs (deadlines and queue timeouts) target different goals. Deadlines are more restrictive than a queue timeout since they constrain the time of the entire call-path (sort of: the enforcement only happens when a Dialogue call is made so an exception isn't thrown exactly when a deadline expires). I think this feature is strictly just constraining the time a request spends in a queue (per attempt).

aldexis: > I'm slightly worried about doing that as part of this change in case there are edge cases where a long queuing time is not actually a problem (e.g. misc background jobs)

I agree - this is beyond the scope of this PR. I was asking for afterwards.

> The integration with deadlines here should be reading the deadline value to configure the queue timeout. I think these two knobs (deadlines and queue timeouts) target different goals. Deadlines are more restrictive than a queue timeout since they constrain the time of the entire call-path (sort of: the enforcement only happens when a Dialogue call is made so an exception isn't thrown exactly when a deadline expires). I think this feature is strictly just constraining the time a request spends in a queue (per attempt).

I was going off of a world where we make deadlines enforcement default on, in which case reading the deadline value would be enough as the default here (we could even already do this, when the deadline is enforced). I agree these are otherwise a bit different, but the deadline gives us a good upper bound for what we should target.

aldexis (dismissed):

aldexis (approved):

PR #2941, opened 2026-04-01
Add apache client reloading on trust and key store file change

atrocities:
Support automatic reloading of keystore and truststore per apache hc5 client upon changes to these files. 

Adds a refresh mechanism similar to the DNS resolution polling feature, and uses hashes of the keystore and truststore contents as a new cache key component. Key material refreshing happens at the DialogueChannel level, and the hashed data is passed down into the apache cache level. A change in key material contents results in invalidation of entries in the apache cache, causing the hc5 clients to reload with the new key/truststore contents. 

Changes in the location and nature of the key material constitute a configuration change, which would also cause a reload.

### DIFF
--- a/dialogue-clients/src/main/java/com/palantir/dialogue/clients/ChannelCache.java
+++ b/dialogue-clients/src/main/java/com/palantir/dialogue/clients/ChannelCache.java
@@ -23,6 +23,7 @@ import com.palantir.conjure.java.api.config.service.ServiceConfiguration;
 import com.palantir.conjure.java.client.config.ClientConfiguration;
 import com.palantir.dialogue.core.DialogueChannel;
 import com.palantir.dialogue.core.DialogueDnsResolver;
+import com.palantir.dialogue.core.SslStoreMetadata;
 import com.palantir.dialogue.core.TargetUri;
 import com.palantir.dialogue.hc5.ApacheHttpClientChannels;
 import com.palantir.logsafe.DoNotLog;
@@ -152,6 +153,9 @@ final class ChannelCache {
 
         ApacheCacheEntry apacheClient = getApacheClient(request);
 
+        Refreshable<SslStoreMetadata> storeMetadata =
+                KeystoreSupport.pollForChanges(channelCacheRequest.serviceConf().security());
+
         Refreshable<List<TargetUri>> targets;
         if (channelCacheRequest.overrideHostIndex().isPresent()) {
             targets = Refreshable.only(
@@ -180,6 +184,7 @@ final class ChannelCache {
                         .uris(channelCacheRequest.serviceConf().uris()) // restore uris
                         .build())
                 .uris(targets)
+                .storeMetadata(storeMetadata)
                 .factory(args -> ApacheHttpClientChannels.createSingleUri(args, apacheClient.client()))
                 .overrideHostIndex(channelCacheRequest.overrideHostIndex().stream()
                         .mapToInt(OverrideHostIndex::index)
--- a/dialogue-clients/src/main/java/com/palantir/dialogue/clients/ChannelCache.java
+++ b/dialogue-clients/src/main/java/com/palantir/dialogue/clients/ChannelCache.java
@@ -146,9 +146,6 @@ final class ChannelCache {
         Refreshable<SslStoreMetadata> storeMetadata =
                 KeystoreSupport.pollForChanges(channelCacheRequest.serviceConf().security());
 
-        ImmutableApacheClientRequest initialApacheRequest = apacheRequest(channelCacheRequest, storeMetadata.current());
-        ApacheCacheEntry initialApacheClient = getApacheClient(initialApacheRequest);
-
         Refreshable<List<TargetUri>> targets;
         if (channelCacheRequest.overrideHostIndex().isPresent()) {
             targets = Refreshable.only(
@@ -170,10 +167,13 @@ final class ChannelCache {
                             dnsResolutionResults.resolvedHosts(),
                             channelCacheRequest.taggedMetrics()));
         }
+
+        ClientConfiguration clientConf =
+                AugmentClientConfig.getClientConf(channelCacheRequest.serviceConf(), channelCacheRequest);
         return DialogueChannel.builder()
                 .channelName(channelCacheRequest.channelName())
                 .clientConfiguration(ClientConfiguration.builder()
-                        .from(initialApacheClient.conf())
+                        .from(clientConf)
                         .uris(channelCacheRequest.serviceConf().uris()) // restore uris
                         .build())
                 .uris(targets)
@@ -197,20 +197,8 @@ final class ChannelCache {
                 .serviceConf(stripUris(channelCacheRequest.serviceConf())) // we strip out uris to maximise cache hits
                 .blockingExecutor(channelCacheRequest.blockingExecutor())
                 .dnsResolver(channelCacheRequest.dnsResolver())
-                .sslStoreHash(sslStoreHash(sslStoreMetadata))
+                .sslStoreHash(sslStoreMetadata.hash())
                 .build();
-    }
-
-    private static String sslStoreHash(SslStoreMetadata sslStoreMetadata) {
-        String trustStoreHash = sslStoreMetadata
-                .trustStore()
-                .map(SslStoreMetadata.StoreFileMetadata::sha256)
-                .orElse("none");
-        String keyStoreHash = sslStoreMetadata
-                .keyStore()
-                .map(SslStoreMetadata.StoreFileMetadata::sha256)
-                .orElse("none");
-        return trustStoreHash + ':' + keyStoreHash;
     }
 
     @VisibleForTesting
--- a/dialogue-clients/src/main/java/com/palantir/dialogue/clients/ChannelCache.java
+++ b/dialogue-clients/src/main/java/com/palantir/dialogue/clients/ChannelCache.java
@@ -144,8 +144,8 @@ final class ChannelCache {
     }
 
     private DialogueChannel createNonLiveReloadingChannel(ChannelCacheKey channelCacheRequest) {
-        Refreshable<SslStoreMetadata> storeMetadata =
-                KeystoreSupport.pollForChanges(channelCacheRequest.serviceConf().security());
+        Refreshable<SslStoreMetadata> storeMetadata = SslStoresSupport.pollForChanges(
+                channelCacheRequest.serviceConf().security());
 
         Refreshable<List<TargetUri>> targets;
         if (channelCacheRequest.overrideHostIndex().isPresent()) {
@@ -318,6 +318,7 @@ final class ChannelCache {
 
         DialogueDnsResolver dnsResolver();
 
+        // This argument is used solely to cause cache misses when the underlying keystores change on disk
         Optional<String> sslStoreHash();
 
         @Value.Check
--- a/dialogue-clients/src/main/java/com/palantir/dialogue/clients/ChannelCache.java
+++ b/dialogue-clients/src/main/java/com/palantir/dialogue/clients/ChannelCache.java
@@ -145,8 +145,7 @@ final class ChannelCache {
 
     private DialogueChannel createNonLiveReloadingChannel(ChannelCacheKey channelCacheRequest) {
         Refreshable<SslStoreMetadata> storeMetadata = SslStoresSupport.pollForChanges(
-                channelCacheRequest.serviceConf().security(),
-                channelCacheRequest.taggedMetrics());
+                channelCacheRequest.serviceConf().security(), channelCacheRequest.taggedMetrics());
 
         Refreshable<List<TargetUri>> targets;
         if (channelCacheRequest.overrideHostIndex().isPresent()) {
--- a/dialogue-clients/src/main/java/com/palantir/dialogue/clients/ChannelCache.java
+++ b/dialogue-clients/src/main/java/com/palantir/dialogue/clients/ChannelCache.java
@@ -22,6 +22,7 @@ import com.google.common.annotations.VisibleForTesting;
 import com.palantir.conjure.java.api.config.service.ServiceConfiguration;
 import com.palantir.conjure.java.client.config.ClientConfiguration;
 import com.palantir.dialogue.core.DialogueChannel;
+import com.palantir.dialogue.core.DialogueChannelFactory.ChannelArgs;
 import com.palantir.dialogue.core.DialogueDnsResolver;
 import com.palantir.dialogue.core.SslStoreMetadata;
 import com.palantir.dialogue.core.TargetUri;
@@ -180,7 +181,7 @@ final class ChannelCache {
                 .storeMetadata(storeMetadata)
                 .factory(args -> ApacheHttpClientChannels.createSingleUri(
                         args,
-                        getApacheClient(apacheRequest(channelCacheRequest, storeMetadata.current()))
+                        getApacheClient(apacheRequest(channelCacheRequest, args))
                                 .client()))
                 .overrideHostIndex(channelCacheRequest.overrideHostIndex().stream()
                         .mapToInt(OverrideHostIndex::index)
@@ -189,15 +190,14 @@ final class ChannelCache {
                 .build();
     }
 
-    private static ImmutableApacheClientRequest apacheRequest(
-            ChannelCacheKey channelCacheRequest, SslStoreMetadata sslStoreMetadata) {
+    private static ImmutableApacheClientRequest apacheRequest(ChannelCacheKey channelCacheRequest, ChannelArgs args) {
         return ImmutableApacheClientRequest.builder()
                 .from(channelCacheRequest)
                 .channelName(channelCacheRequest.channelName())
                 .serviceConf(stripUris(channelCacheRequest.serviceConf())) // we strip out uris to maximise cache hits
                 .blockingExecutor(channelCacheRequest.blockingExecutor())
                 .dnsResolver(channelCacheRequest.dnsResolver())
-                .sslStoreHash(sslStoreMetadata.hash())
+                .sslStoreHash(args.sslStoreHash())
                 .build();
     }
 
@@ -318,10 +318,7 @@ final class ChannelCache {
 
         DialogueDnsResolver dnsResolver();
 
-        @Value.Default
-        default String sslStoreHash() {
-            return "";
-        }
+        Optional<String> sslStoreHash();
 
         @Value.Check
         default void check() {
--- a/dialogue-clients/src/main/java/com/palantir/dialogue/clients/ChannelCache.java
+++ b/dialogue-clients/src/main/java/com/palantir/dialogue/clients/ChannelCache.java
@@ -143,18 +143,11 @@ final class ChannelCache {
     }
 
     private DialogueChannel createNonLiveReloadingChannel(ChannelCacheKey channelCacheRequest) {
-        ImmutableApacheClientRequest request = ImmutableApacheClientRequest.builder()
-                .from(channelCacheRequest)
-                .channelName(channelCacheRequest.channelName())
-                .serviceConf(stripUris(channelCacheRequest.serviceConf())) // we strip out uris to maximise cache hits
-                .blockingExecutor(channelCacheRequest.blockingExecutor())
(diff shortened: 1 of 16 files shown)


bjlaub (commented):
Some initial thoughts:
- what happens to in-flight requests when we detect a change to the key/trust material and reload the hc5 client? I believe we're using pooled connections; I'm not sure what will happen to in-flight requests using connection from the pool if the client is closed before those requests finish. Probably worth adding some tests for this.
- watching for changes to the keystore in addition to the truststore adds a small bit of complexity here; I wonder if it's really necessary?

dialogue-core/src/main/java/com/palantir/dialogue/core/SslStoreMetadata.java
```diff
@@ -0,0 +1,82 @@
...
+            }
+            return new SslStoreMetadata(trustMd, keyMd);
+        } catch (IOException e) {
+            log.warn("Could not read from key material", e);
+            throw new SafeUncheckedIoException(e);
+        }
+    }
+
+    private static HashCode hashFile(Path path) throws IOException {
+        byte[] bytes = Files.readAllBytes(path);
```
bjlaub, line 64: nit: we could possibly avoid pulling the whole file into memory by using guava's `HashingInputStream` instead. I don't have a sense for how large these files will get in practice, but it's probably not a huge concern.

dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueChannelFactory.java
```diff
@@ -34,6 +34,8 @@ interface ChannelArgs {
 
         Optional<InetAddress> resolvedAddress();
 
+        Optional<String> sslStoreHash();
```
bjlaub, line 37: what happens if this is empty?

dialogue-clients/src/main/java/com/palantir/dialogue/clients/ChannelCache.java
```diff
@@ -142,15 +144,8 @@ DialogueChannel getNonReloadingChannel(
...
-                .from(channelCacheRequest)
-                .channelName(channelCacheRequest.channelName())
-                .serviceConf(stripUris(channelCacheRequest.serviceConf())) // we strip out uris to maximise cache hits
-                .blockingExecutor(channelCacheRequest.blockingExecutor())
-                .dnsResolver(channelCacheRequest.dnsResolver())
-                .build();
-
-        ApacheCacheEntry apacheClient = getApacheClient(request);
+        Refreshable<SslStoreMetadata> storeMetadata =
+                KeystoreSupport.pollForChanges(channelCacheRequest.serviceConf().security());
```
bjlaub: I think there's some nuance to this we should think about, particularly since it was adapted from the DNS polling stuff (disclaimer: this code is all a bit confusing and it's taking me a while to wrap my head around it again, so I could be wrong):

- This will create one `MetadataPollingTask` per cached DialogueChannel, where the work of each polling task is tied directly to reading bytes from the configured keystore/truststore on disk. But in practice, we may have many DialogueChannel instances that share the same truststore on disk, so with a large number of channels we end up mostly repeating work at fixed intervals.
- The `DialogueDnsResolutionWorker` operates similarly (we create one worker task per channel and run them at fixed intervals), and also may repeat work for services that share URIs, though this seems less likely (e.g. we are less likely to have 10 difference channels with overlapping host names to resolve, since they likely represent 10 different services with distinct host names, though that's not guaranteed). Another important difference is that we implicitly rely on the JVM DNS cache, so even though we poll at 1-second intervals in the DNS case, lookups may end up being very fast if the JVM has already cached results. Again, not guaranteed, and eventually we _will_ make a network call.

I might be overthinking this, perhaps the cost to compute the hash is cheap enough that many polling tasks over the same files on disk just isn't worth optimizing for. Ideally we might be able to create one polling task per keystore/truststore on disk, and have it update any number of refreshables (connected to the hc5 clients or whatever) to get them to reload trust material when it changes.

dialogue-clients/src/main/java/com/palantir/dialogue/clients/KeystoreSupport.java
```diff
@@ -0,0 +1,124 @@
...
+        cleaner.register(metadataRefreshable, new CleanupTask(future));
+        return metadataRefreshable;
+    }
+
+    private KeystoreSupport() {}
+
+    private static final class MetadataPollingTask implements Runnable {
+
+        private final SslConfiguration sslConfiguration;
+        private final SettableRefreshable<SslStoreMetadata> metadataRefreshable;
```
bjlaub: `DialogueDnsResolutionWorker` stores a [WeakReference](https://github.com/palantir/dialogue/blob/develop/dialogue-clients/src/main/java/com/palantir/dialogue/clients/DialogueDnsResolutionWorker.java#L59) to the output refreshable so that if it has been garbage collected we can avoid doing any work and exit early. Should we do something similar here?

dialogue-core/src/main/java/com/palantir/dialogue/core/SslStoreMetadata.java
atrocities: afaict, the size of these keystore files is usually a couple kb - I haven't found any enormous ones. 

fwiw, this was copied from the way that witchcraft does hashing of configuration files, which is using readAllBytes

dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueChannelFactory.java
atrocities: That would cause the apache cache key to also not contain an sslStoreHash - it's not used in the construction of the actual client itself. 

To that end, whether to include this argument as part of the ChannelArgs is debatable. It's not strictly a 'channel arg', but it feels like the best place for it in the way that things are currently structured.

dialogue-clients/src/main/java/com/palantir/dialogue/clients/ChannelCache.java
atrocities: That is a very valid concern. I'll have a look into deduping existing tasks, because yes - these file operations, while not terribly resource intensive, are also not exactly free. 

If it's not complicated (trust/keystore combos map to multiple DialogueChannels that need to be invalidated), then I do think it's worth this optimization.

atrocities:
Thanks much for having a look @bjlaub . I'll do the following:
- Try to coalesce MetadataPollingTasks by the paths of the key/truststore on disk instead of having one for every DialogueChannel.
- Add a test to demonstrate what happens when the keymaterial changes. My thought is that it ought to allow for any existing requests to finish, and then be eventually closed and cleaned up. On that note - will want to test for that as well.
- Figure out whether or not we're actually using the keystore (not truststore) in dialogue.  
- Investigate WeakReference in KeystoreSupport.

dialogue-clients/src/main/java/com/palantir/dialogue/clients/ChannelCache.java
```diff
@@ -305,6 +318,8 @@ interface ApacheClientRequest extends AugmentClientConfig {
 
         DialogueDnsResolver dnsResolver();
 
+        Optional<String> sslStoreHash();
```
aldexis, line 322: nit: Add a comment here explaining that this exists to force a cache miss when the ssl stores contents change, to force a reload of their values, even if the hash itself isn't directly used. Probably same for `ChannelArgs#sslStoreHash` (or refer to this method from there)

dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueChannel.java
```diff
@@ -189,33 +189,10 @@ public DialogueChannel build() {
...
-                                    UnsafeArg.of("uris", cf.clientConf().uris()),
-                                    SafeArg.of("numUris", cf.clientConf().uris().size()),
-                                    UnsafeArg.of("targets", targetUris),
-                                    SafeArg.of("numTargets", targetUris.size()));
-                            ImmutableList<LimitedChannel> targetChannels =
-                                    createHostChannels(cf, targetUris, Collections.unmodifiableMap(state));
-                            return NodeSelectionStrategyChannel.create(cf, targetChannels);
-                        }
-                    }));
+            LimitedChannel keystoreUpdatingChannel = createKeystoreUpdatingChannel(
```
aldexis: Technically, this is updating for both keystore and truststore, isn't it? If so, thoughts on naming it `createSslStoresUpdatingChannel`?

dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueChannel.java
```diff
@@ -286,6 +267,49 @@ private static ImmutableList<LimitedChannel> createHostChannels(
...
+        private static LimitedChannel createKeystoreUpdatingChannel(
+                Config cf, Meter _reloadMeter, Function<SslStoreMetadata, LimitedChannel> delegateSupplier) {
+            return new SupplierChannel(cf.storeMetadata().map(new Function<SslStoreMetadata, LimitedChannel>() {
+
+                @Override
+                public LimitedChannel apply(SslStoreMetadata storeMetadata) {
+                    return delegateSupplier.apply(storeMetadata);
+                }
+            }));
+        }
```
aldexis, line 300: nit: Fwiw, I think this can be simplified to
```suggestion
        private static LimitedChannel createKeystoreUpdatingChannel(
                Config cf, Meter _reloadMeter, Function<SslStoreMetadata, LimitedChannel> delegateSupplier) {
            return new SupplierChannel(cf.storeMetadata().map(delegateSupplier));
        }
```

dialogue-clients/src/main/java/com/palantir/dialogue/clients/ChannelCache.java
```diff
@@ -142,15 +144,8 @@ DialogueChannel getNonReloadingChannel(
     }
 
     private DialogueChannel createNonLiveReloadingChannel(ChannelCacheKey channelCacheRequest) {
```
aldexis, line 146: Regarding some comments I made last week when discussing this PR, I had a look at other places this was used, and realized that this is used in `ReloadingClientFactory#perHost`. I'm not sure that's correct fwiw, since there are things beyond the uris that may update, which makes me wonder whether we should have this `createNonLiveReloadingChannel` in the first place.

It's also used through the `ReloadingClientFactory#getNonReloading` method, which might be used in a few places, including through the legacy [`DialogueClients#create`](https://github.com/palantir/dialogue/blob/099b0a2c4d86b1b2b1688a47f8801a6ed90df357/dialogue-clients/src/main/java/com/palantir/dialogue/clients/DialogueClients.java#L57-L71)

This also made me realize that we _might_ want to handle https://github.com/palantir/dialogue/blob/099b0a2c4d86b1b2b1688a47f8801a6ed90df357/dialogue-clients/src/main/java/com/palantir/dialogue/clients/ReloadingClientFactory.java#L83-L110? (honestly this whole client creation codepath should get refactored - we shouldn't be creating this many different clients in so many places)

dialogue-clients/src/main/java/com/palantir/dialogue/clients/KeystoreSupport.java
```diff
@@ -0,0 +1,124 @@
...
+import com.palantir.tritium.metrics.registry.SharedTaggedMetricRegistries;
+import java.lang.ref.Cleaner;
+import java.time.Duration;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.function.Supplier;
+
+final class KeystoreSupport {
```
aldexis: This is for both keystore and truststore, right? (e.g. `SslStoresSupport`?)

dialogue-clients/src/main/java/com/palantir/dialogue/clients/ChannelCache.java
aldexis: Since we aren't tying this refreshable to the config refreshable itself, do we have a test covering the case where either of the file locations changes? (it _should_ be fine afaict, but I want to make sure we cover this case)

dialogue-clients/src/main/java/com/palantir/dialogue/clients/KeystoreSupport.java
aldexis: +1 - otherwise, I think the cleaner will also never remove the scheduled future nor the task itself?

dialogue-clients/src/test/java/com/palantir/dialogue/clients/ChannelCacheTest.java
```diff
@@ -115,6 +115,26 @@ void different_dns_resolver_new_instance() {
...
+
+        ChannelCache.ApacheCacheEntry cacheResult2 = cache.getApacheClient(ImmutableApacheClientRequest.builder()
+                .dnsResolver(StubDnsResolver.INSTANCE)
+                .serviceConf(serviceConf)
+                .channelName("channelName")
+                .sslStoreHash("hash-two")
+                .build());
+
+        assertThat(cacheResult).isNotSameAs(cacheResult2);
+        assertThat(cache.toString()).contains("apacheCache.size=1");
```
aldexis, line 135: nit: Rather than relying on the stringified representation, we could expose a VisibleForTesting package-private method giving us the size of the apache cache (not a bit deal, since this is pretty stable anyway)

dialogue-clients/src/test/java/com/palantir/dialogue/clients/KeystoreSupportTest.java
```diff
@@ -0,0 +1,98 @@
...
+        ScheduledExecutorService executorService = Executors.newSingleThreadScheduledExecutor();
+        try {
+            Refreshable<SslStoreMetadata> refreshable =
+                    KeystoreSupport.pollForChanges(sslConfiguration, executorService, Duration.ofMillis(25));
+
+            SslStoreMetadata initial = refreshable.get();
+            HashCode initialTrustHash = initial.trustStore().hash();
+
+            // Corrupts the truststore for the sake of testing updates
+            Files.write(trustStore, new byte[] {1, 0, 0, 0}, java.nio.file.StandardOpenOption.APPEND);
```
aldexis: vnit: use imports
```suggestion
            Files.write(trustStore, new byte[] {1, 0, 0, 0}, StandardOpenOption.APPEND);
```

dialogue-clients/src/test/java/com/palantir/dialogue/clients/SslStoresSupportTest.java
```diff
@@ -0,0 +1,98 @@
...
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import org.awaitility.Awaitility;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+
+class KeystoreSupportTest {
+    @Test
+    void pollForChanges_updates_on_change(@TempDir Path tempDir) throws IOException {
```
aldexis, line 45: This seems to only test changes on the truststore. Should we also write another test for keystore changes?

dialogue-clients/src/test/java/com/palantir/dialogue/clients/SslStoresSupportTest.java
```diff
@@ -0,0 +1,98 @@
...
+            SslStoreMetadata initial = refreshable.get();
+            HashCode initialTrustHash = initial.trustStore().hash();
+
+            // Corrupts the truststore for the sake of testing updates
+            Files.write(trustStore, new byte[] {1, 0, 0, 0}, java.nio.file.StandardOpenOption.APPEND);
+
+            Awaitility.waitAtMost(Duration.ofSeconds(10)).untilAsserted(() -> {
+                SslStoreMetadata updated = refreshable.get();
+                assertThat(updated).isNotEqualTo(initial);
+                assertThat(updated.trustStore().hash()).isNotEqualTo(initialTrustHash);
```
aldexis, line 71: nit: may be interesting to also validate the hash is the one we expect from the file we wrote?

dialogue-clients/src/test/java/com/palantir/dialogue/clients/KeystoreSupportTest.java
```diff
@@ -0,0 +1,98 @@
...
+
+        SslConfiguration sslConfiguration = SslConfiguration.of(trustStore, keyStore, "keystore");
+        ScheduledExecutorService executorService = Executors.newSingleThreadScheduledExecutor();
+        try {
+            Refreshable<SslStoreMetadata> refreshable =
+                    KeystoreSupport.pollForChanges(sslConfiguration, executorService, Duration.ofMillis(25));
+            HashCode initialTrustHash = refreshable.get().trustStore().hash();
+
+            Files.move(trustStore, movedTrustStore);
+            Thread.sleep(150);
```
aldexis: Technically, we aren't actually testing that we did a reload during this period, but I don't think we have any visible way to observe this?
Makes me wonder whether we may want to add a metric for failing to refresh the ssl stores (in which case, we could await the metric increasing, then move the file back and verify it updates)

dialogue-clients/src/test/java/com/palantir/dialogue/clients/KeystoreSupportTest.java
```diff
@@ -0,0 +1,98 @@
...
+
+            Files.move(movedTrustStore, trustStore);
+            Files.write(trustStore, new byte[] {9}, java.nio.file.StandardOpenOption.APPEND);
```
aldexis: vnit: (update before moving back, that way it's atomic - doesn't matter much but is slightly cleaner?)
```suggestion
            Files.write(trustStore, new byte[] {9}, StandardOpenOption.APPEND);
            Files.move(movedTrustStore, trustStore);
```

dialogue-clients/src/test/java/com/palantir/dialogue/otherpackage/DialogueClientsIntegrationTest.java
```diff
@@ -490,6 +540,15 @@ interface StringToVoidRequestPath {
         String client();
     }
 
+    private static void writeTrustStore(Path trustStorePath, X509Certificate certificate) throws Exception {
+        KeyStore trustStore = KeyStore.getInstance("JKS");
+        trustStore.load(null, new char[0]);
+        trustStore.setCertificateEntry("cert", certificate);
+        try (java.io.OutputStream outputStream = Files.newOutputStream(trustStorePath)) {
```
aldexis, line 644:
```suggestion
        try (OutputStream outputStream = Files.newOutputStream(trustStorePath)) {
```

aldexis (commented):
Pretty nice work - I like the tests!

dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueChannel.java
atrocities: Yes, thanks! The reason was that there were previously some debug loglines in the overriden apply.

dialogue-clients/src/main/java/com/palantir/dialogue/clients/KeystoreSupport.java
atrocities: Yes, will rename.

dialogue-clients/src/main/java/com/palantir/dialogue/clients/ChannelCache.java
atrocities: Yeah, this is unaware of the config refreshable - will add a test to ensure that changing the configuration will dtrt

dialogue-clients/src/test/java/com/palantir/dialogue/clients/ChannelCacheTest.java
atrocities: Good point - this was essentially copied from the other tests in this class. I think that this is appropriate as a flup.

dialogue-clients/src/test/java/com/palantir/dialogue/clients/KeystoreSupportTest.java
atrocities: Good idea, will add metrics in the same way as the dns polling does.

PR #2942, opened 2026-04-02
[do-not-merge] [discussion] Queue timeouts - sweep approach

mpritham:
## After this PR

==COMMIT_MSG==
==COMMIT_MSG==

### DIFF
--- a/dialogue-core/src/test/java/com/palantir/dialogue/core/RequestFlowWalkthroughTest.java
+++ b/dialogue-core/src/test/java/com/palantir/dialogue/core/RequestFlowWalkthroughTest.java
@@ -1,0 +1,462 @@
+/*
+ * (c) Copyright 2024 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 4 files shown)


dialogue-core/src/test/java/com/palantir/dialogue/core/RequestFlowWalkthroughTest.java
```diff
@@ -0,0 +1,462 @@
...
+        ListenableFuture<Response>[] queued = new ListenableFuture[5];
+        for (int i = 0; i < 5; i++) {
+            queued[i] = channel.execute(ENDPOINT, REQUEST);
+            assertThat(queued[i]).isNotDone();
+        }
+        assertThat(outerQueue.getQueueSizeForTesting()).isEqualTo(5);
+
+        // Wait for the periodic sweep to fire (sweep interval = timeout / 2 = 50ms,
+        // so after 200ms the sweep has run multiple times and all entries are expired)
+        Thread.sleep(queueTimeout.toMillis() + 100);
```
mpritham, line 444: We should pass in a clock/ticker so this test isn't flakey in CI.

PR #2954, opened 2026-04-13
QueuedChannel: better observability for 'queue full' error

iamdanfox:
## Before this PR

I am seeing a surprising amount of `Unable to make a request (queue is full)` exceptions. I'd like to gather concrete data about precisely _which_ outgoing requests are triggering this, but it's kinda tricky right now because Dialogue sets up many different QueuedChannels and I can't easily tell which one is filling up.

## After this PR
==COMMIT_MSG==
QueuedChannel: better observability for 'queue full' error
==COMMIT_MSG==

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
@@ -39,7 +39,9 @@ import com.palantir.logsafe.logger.SafeLoggerFactory;
 import com.palantir.tracing.CloseableSpan;
 import com.palantir.tracing.DetachedSpan;
 import com.palantir.tracing.TagTranslator;
+import java.util.ArrayList;
 import java.util.Deque;
+import java.util.List;
 import java.util.Optional;
 import java.util.concurrent.CancellationException;
 import java.util.concurrent.ConcurrentLinkedDeque;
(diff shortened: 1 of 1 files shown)


bjlaub (approved):

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
```diff
@@ -106,10 +108,15 @@ final class QueuedChannel implements Channel {
...
-        this.queueSizeCounter = Suppliers.memoize(metrics::requestsQueued);
-        this.queuedTime = metrics.requestQueuedTime();
-        this.limitedResultSupplier = () -> Futures.immediateFailedFuture(new SafeRuntimeException(
-                "Unable to make a request (queue is full)", SafeArg.of("maxQueueSize", maxQueueSize)));
+        this.queueSizeCounter = Suppliers.memoize(instrumentation::requestsQueued);
+        this.queuedTime = instrumentation.requestQueuedTime();
+        this.limitedResultSupplier = () -> {
+            List<SafeArg<?>> safeArgs = new ArrayList<>(instrumentation.queueFullSafeArgs());
+            safeArgs.add(SafeArg.of("maxQueueSize", maxQueueSize));
+            safeArgs.add(SafeArg.of("channelName", channelName));
```
mpritham, line 116: I think it would also be useful to add `queueType` here. There can be endpoint-level and host*endpoint level queues created with the same `channelName`: `channelName` is populated with `Config#channelName`.

PR #2957, opened 2026-04-15
Update `Attachments` to support `remove` and `putIfAbsent`

mpritham:
In #2926 I would like to atomically update the value of the queue timeout request attachment if it's not already set. 

## After this PR

==COMMIT_MSG==
Update `Attachments` to support `remove` and `putIfAbsent`
==COMMIT_MSG==

### DIFF
--- a/dialogue-target/src/main/java/com/palantir/dialogue/Attachments.java
+++ b/dialogue-target/src/main/java/com/palantir/dialogue/Attachments.java
@@ -44,6 +44,24 @@ final class Attachments {
     }
 
     @Nullable
+    <V> V putIfAbsent(AttachmentKey<V> key, V value) {
+        Preconditions.checkNotNull(key, "key");
+        Preconditions.checkNotNull(value, "value");
+        key.checkIsInstance(value);
+        @SuppressWarnings("unchecked")
+        V result = (V) attachments.putIfAbsent(key, value);
+        return result;
+    }
+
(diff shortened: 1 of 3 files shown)


dialogue-target/src/main/java/com/palantir/dialogue/Attachments.java
```diff
@@ -43,6 +43,24 @@ <V> V put(AttachmentKey<V> key, V value) {
...
+        return result;
+    }
+
+    @Nullable
+    <V> V remove(AttachmentKey<V> key) {
+        Preconditions.checkNotNull(key, "key");
+        @SuppressWarnings("unchecked")
+        V result = (V) attachments.remove(key);
+        return result;
+    }
```
aldexis, line 62: Should probably propagate these to [`RequestAttachments`](https://github.com/palantir/dialogue/blob/develop/dialogue-target/src/main/java/com/palantir/dialogue/RequestAttachments.java) as well? (I have no idea why we need both tbh)

aldexis (dismissed):

dialogue-target/src/main/java/com/palantir/dialogue/Attachments.java
mpritham: Yes we need to do that as well. Updated.

aldexis (approved):

PR #2973, opened 2026-04-27
Conjure error parameter is not overridden by custom runtime

mpritham:
## Before this PR
Setting the Conjure error parameter format with `withConjureErrorParameterFormat` creates an interface that extends `ConjureRuntime` where `runtime#bodySerDe#errorParameterFormat` returns the set format. When `withRuntime` is called after the error parameter format is set, the interface created that set the format is replaced. The ordering of `withRuntime` and `withConjureErrorParameterFormat` matters.

## After this PR

The order should not matter.

==COMMIT_MSG==
Conjure error parameter is not overridden by custom runtime
==COMMIT_MSG==

## Possible downsides?

Any client relying on `withRuntime` to clear the previously set error parameter format will now break. This is incorrect and unintended behavior. There are no such callers internally.

### DIFF
--- a/dialogue-clients/src/main/java/com/palantir/dialogue/clients/ReloadingClientFactory.java
+++ b/dialogue-clients/src/main/java/com/palantir/dialogue/clients/ReloadingClientFactory.java
@@ -115,8 +115,18 @@ final class ReloadingClientFactory implements DialogueClients.ReloadingFactory {
         Refreshable<ServicesConfigBlock> scb();
 
         @Value.Default
-        default ConjureRuntime runtime() {
+        default ConjureRuntime baseRuntime() {
             return DefaultConjureRuntime.builder().build();
+        }
+
+        Optional<ConjureErrorParameterFormat> conjureErrorParameterFormat();
+
+        @Value.Derived
+        default ConjureRuntime runtime() {
+            return conjureErrorParameterFormat()
+                    .<ConjureRuntime>map(
+                            format -> new ErrorSerializationFormatSettingConjureRuntime(baseRuntime(), format))
+                    .orElseGet(this::baseRuntime);
         }
 
         @Value.Default
@@ -332,7 +342,7 @@ final class ReloadingClientFactory implements DialogueClients.ReloadingFactory {
 
     @Override
     public DialogueClients.ReloadingFactory withRuntime(ConjureRuntime runtime) {
-        return new ReloadingClientFactory(params.withRuntime(runtime), cache);
+        return new ReloadingClientFactory(params.withBaseRuntime(runtime), cache);
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


dialogue-clients/src/main/java/com/palantir/dialogue/clients/ReloadingClientFactory.java
```diff
@@ -115,10 +115,20 @@ interface ReloadingParams extends AugmentClientConfig {
         Refreshable<ServicesConfigBlock> scb();
 
         @Value.Default
-        default ConjureRuntime runtime() {
+        default ConjureRuntime baseRuntime() {
```
mpritham, line 118: This interface is package-private. This should not be a break.

dialogue-clients/src/main/java/com/palantir/dialogue/clients/ReloadingClientFactory.java
```diff
@@ -115,10 +115,20 @@ interface ReloadingParams extends AugmentClientConfig {
...
             return DefaultConjureRuntime.builder().build();
         }
 
+        Optional<ConjureErrorParameterFormat> conjureErrorParameterFormat();
+
+        @Value.Derived
+        default ConjureRuntime runtime() {
+            return conjureErrorParameterFormat()
+                    .<ConjureRuntime>map(
+                            format -> new ErrorSerializationFormatSettingConjureRuntime(baseRuntime(), format))
```
mpritham, line 130: As a reminder, this whole `ErrorSerializationFormatSettingConjureRuntime` redirection is needed because `ConjureRuntime` is an interface.

dialogue-clients/src/main/java/com/palantir/dialogue/clients/ReloadingClientFactory.java
```diff
@@ -115,10 +115,20 @@ interface ReloadingParams extends AugmentClientConfig {
...
+                    .orElseGet(this::baseRuntime);
+        }
```
aldexis, line 132: This feels so confusing. Does this mean we'd need to do this wrapping for each possible such setting we'd want to add in the future?

This might be worth adding a comment in here (and/or on `ErrorSerializationFormatSettingConjureRuntime`) explaining why we need to do this, because whoever needs to work with this in the future is going to be confused

aldexis (dismissed):
Approving with one comment

dialogue-clients/src/main/java/com/palantir/dialogue/clients/ReloadingClientFactory.java
mpritham: Agree it's confusing. Noted why this is the case in a comment.

aldexis (approved):

PR #2993, opened 2026-05-15
Cache sticky endpoint channels

schlosna:
## Before this PR

Dialogue clients that make heavy use of sticky channels incur a significant construction overhead (e.g. see https://github.com/palantir/conjure-java/pull/2869 ) leading to per-endpoint `TypeMarker`
anonymous classes, `ExceptionDeserializerArgs` instances, and `Deserializer` fields,
causing excessive allocations on high throughput services.

This adds unnecessary overhead for conjure clients that use sticky channels such as ⛰️ clients via e.g. `DataServiceAsync#of(com.palantir.dialogue.EndpointChannelFactory, com.palantir.dialogue.ConjureRuntime)`.

See https://github.com/palantir/dialogue/pull/2647 which was a basis for this PR.

## After this PR

==COMMIT_MSG==
## Summary

- **Introduce `StickyEndpointChannelsFactory` interface** — replaces the `Supplier<Channel>` return type from `StickyEndpointChannels2.create()` and `DialogueChannel.stickyChannels()` with a typed `stickyChannel()` method. `DialogueChannel` now directly implements this interface, and `DialogueChannel.stickyChannels()` is deprecated in favor of `stickyChannel()`.
- **Extract `CachingEndpointChannelFactory`** — encapsulates the Caffeine `LoadingCache<Endpoint, EndpointChannel>` with `delegate::endpoint` as the loader into a reusable `EndpointChannelFactory` implementation. Cache size is configurable via `Config.maxEndpointCacheSize()` (default 1000), and caching is bypassed when set to `<= 0`.
- **Simplify `StickyEndpointChannels2` internals** — removes the `StickyEndpointChannels2EndpointFactorySupplier` and `StickyEndpointChannel` inner classes; queue override attachment and sticky routing are now inlined in `StickyChannel2.endpoint()`, reducing indirection.
- **Update `ReloadingClientFactory`** — `InternalDialogueChannel` now extends `StickyEndpointChannelsFactory` instead of declaring its own `stickyChannels()` method, and callers use the new `stickyChannel()` API directly.

## Test plan

- [x] Existing `StickyEndpointChannels2Test` passes (caching behavior is identical, just encapsulated)
- [x] `SimulationTest` passes and regenerated simulation resources reflect updated sticky channel behavior
- [x] `ReloadingClientFactoryTest` passes with the new `stickyChannel()` API

==COMMIT_MSG==

## Possible downsides?

This encapsulates a cache of `Endpoint -> EndpointChannel` so there will be some memory overhead; however, use of sticky channels implies these routes intend to stick around in memory and amortize overhead for benefit of throughput.


dialogue-core/src/main/java/com/palantir/dialogue/core/Config.java
```diff
@@ -79,6 +79,12 @@ default int maxQueueSize() {
         return 100_000;
     }
 
+    /** Maximum number of cached endpoint channels. Values {@code <= 0} disable caching. */
+    @Value.Default
+    default int maxEndpointCacheSize() {
+        return 1_000;
```
aldexis, line 85: Thoughts on starting with this at 0 to keep parity with existing logic, and testing a different value in some prod environment, before changing the default?

I also wonder whether 1000 is enough/too much.

dialogue-core/src/main/java/com/palantir/dialogue/core/Config.java
```diff
@@ -79,6 +79,12 @@ default int maxQueueSize() {
         return 100_000;
     }
 
+    /** Maximum number of cached endpoint channels. Values {@code <= 0} disable caching. */
+    @Value.Default
+    default int maxEndpointCacheSize() {
```
aldexis, line 84: This seems to be specific to sticky channels, but the javadoc doesn't surface that at all. `cached endpoint channels` is instead pretty generic. It may be worth adding some clarification.

Additionally, I don't think users can modify this config directly, so we may want to pipe it through  `DialogueChannel.Builder`

dialogue-core/src/main/java/com/palantir/dialogue/core/CachingEndpointChannelFactory.java
```diff
@@ -0,0 +1,47 @@
...
+
+    private final LoadingCache<Endpoint, EndpointChannel> cache;
+    private final EndpointChannelFactory delegate;
+
+    CachingEndpointChannelFactory(EndpointChannelFactory delegate, int maxEndpointCacheSize) {
+        this.delegate = delegate;
+        this.cache = Caffeine.newBuilder()
+                .maximumSize(maxEndpointCacheSize)
+                .weakValues()
+                .build(delegate::endpoint);
```
aldexis, line 35: side-note: it may be worth starting to use https://github.com/palantir/cache for new caches like this (which iirc gives us instrumentation by default, which could be interesting to have regardless, since that would help identify cases where the cache size may be inappropriate)

dialogue-core/src/main/java/com/palantir/dialogue/core/StickyEndpointChannels2.java
```diff
@@ -240,24 +228,4 @@ private static ListenableFuture<Response> executeWithStickyTarget(
...
-
-        @Override
-        public ListenableFuture<Response> execute(Request request) {
-            return stickyRouter.execute(request, delegate);
-        }
-
-        @Override
-        public String toString() {
-            return "StickyEndpointChannel{delegate=" + delegate + '}';
-        }
```
aldexis, line 261: Seeing this makes me wonder whether anybody may have been logging the channel, as a way to debug stuff, and whether the loss of this override may cause debugging pain (since you're now returning `request -> router.execute(request, endpointWithQueueOverride);`)

dialogue-core/src/main/java/com/palantir/dialogue/core/StickyEndpointChannelsFactory.java
```diff
@@ -0,0 +1,23 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.dialogue.core;
+
+import com.palantir.dialogue.Channel;
+
+public interface StickyEndpointChannelsFactory {
```
pkoenig10, line 21: Change the package and make the name non-plural to be consistent with `EndpointChannelFactory`.

Also this returns `Channel`, not `EndpointChannel`.

```suggestion
package com.palantir.dialogue;

import com.palantir.dialogue.Channel;

public interface StickyChannelFactory {
```

dialogue-core/src/main/java/com/palantir/dialogue/core/StickyEndpointChannels2.java
```diff
@@ -33,28 +33,34 @@
...
-    static Supplier<Channel> create(Config cf, LimitedChannel nodeSelectionChannel, EndpointChannelFactory delegate) {
+    static StickyEndpointChannelsFactory create(
+            Config cf, LimitedChannel nodeSelectionChannel, EndpointChannelFactory delegate) {
         Supplier<Channel> queueOverrideSupplier = new QueueOverrideSupplier(cf, nodeSelectionChannel);
-        return new StickyEndpointChannels2(
-                new StickyEndpointChannels2EndpointFactorySupplier(queueOverrideSupplier, delegate));
+        int maxEndpointCacheSize = cf.maxEndpointCacheSize();
+        EndpointChannelFactory channelFactory = (maxEndpointCacheSize <= 0)
+                ? delegate
+                : new CachingEndpointChannelFactory(delegate, maxEndpointCacheSize);
```
pkoenig10, line 62: This feels a bit like overkill to me.

If we want to keep this, I'd consider making the `CachingEndpointChannelFactory` constructor private and pushing this logic into a static factory method there, so callers don't need to be concerned with this.

dialogue-core/src/main/java/com/palantir/dialogue/core/StickyEndpointChannels2.java
```diff
@@ -240,24 +228,4 @@ private static ListenableFuture<Response> executeWithStickyTarget(
             return endpointChannel.execute(request);
         }
     }
-
-    private static final class StickyEndpointChannel implements EndpointChannel {
```
pkoenig10, line 244: nit: Can we keep this? Having non-anonymous classes is helpful for telemetry/debugging because it has a nice name in stacktraces and has a `toString` implementation.

simulation/src/test/java/com/palantir/dialogue/core/Strategy.java
```diff
@@ -50,7 +50,8 @@ public Channel getChannel(Simulation simulation, Supplier<Map<String, Simulation
 
     public Supplier<Channel> getSticky2NonReloading(Simulation simulation, Map<String, SimulationServer> servers) {
```
pkoenig10, line 51: This return type should probably be changed to use the new factory interface we've introduced. Then we can just do `return dialogueChannelWithDefaults(...)`.

dialogue-core/src/test/java/com/palantir/dialogue/core/StickyEndpointChannels2Test.java
```diff
@@ -84,9 +84,12 @@ public final class StickyEndpointChannels2Test {
     public void beforeEach() {
         when(config.channelName()).thenReturn("channel");
         when(config.clientConf()).thenReturn(clientConfiguration);
+        when(config.maxEndpointCacheSize()).thenReturn(1_000);
         lenient().when(endpointChannelFactory.endpoint(any())).thenReturn(endpointChannel);
         when(clientConfiguration.taggedMetricRegistry()).thenReturn(new DefaultTaggedMetricRegistry());
-        sticky = StickyEndpointChannels2.create(config, nodeSelectionChannel, endpointChannelFactory);
+        StickyEndpointChannelsFactory stickyEndpointChannelsFactory =
+                StickyEndpointChannels2.create(config, nodeSelectionChannel, endpointChannelFactory);
+        sticky = () -> stickyEndpointChannelsFactory.stickyChannel();
```
pkoenig10, line 92: Same here - we should get rid of this `Supplier` indirection because it just makes the code harder to understand. Tests can just call `stickyChannel()` instead of `get()`.

dialogue-core/src/main/java/com/palantir/dialogue/core/CachingEndpointChannelFactory.java
pkoenig10: That library doesn't support weak values (and I'm not sure I want to support weak values there). Using Caffeine feels more appropriate here.

aldexis:
@schlosna What is the current status of this PR? We've noticed similar things across a number of services, as the largest allocation pressure (especially on smaller stacks), and I'm keen to figure out if we can reduce/remove this

dialogue-clients/src/main/java/com/palantir/dialogue/clients/ReloadingClientFactory.java
aldexis, line 310: @schlosna If I'm correct, our usage looks roughly like
```java
Supplier<DialogueService> serviceSupplier = () -> {
    DialogueClients.StickyChannelSession stickySession = stickyChannelFactory2.session();
    return stickySession.sticky(DialogueService.class);
};
```
and most of the allocations actually from the `ExceptionDeserializerArgs` that get created in the dialogue clients upon instantiation.

I don't think this PR would actually change that fact, even if we cache the channel, because the client itself would be recreated everytime?

aldexis: I've added a benchmark in https://github.com/palantir/dialogue/pull/3021 showing that
- Exception mapper make up ~75% of the allocations (probably lower/higher depending on how many errors actually exist - this is for ~100 errors)
- Channels make up another ~25%
- The rest (including TypeMarker and deserializer allocations) seem to make up ~1.5% of the current allocation profile

Seems like we need both this and https://github.com/palantir/conjure-java/pull/2869 or instead cache the dialogue client creation itself (which maybe we should do? Creating tons of dialogue clients isn't particularly efficient either)

PR #3016, opened 2026-06-02
Provide more metadata in CIAD debug logs

mpritham:
## Before this PR

The debug logs in `CautiousIncreaseAggressiveDecreaseConcurrencyLimiter` [do not contain any metadata](https://github.com/palantir/dialogue/blob/618c94b99be0d1e0e1cdf79c8a77dee2da447aee/dialogue-core/src/main/java/com/palantir/dialogue/core/CautiousIncreaseAggressiveDecreaseConcurrencyLimiter.java#L222) pertaining to the channel/limiter updated.

We should add some metadata to make it easier to debug our CIAD logic.

## After this PR

==COMMIT_MSG==
Provide more metadata in CIAD debug logs
==COMMIT_MSG==

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/CautiousIncreaseAggressiveDecreaseConcurrencyLimiter.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/CautiousIncreaseAggressiveDecreaseConcurrencyLimiter.java
@@ -54,8 +54,16 @@ final class CautiousIncreaseAggressiveDecreaseConcurrencyLimiter {
 
     private final Behavior behavior;
 
+    // Used strictly for providing the channel name in debug logs.
+    // Volatile because it may be set from a different thread than the one emitting the log line.
+    private volatile String channelNameForLogging = "unknown";
+
     CautiousIncreaseAggressiveDecreaseConcurrencyLimiter(Behavior behavior) {
         this.behavior = behavior;
+    }
+
+    void setChannelNameForLogging(String value) {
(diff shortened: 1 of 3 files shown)


dialogue-core/src/main/java/com/palantir/dialogue/core/CautiousIncreaseAggressiveDecreaseConcurrencyLimiter.java
```diff
@@ -54,10 +54,18 @@ final class CautiousIncreaseAggressiveDecreaseConcurrencyLimiter {
 
     private final Behavior behavior;
 
+    // Used strictly for providing the channel name in debug logs.
+    // Volatile because it may be set from a different thread than the one emitting the log line.
+    private volatile String channelNameForLogging = "unknown";
```
mpritham, line 59: Volatile is not strictly required here but it's cheap and provides correctness in case the details of our channel naming were to be modified so I'm opting to keep it. 

Here is a potential race that using volatile protects against:
1. A response is received and hits the `success`/`dropped` logic in this class where the debug log would be created
2. URIs are reloaded and the shared limiter's channel name is changed.

This race does not exist currently: there is no scenario where we rename channels for the same URI upon reload.

I'd prefer to keep it because it's very cheap: reads only happen when debug logs are enabled and writes only happen at channel creation. It's also one less thing a future reviewer needs to consider when/if calling the `setChannelNameForLogging` method.

parker89 (approved):

PR #3022, opened 2026-06-08
[test only][refactoring] Simulations

mpritham:
## Before this PR

A minor refactoring of some of the simulation code to make it more readable and remove dead code.

## After this PR

==COMMIT_MSG==
[test only][refactoring] Simulations
==COMMIT_MSG==

### DIFF
--- a/simulation/src/main/java/com/palantir/dialogue/core/Benchmark.java
+++ b/simulation/src/main/java/com/palantir/dialogue/core/Benchmark.java
@@ -365,13 +365,6 @@ public final class Benchmark {
         BenchmarkTimingEndpointChannel channel();
     }
 
-    @Value.Immutable
-    interface LimitedStream {
-        long limit();
-
-        Stream<ScheduledRequest> stream();
-    }
-
     /**
      * Determines when the benchmark terminates - useful when a server is behaving like a black hole (not returning).
      */
--- a/simulation/src/test/java/com/palantir/dialogue/core/SimulationTest.java
+++ b/simulation/src/test/java/com/palantir/dialogue/core/SimulationTest.java
@@ -20,7 +20,6 @@ import static com.palantir.dialogue.core.Benchmark.DEFAULT_ENDPOINT;
 import static org.assertj.core.api.Assertions.assertThat;
 
 import com.codahale.metrics.Meter;
-import com.google.common.base.Stopwatch;
 import com.google.common.base.Suppliers;
 import com.palantir.dialogue.Channel;
 import com.palantir.dialogue.Endpoint;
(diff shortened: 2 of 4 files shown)


simulation/src/main/java/com/palantir/dialogue/core/Benchmark.java
```diff
@@ -365,13 +365,6 @@ interface NamedEndpointChannel {
         BenchmarkTimingEndpointChannel channel();
     }
 
-    @Value.Immutable
-    interface LimitedStream {
```
mpritham, line 369: This was unused.

simulation/src/test/java/com/palantir/dialogue/core/Strategy.java
```diff
@@ -66,7 +63,6 @@ private static void pinUntilError(ClientConfiguration.Builder configBuilder) {
     private static void unlimitedRoundRobin(ClientConfiguration.Builder configBuilder) {
         configBuilder
                 .nodeSelectionStrategy(NodeSelectionStrategy.ROUND_ROBIN)
-                .failedUrlCooldown(Duration.ofMillis(200))
```
mpritham, line 69: [This config does not do anything in Dialogue](https://github.com/palantir/dialogue/blob/62c8eecd2bff82d4a28ab72cb16fbf183a108a6d/dialogue-clients/src/main/java/com/palantir/dialogue/clients/ReloadingClientFactory.java#L367)

simulation/src/test/java/com/palantir/dialogue/core/SimulationTest.java
```diff
@@ -112,375 +96,334 @@ final class SimulationTest {
...
 
-    private Strategy st;
-
     private final Simulation simulation = new Simulation();
-    private Supplier<Map<String, SimulationServer>> servers;
-    private Benchmark.BenchmarkResult result;
+    private final SimulationReport report = new SimulationReport(simulation);
+    private Outcome outcome;
+
+    private record Outcome(Strategy strategy, Benchmark.BenchmarkResult result) {}
```
mpritham, line 103: This could be a package-private class so that `SimulationReport` can use it too. No strong feelings.

simulation/src/test/java/com/palantir/dialogue/core/Strategy.java
aldexis: Would it make sense to update the javadoc in https://github.com/palantir/conjure-java-runtime/blob/0af5ccdb5f4600318eb84c30578c8c9d195d9a50/client-config/src/main/java/com/palantir/conjure/java/client/config/ClientConfiguration.java#L96-L100? I know this config is used in other places (namely okhttp-clients), but it seems misleading since we recommend dialogue usage across the board.

simulation/src/test/java/com/palantir/dialogue/core/SimulationTest.java
```diff
@@ -112,375 +96,334 @@ final class SimulationTest {
             name = ParameterizedTest.DISPLAY_NAME_PLACEHOLDER + "[" + ParameterizedTest.ARGUMENTS_PLACEHOLDER + "]")
     @interface SimulationCase {}
 
-    private Strategy st;
-
     private final Simulation simulation = new Simulation();
-    private Supplier<Map<String, SimulationServer>> servers;
-    private Benchmark.BenchmarkResult result;
+    private final SimulationReport report = new SimulationReport(simulation);
+    private Outcome outcome;
```
aldexis, line 101: This isn't exactly something new to this PR, but it seems a bit weird that we have to store a specific outcome/result at the end of each test.

I wonder whether we could simply have `TestInfo` passed to the test itself (or the method name, since that's all we use it for), then just call `report.record` directly at the end of the test.

Not a big deal regardless - just something that seemed a bit strange in the setup

aldexis (approved):
Thanks for the refactor!

simulation/src/test/java/com/palantir/dialogue/core/Strategy.java
mpritham: Yeah, happy to add a small note that it's not required to be enforced by all clients (e.g. dialogue), there.

PR #3046, opened 2026-06-19
Support setting a per-client response size limit

aldexis:
## Before this PR
Clients want to be able to protect themselves against misbehaving servers, which may return responses far too large compared to what they'd expect, causing them to OOM when trying to deserialize them

## After this PR

==COMMIT_MSG==
Support setting a per-client response size limit
==COMMIT_MSG==

### DIFF
--- a/dialogue-clients/src/main/java/com/palantir/dialogue/clients/DialogueClients.java
+++ b/dialogue-clients/src/main/java/com/palantir/dialogue/clients/DialogueClients.java
@@ -209,6 +209,12 @@ public final class DialogueClients {
          */
         ReloadingFactory withDeadlineEnforcement(boolean deadlineEnforcementEnabled);
 
+        /**
+         * Configures the maximum size of responses that will be accepted.
+         * Any response received by a server, which exceeds this size, will throw an exception.
+         */
+        ReloadingFactory withMaxResponseSize(long maxResponseSize);
+
         StickyChannelFactory getStickyChannels(String serviceName);
 
         @Beta
--- a/dialogue-clients/src/main/java/com/palantir/dialogue/clients/DialogueClients.java
+++ b/dialogue-clients/src/main/java/com/palantir/dialogue/clients/DialogueClients.java
@@ -28,6 +28,7 @@ import com.palantir.dialogue.ConjureRuntime;
 import com.palantir.dialogue.core.DialogueChannel;
 import com.palantir.dialogue.core.DialogueDnsResolver;
 import com.palantir.refreshable.Refreshable;
+import java.io.InputStream;
 import java.time.Duration;
 import java.util.List;
 import java.util.Map;
@@ -211,7 +212,11 @@ public final class DialogueClients {
 
         /**
          * Configures the maximum size of responses that will be accepted.
-         * Any response received by a server, which exceeds this size, will throw an exception.
+         * Any response received by a server exceeding this size and which dialogue attempts to deserialize will throw
+         * a {@link ResponseSizeTooLargeException}.
+         *
+         * This does not impact responses returning {@link InputStream},
+         * which are not actually deserialized by dialogue.
          */
         ReloadingFactory withMaxResponseSize(long maxResponseSize);
 
--- a/dialogue-clients/src/main/java/com/palantir/dialogue/clients/DialogueClients.java
+++ b/dialogue-clients/src/main/java/com/palantir/dialogue/clients/DialogueClients.java
@@ -211,7 +211,9 @@ public final class DialogueClients {
         ReloadingFactory withDeadlineEnforcement(boolean deadlineEnforcementEnabled);
 
         /**
-         * Configures the maximum size of responses that will be accepted.
+         * Configures the maximum size of response bodies (i.e. headers and other parts of the response are not limited)
+         * that will be accepted.
+         *
          * Any response received by a server exceeding this size and which dialogue attempts to deserialize will throw
          * a {@link ResponseSizeTooLargeException}.
          *
--- a/dialogue-clients/src/main/java/com/palantir/dialogue/clients/ReloadingClientFactory.java
+++ b/dialogue-clients/src/main/java/com/palantir/dialogue/clients/ReloadingClientFactory.java
@@ -115,6 +115,11 @@ final class ReloadingClientFactory implements DialogueClients.ReloadingFactory {
         Refreshable<ServicesConfigBlock> scb();
 
         @Value.Default
+        default long maxResponseSize() {
+            return -1L;
+        }
+
+        @Value.Default
         default ConjureRuntime baseRuntime() {
             return DefaultConjureRuntime.builder().build();
         }
@@ -125,10 +130,15 @@ final class ReloadingClientFactory implements DialogueClients.ReloadingFactory {
         // error serialization format.
         @Value.Derived
         default ConjureRuntime runtime() {
-            return conjureErrorParameterFormat()
-                    .<ConjureRuntime>map(
-                            format -> new ErrorSerializationFormatSettingConjureRuntime(baseRuntime(), format))
-                    .orElseGet(this::baseRuntime);
+            ConjureRuntime runtime = baseRuntime();
+            if (maxResponseSize() > -1) {
+                runtime = new ResponseSizeLimitingConjureRuntime(runtime, maxResponseSize());
+            }
+            if (conjureErrorParameterFormat().isPresent()) {
+                runtime = new ErrorSerializationFormatSettingConjureRuntime(
+                        runtime, conjureErrorParameterFormat().get());
+            }
+            return runtime;
         }
 
         @Value.Default
@@ -416,6 +426,11 @@ final class ReloadingClientFactory implements DialogueClients.ReloadingFactory {
     @Override
     public ReloadingFactory withDeadlineEnforcement(boolean deadlineEnforcementEnabled) {
         return new ReloadingClientFactory(params.withDeadlineEnforcement(deadlineEnforcementEnabled), cache);
+    }
+
+    @Override
+    public ReloadingFactory withMaxResponseSize(long maxResponseSize) {
+        return new ReloadingClientFactory(params.withMaxResponseSize(maxResponseSize), cache);
     }
 
     @Override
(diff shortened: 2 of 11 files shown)


dialogue-clients/src/main/java/com/palantir/dialogue/clients/DialogueClients.java
```diff
@@ -209,6 +209,12 @@ public interface ReloadingFactory
          */
         ReloadingFactory withDeadlineEnforcement(boolean deadlineEnforcementEnabled);
 
+        /**
+         * Configures the maximum size of responses that will be accepted.
+         * Any response received by a server, which exceeds this size, will throw an exception.
```
mpritham: nit: I think we should mention the type of exception in the docs, so devs can know what to catch

dialogue-clients/src/main/java/com/palantir/dialogue/clients/DialogueClients.java
```diff
@@ -209,6 +209,12 @@ public interface ReloadingFactory
...
+         */
+        ReloadingFactory withMaxResponseSize(long maxResponseSize);
```
mpritham, line 223: Should we use a human readable type here? [(e.g.)](https://github.com/palantir/human-readable-types/blob/bc5a4683cc5bf4535776e4befb804aab4fb928ac/human-readable-types/src/main/java/com/palantir/humanreadabletypes/HumanReadableByteCount.java#L42)

dialogue-clients/src/main/java/com/palantir/dialogue/clients/ReloadingClientFactory.java
```diff
@@ -114,21 +114,33 @@ interface ReloadingParams extends AugmentClientConfig {
 
         Refreshable<ServicesConfigBlock> scb();
 
+        @Value.Default
+        default long maxResponseSize() {
```
mpritham: Can we use an `OptionalLong` or `Optional<HumanReadableBytes>`?

dialogue-clients/src/main/java/com/palantir/dialogue/clients/ResponseSizeLimitingConjureRuntime.java
```diff
@@ -0,0 +1,199 @@
+/*
+ * (c) Copyright 2025 Palantir Technologies Inc. All rights reserved.
```
mpritham: nit: 
```suggestion
 * (c) Copyright 2026 Palantir Technologies Inc. All rights reserved.
```

dialogue-clients/src/main/java/com/palantir/dialogue/clients/SizeLimitedInputStream.java
```diff
@@ -0,0 +1,69 @@
...
+        return read;
+    }
+
+    @Override
+    public int read(byte[] buf, int off, int len) throws IOException {
+        int count = super.read(buf, off, len);
+        if (count > 0) {
+            bytesRead += count;
+            if (bytesRead > maxBytes) {
+                throw new SafeIllegalStateException(
```
mpritham: I think we should use`SafeIoException` instead. Also, is it expected that folks will catch this exception? Should we create a custom `ResponseSizeTooLarge` exception?

dialogue-clients/src/main/java/com/palantir/dialogue/clients/DialogueClients.java
mpritham: Also we should mention that input stream deserializers are not limited.

mpritham: Also errors are exempt from this bound.

dialogue-clients/src/main/java/com/palantir/dialogue/clients/ReloadingClientFactory.java
```diff
@@ -114,21 +114,33 @@ interface ReloadingParams extends AugmentClientConfig {
...
+        // We also can't actually set it directly in baseRuntime, because it does not get recomputed when
+        //   withXXX is called on the immutable params class
         @Value.Derived
         default ConjureRuntime runtime() {
-            return conjureErrorParameterFormat()
-                    .<ConjureRuntime>map(
-                            format -> new ErrorSerializationFormatSettingConjureRuntime(baseRuntime(), format))
-                    .orElseGet(this::baseRuntime);
+            ConjureRuntime runtime = baseRuntime();
+            if (maxResponseSize() > -1) {
```
mpritham: Should we allow 0? This will just reject all responses.

dialogue-clients/src/main/java/com/palantir/dialogue/clients/DialogueClients.java
aldexis: dialogue doesn't currently have this dependency, and I'd rather not add it just for this. I think it makes sense for human-written configs to deserialize with that type, but that's effectively downstream of dialogue

dialogue-clients/src/main/java/com/palantir/dialogue/clients/DialogueClients.java
aldexis: Adding for the exception and the input stream deserializers. I think the errors actually include this bound (and should do so), so I'll add a test instead

dialogue-clients/src/main/java/com/palantir/dialogue/clients/DialogueClients.java
bjlaub: is this the uncompressed response size? if so, we should clarify that in the javadoc.

EDIT: does the response size also include headers, or just the response body? (i assume the latter, but again maybe worth just clarifying here)

dialogue-clients/src/main/java/com/palantir/dialogue/clients/ResponseSizeLimitingConjureRuntime.java
```diff
@@ -0,0 +1,199 @@
...
+
+        @Override
+        public <T> Deserializer<T> deserializer(ExceptionDeserializerArgs<T> exceptionDeserializerArgs) {
+            return new ResponseSizeLimitingDeserializer<>(
+                    delegate.deserializer(exceptionDeserializerArgs), maxResponseSize);
+        }
+
+        @Override
+        public Deserializer<Void> emptyBodyDeserializer() {
+            return new ResponseSizeLimitingDeserializer<>(delegate.emptyBodyDeserializer(), maxResponseSize);
```
bjlaub: why is it necessary to wrap this in a `ResponseSizeLimitingDeserializer`?

dialogue-clients/src/main/java/com/palantir/dialogue/clients/SizeLimitedInputStream.java
```diff
@@ -0,0 +1,67 @@
...
+        return read;
+    }
+
+    @Override
+    public int read(byte[] buf, int off, int len) throws IOException {
+        int count = super.read(buf, off, len);
+        if (count > 0) {
+            bytesRead += count;
+            if (bytesRead > maxBytes) {
+                throw new ResponseSizeTooLargeException(maxBytes);
```
bjlaub, line 62: do we want to also emit a metric here?

aldexis: I tried to add the following two metrics
```yaml
  client.response:
    docs: Dialogue response metrics.
    tags: [channel-name, service-name, endpoint]
    metrics:
      size:
        type: histogram
        docs: Measures the size of the response body in bytes.
      size.exceeded:
        type: meter
        docs: Meter measuring the number of times the response body size exceeded the configured limit.
```
but the issue is that we don't have access to any of the tags which would make sense here. Without them, I think the metrics themselves are far less useful, so I'm inclined to not add anything 🤔

dialogue-clients/src/main/java/com/palantir/dialogue/clients/ResponseSizeLimitingConjureRuntime.java
aldexis: At first, I thought it was unnecessary, but it turns out the responses may be arbitrarily large as well and could OOM in the same way. This actually also impacts the input stream deserializers, so I'll wrap them, but only limit on errors, and expand the tests to cover all the deserializers

dialogue-clients/src/main/java/com/palantir/dialogue/clients/DialogueClients.java
aldexis: Can confirm this is the uncompressed size, and also does not include headers, but just the request body. I've added integration tests to confirm these

dialogue-clients/src/main/java/com/palantir/dialogue/clients/ResponseSizeLimitingConjureRuntime.java
```diff
@@ -0,0 +1,222 @@
...
+        public T deserialize(Response response) {
+            if (onlyLimitErrors && !ErrorDecoder.INSTANCE.isError(response)) {
+                return delegate.deserialize(response);
+            } else {
+                return delegate.deserialize(new ResponseSizeLimitingResponse(response, maxResponseSize));
+            }
+        }
+
+        @Override
+        public Optional<String> accepts() {
```
atrocities, line 176: Any particular reason why this doesn't call the delegate? The other example of a delegating deserializer (ErrorHandlingVoidDeserializer) seems to do this.

dialogue-clients/src/main/java/com/palantir/dialogue/clients/SizeLimitedInputStream.java
```diff
@@ -0,0 +1,68 @@
...
+package com.palantir.dialogue.clients;
+
+import com.palantir.logsafe.exceptions.SafeIllegalStateException;
+import java.io.FilterInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+
+/**
+ * Wrapper class that limits the number of bytes read from an underlying {@link InputStream}.
+ * Upon reading more bytes than the specified limit, a {@link SafeIllegalStateException} is thrown.
```
atrocities: Update this to ResponseSizeTooLargeException too

atrocities (dismissed):
Nit and request for clarification. Approach looks fine otherwise. Going to mark as approve.

dialogue-clients/src/main/java/com/palantir/dialogue/clients/ResponseSizeLimitingConjureRuntime.java
aldexis: That's an oversight - thanks for catching!

aldexis:
Thanks for the review @atrocities - I've addressed your comments, which dismissed your review, so if you could re-approve, this would be great! ❤️

atrocities (approved):

atrocities:
Re-approved. Alas it looks like my approval doesn't count though.

PR #3052, opened 2026-06-23
Add opt-in exponential ramp concurrency limiter

mpritham:
## Before this PR

Dialogue uses AIMD to update the max number of concurrent connections to servers and endpoints. When there is a spike in load AIMD grows the permit count too slowly despite not getting any negative signals (429, 503, etc. - see the `Behavior` enum) from servers/endpoints. Once it has found an appropriate concurrency limit, AIMD cautiously probes for additional capacity. ["Successful" requests](https://github.com/palantir/dialogue/blob/42c86f722128a15f09ec52d7d9027a9af7e15d42/dialogue-core/src/main/java/com/palantir/dialogue/core/CautiousIncreaseAggressiveDecreaseConcurrencyLimiter.java#L99-L145) increase the limit by `1 / limit`, producing approximately one additional permit per round trip. While this is safe at steady state, convergence is slow when request load suddenly increases. A client starting at a limit of 20 may continue admitting substantially less work than a healthy downstream server can process. Requests queue while the limiter slowly discovers capacity, increasing latency even though the downstream has not asked the client to back off.

This has been observed in real workloads (and has led to #2950): the client limiter remained saturated and accumulated substantial queue time while the downstream retained worker headroom and emitted effectively no load-shedding signals. Completely disabling concurrency limiting avoids the slow ramp, but also removes the protection that prevents clients from overwhelming a genuinely saturated downstream.

This PR:
1) introduces a new experimental update function called "exponential ramp" that is inspired by [slow start](https://developer.mozilla.org/en-US/docs/Glossary/TCP_slow_start) to exponentially increase permit count until a `dropped` signal is received.
2) adds a simulation scenario for the "spiky load" workflow
3) adds a load-sweep simulation

## After this PR

==COMMIT_MSG==
Add an opt-in exponential ramp concurrency limiter that discovers available downstream capacity more quickly after sudden load increases, while retaining AIMD behavior after the first decrease signal.
==COMMIT_MSG==

The existing AIMD limiter remains the default.

Clients can opt into the experimental limiter using:
```java
  DialogueClients.create(servicesConfig)
          .withConcurrencyLimiterSlowStart(true);
```

There is no accompanying change to the `conjure-java-runtime-api`. At this stage, I would prefer that users explicitly constructed clients enabling this, not relying on any web-server framework utilities to automatically create clients with this flag enabled.

The exponential ramp limiter retains the existing utilization gate: it only increases the limit when the request acquired its permit at 90% or greater utilization. Below that threshold, the limit does not grow.

While the limiter is saturated and has not observed a `dropped()` signal:
  - Each `success()` increases the limit by 1.
  - With approximately one successful completion per permit per round trip, the limit grows multiplicatively.
  - This allows the client to discover available capacity substantially faster than AIMD’s approximately +1 permit per round trip.

The first `dropped` signal permanently ends exponential ramp:
  - The limit is immediately reduced by 10%, matching the existing limiter.
  - All subsequent qualifying successes use the existing AIMD increase of 1 / limit.
  - Exponential ramp cannot resume later in the limiter’s lifetime.

Sticky-channel limiting intentionally continues to use the existing behavior because it does not receive the host/endpoint congestion signals required to end exponential ramp.

#### A burst against a healthy, high-capacity server
This scenario sends 1,000 requests in approximately one millisecond, followed by a low-volume stream. The simulated server has constant response latency and enough capacity to accept the faster ramp.

  <details>
  <summary>Charts</summary>
  <table>
  <tr>
  <th>AIMD</th>
  <th>Exponential Ramp</th>
  </tr>
  <tr>
  <td></td>
  <td></td>
  </tr>
  </table>
  </details>

  AIMD gradually reaches approximately 45 active requests and drains the burst in approximately 2.8 seconds. Exponential ramp reaches approximately 130 active requests and drains it in approximately 1.8 seconds.

  #### A single large spike

  <details>
  <summary>Charts</summary>
  <table>
  <tr>
  <th>AIMD</th>
  <th>Exponential Ramp</th>
  </tr>
  <tr>
  <td></td>
  <td></td>
  </tr>
  </table>
  </details>

  #### Gradual saturation and downstream protection

A separate load sweep gradually increases offered load across a server with a modeled slowdown and load shedding.

The AIMD and exponential ramp results are almost identical:
  - Below the server’s capacity ("knee"), both deliver the offered load with similar latency
  - Once the server signals congestion, exponential ramp exits and behaves like AIMD
  - Both limited variants retain 100% eventual success in this simulation
  - The unlimited control collapses to 0.9–5.7% success above the capacity knee

This only demonstrates that the exponential ramp limiter does better than disabling the limiter when downstream servers are providing signal that they are overwhelmed.

<details>
  <summary>Charts</summary>  
<table>
  <tr>
  <th>AIMD</th>
  <th>exponential ramp</th>
  </tr>
  <tr>
  <td></td>
  <td></td>
  </tr>
  </table>
</details>

Some simulations have also been run internally with a load testing framework on a development environment [with similar findings.](https://pl.ntr/2CY)

## Possible downsides?

This limiter deliberately admits concurrency faster than AIMD before the first `dropped()` signal. If downstream capacity has a sharp cliff or feedback arrives slowly, it may temporarily overshoot and create more concurrent work than the existing limiter.

The implementation bounds this risk:
- It is experimental, explicit, and disabled by default.
- Growth still requires at least 90% limiter utilization.
- The first host- or endpoint-level congestion signal immediately reduces the limit.
- After that signal, the limiter permanently uses the existing AIMD behavior.

The simulations use very simplified server models. They demonstrate the expected limiter dynamics but cannot accurately represent all possible production service behavior. Consumers should enable this selectively for bursty clients whose downstream servers have reliable load-shedding mechanisms in place.


dialogue-core/src/main/java/com/palantir/dialogue/core/SlowStartConcurrencyLimiter.java
```diff
@@ -0,0 +1,200 @@
...
+    private static final double MIN_LIMIT = 1;
+    private static final double MAX_LIMIT = 1_000_000D;
+
+    private final AtomicDouble limit = new AtomicDouble(INITIAL_LIMIT);
+
+    /**
+     * The limit below which we ramp up exponentially and at/above which we probe cautiously (AIMD).
+     * Starts at {@link #MAX_LIMIT} so the very first ramp grows quickly until the first {@code dropped()} signal.
+     */
+    private final AtomicDouble slowStartThreshold = new AtomicDouble(MAX_LIMIT);
```
mpritham: We could make this a boolean which is set after the first `dropped` signal. But for testing, I want to keep the option open for extension where we set the `slowStartThreshold` to be a bit higher than the `newLimit` on decrease.

bjlaub (dismissed):
approving for RC

dialogue-core/src/main/java/com/palantir/dialogue/core/Behavior.java
```diff
@@ -0,0 +1,106 @@
...
+                control.dropped();
+            } else {
+                control.ignore();
+            }
+        }
+    },
+    ENDPOINT_LEVEL() {
+        @Override
+        void onSuccess(Response result, PermitControl control) {
+            if ((Responses.isTooManyRequests(result) && !Responses.isQosDueToCustom(result))
```
atrocities, line 52: Why is this specifically excluding custom QoS responses? Wouldn't this effectively make custom responses count as successes, since they are not dropped, and not ignored?

(Looking farther down I can see that this is a refactor, but would you happen to know?)

dialogue-clients/src/main/java/com/palantir/dialogue/clients/DialogueClients.java
```diff
@@ -222,6 +222,15 @@ public interface ReloadingFactory
...
         ReloadingFactory withMaxResponseSize(long maxResponseSize);
 
+        /**
+         * Experimental: opt into the slow-start concurrency limiter, which ramps the per-host and per-endpoint
+         * concurrency limits up exponentially from a cold start (until the server first pushes back) rather than the
+         * default AIMD limiter's cautious linear probe. Intended for clients that suffer queueing during cold-start
+         * bursts.
+         */
+        @Beta
+        ReloadingFactory withConcurrencyLimiterSlowStart(boolean enabled);
```
atrocities, line 232: nit on nomenclature here - I can see that this comes from TCP slow start, but in this context, it's counterintuitive because it starts faster than AIMD. Browsing through this PR I frequently have to remind myself that 'slow start' is actually the more rapid one.

atrocities (commented):
Explanation and implementation look good. 

Have we considered pathological situations in which many dialogue clients, all using slow-start, would produce problematic amounts of load on certain endpoints, where aimd would not? 

One example would be the case in which DNS changes (e.g. new server node is created), cause DialogueChannel to create new ChannelStates across multiple nodes. The server side would reach saturation, and start 429/503ing more quickly under slow-start than under aimd. This can cause the clients to discover an erroneously lower limit. 

Practically speaking this might not be very severe, as the DNS polling interval will likely spread the limiter initialization across a couple seconds. The aimd recovery after any issues with the initial simultaneous slow-start would mean that any issues would quickly disappear.

Let me see if I can cook up a simulation example

simulation/src/test/java/com/palantir/dialogue/core/SimulationTest.java
```diff
@@ -506,11 +515,117 @@ public void one_big_spike(Strategy strategy) {
...
+    public void burst_then_low_load(Strategy strategy) {
+        burstThenLowLoad(strategy, false);
+    }
+
+    @SimulationCase
+    public void burst_then_low_load_exponential_ramp(Strategy strategy) {
+        burstThenLowLoad(strategy, true);
+    }
+
+    private void burstThenLowLoad(Strategy strategy, boolean exponentialRampEnabled) {
```
mpritham, line 588: This scenario highlights best the usage of this new limiter.

This scenario sends:
  - an initial burst of 1,000 requests
  - followed by 200 requests at 10 RPS
  - against a constant-latency server that does not emit congestion signals

The initial burst and later trickle share the same channel and limiter state. This makes queued work from the burst visible in the latency of later requests.

From the graphs:

With AIMD, active concurrency peaks around 45 and the burst takes roughly 3 seconds to drain. The later low-rate requests therefore spend time behind the original backlog. With exponential ramp, concurrency rises to roughly 130 and the burst drains in approximately 1.7 seconds.

dialogue-clients/src/main/java/com/palantir/dialogue/clients/DialogueClients.java
mpritham: Updated the name to `ExponentialRamp` to make it a bit clearer what it's actually doing.

mpritham:
> Have we considered pathological situations in which many dialogue clients, all using slow-start, would produce problematic amounts of load on certain endpoints, where aimd would not?

I added a simulation for this called "thundering herd". I think the failure case here is that the server rejects more requests than it would've with AIMD. The risk of this limiter is when servers are _not_ protective enough about shedding load and actually do get overwhelmed by clients. While our server-side load shedding is becoming smarter too (with HPL for e.g.) I think we should be careful rolling this out widely and start with a few services that we know are experiencing overly conservative client-side limiting.

atrocities (approved):
Approved for RC

atrocities (dismissed):
Rename to exponential ramp sounds good. There are some spots in the PR description that still reference slow-start (and not just referencing the TCP mechanism) that I think we should fix up though. Approved otherwise

dialogue-core/src/main/java/com/palantir/dialogue/core/Behavior.java
mpritham: Sorry I missed this initially. 

We wanted a QoS reason that would be ignored by Dialogue so that libraries/services which return QoSes due to per-user/enrollment/org limits being exceeded do not affect other users/enrollments/orgs that a Dialogue client could be making requests for. The "custom" reason is that.

atrocities (approved):

PR #3085, opened 2026-07-14
Add load sweep simulations

mpritham:
## Before this PR
Our simulations currently do not tell us how Dialogue's concurrency limiters behave at various levels of offered load. As we make changes to our concurrency limiting logic, we should check in a "load-sweep" simulation that plots the offered load (the client side RPS) against {latency, goodput, success %}. We should expect to find a point on these plots where additional offered load leads to lower latency, decreased goodput, and decreased success %.

## After this PR

==COMMIT_MSG==
Add load sweep simulations
==COMMIT_MSG==

## Possible downsides?

The simulated server is naive. While the number of concurrent requests is less than 20, it responds with 200s in 60ms. When there are > 100 concurrent requests, it responds in 300ms. Between the 20 and 100 concurrent requests, it linearly interpolates a latency from 60ms to 300ms weighted by the number of concurrent requests: (30 + (30*in_flight)/10).


simulation/src/test/java/com/palantir/dialogue/core/LoadSweepTest.java
```diff
@@ -0,0 +1,209 @@
...
+ *  - goodput - successful (200) responses per second
+ *  - success rate as a percentage
+ */
+final class LoadSweepTest {
+    private static final Logger log = LoggerFactory.getLogger(LoadSweepTest.class);
+
+    // In-flight requests a single node serves before its response time cliffs to 5x.
+    private static final int SLOWDOWN_CAPACITY_PER_NODE = 10;
+
+    // In-flight requests a single node serves before it starts shedding with 429s.a
```
aldexis:
```suggestion
    // In-flight requests a single node serves before it starts shedding with 429s.
```

simulation/src/test/java/com/palantir/dialogue/core/LoadSweepTest.java
```diff
@@ -0,0 +1,209 @@
...
+                ? new String(Files.readAllBytes(Paths.get(txtPath)), StandardCharsets.UTF_8)
+                : "";
+
+        if (System.getenv().containsKey("CI")) {
+            // Don't regenerate files on CI: the summary is asserted against the checked-in .txt, but the rendered .png
+            // may not be byte-stable across machines, so rewriting it would dirty the working tree and break
+            // publishing.
+            assertThat(onDisk)
+                    .describedAs("Run tests locally to update checked-in file: %s", txtPath)
+                    .isEqualTo(summary);
```
aldexis, line 133: Isn't it unlikely that we have the same p99 across multiple runs and machines?

simulation/src/test/java/com/palantir/dialogue/core/SimulationTest.java
```diff
@@ -772,7 +772,8 @@ public static void afterClass() throws IOException {
         String txtSection = buildTxtSection();
         String images = buildImagesTable();
         String report = String.format(
-                "# Report%n<!-- Run SimulationTest to regenerate this report. -->%n%s%n%n%s%n", txtSection, images);
+                "# Report%n<!-- Run SimulationTest and LoadSweepTest to regenerate this report. -->%n%s%n%n%s%n",
+                txtSection, images);
```
aldexis: We'd need to run `LoadSweepTest` _then_ `SimulationTest`, correct? Is there a reason we didn't fold `LoadSweepTest` into another test here?

aldexis (dismissed):

simulation/src/test/java/com/palantir/dialogue/core/LoadSweepTest.java
mpritham: The figures in these simulations should be the same across runs. We use [`NanosecondPrecisionDeterministicScheduler`](https://github.com/palantir/dialogue/blob/d94d8713af1182c6ec6cf3be9c68e24ebb879088/simulation/src/main/java/com/palantir/dialogue/core/NanosecondPrecisionDeterministicScheduler.java#L80) and [`TestCaffeineTicker` ](https://github.com/palantir/dialogue/blob/d94d8713af1182c6ec6cf3be9c68e24ebb879088/simulation/src/main/java/com/palantir/dialogue/core/TestCaffeineTicker.java#L23)

simulation/src/test/java/com/palantir/dialogue/core/SimulationTest.java
mpritham: Refactored this to create a new file that shares this logic. I didn't just fold this into `SimulationTest` because the graphs generated by this test look different from what is measured in `SimulationTest`. I intend to expand the tests in `LoadSweepTest` as well.

simulation/src/test/java/com/palantir/dialogue/core/LoadSweepTest.java
aldexis: Aaah, nice!

aldexis (approved):

PR #3104, opened 2026-07-30
Queue timeouts respect enforced deadline values

mpritham:
## Before this PR
Queue timeout enforcement only considered the client’s configured `queueTimeout`.

For example, a request with five seconds remaining on its enforced deadline and a one-minute `queueTimeout` could remain queued for the full minute, despite no longer being able to complete successfully. Without a configured `queueTimeout`, it could remain queued indefinitely.

## After this PR
==COMMIT_MSG==
Fail queued requests when their enforced deadline expires.
==COMMIT_MSG==

Dialogue resolves the enforced deadline once per logical request and shares its absolute expiration across queue layers and retries. Each queue uses the earlier of:
  - The remaining request deadline.
  - The configured `queueTimeout`

The configured queue-timeout budget continues to reset for each retry, while the request deadline remains fixed across the entire logical request.

The `dialogue.client.request.queue.timeout` metric now includes a `source` tag with either `deadline` or `configured`.

## Possible downsides?
Requests may now fail while queued with a `QueueTimeoutException` when their enforced deadline expires, rather than remaining queued until capacity becomes available. This can be alarming to some clients, but with deadlines enforced, this is the right behavior. Currently, a deadline expiration while in a Dialogue queue does not throw an exception pre-emptively.

### DIFF
--- a/dialogue-core/src/main/java/com/palantir/dialogue/core/QueueTimeoutAttachments.java
+++ b/dialogue-core/src/main/java/com/palantir/dialogue/core/QueueTimeoutAttachments.java
@@ -21,47 +21,39 @@ import com.palantir.dialogue.Request;
 import com.palantir.dialogue.RequestAttachmentKey;
 import javax.annotation.Nullable;
 
-/**
- * Attachment for sharing a queue timeout expiration across both the channel-level and endpoint-level
- * {@link QueuedChannel} instances. The first queue to enqueue a request stamps an absolute expiration time (nanos from
- * {@link com.github.benmanes.caffeine.cache.Ticker#read()}). The second queue reads the existing expiration and uses
- * the remaining budget.
- */
+/** Shares configured queue-timeout and deadline expirations across {@link QueuedChannel} instances. */
 final class QueueTimeoutAttachments {
-    private static final RequestAttachmentKey<Long> QUEUE_EXPIRATION_NANOS = RequestAttachmentKey.create(Long.class);
-    private static final RequestAttachmentKey<DeadlineExpiration> DEADLINE_EXPIRATION =
-            RequestAttachmentKey.create(DeadlineExpiration.class);
+    private static final RequestAttachmentKey<Long> CONFIGURED_EXPIRATION_NANOS =
+            RequestAttachmentKey.create(Long.class);
+    private static final RequestAttachmentKey<Long> DEADLINE_EXPIRATION_NANOS = RequestAttachmentKey.create(Long.class);
 
     private QueueTimeoutAttachments() {}
 
-    /** Clears the expiration so the next queue stamps a fresh budget (used on retry). */
-    static void clearExpiration(Request request) {
-        request.attachments().remove(QUEUE_EXPIRATION_NANOS);
+    /** Clears the configured expiration so the next queue stamps a fresh budget (used on retry). */
+    static void clearConfiguredExpiration(Request request) {
+        request.attachments().remove(CONFIGURED_EXPIRATION_NANOS);
     }
 
-    /** Sets the expiration only if one hasn't been set yet. */
-    static void setExpirationIfAbsent(Request request, long expirationNanos) {
-        request.attachments().putIfAbsent(QUEUE_EXPIRATION_NANOS, expirationNanos);
+    /** Returns the configured expiration, initializing it to the candidate value if absent. */
+    static long getOrInitializeConfiguredExpiration(Request request, long candidateExpirationNanos) {
+        Long existing = request.attachments().putIfAbsent(CONFIGURED_EXPIRATION_NANOS, candidateExpirationNanos);
+        return existing != null ? existing : candidateExpirationNanos;
     }
 
     @VisibleForTesting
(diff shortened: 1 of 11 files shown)


kunalrkak (dismissed):
+1 for RC, will review shortly

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
```diff
@@ -375,12 +376,15 @@ private ScheduledFuture<?> scheduleTimeoutFromExpiration(
...
     @VisibleForTesting
     void failWithQueueTimeout(
+            QueueTimeoutSource source,
             SettableFuture<Response> responseFuture,
             DetachedSpan span,
             IdempotentTimerContext timer,
             QueueSizeAccounting accounting) {
-        if (responseFuture.setException(new QueueTimeoutException(channelName, queueTimeoutNanos))) {
-            queueTimeoutCounter.get().inc();
+        if (responseFuture.setException(new QueueTimeoutException(channelName, source, queueTimeoutNanos))) {
```
kunalrkak, line 384: Thoughts on whether to use a DeadlineExpiredException if the resolved timeout came from the deadline state?

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
```diff
@@ -810,8 +834,8 @@ public List<SafeArg<?>> queueFullSafeArgs() {
         }
 
         @Override
-        public Counter requestQueueTimeout() {
-            return requestQueueTimeoutSupplier.get();
+        public Counter requestQueueTimeout(QueueTimeoutSource source) {
+            return (source == QueueTimeoutSource.DEADLINE ? deadlineTimeoutSupplier : configuredTimeoutSupplier).get();
```
kunalrkak, line 838: nit: we could switch over both enum values here

dialogue-core/src/main/java/com/palantir/dialogue/core/QueueTimeoutAttachments.java
```diff
@@ -21,29 +21,39 @@
...
+    /** Clears the configured expiration so the next queue stamps a fresh budget (used on retry). */
+    static void clearConfiguredExpiration(Request request) {
+        request.attachments().remove(CONFIGURED_EXPIRATION_NANOS);
     }
 
-    /** Sets the expiration only if one hasn't been set yet. */
-    static void setExpirationIfAbsent(Request request, long expirationNanos) {
-        request.attachments().putIfAbsent(QUEUE_EXPIRATION_NANOS, expirationNanos);
+    /** Returns the configured expiration, initializing it to the candidate value if absent. */
+    static long getOrInitializeConfiguredExpiration(Request request, long candidateExpirationNanos) {
```
kunalrkak, line 38: nit: I think the old naming style was better, why not `setConfiguredExpirationIfAbsent`, and let the get be implied from the return?

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
```diff
@@ -324,45 +323,47 @@ private ScheduledFuture<?> scheduleQueueTimeout(
...
         // reflects only live requests rather than lingering until the next scheduleNextTask() drain pops the entry.
-        return scheduler.schedule(
-                () -> failWithQueueTimeout(responseFuture, span, timer, accounting), delayNanos, TimeUnit.NANOSECONDS);
+        return timeoutScheduler.schedule(
+                () -> failWithQueueTimeout(expiration.source(), responseFuture, span, timer, accounting),
+                delayNanos,
+                TimeUnit.NANOSECONDS);
+    }
+
+    private Optional<EffectiveExpiration> initializeAndGetEffectiveExpiration(Request request) {
```
kunalrkak, line 349: maybe worth writing a quick javadoc here explaining this is just taking the min of the two values

kunalrkak (commented):
few comments, overall lgtm!

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
mpritham: I think we do want to surface this as a `DeadlineExpiredException` when the source of the timeout is from deadlines. I don't think we expose whether the deadline state is "internal" or "external" in any API though.

dialogue-core/src/main/java/com/palantir/dialogue/core/QueuedChannel.java
```diff
@@ -171,12 +176,17 @@ private static OptionalLong toNanos(Optional<Duration> queueTimeout) {
         this.queuedCalls = new ProtectedConcurrentLinkedDeque<>();
         this.maxQueueSize = maxQueueSize;
         this.queueTimeoutNanos = queueTimeoutNanos;
+        this.clientEnforcement = deadlineEnforcement
```
mpritham, line 179: This is derived from the client configuration, not the TraceLocal.

parker89 (approved):

PR #3154, opened 2026-08-21
Allow building a client with a specific channel name

Daniel-Davies:
We want to hand roll a dialogue client while we wait for https://github.com/palantir/conjure-java-runtime/issues/3585 to resolve.

We will supply a ClientConfiguration with a proxy option to achieve this, which we can pass to the DialogueClients builder. Unfortunately, there is no way to supply a ServiceName to the getNonReloading method when supplying your own ClientConfiguration. This matters for us because metrics are tagged with the service name, and the default is arguably quite non-intuitive (using the client class name, which can be detached from the destination that the client is trying to reach. Though this is probably a limitation of ClientConfiguration not carrying a channel name).

### Alternatives considered

This is the lowest lift solution. Another option here would be to add a withProxySelector option onto the reloading client factory that dialogue publishes ([here](https://github.com/palantir/dialogue/blob/459bc881aaa58d68fb786c76161e66a704114801/dialogue-clients/src/main/java/com/palantir/dialogue/clients/ReloadingClientFactory.java#L74)). Comparison PR [here](https://github.com/palantir/dialogue/pull/3155).

### DIFF
--- a/dialogue-clients/src/main/java/com/palantir/dialogue/clients/DialogueClients.java
+++ b/dialogue-clients/src/main/java/com/palantir/dialogue/clients/DialogueClients.java
@@ -113,20 +113,14 @@ public final class DialogueClients {
         <T> T getNonReloading(Class<T> clientInterface, ClientConfiguration clientConfiguration);
 
         /**
-         * Equivalent to {@link #getNonReloading(Class, ClientConfiguration)}, but identifies the client in metrics
-         * using the provided {@code serviceName} rather than the name of {@code clientInterface}.
-         *
-         * This is intended for callers which build many clients from a single generated interface, one per upstream,
-         * where the interface describes a protocol rather than a specific service. A dispatcher calling out to a set
-         * of configured webhooks is the motivating example: the derived name is identical for every such client, so
-         * their metrics collapse into a single series even though the upstreams are unrelated.
+         * Equivalent to {@link #getNonReloading(Class, ClientConfiguration)}, but identifies the upstream using the
+         * provided {@code serviceName} rather than the name of {@code clientInterface}.
          *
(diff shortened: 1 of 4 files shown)


dialogue-clients/src/main/java/com/palantir/dialogue/clients/DialogueClients.java
```diff
@@ -111,6 +111,23 @@ public interface ClientConfigurationNonReloadingClientFactory {
...
+         * of configured webhooks is the motivating example: the derived name is identical for every such client, so
+         * their metrics collapse into a single series even though the upstreams are unrelated.
+         *
+         * Behaviour is undefined if {@code clientConfiguration} contains no URIs.
+         *
+         * @param clientInterface Dialogue client interface annotated with {@link com.palantir.dialogue.DialogueService}
+         * @param serviceName Name of the conceptual upstream, used to identify this client in metrics
+         * @param clientConfiguration Configuration which <b>must</b> contain a user-agent
+         */
+        <T> T getNonReloading(Class<T> clientInterface, String serviceName, ClientConfiguration clientConfiguration);
```
atrocities, line 124: Since it's equivalent to the deprecated method above, shouldn't this also effectively be deprecated?

Daniel-Davies: Happy to make this deprecated from the get-go 👍

atrocities (approved):
Approved for the sake of unblocking. Please do be aware that getNonReloading with ClientConfiguration does not go through the ChannelCache at all - be aware of the implications if this ever does wind up being on a hot path and no other caching mechanisms wrap the call.

PR #3172, opened 2026-08-31
Add TypeMarker factory methods for simple types and common parameterized types

atrocities:
This is part of a broader effort to eliminate most TypeMarker anonymous classes for Conjure generated code. These anonymous classes pollute the classpath and unnecessarily consume memory at runtime, and aren't strictly necessary a vast majority of the time. 

Introduce static factories into TypeMarker for the parameterized types that Conjure generates: Set, Map, Optional, and List, and change WrappingTypeMarker to be genericized, which allows adding a new type-safe non-parameterized factory method. Use these static factories in the annotation processor instead of generaeting anon classes for the applicable types. Parameterized base types not covered by the static factories and complicated nested types will still generate anon classes. 

Conjure code generation will eventually use these static factories as well for generated Dialogue code and exception markers.

### DIFF
--- a/dialogue-target/src/main/java/com/palantir/dialogue/TypeMarker.java
+++ b/dialogue-target/src/main/java/com/palantir/dialogue/TypeMarker.java
@@ -133,6 +133,10 @@ public abstract class TypeMarker<T> {
             this.actualTypeArguments = actualTypeArguments.clone();
             for (Type actualTypeArgument : this.actualTypeArguments) {
                 Preconditions.checkNotNull(actualTypeArgument, "Type argument is required");
+                Preconditions.checkArgument(
+                        !(actualTypeArgument instanceof Class<?> argumentClass && argumentClass.isPrimitive()),
+                        "Primitive types cannot be used as type arguments",
+                        SafeArg.of("typeArgument", actualTypeArgument));
             }
         }
 
--- a/dialogue-target/src/main/java/com/palantir/dialogue/TypeMarker.java
+++ b/dialogue-target/src/main/java/com/palantir/dialogue/TypeMarker.java
@@ -22,11 +22,23 @@ import com.palantir.logsafe.exceptions.SafeIllegalArgumentException;
 import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
 import java.lang.reflect.TypeVariable;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.Set;
+import org.jspecify.annotations.Nullable;
 
 /**
  * Captures generic type information.
  *
- * Usage example: <pre>new TypeMarker&lt;List&lt;Integer&gt;() {}</pre>.
+ * <p>Usage examples:
+ * <pre>{@code
+ * TypeMarker<String> stringMarker = TypeMarker.of(String.class);
+ * TypeMarker<List<String>> listMarker = TypeMarker.listOf(String.class);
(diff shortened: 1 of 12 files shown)


atrocities:
For reviewers - this change should only affect the behavior of the Dialogue annotation processor, for which I'm testing on some internal repos. All other changes should be purely additive.

dialogue-target/src/main/java/com/palantir/dialogue/TypeMarker.java
```diff
@@ -78,12 +90,104 @@ public final String toString() {
...
+                return Objects.equals(getOwnerType(), that.getOwnerType())
+                        && rawType.equals(that.getRawType())
+                        && Arrays.equals(actualTypeArguments, that.getActualTypeArguments());
+            }
+            return false;
+        }
+
+        @Override
+        public int hashCode() {
+            return Arrays.hashCode(actualTypeArguments) ^ Objects.hashCode(getOwnerType()) ^ Objects.hashCode(rawType);
```
aldexis, line 173: I've never seen this way of building the hashcode previously (I'm more familiar with the immutables style, e.g.
```
    int h = 5381;
    h += (h << 5) + <field1 hashcode>;
    h += (h << 5) + <field2 hashcode>;
    return h;
```
etc

Searching for it, I now see it in some other places, but curious if you know of specific benefits

dialogue-target/src/main/java/com/palantir/dialogue/TypeMarker.java
```diff
@@ -78,12 +90,104 @@ public final String toString() {
...
+
+        @Override
+        public Type getRawType() {
+            return rawType;
+        }
+
+        @Override
+        public @Nullable Type getOwnerType() {
+            return null;
+        }
```
aldexis, line 156: Is this actually always null? Or is it simply because of the special cases we use this with? (e.g. if in the future we want to handle more generic types)

aldexis (commented):
Overall looks good to me, but I'd like to double check:
- Have we tested this in real conditions? e.g. have we tried to use these new type markers to deserialize in jackson and/or have tested this with a dialogue annotations processor client?
- Have we tested the updated typemarkers class with an existing, already generated code? (revapi didn't complain, so it's likely fine, but I want to double check there aren't weird things going on with jackson)

dialogue-target/src/main/java/com/palantir/dialogue/TypeMarker.java
atrocities: No specific benefits, this just happened to be from [sun.reflect.generics.reflectiveObjects.ParameterizedTypeImpl](https://github.com/JetBrains/jdk8u_jdk/blob/master/src/share/classes/sun/reflect/generics/reflectiveObjects/ParameterizedTypeImpl.java#L198-L204), which this is based off of.

dialogue-target/src/main/java/com/palantir/dialogue/TypeMarker.java
atrocities: Yes, because this is only used for the 4 outer parameterized types in conjure, I think it will always be null. I'd imagine that this is a reasonably safe assumption because should we ever need to add more generic types, there would need to be an explicit change to make them use these static factories instead of anon classes, at which point I'd expect this implementation of ParameterizedType to be revisited.

atrocities:
@aldexis Messaged internally about testing. 

Re: TypeMarkers and potential breaks with existing code - I don't think there will be issues with Jackson: references in dialogue only seem to care about the Type field, and I don't see any evidence that Jackson cares about the new static factories. Empirically, integration testing with the internal repo reveals no issues either.

aldexis (approved):

Issue #3180, opened 2026-09-04
Null-check `RequestBody` values passed into `Request.body(Optional<? extends RequestBody> value)`


mpritham:
Currently, we do not ensure that the `RequestBody` value passed into [this method](https://github.com/palantir/dialogue/blob/70196eea62323d603264dd9de2359b804ff795d9/dialogue-target/src/main/java/com/palantir/dialogue/Request.java#L289-L292) is non-null. This allows callers to create an invalid `Request` whose `body()` method returns `null` and can lead to confusing NPEs.

We should null-check the value provided is not `null`. Note we are checking that the argument is non-null, not the enclosing value of the `Optional`: the constructors to [`Optional`](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/Optional.html#of(T)) throw NPE exceptions if the value being encapsulated is `null`.

When making this change, we should 1) find all of the usages of this method and ensure that there are no callers that expect to be able to pass a null value (this would be very surprising since requests would fail, but [you never know](https://www.hyrumslaw.com/)), 2) add a test to ensure we get a `SafeNullPointerException` when the method is passed a `null` value.

PR #3183, opened 2026-09-08
Ensure that RequestBody value is non-null

ericwang1409:
## Before this PR
Addressing #3180. Currently there is no null-check for the RequestBody value passed into [this method](https://github.com/palantir/dialogue/blob/70196eea62323d603264dd9de2359b804ff795d9/dialogue-target/src/main/java/com/palantir/dialogue/Request.java). 

## After this PR
Found all usages of this method to ensure that there are no callers that expect to be able to pass a null value and added a test to ensure we get a SafeNullPointerException when the method is passed a null value.

## Possible downsides?
Possibly could have missed a usage of this method that expected a null value since the search was done manually.

### DIFF
--- a/dialogue-target/src/main/java/com/palantir/dialogue/Request.java
+++ b/dialogue-target/src/main/java/com/palantir/dialogue/Request.java
@@ -287,7 +287,7 @@ public final class Request {
 
         @SuppressWarnings("unchecked")
         public Request.Builder body(Optional<? extends RequestBody> value) {
-            body = (Optional<RequestBody>) value;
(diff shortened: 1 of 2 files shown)


palantirtech:
Thanks for your interest in [palantir/dialogue](https://github.com/palantir/dialogue), @ericwang1409! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

mpritham (dismissed):
Thanks!

mpritham (approved):

